PR #35, opened 2020-05-18
Add Polykey CLI

### LINKED ISSUE #23: CLI app architecture
I'm thinking of using the following libraries in building the CLI app:

For building the core of the CLI:
https://www.npmjs.com/package/oclif

For text styling:
https://www.npmjs.com/package/chalk

For text output and formatting:
https://www.npmjs.com/package/winston

There were some other contenders such as [commander](https://www.npmjs.com/package/commander) and [yargs](https://www.npmjs.com/package/yargs), both of which have a much larger user base than oclif, which I'm mainly putting down to having mature codebase. 

The reasons I decided to go with oclif are: 

1. It's modern
2. Built by Heruko (Heruko CLI uses oclif)
3. Coded in typscript so it will integrate well with our codebase
4. Good multi/sub-command support
5. Clean api and good docs

### LINKED ISSUE #33: Port codebase to Typescript
The codebase currently utilises flow type annotations. We want to migrate to the more modern typescript for Polykey. This isn't a big chunk of work since, the code base isn't too large at this point.

### LINKED ISSUE #38: Passphrase Strength in Key Generation
I think we should throw an error for a weak passphrase passed into the kbpgp key generation. We can validate password strength with [zxcvbn](https://github.com/dropbox/zxcvbn). They define anything under a score of 2 to be somewhat guessable (guesses < 10^8) so I think its reasonable to error if the user provided passphrase is below this. 

Note: this issue is solved in the upcoming PR #35, but I thought it best to document our decision here.

robert-cronin:
There is additional development that is happening on the `development` branch but that is more general so I thought I should start breaking it into separate, manageable PRs and this is the first.

Fixes #23
Fixes https://github.com/MatrixAI/PolyKey/issues/33
Fixes #38


CMCDragonkai:
Does this address https://github.com/MatrixAI/PolyKey/issues/33? Since you're already starting to write typescript.

CMCDragonkai:
I'd like a quick schema of all the CLI commands in the README.md.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+import { program } from 'commander'
+import inquirer from 'inquirer'
+import PolyKey from '../src/Polykey'
+import chalk from 'chalk'
+import os from 'os'
+import fs from 'fs'
+
+let pk: PolyKey
+// Write a `file`
+let tempDir: string
```
CMCDragonkai: Why is this defined here? It seems to be only used in one place. Unless the `tempDir` is meant to be shared across the place. But I wonder if it's better to write your code using modular functions and push down the `tempDir` instead of having global mutables. Usually global mutables are not a good idea except in special cases. This goes for the `pk` variable too.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+/*******************************************/
+// initialization
+function initPolyKey(
+    password: string | null,
+    polykeyPath: string | undefined = undefined,
```
CMCDragonkai: I believe we should be using typescript's `?` operator for optional parameters right? That should simplify things, and there's no need to default the parameters to `undefined`, since that's by default.

Although for JS/TS. I prefer using `null` for explicit "emptiness". Whereas `undefined` is more for implicit emptiness. If you use `null` you must explicitly say `param: type | null = null`. Whereas if you are going to leave it for implicit emptiness, then just `?param: type` is fine.

When something is `null`, that means somebody meant for the thing to exist, and be null. Whereas when using `undefined` that means nobody even meant for the thing to exist.

Thus I don't think default parameter of `undefined` is correct.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+    publicKeyPath: string | undefined = undefined,
+    privateKeyPath: string | undefined = undefined,
+    privatePassphrase: string | undefined = undefined,
+    verbose: boolean | undefined = undefined
+) {
+    try {
+        // Initialize polykey
+        if (password !== null) {
+            pk = new PolyKey(Buffer.from(password), undefined, undefined, polykeyPath)
```
CMCDragonkai: I'm wondering if a `config` object as the parameter for `Polykey` might be better.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+        } else if (process.env.PK_KEY !== undefined) {
+            pk = new PolyKey(Buffer.from(process.env.PK_KEY!), undefined, undefined, polykeyPath)
+        } else {
+            console.log(chalk.red('failed to initialize polykey: password was null'));
+        }
+        // import keys if provided
+        if (publicKeyPath !== undefined) {
```
CMCDragonkai: Please create a bootstrapping diagram/docs in the README.md that explains the setup path of these things.

CMCDragonkai: Similar to our bootstrap diagram in Emergence nodes.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+            pk.km.loadPublicKey(publicKeyPath)
+            if (verbose) {
+                console.log(chalk.green('public key successfully imported'))
```
CMCDragonkai: I'm wondering if its better to create our own special `logger` here, and then use an enum to choose colours. This sort of `chalk.green` is a cross cutting concern, and we shouldn't need to specify this inline. But instead control somewhere else. Then like Python, we can have different classes of messages and colour them appropriately.

But again I'd say this is not important atm, and I'd remove this sort of code atm, and we can design this later.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+            }
+        }
+        if (privateKeyPath !== undefined) {
+            if (privatePassphrase !== undefined) {
+                try {
+                    pk.km.loadPrivateKey(privateKeyPath)
+                    if (verbose) {
+                        console.log(chalk.green('private key successfully imported'))
+                    }
+                } catch (err) {
```
CMCDragonkai: For the application entrypoint, you should be doing it similar to in Loja. In Loja we have 1 catch all exception handler that is the root exception handler. At that point, the program should exit with a fatal error and report it.

All other exceptions should be caught and handled. Also it doesn't make sense to handle it only by logging out an error. That should only happen at the root exception handler. Because any other exception handler should actually **handle** the error so there's no more error!

See loja's server/index.ts.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+/*******************************************/
+const polykey = new program.Command()
+polykey
+    .version(require('../package.json').version, '--version', 'output the current version')
+
+/*******************************************/
+// init
+const config = polykey.command('config')
+    .description('configure polykey')
+    .option('--password <password>', 'provide the password to polykey')
```
CMCDragonkai: Do these provide short options as well? See Python's argparse, we should have both short options and long options.

bin/polykey.ts
```diff
@@ -0,0 +1,373 @@
...
+            }
+        } catch (err) {
+            console.log(`Failed to list secrets for vault '${vaultName}': ${err.message}`)
+        }
+    })
+
+
+/*******************************************/
+// secrets add
+const secretAdd = secrets.command('add')
```
CMCDragonkai: This file is getting pretty long. Are we able to split up the commands so that way the entrypoint into the program can be a high level listing of the subcommands, and then each sub command can have their own file?

jest.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  "roots": [
+    "<rootDir>/tests",
+    "<rootDir>/src"
```
CMCDragonkai: Don't need this line anymore.

src/Identity.ts
```diff
@@ -1,8 +1,76 @@
-// @flow
+// // this module should be used to represent an identity
```
CMCDragonkai: This is all commented out?

src/KeyManager.ts
```diff
@@ -1,13 +1,525 @@
...
+
+// We want a consistent and uniform library. So even if we change crypto libraries for symm or asymm, not much
+// changes for compositional code. KeyManager, returns a 'Key' instance. There can be difference classes for both
+// symm and asymm. Cryptor can then take these Key classes, which expose the crytpo functions. So even if the
+// crypto library changes, Cryptor doesn't have to cahnge and neither does polykey.
+
+
+export default class KeyManager {
+  // TODO: wouldn't keymanager have many sym keys keys to look after?
+  private _keyPair: KeyPair = {private: '', public: '', passphrase: ''}
```
CMCDragonkai: I don't think we need to use `_` prefix. Since you have a real `private` scope now.

src/KeyManager.ts
```diff
@@ -1,13 +1,525 @@
...
+  private _identity: Object | undefined = undefined
+  private _key!: Buffer
+  private _salt!: Buffer
+  private _passphrase!: string
+  private _storePath: string
+  private _fs: typeof fs
+  constructor(
```
CMCDragonkai, line 25: Can you write up a spec about what this class is supposed to do, and even a plantuml diagram that shows the interaction between these pieces.

For complex pieces of software like this, it's always better to start with a design spec in an issue or markdown file somewhere, so I can review according to the logic/spec, instead of just looking at syntax. It's like "reviewing in the small vs reviewing in the large".

src/util.ts
```diff
@@ -0,0 +1,5 @@
+export type KeyPair = {
```
CMCDragonkai: This doesn't look like something that should be in `utils`. Also it should be `utils.ts`, not `util.ts`. This looks like a domain specific type, it should be in some relevant module.

tests/Polykey.unit.test.ts
```diff
@@ -0,0 +1,155 @@
...
+	})
+
+	afterAll(() => {
+		fs.rmdirSync(`${tempDir}`)
+	})
+
+	////////////
+	// Vaults //
+	////////////
+	describe('vaults', () => {
```
CMCDragonkai: I think `vault` tests should be in a separate file entirely.

You want to test a couple things. Start with the domain models first. And then we wrap it like an onion and then testing larger things, to eventually the entire CLI commands.

CMCDragonkai (changes requested):
We need to ensure that we are coding to a design spec, so its easier to verify what is going on. At some point, I'm not supposed to be reviewing the syntax/coding patterns anymore, but at a more higher level.

bin/polykey.ts
robert-cronin: oh yes this was left in by mistake, I will take it out

bin/polykey.ts
robert-cronin: Done.

src/Identity.ts
robert-cronin: This isn't used for anything so I've removed it

jest.config.js
robert-cronin: Removed.

src/KeyManager.ts
robert-cronin: Done.

src/util.ts
robert-cronin: I've put it into the KeyManager class as an exported type

tests/Polykey.unit.test.ts
robert-cronin: Vault tests have been moved to a separate file

bin/polykey.ts
robert-cronin: I have added a short syntax for every option

~/.polykey/.polykey/metadata
```diff
@@ -0,0 +1 @@
+{"vaults":{},"publicKeyPath":null,"privateKeyPath":null,"passphrase":null}
```
CMCDragonkai: What's this file?

bin/polykey.ts
robert-cronin: yeah we could do that but I've realised there are only two things that need to be parameters; the KeyManager and the home directory. Every thing else was moved to the KeyManager itself.

bin/polykey.ts
robert-cronin: I agree, I have replaced it with the more explicit null operator `?`

bin/polykey.ts
robert-cronin: I've added global logger for CLI with enum for message types and corresponding colours.

bin/polykey.ts
robert-cronin: I have created a root exception handler now.

~/.polykey/.polykey/metadata
robert-cronin: this was a file left in by mistake, it h as been removed

src/KeyManager.ts
robert-cronin: I have  put an outline of what the KeyManager class is supposed to do in the README

bin/polykey.ts
robert-cronin: I have put docs in the README that explains this setup, but no diagram as yet, hopefully this should be enough for now

robert-cronin:
> Does this address [MatrixAI/PolyKey#33](https://github.com/MatrixAI/PolyKey/issues/33)? Since you're already starting to write typescript.

Yeah I think that covers it, we have already ported EFS and this library, I was saving the issue in case there were any other libraries yet to be converted like VFS but I remember you mentioned VFS should stay in JavaScript.

bin/polykey.ts
robert-cronin: I've left it as just standard non-config object style parameters, let me know if you think we should have the config object and I will put it in. I left it out because I reduced the parameters required to 2

CMCDragonkai: Ok that's fine.

bin/config/Config.ts
```diff
@@ -0,0 +1,49 @@
+import commander from 'commander'
+import Configstore from 'configstore'
+import { actionRunner, pkLogger, PKMessageType } from '../polykey'
```
CMCDragonkai, line 3: Can we use `@polykey`?

src/KeyManager.ts
```diff
@@ -1,13 +1,454 @@
...
-// kbpgp.const.openpgp
-// so that's good
+
+
+// js imports
+const kbpgp = require('kbpgp')
+var F = kbpgp["const"].openpgp
+const zxcvbn = require('zxcvbn')
+
+// TODO: remember that the symmetric keys will just be stored in vault directories.
```
CMCDragonkai: Can you put these comments into the relevant issue. I don't really like having important notes in the code, especially if they haven't been reified to the code.

src/KeyManager.ts
```diff
@@ -1,13 +1,454 @@
...
+  async generateKeyPair(name: string, email: string, passphrase: string, numBits: number = 4096): Promise<KeyPair> {
+    // Validate passphrase
+    const passValidation = zxcvbn(passphrase)
+    // The following is an arbitrary delineation of desirable scores
+    if (passValidation.score < 2) {
+      console.log(chalk.red(`passphrase score for new keypair is below 2!`))
+    } else if (passValidation.score <4) {
+      console.log(chalk.yellow(`passphrase score for new keypair is below 4!`))
+    } else {
+      console.log(chalk.green(`passphrase score for new keypair is 4 or above.`))
```
CMCDragonkai: Hey you still using `chalk` here!

tests/Polykey.test.ts
```diff
@@ -0,0 +1,96 @@
...
+import fs from 'fs'
+import os from 'os'
+import Polykey from '../src/Polykey'
+
+// js imports
+const kbpgp = require('kbpgp')
+const vfs = require('virtualfs')
+
+function randomString(): string {
+	return Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 5)
```
CMCDragonkai: This should be in our `utils` library.

CMCDragonkai:
The number of tests atm seem pretty low. Are we expecting more tests? There's alot of functionality here or perhaps they are in the alter PRs?

CMCDragonkai:
Let's have a meeting on saturday to discuss this in detail...

robert-cronin:
> The number of tests atm seem pretty low. Are we expecting more tests? There's alot of functionality here or perhaps they are in the alter PRs?

Yeah I was hoping to have a PR specifically for testing so I can spend some dedicated time on it.

tests/Polykey.test.ts
robert-cronin: Is this a dedicated library we publish to npm? for now I just put it in `utils.ts` as we might use it later for something.

src/KeyManager.ts
robert-cronin: oops! good spot, I've taken it out. This would be good to feed into our logger, but that is specific to our CLI. I think I can just remove the notifying of a good passphrase and make a bad passphrase throw an error.

bin/config/Config.ts
robert-cronin: I didn't think you could do that outside the `src` directory, it throws an error for me.

robert-cronin: I've added `@` relative imports to the src directory.

robert-cronin: Yeah I think we wouldn't be able to have relative imports outside of `src` unless we change our baseUrl.

src/KeyManager.ts
robert-cronin: One of these comments was Aashwin (I think?) talking about changing crypto libraries and making it able to support both asymm and symm encryption but I think we no longer want that, we are integrating kbpgp into our code and depending on asymmetric pubkey encryption.

The other comments I feel are already captured in existing issues.

There were also some TODO comments that seemed more like statements so I've take those out as well.

PR #40, opened 2020-06-04
Add Peer Discovery Code

### LINKED ISSUE #1: Using ts-node with @ imports
Note: this issue is just for documentation and has a solution already:

The issue in question is an error that occurs when using imports that use the `@` symbol to denote the module (e.g. `import PolyKey from @polykey/PolyKey.ts`). It turns out `ts-node` doesn't like this and throws an error. There is a whole issue about [tsconfig.json/paths not working with ts-node](https://github.com/TypeStrong/ts-node/issues/138). The solution is to use [tsconfig-paths](https://github.com/dividab/tsconfig-paths) and then suffix all your calls with `ts-node -r tsconfig-paths/register <file>`. One can do this with a handy alias: `alias ts-node='ts-node -r tsconfig-paths/register'`.

For webpack, there is also a module for this: https://github.com/dividab/tsconfig-paths-webpack-plugin

robert-cronin:
This PR will include a `PeerStore` for keeping track of `PeerInfo`. A multicast broadcasting class for peer discovery on the same LAN and various other bug fixes (e.g. relative `@` imports).

Edit: this PR now also includes a mechanism for discovering a social user via a handle (`robert-cronin`) and a service (`github`) - relates to #41 

Fixes https://github.com/MatrixAI/Typescript-Demo-Lib/issues/1


bin/initPolykey.ts
```diff
@@ -1,10 +1,12 @@
 import fs from 'fs'
```
CMCDragonkai, line 1: Everything in `bin` should actually be a real command. I'm pretty sure `initPolykey` is not user-friendly command. I remember writing in the design repo that we should be using `pk ...` commands. Sub commands as good.

bin/initPolykey.ts
```diff
@@ -157,7 +159,19 @@ async function initPolyKey(): Promise<PolyKey> {
       const keyGenerationName = configStore.get('keyGenerationName')
       const keyGenerationEmail = configStore.get('keyGenerationEmail')
       const keyGenerationPassphrase = configStore.get('keyGenerationPassphrase')
-      await keyManager.generateKeyPair(keyGenerationName, keyGenerationEmail, keyGenerationPassphrase)
+
+      // Validate passphrase
+      const passValidation = zxcvbn(keyGenerationPassphrase)
+      // The following is an arbitrary delineation of desirable scores
+      if (passValidation.score < 2) {
+        pkLogger('passphrase score for new keypair is below 2!', PKMessageType.WARNING)
```
CMCDragonkai, line 167: Should we allow an opportunity for the user to redo this? Of course we don't want to limit the user in this case. But in GUI, it's usually a UI info message. For CLI, we just do a warning.

bin/keymanager/KeyManager.ts
```diff
@@ -1,6 +1,6 @@
 import commander from 'commander'
-import { actionRunner, pkLogger, PKMessageType } from '../polykey'
 import initPolyKey from '../initPolykey'
+import { actionRunner, pkLogger, PKMessageType } from '../polykey'
```
CMCDragonkai: Use `@polykey`?

dist/P2P/MulticastPeerDiscovery.js
```diff
@@ -0,0 +1,175 @@
+// import PeerStore from '../PeerStore/PeerStore'
```
CMCDragonkai: This is all commented out? It should be an in issue if it's not yet ready?

src/KeyManager.ts
```diff
@@ -5,8 +5,7 @@ import { promisify } from 'util'
 
 // js imports
 const kbpgp = require('kbpgp')
```
CMCDragonkai: Why are we not using typescript `import ...`?

src/KeyManager.ts
```diff
@@ -331,6 +331,60 @@ class KeyManager {
...
+      } else {
+        throw(Error('no identity available for signing'))
+      }
+
+      ring.add_key_manager(resolvedIdentity)
+      const params = {
+        armored: data,
+        keyfetch: ring
+      }
+      kbpgp.unbox(params, (err, literals) => {
```
CMCDragonkai: If kbpgp takes longer, perhaps we can use the threads.js like we did in js-encryptedfs?

src/P2P/MulticastBroadcaster.ts
```diff
@@ -0,0 +1,193 @@
...
+// This module was also generated with the help of:
+// https://nrempel.com/using-udp-multicast-with-node-js/
+//
+// """
+// In computer networking, the multicast DNS (mDNS) protocol
+// resolves hostnames to IP addresses within small networks
+// that do not include a local name server
+// """
+
+const UDP_MULTICAST_PORT = 5353;
```
CMCDragonkai: Who made this port choice? We can do it as an environment variable? Although I guess we can hard code our own "service" registry.

src/P2P/MulticastBroadcaster.ts
```diff
@@ -0,0 +1,193 @@
...
+    query()
+    return setInterval(query, this.interval)
+  }
+
+  /**
+   * Start sending queries to the LAN.
+   *
+   * @returns {void}
+   */
+  async start(): Promise<void> {
```
CMCDragonkai: Please keep return types consistent. I think some of your methods have no return type. If we specify return types, specify it for all.

src/P2P/MulticastBroadcaster.ts
```diff
@@ -0,0 +1,193 @@
...
+    }).bind(this));
+
+    // Handle messages
+    this.socket.on("message", this.handleMessages.bind(this));
+
+    // Start the query process
+    this.queryInterval = this.queryLAN()
+  }
+
+  private async handleMessages(message: any, rinfo: any) {
```
CMCDragonkai: I'm thinking that this class handles a specific domain of messages. Not just any message at all. Perhaps instead we name this more specifically.

src/P2P/MulticastBroadcaster.ts
```diff
@@ -0,0 +1,193 @@
...
+      encryptedRandomMessage: Buffer.from(encryptedRandomMessage)
+    })
+  }
+
+  /**
+   * Stop sending queries to the LAN.
+   *
+   * @returns {Promise}
+   */
+  async stop(): Promise<void> {
```
CMCDragonkai: Perhaps we can use RAII for this? js-virtualfs had a similar issue here with cyclic dependencies, and thus had explicit constructors and deconstructors. But if we can model the construction/deconstruction of side-effects with the lifetime of objects, this keeps our implementation more cohesive with our programming style.

src/P2P/PeerDiscovery.ts
```diff
@@ -0,0 +1,63 @@
...
+      this.multicastBroadcaster.on('timeout', (timedOutPubKey: string) => {
+        if (timedOutPubKey == pubKey) {
+          reject('The broadcaster stopped looking')
+        }
+      })
+    })
+  }
+
+  async findSocialUser(handle: string, service: 'username' | 'github' | 'twitter' | 'reddit' | 'hackernews'): Promise<PeerInfo> {
+    const url = `https://keybase.io/_/api/1.0/user/lookup.json?${service}=${handle}`
```
CMCDragonkai: Keybase is only one of the ways we can do this. I recommend using a strategy pattern for this. So we can extend it with more social discovery providers.

src/PeerStore/PeerStore.ts
```diff
@@ -0,0 +1,56 @@
+import PeerInfo from "@polykey/PeerStore/PeerInfo"
+
+/**
+ * Responsible for managing known peers, as well as their addresses and metadata
+ */
+class PeerStore {
```
CMCDragonkai: If this is sort of modelled off libp2p, then our peerstore and similar modules can be separate libraries, and even libp2p libraries like multiaddr and multihash.

BTW have you looked at our hs-multiaddr and hs-multihash?

src/Polykey.ts
```diff
@@ -394,22 +358,33 @@ export default class Polykey {
...
+        peerInfo: RPCMessage.encodePeerInfo(peerInfo)
       }
       jsonfile.writeFileSync(this.metadataPath, metadataTemplate)
       this.metadata = metadataTemplate
-    } else if (this.fs.existsSync(this.metadataPath)) {
-      this.metadata = jsonfile.readFileSync(this.metadataPath)
     }
 
+    this.peerStore = new PeerStore(RPCMessage.decodePeerInfo(this.metadata.peerInfo))
+    this.peerDiscovery = new PeerDiscovery(this.peerStore, this.keyManager)
```
CMCDragonkai: Similar issue to the git PR, I think the `Polykey` class is doing too many things. We can make it more cohesive.

src/RPC/PeerInfoMessage.proto
```diff
@@ -0,0 +1,9 @@
+// awesome.proto
```
CMCDragonkai: Are you using GRPC for this? Or just protobuf? If we can make use of GRPC, that would be great, as that aligns it with the rest of Matrix OS.

tests/RPC/RPCMessage.test.ts
```diff
@@ -0,0 +1,23 @@
+import PeerInfo from '@polykey/PeerStore/PeerInfo'
+import RPCMessage from '@polykey/RPC/RPCMessage'
+
+// TODO add tests as part of testing PR
```
CMCDragonkai, line 5: More tests needed!! But after we get the domain modelling correct.

CMCDragonkai (changes requested):
As a general note. This is awesome. But we need to discuss domain modelling details.

bin/initPolykey.ts
robert-cronin: `initPolykey` was more intended to be a function used to get a polykey instance up and running for each command called by the user. It's not exposed to the end user. It also prompts the user in case there are any things missing from polykey's config like private key or passphrase or location. We could put it in the main `polykey.ts` file perhaps?

bin/keymanager/KeyManager.ts
robert-cronin: I got stuck here, I'm not sure how to use the module level imports for this one, at the moment `@polykey` resolves to the `src` directory. Perhaps we could have another `tsconfig path` for the bin folder? something like `@polykey-cli` perhaps?

dist/P2P/MulticastPeerDiscovery.js
robert-cronin: I have just forgotten to build into dist before I committed.

src/KeyManager.ts
robert-cronin: These js-imports have been changed to import statements

src/P2P/MulticastBroadcaster.ts
robert-cronin: Sure we can make it an environment variable, I think that would work well for both electron and nativescript

CMCDragonkai:
Things inside bin should not be imported. Bin things should be completely independent, self-contained programs that each result in an CLI executable.

On 11 June 2020 12:09:35 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@robert-cronin commented on this pull request.
>
>
>
>>  import initPolyKey from '../initPolykey'
>+import { actionRunner, pkLogger, PKMessageType } from '../polykey'
>
>I got stuck here, I'm not sure how to use the module level imports for
>this one, at the moment `@polykey` resolves to the `src` directory.
>Perhaps we could have another `tsconfig path` for the bin folder?
>something like `@polykey-cli` perhaps?
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/40#discussion_r438503630

-- 

CMCDragonkai:
Yea this shouldn't be in bin.

On 11 June 2020 12:06:53 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@robert-cronin commented on this pull request.
>
>
>
>> @@ -1,10 +1,12 @@
> import fs from 'fs'
>
>`initPolykey` was more intended to be a function used to get a polykey
>instance up and running for each command called by the user. It's not
>exposed to the end user. It also prompts the user in case there are any
>things missing from polykey's config like private key or passphrase or
>location. We could put it in the main `polykey.ts` file perhaps?
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/40#discussion_r438503008

-- 

src/PeerStore/PeerStore.ts
robert-cronin: I have taken a look at them as yet, but yeah I think PeerStore could be renamed to something like `PubKeyStore` or `js-pubkeystore` and made as a separate library. Our implementation is more specific than libp2p's `PeerStore` so this makes sense.

tests/RPC/RPCMessage.test.ts
robert-cronin: yeah definitely, I think the amount of testing required for this library requires a separate PR? I  can add it after these two are merged, I will add more tests now too :+1:

src/RPC/PeerInfoMessage.proto
robert-cronin: yeah I hadn't got around to making use of gRPC, the use of `protobufjs` was more as a `.proto` file loader as `grpc-js` doesn't have that functionality. They recommend the `@grpc/proto-loader` library so I will switch to that :+1:

src/P2P/PeerDiscovery.ts
robert-cronin: Sure, I think we could provide and interface for this, so something like:
```
interface SocialDiscovery {
    findUser(handle, service): PubKey | null { ... }
}
```
Then the end user can create as many `SocialDiscovery` objects they want and perhaps pass them into `Polykey` class constructor.

src/P2P/MulticastBroadcaster.ts
robert-cronin: I have made all return types explicit :+1:

src/RPC/PeerInfoMessage.proto
robert-cronin: It turns our `@grpc/proto-loader` uses `protobufjs` under the hood. I couldn't find any way to just purely encode an object into a Buffer in `gRPC`, unless I've missed something. From what I understand, `grpc` is not designed for byo transport/byo socket and it doesn't look like it supports UDP or broadcasting to multiple clients as in Multicast Service Discovery.

src/Polykey.ts
robert-cronin: Well there are some things that we can remove from the `Polykey` class, the vault operations for example (e.g. secret opperations

robert-cronin: Well there are some things that we can remove from the `Polykey` class, the vault operations for example (e.g. secret operations). It makes sense that the `Polykey` class should only be responsible for the highest order operations like creating/destroying vaults. I'm not sure how we would separate out peer discovery though, perhaps we could make it a separate class like the `KeyManager`?

src/P2P/PeerDiscovery.ts
robert-cronin: This is now implemented as an extra parameter in the constructor with a default of just the keybase social discovery API.

src/KeyManager.ts
robert-cronin: I'm not sure web workers would help us with the initial key generation (in polykey initialization) as its really just a one off. I never had an issue with speed on signing and verifying files but we can put it in there :+1: it will probably come in handy later for huge networks as they will be relying on encrypted communication with the users pubkey.

src/P2P/MulticastBroadcaster.ts
robert-cronin: These have been made environment variables with defaults set to what they were before.

src/P2P/MulticastBroadcaster.ts
robert-cronin: I agree, I have renamed it to `handleHandshakeMessages`, this could probably be more specific as well, but I think 'handshake' in the context of polykey will probably mean the same thing if it's multicast peer discovery or some other way to handshake (i.e. they will probably need the same fields for the handshake)

bin/keymanager/KeyManager.ts
robert-cronin: The cli source files have been moved to `src` directory so this is no longer an issue.
The polykey library is contained within `src/lib/` and has a path of `@polykey/...`.
The polykey cli is contained within `src/cli/` and has a path of `@polykey-cli/...`

bin/initPolykey.ts
robert-cronin: There is now only one file in the `bin` directory, everything else is in the `src` folder.

bin/initPolykey.ts
robert-cronin: yeah since this is now only as a warning in the CLI, and not actually in the polykey library itself, we can redo it in the polykey GUI

src/PeerStore/PeerStore.ts
robert-cronin: As discussed elsewhere, this will stay inside the polykey library repo

src/Polykey.ts
robert-cronin: I have removed vault operations from Polykey to reduce its domain model scope. Now one has to `getVault` and then can do things like `vault.addSecret(...)`

src/P2P/MulticastBroadcaster.ts
robert-cronin: I'm not sure how we would do that, perhaps we can discuss that at our next meeting?

CMCDragonkai:
* https://blog.acolyer.org/2016/02/16/capability-myths-demolished/
* https://srl.cs.jhu.edu/pubs/SRL2003-02.pdf
* https://en.wikipedia.org/wiki/Proof_of_work

src/KeyManager.ts
robert-cronin: A key manager worker has been added in the same way we added one to EFS, it is passed in to the KeyManager class at initialisation via a worker pool (threadsjs)

robert-cronin: Note this only includes sign/verify and encrypt/decrypt actions for now.

robert-cronin:
So from our previous meeting I have removed a lot of methods from the Polykey class and put them in their respective classes according to which domain model they belong to.

src/P2P/MulticastBroadcaster.ts
robert-cronin: I have removed the stop function and moved the start functionality into the initialisation of the MulticastBroadcaster class.

CMCDragonkai:
Can you give a brief description of what was done so I know what was applied on a high level from our discussion.

webpack.config.js
```diff
@@ -37,9 +86,24 @@ module.exports = {
...
+    new CopyPlugin({
+      patterns: [
+        {
+          from: 'src/**/*.proto',
+          to: '',
+          flatten: true
+        },
+      ]
+    }),
+    new NpmDtsPlugin({
```
CMCDragonkai: What's this for? And if this is what I think it is for, why wasn't this used in js-encryptedfs?

webpack.config.js
```diff
@@ -14,14 +65,12 @@ module.exports = {
...
-      '@': path.resolve('src')
-    }
+    extensions: ['.ts', '.js', '.proto'],
+    plugins: [new TsconfigPathsPlugin()]
   },
-  node: {
-    fs: 'empty'
+  target:'node',
+  externals: {
+    fs: "commonjs fs"
```
CMCDragonkai, line 72: What does `fs: "commonjs fs"` mean?

robert-cronin: This tells webpack that `fs` is a library that is not resolved by webpack, but are dependencies of the output (i.e. dependencies of `dist/polykey.js`). The use of `commonjs` here just tells webpack what the module context is.

This probably needs some extra care to make it isomorphic like `EncryptedFS` is. For Electron, this is not an issue with the following code:
```
app.on('ready', () => {
    mainWindow = new BrowserWindow({
        webPreferences: {
            nodeIntegration: true, // <--- flag
        }
    });
});
```

webpack.config.js
robert-cronin: This was for creating a single definition file `index.d.ts` to simplify things but I don't think it really works, I have taken it out.

robert-cronin:
> Can you give a brief description of what was done so I know what was applied on a high level from our discussion.

So the main changes are:
- Move domain model concerns to relevant classes (i.e. vault operations to `Vault.ts`, crypto operations to `KeyManager.ts` etc.)
- Added a web worker for the crypto operations (`KeyManagerWorker.ts`)
- Tied the starting of the multicast peer discovery server to its initialisation
- Made UDP multicast port optional environment variables
- Added ability to add more social discovery services to `PeerDiscovery` class

Note: I think it's probably better if I do WIP commits with good descriptions from now on and leave the squashing to the end for more clarity for the reviewer.

webpack.config.js
robert-cronin: This was also an attempt at trying to generate normal typings with the tsconfig paths [issue](https://github.com/MatrixAI/Typescript-Demo-Lib/issues/1) which made it much harder since we have a CLI and a LIB in one repo. I still would rather just forget about this as it's been a huge effort to try and integrate it into our project for what seems like minimal benefit.

robert-cronin: now if I put `"declaration": true` into `tsconfig.json`, it generates typings for the test folder as well, since I've had to 'include' the test folder in order to get `@` imports working from the test folder.
https://github.com/MatrixAI/Typescript-Demo-Lib/issues/1#issuecomment-643945531

CMCDragonkai: Hmm the CLI stuff shouldn't generate any kind of typings since it's not meant to be used as a library. It shouldn't even be importable at all.

Perhaps the CLI portion of the code should be moved out into its own directory? That is outside of the `src` directory completely.

But yea if you cannot use `@polykey` in the CLI code src, then so be it.

CMCDragonkai:
I'd like to see more integration tests in the near future for this kind of code especially as it interacts with the network. But also all the edgecases, we'll have to plan that out.

I'm going to approve for merging so you can start synchronizing the other work.

CMCDragonkai (approved):

CMCDragonkai:
Woops I was trying the Github merge options. But I think it just left the WIP commits there. Not a big deal, but need to remember to use the squash option, not rebase option.

CMCDragonkai:
I've rebased and squashed so the master is clean again.

CMCDragonkai:
Delete the above branch if you don't need it again.

PR #43, opened 2020-06-05
Add Git Backend

### LINKED ISSUE #8: Vault sharing via Git
We want to harness git's abilities to push and pull repositories (i.e. secret bundles). We will only want to implement a pull based model for now. But want to ensure the the keynode pull the repository has is authenticated to access that secret bundle. This can be implementing using PGP. The node that owns the keynode can encrypt the bundle with requesting node's pub key and the node pulling the bundle will eventually get access with the corresponding private key. Bundles should also be tagged with keys. For example a key tag should list all the bundles currently shared with that node.

### LINKED ISSUE #16: Automatic Commits for Vault Changes
Changes to any vault should involve an automatic commit to our Git version control system.

We should try to make this system write sensible messages. And we need to carefully manage the Git commit history and consider issues like diverging histories and merge strategies dealing with incompatible histories. We need to make this as automatic as possible so that end-users do not have to do a manual merge like they would for source control.

Issue regarding redundant commit messages: https://github.com/MatrixAI/js-polykey/issues/5 - we may need to be able to transactionally modify things. So edits to multiple files should be 1 commit if they represent the same change in secrets.

Initial experiment here: https://github.com/MatrixAI/js-polykey/commit/aa8d4aeac26477a71c8cdd15bc420c06ee9ad51f

### LINKED ISSUE #22: Encryption of vault keys
Each Vault will have it's own individual symmetric key for aes encryption. This needs to be secret. We also need to determine where to store the encrypted vault key so it can be retrieved easily when encrypting and decryption secrets.

To tackle the first problem, the vault key can be encrypted with the Keynode's public key so only it can only be decrypted with it's passphrase protected private key. Another option is to simply have a password derived symmetric key which will encrypt the vault key. This key need not be stored anywhere it can be recreated each time from the user's password and the salt which can be persisted upon creation of the keynode. 

As to where the vault keys will be stored. Each Vault can have a hidden '.vault' folder at the root which can contain the encrypted vault key as well as other metadata.

### LINKED ISSUE #31: Git backed server for sharing vaults
TBD.

Should be lower level pull/push style.

However we can also have a higher level sync. But that can also be automated via the polykey agent.

Relevant to MatrixAI/PolyKey#31

### LINKED ISSUE #32: Peer to Peer Architecture
Inspirations:

* https://github.com/sit-fyi/sit
* Dat P2P
* Libp2p (js libraries available)
* Kademlia

The peer to peer architecture is the architecture in which Polykey Keynodes can discover each other and connect to each other. Note that communication between Polykey keynodes happens over this p2p network. But social discovery occurs via social providers like Keybase which facilitate discovery of human profiles to automated keynodes.

robert-cronin:
This PR will add a way to serve the vaults over http as git repos by exposing them via a git upload pack (stitched together from `isomorphic-git/http-receiver`). All it needs is an address (`ip` and `port`) to listen on (which is auto-generated).  The other side is pulling vaults, this also needs an address (`ip` and `port`) which is provided by the peer discovery mechanism being implemented concurrently as a part of PR #40 

Vault store is also implemented to keep track of which pubKeys have access to which vaults. This is also a convenience to pass into `GitServer` to retrieve the specific vault that git needs in order to use it's EFS instance to serve the vault from it's `upperDir`.

Fixes #31 
This also in part solves #16, in terms of automatic commits. However, there is still the issue of sensible commit messages when committing more than 1 file (i.e. changing more than one secret at a time).

Fixes #22
Fixes #16 
Fixes #32


CMCDragonkai:
To be precise this enables "pull" behaviour. But not "push" behaviour. Eventually we need to apply capability theory to the connections between different Polykeys and vaults. That is what PK agents are allowed to pull from my vault, and what PK agents are allowed to push to my vault.

However in doing so, we then have to deal with merge conflicts. There are ways to resolve this automatically however.

CMCDragonkai:
> Vault store is also implemented to keep track of which pubKeys have access to which vaults. This is also a convenience to pass into GitServer to retrieve the specific vault that git needs in order to use it's EFS instance to serve the vault from it's upperDir.

I originally thought that the above could be implemented through hardlinks or something else. A pub key may relate to multiple vaults and indicate some sort of relationship. These relationships are conceptualized as "capabilities".

CMCDragonkai:
Automatic commit messages are for convenience, we don't expect users of Polykey to understand Git. So we have to "specialize" the behaviour of git for secrets management. I'm sure they will appreciate it.

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+import url from 'url'
+import Path from 'path'
+import http from 'http'
+import { promisify } from "util";
+import { AddressInfo } from "net";
+import { Readable } from "stream";
+import { parse } from 'querystring'
+import { HttpDuplex } from "./HttpDuplex";
+import uploadPack from "./upload-pack/UploadPack";
+import VaultStore from "@polykey/VaultStore/VaultStore";
```
CMCDragonkai: If VaultStore is for storing relationships between public keys and vaults, then this is a strange name for it. I'd prefer that we talk little about capability theory.

CMCDragonkai:
I'm thinking everything under `GitServer/` should just be under `git/`. I suspect there's a lot more to managing the git stuff than just the serving of git. I suppose if we maintain directory names to be capitalized, then `Git/` is fine too. But I've actually stopped doing that for my other projects. I know I did that for js-virtualfs, but we can leave that as so. So lower case directory names is fine. Upper case file names for actual class files. Lower case file names for non-classes. Also this depends on what the default export is. For some files, the default export is a class, but it has ancillary  exports, and upper case name is fine here. In other cases, a file exports lots of little things, or the default export is a function or an object/dictionary/array. Then lowercase name.

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+
+// Here is the protocol git outlines for sending pack files over http:
+// https://git-scm.com/docs/pack-protocol/2.17.0
+// https://github.com/git/git/blob/master/Documentation/technical/pack-protocol.txt
```
CMCDragonkai: General question, if the pack file format has version numbers. Or does it just evolve as git evolves?

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+// This should be consulted in developing our upload pack implementation
+
+// This git backend (as well as HttpDuplex class) is heavily inspired by node-git-server:
+// https://github.com/gabrielcsapo/node-git-server
+
+const services = ['upload-pack', 'receive-pack']
+
+class GitServer {
+  repoDir: string;
```
CMCDragonkai: Are we using `private` here?

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+  ) {
+    this.repoDir = repoDir
+    this.vaultStore = vaultStore
+  }
+
+  /**
+   * Find out whether `repoName` exists.
+   */
+  exists(repo: string) {
+    // TODO: consider if vault has been shared
```
CMCDragonkai: What does this comment mean? What relationship does the existence of a repo has with whether a vault is shared or not?

Or one should ensure that vaults that are not "shared" should not be queryable for existence.

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+      case 'receive-pack':
+        return 'push'
+      default:
+        return 'unknown'
+    }
+  }
+
+
+  // alternatively you can also pass authenticate a promise
+  authenticate(
```
CMCDragonkai: Is this HTTP basic authentication?

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+    }
+    return false
+  }
+
+  /**
+   * Handle incoming HTTP requests with a connect-style middleware
+   */
+  handle(req: http.IncomingMessage, res: http.ServerResponse) {
+
+    const infoHandler = (req: http.IncomingMessage, res: http.ServerResponse): boolean => {
```
CMCDragonkai: Perhaps these handlers should be abstracted in the MVC style like loja-api? Not sure if you copied this structure from node-git-server?

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+        res.end('service not available')
+        return true
+      }
+
+      const repoName = this.parseGitName(m[1])
+
+      // check if the repo is authenticated
+      const type = this.getType(service)
+      const headers = req.headers
+      const { username, password } = {username: 'foo', password: 's'}
```
CMCDragonkai: These are definitely stubs right?

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+        res.end('service not available')
+        return true
+      }
+
+      res.setHeader('content-type', 'application/x-git-' + service + '-result')
+      this.noCache(res)
+
+      const repoDir = Path.join(this.repoDir, repo)
+
+      const fileSystem = this.vaultStore.getVault(repo)?.efs
```
CMCDragonkai: What happens if the vault store doesn't have that repo? You used the `?` here to ignore the null. But where is it confirmed that such a repo indeed exists prior?

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  /**
+   * starts a git server on the given port
+   */
+  listen(port?: number): AddressInfo {
+    this.server = http.createServer((req, res) => {
+      this.handle(req, res)
+    })
+
+    this.server.listen(port)
```
CMCDragonkai: If the port is optional , does this mean this finds a random port? I'd prefer for this to be explicit with `port: number = 0`.

src/GitServer/GitServer.ts
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  /**
+   * stops the server instance
+   */
+  async stop() {
+    try {
+      await promisify(this.server.close)()
+    } catch (error) {
+      throw(error)
```
CMCDragonkai: Why catch only to throw here?

src/GitServer/HttpDuplex.ts
```diff
@@ -0,0 +1,178 @@
...
+import { EncryptedFS } from 'encryptedfs';
+import GitSideBand from "./side-band/GitSideBand";
+import packObjects from './pack-objects/PackObjects';
+
+const headerRE = {
+  'receive-pack': '([0-9a-fA-F]+) ([0-9a-fA-F]+) refs\/(heads|tags)\/(.*?)( |00|\u0000)|^(0000)$', // eslint-disable-line
+  'upload-pack': '^\\S+ ([0-9a-fA-F]+)'
+};
+
+export class HttpDuplex {
```
CMCDragonkai: What is the purpose of this class?

src/GitServer/pack-objects/BufferCursor.ts
```diff
@@ -0,0 +1,64 @@
+// Modeled after https://github.com/tjfontaine/node-buffercursor
+// but with the goal of being much lighter weight.
+export class BufferCursor {
```
CMCDragonkai: Is this really needed? Seems like a simpler abstraction is possible.

CMCDragonkai: Also as a general style, define your class by itself, and run `export default...` at the end.

src/GitServer/pack-objects/GitCommit.ts
```diff
@@ -0,0 +1,215 @@
...
+function outdent (str) {
+  return str
+    .split('\n')
+    .map(x => x.replace(/^ /, ''))
+    .join('\n')
+}
+
+// TODO: Make all functions have static async signature?
+
+class GitCommit {
```
CMCDragonkai: Is this something that can be a separate library? It seems like something that isomorphic-git should be doing.

src/GitServer/pack-objects/GitObjectManager.ts
```diff
@@ -0,0 +1,81 @@
...
+import path from 'path'
+import fs from 'fs'
+
+import GitObject from './GitObject'
+import { GitPackIndex } from './GitPackIndex'
+import { EncryptedFS } from 'encryptedfs'
+
+const PackfileCache = new Map()
+
+class GitObjectManager {
```
CMCDragonkai: With this amount of complexity, I'm hoping there should be tests for all of this.

src/Polykey.ts
```diff
@@ -77,7 +79,17 @@ export default class Polykey {
     let vault: Vault
     try {
       vault = await this.getVault(vaultName)
-      vault.addSecret(secretName, secret)
+      await vault.addSecret(secretName, secret)
+    } catch(err) {
+      throw err
```
CMCDragonkai: Why catch just to throw?

src/Polykey.ts
```diff
@@ -418,7 +399,7 @@ export default class Polykey {
           try {
             const vaultKey = Buffer.from(this.metadata.vaults[vaultName].key)
             const vault = new Vault(vaultName, vaultKey, this.polykeyPath)
-            this.vaults.set(vaultName, vault)
+            this.vaultStore.setVault(vaultName, vault)
```
CMCDragonkai: I can see that `Polykey` is one of the primary classes in this library. However its domain modelling is too broad. `Polykey` as a class shouldn't be dealing with secrets directly. That should be in a subdomain of  vaults.

Instead `Polykey` should be managing vaults, and capabilities, and peers. Each of which has their own subdomain. We can discuss this more later.

src/VaultStore/VaultStore.ts
```diff
@@ -0,0 +1,52 @@
+import Vault from "@polykey/VaultStore/Vault";
+
+class VaultStore {
```
CMCDragonkai: This should be better represented with capability theory.

CMCDragonkai (changes requested):
A quick review of this PR. I have a feeling that the git portion of it has a lot of complexity, and it would be better to have that in a separate library, or at least extensively tested. It also has a bunch of code smells. And I'm not sure if alot of that code was just derived from some upstream library.

CMCDragonkai:
Needs rebase on top of master.

And most importantly lots more tests for this git stuff.

src/VaultStore/VaultStore.ts
CMCDragonkai: Might not need to be anymore...

src/Polykey.ts
CMCDragonkai: Solved in PR 40.

src/GitServer/GitServer.ts
robert-cronin: yeah this was originally just implicitly null, but we can make it explicitly random with port of `0`

src/GitServer/GitServer.ts
robert-cronin: this was redundant and has been removed

src/GitServer/pack-objects/GitCommit.ts
robert-cronin: Ideally it would be something that isogit is doing, and there is a [branch](https://github.com/isomorphic-git/isomorphic-git/tree/http-receiver) and [issue](https://github.com/isomorphic-git/isomorphic-git/issues/1071) for it, but it looks like progress has stalled recently.

src/GitServer/pack-objects/GitObjectManager.ts
robert-cronin: I will be reducing complexity to cater to our specific needs as well as including some tests as a part of this PR

src/Polykey.ts
robert-cronin: This has been taken out :+1: and the whole method moved to the `Vault` class

src/VaultStore/VaultStore.ts
robert-cronin: Well we could remove the vault store and just store which pubkeys have access to the vault in the vault class itself.

This is also related to a thread above:
> If VaultStore is for storing relationships between public keys and vaults, then this is a strange name for it. I'd prefer that we talk little about capability theory.

src/GitServer/GitServer.ts
robert-cronin: yes all of these can be private

src/GitServer/GitServer.ts
robert-cronin: yeah these were just stubs, I am yet to dive deep into mutual TLS authentication. I will begin looking into PKIjs for a self signed certificate. I think its good if we can add it as a part of this PR. Then later if we want to give users the ability to sign with their own internal CA, then we can do that.

CMCDragonkai:
Surely we don't have roll our own tls. This us very common.

On 16 June 2020 17:23:44 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@robert-cronin commented on this pull request.
>
>
>
>> +        return true
>+      }
>+
>+      const service = (<string>params.service).replace(/^git-/, '')
>+      if (services.indexOf(service) < 0) {
>+        res.statusCode = 405
>+        res.end('service not available')
>+        return true
>+      }
>+
>+      const repoName = this.parseGitName(m[1])
>+
>+      // check if the repo is authenticated
>+      const type = this.getType(service)
>+      const headers = req.headers
>+      const { username, password } = {username: 'foo', password: 's'}
>
>yeah these were just stubs, I am yet to dive deep into mutual TLS
>authentication. I will begin looking into PKIjs for a self signed
>certificate. I think its good if we can add it as a part of this PR.
>Then later if we want to give users the ability to sign with their own
>internal CA, then we can do that.
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/43#discussion_r440638733

-- 

robert-cronin:
> Surely we don't have roll our own tls. This us very common.

Not sure what you mean by rolling our own TLS, we can use whatever HTTPS library is provided by the environment polykey is deployed in and the server `certificate.pem` and `key.pem` we generate.

I've also found PKI.js is very complicated but I've found something much easier in `node-forge` it comes with the ability to use just pure JavaScript or node if we're in that environment. Perhaps we can use this instead? PKIjs examples can extend to 900 lines of code to generate a certificate.

If you're talking about the self-signing part, this isn't intended on being a long term solution, I was planning on looking into the Web of trust concepts for peers acting as CAs.

Another thing is, I'm not entirely sure yet how to incorporate keybase keys into this whole thing. They are PGP blocks so technically we can extract the RSA keys out of them to use in the X.509 certificates but I don't see any way to do this in JavaScript as yet. Currently I am generating a new RSA keypair for the X.509 certificates.

CMCDragonkai:
PKI is an additional feature that is rather complicated, I want to have a discussion about this before going ahead. Any web of trust should right now be focusing on PGP only.

src/GitServer/GitServer.ts
robert-cronin: yeah this structure is directly from `node-git-sever`. I will try to simplify this along with the HttpDuplex class

src/GitServer/GitServer.ts
robert-cronin: As discussed, PKI can come later but for now I was looking into having a TLS connection with a self-signed certificate and I have created a new `ConnectionManager` class for this. The idea being we have one connection and multiplex the git http stuff over that. I have had some trouble trying to get http working over an existing socket (i.e. a tls socket) but I haven't found an existing way to do it. I suspect we probably want to build our own HTTP parser or use an existing one. I will create an issue for this.

src/GitServer/GitServer.ts
robert-cronin: It definitely has version numbers: https://git-scm.com/docs/pack-format

It looks like the pack file header contains the 4-byte version number

src/GitServer/GitServer.ts
robert-cronin: Here is the issue: https://github.com/MatrixAI/js-polykey/issues/46
I've taken the authentication stubs out for now because we will be doing authentication via mutual TLS in the future

src/GitServer/GitServer.ts
robert-cronin: yes, this was intended to act as both authentication and authorisation (who are you via https and what are you able to do via the vault) but I've now taken this out as we will have an existing (authenticated) TLS connection and authorisation will be handled by the vault depending on which public key is presented.

src/GitServer/GitServer.ts
robert-cronin: yeah this was reserved until we had a discussion about authorisation/capability theory. I think from our discussions, this should be implemented by consulting the corresponding vault if a particular peer is able to access the vault. Once the git server is using the authenticated TLS connection between two peers, we can just worry about authorisation.

src/GitServer/GitServer.ts
robert-cronin: I've decided that vaults should determine who/what is able to access their methods. In our last discussion, we spoke about domain modelling and capability theory, but there were some points about the security concern of passing around cryptographic tokens for capability access (i.e. we might want to limit transitive access in lieu of direct, identity based access)

For now the ability of the vault store to determine which pubkey has access to which vault is replaced by a map in each individual vault.

src/GitServer/GitServer.ts
robert-cronin: I've added new code to the exists method, it should check 2 things;
1) does the vault exist? (this includes if it exists on file as well as if polykey has it in its stores, both need to be true); and
2) is the peer authorised to know it exists?

The last one is yet to be implemented and depends on getting the secure TLS connection right so we know who is connecting (client authentication)

src/GitServer/GitServer.ts
robert-cronin: Eventually, this will be listening on a pre-made secure connection and we won't need to provide a port.

src/GitServer/GitServer.ts
robert-cronin: VaultStore has now been removed

src/GitServer/HttpDuplex.ts
robert-cronin: This is to Handle the response when the client asks for a particular pack, to handle `POST` requests in other words. We need to be able to handle incoming and outgoing messages at the same time. HttpDuplex class accumulates the request data (git want messages) and implements a stream that handles that data sequentially (i.e. writing the packed objects into the response stream)

src/GitServer/pack-objects/BufferCursor.ts
robert-cronin: > Is this really needed? Seems like a simpler abstraction is possible.

Yes this should be possible with the [feross/buffer](https://www.npmjs.com/package/buffer) and handcoding the methods this `BufferCursor` class uses.

> Also as a general style, define your class by itself, and run `export default...` at the end.

Oh yes, I forgot this for the imported code. This has been done for all classes :+1:

robert-cronin: Upon further inspection, this `BufferCursor` class provides some really helpful methods and I think if we remove it, we would make the `GitPackIndex` class much more complicated. I have left it in but removed a lot of unused methods.

src/GitServer/HttpDuplex.ts
robert-cronin: I've been able to reduce complexity enough to embed the whole of git server in a single `request.on('data', (data) => {})` method within the GitServer itself. So the `HttpDuplex class is now redundant and has been removed.

src/GitServer/pack-objects/GitObjectManager.ts
robert-cronin: I've been able to reduce complexity a little within the imported code but there is still a lot there just due to the fact that we are handling the git backend on the lowest level (directly on the git protocol level with network packets). It would be great if there was a library that could handle all this but to the best of my knowledge, a git backend in pure JavaScript has never been done in production level code.

robert-cronin: I am adding several tests on the macro level (vaults pulling from peer vaults) and I think this might be enough for now just to have initial confidence that it will work the way we intend, but more comprehensive testing is definitely needed for the git backend code (`packObjects`, `GitSideBand` and `uploadPack`). I plan on diving deep into all of that in future so we can really make that code our own and not just a copy from isomorphic-git.

src/GitServer/GitServer.ts
robert-cronin: I'm not sure what you mean by the MVC style from loja-api, these are just for handling different GET or POST requests, there aren't really any `Models` as far as the MVC architecture defines. I think I can understand what you're thinking, perhaps we can have a single handler that determines what type of action is needed (the controller) and then have different methods defined for handling the IncomingMessage (the models). This would simplify things a bit :+1:

robert-cronin: This has now been simplified and I was able to remove quite a bit of code. I think its a little easier to read now :+1:

package.json
```diff
@@ -40,23 +40,34 @@
...
     "futoin-hkdf": "^1.3.2",
+    "git-list-pack": "0.0.10",
+    "http-parser-js": "^0.5.1",
     "inquirer": "^7.1.0",
     "inquirer-fuzzy-path": "^2.3.0",
-    "isomorphic-git": "^0.35.1",
+    "isomorphic-git": "^1.5.0",
     "jsonfile": "^6.0.1",
     "kbpgp": "^2.0.82",
+    "marky": "^1.2.1",
```
CMCDragonkai: This looks like a dev dependency?

package.json
```diff
@@ -40,23 +40,34 @@
     "cli-progress": "^3.8.2",
     "commander": "^5.1.0",
     "configstore": "^5.0.1",
+    "crc": "^3.8.0",
     "encryptedfs": "^0.1.0",
     "fs-extra": "^9.0.0",
     "futoin-hkdf": "^1.3.2",
+    "git-list-pack": "0.0.10",
```
CMCDragonkai: Can we synthesize this code instead of relying on such an old third party dependency.

package.json
```diff
@@ -40,23 +40,34 @@
...
     "inquirer-fuzzy-path": "^2.3.0",
-    "isomorphic-git": "^0.35.1",
+    "isomorphic-git": "^1.5.0",
     "jsonfile": "^6.0.1",
     "kbpgp": "^2.0.82",
+    "marky": "^1.2.1",
     "moniker": "^0.1.2",
     "node-fetch": "^2.6.0",
+    "node-forge": "^0.9.1",
+    "pad": "^3.2.0",
```
CMCDragonkai: We should not be using libraries like these. These utility functions should be embedded in Polykey.

package.json
```diff
@@ -40,23 +40,34 @@
...
+    "simple-get": "^4.0.0",
```
CMCDragonkai: Can we just use the native fetch API without these abstractions?

package.json
```diff
@@ -40,23 +40,34 @@
...
+    "split-buffer": "^1.0.0",
```
CMCDragonkai: Another library that should be an embedded utility function.

package.json
```diff
@@ -40,23 +40,34 @@
...
+    "stream-source": "^0.3.5",
```
CMCDragonkai: Is this really needed? I imagine native streams API is fine. We don't always need to use every kind of library. Sometimes a little bit of verbosity is fine. And there's always `readable-streams` as used in VFS.

package.json
```diff
@@ -40,23 +40,34 @@
...
     "threads": "^1.6.0",
+    "url": "^0.11.0",
```
CMCDragonkai: Just to be clear, is this because we want to use Node URL but in other platforms? Maybe in that case, we should just have our own utility function.

src/lib/KeyManager.ts
```diff
@@ -308,8 +308,8 @@ class KeyManager {
   }
 
   // Verify data
-  verifyData(data: Buffer | string, signature: Buffer, withKey?: Buffer): Promise<string> {
-    return new Promise<string>(async (resolve, reject) => {
+  verifyData(data: Buffer | string, signature: Buffer, withKey?: Buffer): Promise<boolean> {
+    return new Promise<boolean>(async (resolve, reject) => {
```
CMCDragonkai: Why are we not using async await here?

src/lib/KeyManager.ts
```diff
@@ -342,16 +342,21 @@ class KeyManager {
...
-            resolve(km.get_pgp_fingerprint().toString('hex'));
+            if (km.get_pgp_fingerprint()) {
+              resolve(true)
+            } else {
+              resolve(false)
+            }
+            resolve(km.get_pgp_fingerprint());
           } else {
-            reject(Error('could not verify file'))
+            resolve(false)
```
CMCDragonkai: Same here. I thought with async await, we would't need to use CPS.

src/lib/Polykey.ts
```diff
@@ -24,17 +28,17 @@ type Metadata = {
...
   polykeyPath: string
   private fs: typeof fs
-  private vaults:Map<string, Vault>
   private metadata: Metadata
   private metadataPath: string
 
   keyManager: KeyManager
-  peerStore: PeerStore
-  peerDiscovery: PeerDiscovery
+  vaults: Map<string, Vault>
```
CMCDragonkai: Looks like vaults should have a "manager" type as well?

src/lib/Vault.ts
```diff
@@ -1,30 +1,48 @@
...
+  sharedPubKeys: string[]
+}
 
 class Vault {
 
   private key: Buffer
   private keyLen: number
   name: string
-  private fs: EncryptedFS
+  efs: EncryptedFS
```
CMCDragonkai: Are these meant to be private properties?

src/lib/git/upload-pack/GitPktLine.ts
```diff
@@ -0,0 +1,96 @@
...
+
+function padHex(b, n) {
+  const s = n.toString(16)
+  return '0'.repeat(b - s.length) + s
+}
+
+// I'm really using this more as a namespace.
+// There's not a lot of "state" in a pkt-line
+
+export class GitPktLine {
```
CMCDragonkai: Can we ensure that we use `export default` or `export { ... }` pattern for consistency.

src/lib/pki/PublicKeyInfrastructure.ts
```diff
@@ -0,0 +1,84 @@
+import forge from 'node-forge'
+
+function createX509Certificate() {
+  const pki = forge.pki;
+
+  // generate a keypair and create an X.509v3 certificate
+  const keys = pki.rsa.generateKeyPair(2048);
```
CMCDragonkai: Is it wise to hardcode these things? I'm not sure but I want to have a meeting to discuss PKI usage. I want to know what you're thinking regards PKI. You can writie it into the PR comments.

CMCDragonkai (changes requested):
Some preliminary review, mostly focusing on the overabundance of dependencies that shouldn't really be there. For a strong crypto app, the less dependencies the better. It becomes easier to audit, and make sure we are not bringing any unknown vulnerabilities.

package.json
robert-cronin: I originally thought this was for providing information to the progress hook on the client side but in any case, I've removed it, if we want progress shown on the pulling side, we could probably come up with a better way to do progress

robert-cronin: I've taken it out for now

package.json
robert-cronin: So it turns out that the git backend works just fine without this so I've taken it out.

package.json
robert-cronin: Done. This was quite simple to replace in code with string addition.

package.json
robert-cronin: I've taken this out in favour of the node fetch module :+1:

package.json
robert-cronin: This has been embedded in GitSideBand where it belongs

package.json
robert-cronin: I have removed this in favour of the native streams API.

package.json
robert-cronin: This has been replaced by a simple string split and regex for checking the url path.

src/lib/KeyManager.ts
robert-cronin: we can't do the standard async await here because kbpgp is callback based. I thought it was redundant to add the async modifier since we are returning an explicit promise but I've added it in anyway.

src/lib/KeyManager.ts
robert-cronin: Yes, well technically, we don't actually use CPS here. The resolve method is just for turning the kbpgp method into a promise based call, its not a callback you can pass into the `verifyData` method.

src/lib/Polykey.ts
robert-cronin: Yeah we can have a vault manager, makes sense from a domain modelling perspective. That's kind of what I was trying to achieve with the `VaultStore` but I think I misinterpreted the need to rename it with a proper name with it being redundant so I took it out completely. Also it's probably more consistent to have a vault manager instead of just a map of vaults :+1:

src/lib/Vault.ts
robert-cronin: this should be accessible in some way for the `GitServer` but they definitely shouldn't be settable. I have changed this to private with a public getter :+1:

src/lib/git/upload-pack/GitPktLine.ts
robert-cronin: This was a redundant export and has been removed.

src/lib/pki/PublicKeyInfrastructure.ts
robert-cronin: not sure how far you wanted me to go, but I've added some defaults to the cert generation method.

src/lib/Polykey.ts
robert-cronin: This has been refactored into a new `VaultManager` class that takes care of everything under the `Vault` domain model.

robert-cronin:
This PR is ready for review. Among the changes are a reduction of external dependencies, creating local methods for any minimally used packages and even removing some that weren't even used. Additionally, the domain modelling of the whole lib has been strengthened with the creation of a manager for each domain that polykey oversees. The polykey class now is much more like a convenience class that sets everything up for you than anything that manages `Vaults`/`Keys`/`Peers`. The latter have been delegated to their respective `Manager` classes.

metadata
```diff
@@ -0,0 +1 @@
+[["1",{"type":"Buffer","data":[49,32,97,104,97,104,97,104]}],["2",{"type":"Buffer","data":[50,32,97,104,97,104,97,104]}]]
```
CMCDragonkai: What is this `metadata` file left in the project root? You should use `./tmp` for this stuff or add some stuff into `.gitignore`.

src/cli/initPolykey.ts
```diff
@@ -5,7 +5,7 @@ import inquirer from 'inquirer'
 import Configstore from 'configstore'
 import PolyKey from '../lib/Polykey'
 import cliProgress from 'cli-progress'
-import KeyManager from '@polykey/KeyManager'
+import KeyManager from '@polykey/keys/KeyManager'
 import { actionRunner, pkLogger, PKMessageType } from './polykey'
```
CMCDragonkai, line 10: In `package.json`, there's a `bin/pk.js` used as the actual command. If you see in the JS demo repo, we should actually be using an alias:

```
  "bin": {
    "javascript-demo": "dist/bin/javascript-demo"
  }
```

So the command is therefore `javascript-demo`. And multiple commands can exist, just by adding another key value pair.

Furthermore, the utilisation of `#!/usr/bin/env node` would ensure that the script doesn't require a `.js` extension.

Therefore a question is, this `cli` part of the source is being required/imported by the CLI script right? Or perhaps the `bin/*` is being generated by the webpack. I'm not sure cause I didn't see any documentation as to how the CLI entrypoint is actually generated/produced. This behaviour should be written into the README.md and replicated in the comments of this PR.

Finally, it's a good idea to use the `index.ts` or `index.js` as the main coordinator/entrypoint for any directory (where a directory can represent a module in JS projects).

src/lib/Polykey.ts
```diff
@@ -1,232 +1,39 @@
...
+import VaultManager from '@polykey/vaults/VaultManager'
 
 class Polykey {
   polykeyPath: string
-  private fs: typeof fs
-  private vaults:Map<string, Vault>
-  private metadata: Metadata
-  private metadataPath: string
 
+  vaultManager: VaultManager
```
CMCDragonkai, line 11: Did you mean to remove the `private` scoping?

Remember due to JS binding, if any user were to use your properties outside of the object context, they will lose the this binding.

Which is why I use arrow methods in the loja-api, but only because I know those methods will be referred to without the object context.

src/lib/git/GitServer.ts
```diff
@@ -0,0 +1,281 @@
+import Path from 'path'
+import http from 'http'
+import through from 'through';
```
CMCDragonkai: Do we need this? I thought `readable-streams` is enough? It's used in js-virtualfs.

src/lib/git/GitServer.ts
```diff
@@ -0,0 +1,281 @@
...
+// This should be consulted in developing our upload pack implementation
+
+// This git backend (as well as HttpDuplex class) is heavily inspired by node-git-server:
+// https://github.com/gabrielcsapo/node-git-server
+
+// We need someway to notify other agents about what vaults we have based on some type of authorisation because they don't explicitly know about them
+
+const services = ['upload-pack', 'receive-pack']
+
+class GitServer {
```
CMCDragonkai, line 24: It's at this point, I reckon, classes should have inline comments on what the purpose of the class is. I used to use jsdoc. Perhaps it's still the same for TS?

src/lib/git/GitServer.ts
```diff
@@ -0,0 +1,281 @@
...
+    await promisify(this.server.close)()
+  }
+
+  // ============ Helper functions ============ //
+  /**
+   * parses a git string and returns the repo name
+   */
+  parseGitName(repo: string): string {
+    const locationOfGit = repo.lastIndexOf('.git')
+    return repo.substr(0, locationOfGit > 0 ? locationOfGit : repo.length)
```
CMCDragonkai: I think this could be clearer if you just explicitly check for whether the `lastIndexOf` is found, and if so, substr it, if not just return the entire repo as normal. No need to substr it at all.

src/lib/git/GitServer.ts
```diff
@@ -0,0 +1,281 @@
...
+      res.end('method not supported')
+    }
+  }
+
+  private notFoundResponse(res: http.ServerResponse) {
+    res.statusCode = 404
+    res.end('not found')
+  }
+
+  private handleInfoRequests(req: http.IncomingMessage, res: http.ServerResponse) {
```
CMCDragonkai: This method should have an comment doc.

CMCDragonkai: Say it looks like a node request handler like a connect middleware with req and res.

src/lib/git/GitServer.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (services.indexOf(service) < 0) {
+      res.statusCode = 405
+      res.end('service not available')
+      return
+    }
+
+    this.infoResponse(repo, service, req, res)
+  }
+
+  private handlePackRequests(req: http.IncomingMessage, res: http.ServerResponse) {
```
CMCDragonkai: I'm getting it now, these are the different requests that a git server has to handle. Assuming these are HTTP requests. Are requests coming over SSH in the same form?

A general question, do these correspond to HTTP GET requests or are they a subtype or do they have a custom HTTP type?

src/lib/git/pack-objects/BufferCursor.ts
```diff
@@ -0,0 +1,33 @@
+// Modeled after https://github.com/tjfontaine/node-buffercursor
+// but with the goal of being much lighter weight.
+class BufferCursor {
```
CMCDragonkai: From my understanding, the buffer cursor is about giving buffers a sort of file descriptor interface. With the ability to tell, seek and so on. But all of this seems rather unnecessary.

A buffer is always a in-memory block of bytes. You don't need to seek anything, you just go to the offset you want and slice the section you want.

If you want read an unsigned 8 bit integer, you just slice 8 bits.

Now if you were intending to use this for parsing, I have a feeling there's a better packet/binary parser library to use.

However this is not a big deal, if we use this for now, that's fine. But if you look at the Haskell code we have, we make use of specific parsing libraries

As a long term thing, we could use something like https://kaitai.io/

CMCDragonkai: And we would have to unify our parsing of grpc/protobuf/git... etc whatever special protocol of messages we needed to have. Although grpc and friends probably have their own libraries.

CMCDragonkai: This is something to create an issue for. Standardized binary parsing library for our Typescript projects.

CMCDragonkai: More resources here: https://github.com/dloss/binary-parsing

At any case, I haven't seen the full usage yet, but I reckon I would have just created utility functions for reading a uint8 and whatever and just did direct operations on the buffer instead of creating a `BufferCursor` class.

src/lib/git/pack-objects/GitCommit.ts
```diff
@@ -0,0 +1,215 @@
+// The amount of work that went into crafting these cases to handl
+// -0 (just so we don't lose that information when parsing and reconstructing)
```
CMCDragonkai, line 2: What does this mean? `-0`?

src/lib/httpRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+        } else if (queue.length > 0) {
+          return resolve({ value: queue.shift(), done: false })
+        } else if (queue.length === 0 && !ended) {
+          defer = { resolve, reject }
+        }
+      })
+    },
+  }
+}
+function httpRequest(getSocket: (address: Address) => net.Socket, address: Address) {
```
CMCDragonkai: Should have a newline above.

src/lib/httpRequest.ts
```diff
@@ -0,0 +1,84 @@
+import net from 'net'
+import http from 'http'
+import { Address } from '@polykey/peers/PeerInfo'
+
+function fromNodeStream(stream) {
```
CMCDragonkai: What's the type of `stream`? Is there a generic TS type for the stream object?

src/lib/keys/KeyManager.ts
```diff
@@ -1,50 +1,72 @@
...
   }
 
   // return {private: string, public: string}
   // The replacePrimary parameter will tell KeyManager to replace the
   // existing identity with one derived from the new keypair
-  async generateKeyPair(name: string, email: string, passphrase: string, replacePrimary: boolean = false, progressCallback?: (info) => void): Promise<KeyPair> {
+  async generateKeyPair(name: string, email: string, passphrase: string, nbits: number = 4096, replacePrimary: boolean = false, progressCallback?: (info) => void): Promise<KeyPair> {
 
     // Define options
     const F = kbpgp["const"].openpgp
```
CMCDragonkai: Is this supposed to mean `F` for flags? Can we use a better name?

src/lib/keys/KeyManager.ts
```diff
@@ -459,16 +495,31 @@ class KeyManager {
...
 
-  isLoaded(): boolean {
+  isLoaded(name: string): boolean {
     if (this.derivedKeys[name]) {
       return true
     }
     return false
   }
+
+  private writeMetadata(): void {
```
CMCDragonkai, line 563: I think the `fs` here is being hardcoded, and in preparation for nativescript we should actually abstract the `fs` into a dependency injected property of these Polykey classes.

That's step 1.

The other issue is resource management.

Resource management of external resources is similar to memory management. If we use RAII, we tie resources to memory management and traditional OOP garbage collection.

If not, then we should use the analogue of "manual memory management" but for resources. And in JS that means `try` plus `finally` clauses to open and close descriptors.

In fact in other languages there will be generally a higher level `file` object, and then you can use that instead. This exists in Python and C actually, with the lower level fd object existing internally.

CMCDragonkai: The below functions are side-effectful convenience functions, but they don't compose really well. Firstly they are synchronous thus blocking the entire thread when reading/writing files. But also they assume alot, the assume the existence of `fs` that matches that API that is available in the outside static context, and then they assume that the file they are writing to can be opened/closed and nothing else cares about that file afterwards. What if you needed to create that file ahead of time with specific permissions first?

In the name of composition, you should think about what you can do abstract some of this.

Of course in the name of practice, you eventually need to close the loop, tie the knot so to speak. Usually that's the IO layer in haskell. And so that's when all the loose ends must have closure, and you plug in the real world.

Finally doing this makes it easier to test/mock.

src/lib/keys/KeyManagerWorker.ts
```diff
@@ -38,9 +38,14 @@ const keyManagerWorker = {
           km = ds.get_key_manager()
         }
         if (km) {
-          resolve(km.get_pgp_fingerprint().toString('hex'));
+          if (km.get_pgp_fingerprint()) {
```
CMCDragonkai: You need to add more inline comments to express the "human" logic of what you are trying to do here. And also "WHY". The order is always: `Why` -> `How` -> `What.

src/lib/peers/PeerManager.ts
```diff
@@ -0,0 +1,237 @@
...
+import RPCMessage from '@polykey/rpc/RPCMessage';
+import createX509Certificate from '@polykey/pki/PublicKeyInfrastructure'
+
+interface SocialDiscovery {
+  // Must return a public pgp key
+  name: string
+  findUser(handle: string, service: string): Promise<string>
+}
+
+const keybaseLookup = async (handle: string, service: string): Promise<string> => {
```
CMCDragonkai: Should eventually be abstracted into a strategy pattern. Not all the time we will be using keybase.

src/lib/utils.ts
```diff
@@ -9,5 +12,114 @@ function invertPromise<T>(p: Promise<T>): Promise<T> {
...
+  symlink(target, path, type, callback)
+  chmod(path, mode, callback)
+  // For git server
+  exists(path, callback)
+  existsSync(path)
+  mkdirSync(path, options?)
+  createReadStream(path, options)
+}
+
+function efsCallbackWrapper(efs: EncryptedFS): GitFS {
```
CMCDragonkai: Wait, this whole function is to enable a callback-version of EFS? That's pretty strange. Didn't VFS have this ability, and then we promisified it?

CMCDragonkai (changes requested):
I want to merge this so I can review the code in its entirety (including nix-shell and execution) and then meetup later to discuss some finer details. But I reckon, you need to add more comments and function documentation (mainly on the the algorithmic details and data flow logic). Anything that is more likely to change and not stabalise you can leave it not commented, but must be documented in the MR or issue, or new issues should be created to address TODO elements.

I haven't got through all of it yet. But this is review number 2!

But it's looking alot more cleaner and structured! :D

src/lib/pki/PublicKeyInfrastructure.ts
CMCDragonkai: If you are going to to leave some defaults, can we get the defaults either at the class parameter/function parameter level, or if there's alot, then a `config` or `defaults` module file should be created to centralize it. It will make it easier to iterate on the defaults.

src/lib/KeyManager.ts
CMCDragonkai: Can we use the `promisify` and `callbackify` to unify what we need? https://medium.com/trabe/understanding-nodes-promisify-and-callbackify-d2b04efde0e0

package.json
CMCDragonkai: This is actually something I reckon we can build ourselves (at least from a timing perspective). And I also think the visual aspect of progress bar componentry would probably be in Polykey repo and not here.

src/lib/KeyManager.ts
robert-cronin: I thought we couldn't do the standard promisify pattern because kbpgp had multiple return types but it turns out promisify just takes the first one anyway so I've changed it to using this :+1:

metadata
robert-cronin: this was added in by mistake, it has been removed.

src/cli/initPolykey.ts
robert-cronin: I've added in a new key value pair for polykey cli.

The `bin/pk.js` file was generated by hand, I am not sure how to get it to be generated by webpack. I can set it as an entry and then have it compiled but that would compile the rest of the code into that file as well since webpack follows `require` imports.

Perhaps we should also add another key for our alias `pk`. I've added it in for now, otherwise we can make that something that happens with our desktop app installation process.

CMCDragonkai:
We should have `pk` and `polykey` as the main cli aliases. Gui app can be `pk gui` or something. Although we also need a libexec command for starting the daemon. Like `pk agent`.

Will discuss this with you soon. Suggest starting a spec somewhere in Polykey issue of all the command workflows.

On 29 June 2020 13:07:12 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@robert-cronin commented on this pull request.
>
>
>
>> @@ -5,7 +5,7 @@ import inquirer from 'inquirer'
> import Configstore from 'configstore'
> import PolyKey from '../lib/Polykey'
> import cliProgress from 'cli-progress'
>-import KeyManager from '@polykey/KeyManager'
>+import KeyManager from '@polykey/keys/KeyManager'
> import { actionRunner, pkLogger, PKMessageType } from './polykey'
> 
>
>I've added in a new key value pair for polykey cli.
>
>The `bin/pk.js` file was generated by hand, I am not sure how to get it
>to be generated by webpack. I can set it as an entry and then have it
>compiled but that would compile the rest of the code into that file as
>well since webpack follows `require` imports.
>
>Perhaps we should also add another key for our alias `pk`. I've added
>it in for now, otherwise we can make that something that happens with
>our desktop app installation process.
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/43#discussion_r446748129

-- 

src/lib/git/GitServer.ts
robert-cronin: yes, I wasn't sure whether there was anything in `through` that we particularly needed from the `node-git-http-backend` code but it doesn't look like it. So this has been removed in favour of the `readable-stream` lib :+1:

src/lib/git/GitServer.ts
robert-cronin: Yeah to be honest, I haven't given inline documentation much thought until now. I will go ahead and add it for the rest of the code as well :+1:

src/lib/keys/KeyManager.ts
robert-cronin: I have changed this to the lowercase `flags`

src/lib/git/GitServer.ts
CMCDragonkai: Anything that's non-obvious.

src/cli/initPolykey.ts
CMCDragonkai: Yea that's fine to generate by hand.

src/lib/pki/PublicKeyInfrastructure.ts
robert-cronin: These defaults have been added to the class level as static parameters. I figured there isn't enough defaults to justify a defaults module file yet. Also I actually don't yet know what sensible defaults for the certificates should be but at least they are at the class level now :+1:

src/lib/Polykey.ts
robert-cronin: I deliberately removed the private scoping to match what was there for the `KeyManager` class as well as make it easy for users to access the different managers. I have had no issues with losing scope because there are no methods in the `PolyKey` class that need binding.

src/lib/git/GitServer.ts
robert-cronin: I've added documentation to all the public methods available and additional doc strings to methods that I feel are non-obvious :+1:

src/lib/git/GitServer.ts
robert-cronin: This turned out to be a redundant method after all! it has been removed. All our repos/vaults are non-bare repos and as such don't have the `.git` prefix.

src/lib/git/GitServer.ts
robert-cronin: This method has been documented.

src/lib/git/GitServer.ts
robert-cronin: Yes these are the HTTP requests that underlie the Git protocol. For `handleInfoRequest`, the client sends a GET request and specifies what it wants in the url: `<hostname>/<vaultname>/info/refs` and the git server replies with its info. Then the client can further request a git pack with a POST request and in the body it specifies what it wants and what is has, then the git server takes the difference and bundles it into a packfile that is streamed back to the client. This is how the `handlePackRequests` is supposed to work.

I am actually not sure about SSH yet, but I suspect the way git communicates would be similar and just how the requests and responses are wrapped and sent over the wire is different. In the gRPC PR, I am trying to separate the way git talks from how it is transported.

src/lib/git/pack-objects/BufferCursor.ts
robert-cronin: I've managed to completely take out the dependency on `BufferCursor`. It turned out the part of the code that was using it wasn't being called at all. I'm not sure why I didn't see this before. Shall I still create an issue for this?

src/lib/httpRequest.ts
robert-cronin: This function and its corresponding helper method has been grouped into a class that makes the whole thing a little easier to handle. Newlines have also been added above each method declaration.

src/lib/httpRequest.ts
robert-cronin: The stream type is a http Incoming message and its not really a good name for it. So I have renamed `stream` to `message` and `fromNodeStream` to `httpMessageToIter`.

src/lib/keys/KeyManagerWorker.ts
robert-cronin: I've added some comments in to make it a little more clear why we verify the data only if there is a pgp finger print.

src/lib/utils.ts
robert-cronin: Yeah so this is something I didn't realise until I got to working on the git integration in polykey. Isomorphic git expects the `fs` client to either be a callback or promise based agent and in the case of promises, it expects the methods to be contained in `fs`.`promises` which is something EFS does not do yet. There are also some other discrepancies in so far as `efs` doesn't exactly conform to the `fs` interface. I have been meaning to create an issue for this.

robert-cronin: I have created an issue here: https://github.com/MatrixAI/js-encryptedfs/issues/32

src/lib/peers/PeerManager.ts
robert-cronin: There is a strategy pattern just above this function in the form of an interface for `SocialDiscovery`. The keybase was set as a default `SocialDiscovery` but it is left to the user to add more.

src/cli/initPolykey.ts
robert-cronin: awesome, I shall resolve this then :+1:

src/lib/git/pack-objects/GitCommit.ts
robert-cronin: I believe this has something to do with the timezone offset, there is a function called `simpleSign` that converts a `-0` which represents a negative sign into a `-1`

src/lib/keys/KeyManager.ts
robert-cronin: All of these methods now use the file system that is passed into the class in the same way that `EncryptedFS` takes a file system parameter. This should cover us for nativescript as we can just build an object that can wrap around the Nativescript API or just used [nativescript-nodeify](https://www.npmjs.com/package/nativescript-nodeify) as nodejs fs is a particular node package they say is supported.

robert-cronin:
Captains log, star-date 73962.7.
I have also added a fix for https://github.com/MatrixAI/js-polykey/issues/22 to the PR. This is explained at the bottom of the linked issue.

src/lib/utils.ts
robert-cronin: and a PR here: https://github.com/MatrixAI/js-encryptedfs/pull/33
I don't think there is anything stopping us from merging this PR, the callback wrapper was intended to just be a stop-gap until we could just pass in the naked efs object.

src/lib/git/GitServer.ts
robert-cronin: I have left out the underlying, low-level git functions as I'm not sure they would ever need to be understood by the user.

src/lib/keys/KeyManager.ts
CMCDragonkai: Great awesome.

src/lib/peers/PeerManager.ts
CMCDragonkai: I was expecting a directory with different strategies in different files. But it's cool for now.

src/lib/git/pack-objects/GitCommit.ts
CMCDragonkai: This is about git commits? I don't entirely understand the context of this problem. I may ask you later during our meeting.

CMCDragonkai:
Time to merge this!

We'll move on from the git problem. And now start on the IPC architecture and start to draft out how Electron, CLI, Agent, OS platforms and nativescript will all work together.

New issues for IPC required, @robert-cronin please link them at the end of this PR. And just squash/clean the commits.

robert-cronin:
Yes time to merge indeed!

I've added two issues (https://github.com/MatrixAI/js-polykey/issues/51 and https://github.com/MatrixAI/js-polykey/issues/50) from our discussion yesterday that I will also create a new branch and PR for.

PR #47, opened 2020-06-29
Add grpc server

### LINKED ISSUE #46: Git Communication over a gRPC connection
Currently we have the git server working over an insecure http connection (using nodejs http lib) but we want to eventually have the git server working over an existing connection that has already been secured between two peers. 

I have implemented a `ConnectionManager` that creates a single secure connection between 2 peers (e.g. server from peer A and socket from peer B) using mutual TLS, but the question now is how do we make the git server use that connection?

The request (client connection) can be done fairly easily using the `createConnection` option. The server has been more challenging and that is what this issue is meant to address.

robert-cronin:
This PR is specifically for making the git backend use GRPC for secure communications.

Fixes #46.


CMCDragonkai:
@Zachaccino can you have a look at this regarding GRPC.

@robert-cronin can you point to which specific file has the GRPC implementation?

robert-cronin:
The specific files that have gRPC are [git_server.proto](https://github.com/MatrixAI/js-polykey/blob/add-grpc-server/src/proto/git_server.proto), [PeerManager.ts:Line98](https://github.com/MatrixAI/js-polykey/blob/a0d961b41091bfd09cdf6ce08dd9f63efee3ccdf/src/lib/peers/PeerManager.ts#L98) and [GitClient.ts](https://github.com/MatrixAI/js-polykey/blob/add-grpc-server/src/lib/git/GitClient.ts). Any insights you could share would be appreciated, I've only had limited experience with using gRPC in production. I also am still yet to figure out the best way to implement the secure credentials for [SSL](https://github.com/MatrixAI/js-polykey/blob/a0d961b41091bfd09cdf6ce08dd9f63efee3ccdf/src/lib/git/GitClient.ts#L26) mostly the CA.

Zachaccino:
Looks good to me, that said, I don't have much experience working with GRPC in production either.

robert-cronin:
This PR is closed in favour of PR #52

PR #52, opened 2020-07-03
Add grpc server

### LINKED ISSUE #46: Git Communication over a gRPC connection

robert-cronin:
Since this is a relatively small PR, I've decided instead of rebasing the other PR onto master, I would just create a new branch with the small changes to save time. I will close PR #47 in favour of this one.

This PR is specifically for making the git backend use GRPC for secure communications.

Fixes #46.


src/lib/HttpRequest.ts
```diff
@@ -1,101 +1,101 @@
-import net from 'net'
-import http from 'http'
-import { Address } from './peers/PeerInfo'
+// import net from 'net'
```
CMCDragonkai: Why is this all commented out? Are we no longer using HTTP?

robert-cronin: Oh wow that was quick! Actually the HttpRequest is no longer used so I've removed it. See latest file :+1:

robert-cronin: It is replaced by `GitClient`, new name is because we're no longer using HTTP protocol, instead we are using a gRPC service :+1:

src/lib/git/GitClient.ts
```diff
@@ -0,0 +1,132 @@
+import grpc from 'grpc';
+import { Address } from '../peers/PeerInfo';
+
+class GitClient {
```
robert-cronin, line 10: This also probably needs some documenting to explain exactly what it's for

src/lib/pki/PublicKeyInfrastructure.ts
```diff
@@ -1,16 +1,43 @@
-import forge from 'node-forge'
+import forge, { pki } from 'node-forge'
 
 class PublicKeyInfrastructure {
```
robert-cronin, line 6: This requires documenting as well.

CMCDragonkai:
Are we able to separate PKI CA functionality from the TLS requirements of GRPC? Because the design of the CA is needed for other things, whereas GRPC just need a TLS certificate (a self-signed cert is fine). I think alot of apps even just expect you to give them a cert, and if it doesn't exist, will generate one like jupyter notebook which auto creates HTTPS server.

robert-cronin:
> Are we able to separate PKI CA functionality from the TLS requirements of GRPC? Because the design of the CA is needed for other things, whereas GRPC just need a TLS certificate (a self-signed cert is fine). I think alot of apps even just expect you to give them a cert, and if it doesn't exist, will generate one like jupyter notebook which auto creates HTTPS server.

Certainly, we can make the PKI just a convenience class and have polykey take a parameter of cert file, key file and a list of root certs for the trusted CAs. that would satisfy the separation requirements I think

CMCDragonkai:
I believe that the cert file and key file are the ones necessary for the server side. As for the client side, if we are talking about pk clients, they should accept self-signed certs, or not verify against CA roots. Although we could pre-generate certs for polykey when we distribute the application from Matrix AI... but this will require renewal.

If we can bootstrap trust from OS capabilities, then pk client should just trust any certificate presented by the server in order to establish an secure connection. Verifying the certificate is only necessary if you're not sure about the identity of the pk agent/server you are connecting to. If you are on the same OS, same user, same network context... then we assume that the port being used by the pk server is the right one. Similar to how gpg client trusts the gpg agent it connects to through the unix domain socket created in `/run/user/1000`... etc.

Do they consider the possibility that another application must also create a `/run/user/1000/gnupg/...` to masquerade like the gpg-agent? Is this even a thing to worry about?

robert-cronin:
There was the issue of enforced certificate checking by node grpc so self signed certs don't actually work. Could we just make the agent a trusted CA? Since the agent is already a known location for each user context, it could act as the CA for each newly created polykey agent. But like you said, in the same user context, then it's a non-issue. So the agent as a CA would be more for the issue that we cant self sign certs in gRPC. 

Also since there is only ever one agent per user context, this would be the bootstrap as in each of the running pk agents know to trust the agent so can trust each other in the same user context. 

I'm not sure it is an issue that other agents could masquerade as another agent because keynodes would all have the same agent and know to look in the one place: `/run/user/1000/polykey/S.polykey-agent` and there is only ever one of these socket files.

CMCDragonkai:
I meant other non-polykey programs masquerading as a polykey agent. But I'm not sure if this is even a problem since it seems possible for gpg agent to act like an ssh agent.

On 4 July 2020 01:20:52 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>There was the issue of enforced certificate checking by node grpc so
>self signed certs don't actually work. Could we just make the agent a
>trusted CA? Since the agent is already a known location for each user
>context, it could act as the CA for each newly created polykey agent.
>But like you said, in the same user context, then it's a non-issue. So
>the agent as a CA would be more for the issue that we cant self sign
>certs in gRPC. 
>
>Also since there is only ever one agent per user context, this would be
>the bootstrap as in each of the running pk agents know to trust the
>agent so can trust each other in the same user context. 
>
>I'm not sure it is an issue that other agents could masquerade as
>another agent because keynodes would all have the same agent and know
>to look in the one place: `/run/user/1000/polykey/S.polykey-agent` and
>there is only ever one of these socket files. 
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/52#issuecomment-653594003

-- 

CMCDragonkai:
Are you sure that node GRPC enforces CA verification? That seems strange. Usually there should be an option for this.

robert-cronin:
I've added PKI functionality to the proper domain, which as I understand it is the key manager. Also, the CA requirements are no completely separate and the key manager just accepts certs via a load cert function.

src/lib/Polykey.ts
```diff
@@ -1,24 +1,23 @@
...
-  private gitServer: GitServer
 
   constructor(
     polykeyPath: string = `${os.homedir()}/.polykey`,
     fileSystem: typeof fs,
     keyManager?: KeyManager,
     vaultManager?: VaultManager,
-    peerManager?: PeerManager
+    peerManager?: PeerManager,
+    port?: number
```
robert-cronin: This is unnecessary, delete.

src/lib/Polykey.ts
```diff
@@ -29,11 +28,15 @@ class Polykey {
...
-    this.gitServer = new GitServer(this.polykeyPath, this.vaultManager)
-    this.peerManager.connectLocalPeerInfo(this.gitServer.address)
+    this.peerManager = peerManager ?? new PeerManager(
+      this.polykeyPath,
+      fileSystem,
+      this.keyManager,
+      this.vaultManager,
+      undefined,
+      undefined,
+      port
```
robert-cronin: Same here.

src/lib/git/GitClient.ts
```diff
@@ -0,0 +1,170 @@
...
+    });
+    const git_server_proto = grpc.loadPackageDefinition(packageDefinition);
+
+    const pkiInfo = keyManager.PKIInfo
+    if (pkiInfo.caCert && pkiInfo.cert && pkiInfo.key) {
+      this.credentials = grpc.credentials.createSsl(
+        pkiInfo.caCert,
+        pkiInfo.key,
+        pkiInfo.cert,
+        // { checkServerIdentity: this.checkServerIdentity }
```
robert-cronin: No comments left in code.

src/lib/git/GitClient.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    this.client = new (git_server_proto.GitServer as any)(address.toString(), this.credentials);
+  }
+
+  // checkServerIdentity(hostname: string, cert: grpc.Certificate): Error | undefined {
+  //   console.log(hostname);
+  //   console.log(cert);
+
+  //   return
+  // }
```
robert-cronin: Delete.

src/lib/keys/KeyManager.ts
```diff
@@ -6,16 +6,22 @@ import crypto from 'crypto'
...
+  pkiKeyPath: string | null,
+  pkiCertPath: string | null,
+  caCertPath: string | null,
 }
 
 type KeyPair = {
   private: string | null,
   public: string | null
 }
+type PKInfo = { key: Buffer | null, cert: Buffer | null, caCert: Buffer | null }
```
robert-cronin, line 23: Format file.

PR #53, opened 2020-07-03
Add polykey agent

### LINKED ISSUE #50: IPC communication between agent and client (CLI/GUI)
The next problem to figure out is how the client and agent, it looks like there are a couple of ways to achieve this:

- Via [gRPC](https://stackoverflow.com/questions/54844882/grpc-server-in-python-with-unix-domain-socket)
- Via NodeJS net module which has support for [windows and unix]( https://nodejs.org/api/net.html#net_ipc_support)

Discovery between the agent and the client also needs to be figured out, here are some ideas:

- Rely on a shared path (this shared path must be user-specific) like `/run/user/1000/polykey/socket` (this only exists on Linux, find equivalent on Mac/Windows).
- Rely on a unix domain socket which is also on a shared path that is user-specific. This relies on a shared path existing. 

Using UDS is really about getting access to the OS security capabilities to securing the comms. The nodejs TLS module can also do UDS as its an abstraction over the net module, you just have to specify the path in [tls.connect](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback). I'm just not sure about the server and tls.

### LINKED ISSUE #51: Lockfiles to constrain keynodes to a single agent
Also it is intended to have a single agent per user that can manage multiple keynodes just like a single ssh-agent per user can manage multiple keys. For any particular keynode, it can be registered against a particular agent that represents the authentication context for that keynode. 

To prevent multiple agents (i.e. muiltiple users on the same system) from managing the same keynode, we could create a lockfile that constrains the keynode to a particular running agent.

### LINKED ISSUE #54: Prevent webpack from compiling lib code thrice for agent/cli/lib
Webpack currently has three outputs: a file each for the lib, cli and agent. The core lib code is compiled in all three of these outputs and that is unnecessary. we need to find some way to separate them out.
Here is a useful gist: https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693

### LINKED ISSUE #59: grpc-js: No connection established
This error is a little hard to decipher but it occurs when you try to run the server without also running the CA server that they were issued from. Best way to fix is to re-issue the ssl certs with `./scripts/generate_ssl_certs.sh`.

I shall put something in the README along these lines in the upcoming agent PR

robert-cronin:
This PR will add the pk agent and IPC communications between the agent and any clients. It will also introduce lockfiles (**not done**) for preventing two pk agents from operating on the same keynode at once.

Fixes #50 
Fixes #54 
Fixes #59


robert-cronin:
branch has been rebased onto master

robert-cronin:
It seems like there is a lot of redundancy in defining the communication protocol between cli and agent. At first I tried to simply define the cli on the agent and then pass the arguments via the unix domain socket but I ran into some issues especially with controlling the UX on the cli so I went with defining a messaging protocol between the agent and any potential front end. I ended up having what seems to be a lot of redundancy so if you have any tips on how to reduce this, it would be greatly appreciated, unless this is the best way to do it?

CMCDragonkai:
What do you mean by redundancy here? I suppose grpc has protoc files? Or do you mean you want to "generate" the marshalling code. There are libraries to do this.

LICENSE.sig
```diff
@@ -0,0 +1,95 @@
+-----BEGIN PGP MESSAGE-----
```
CMCDragonkai: What is this file?

CMCDragonkai:
This MR is too big for me to review. I mean that literally, my browser doesn't load it properly. Won't be able to review from this interface. Are you able to either split it up into separate MRs, if not, we can instead perhaps do this by just merging and it and going through the pieces.

LICENSE.sig
robert-cronin: this was definitely not meant to be committed, I was testing the agents ability to sign something. This has been removed.

robert-cronin:
> What do you mean by redundancy here? I suppose grpc has protoc files? Or do you mean you want to "generate" the marshalling code. There are libraries to do this.

Yes I just mean there is a lot of code to convert proto files into javascript and back again. I am going to take a look into `pbts` which is the `protobufjs` cli for generating bundled type definition and javascript files that describe the proto files directly.

robert-cronin:
> This MR is too big for me to review. I mean that literally, my browser doesn't load it properly. Won't be able to review from this interface. Are you able to either split it up into separate MRs, if not, we can instead perhaps do this by just merging and it and going through the pieces.

I didn't realize it was this big! there are a few things I can do to make it smaller.
Using `pbts` to replace all the code for converting the proto files into javascript should reduce it significantly.

CMCDragonkai:
There's a conflict in package.json. This commit will need to be rebased.

.env.example
```diff
@@ -1,4 +1,4 @@
-STEP_ROOT='tmp/CA/root_ca.crt'
+STEP_ROOT='tmp/secrets/CA/root_ca.crt'
 STEP_CA_URL='https://localhost:55555'
```
CMCDragonkai, line 2: I don't think we need any CA env variables for agent. The step-ca stuff is only for experimenting with CA features, we are meant to create our own PKI eventually. So embedding the env variables here shouldn't be necessary. I suppose you could put it in your own local `.env`. And update the README.md for what should be done if we want to test some PKI feature.

README.md
```diff
@@ -188,3 +188,11 @@ pk crypto verify ./signed_file --detach-sig='./signed_file.sig'
 
 ## Verbosity
 TODO: explain verbosity levels when it is implemented
+
+# Build
+## Proto Files
+All `.proto` files are stored in the the `proto` directory. JavaScript and type definition files are build using the following command:
+
+```
+npm run build:proto
```
CMCDragonkai, line 203: Should mention how this is integrated into the webpack build system?

src/lib/peers/PeerManager.ts
```diff
@@ -146,11 +150,15 @@ class PeerManager {
       this.credentials = grpc.ServerCredentials.createInsecure()
     }
 
-    // Bind server and set address
-    const port = this.server.bind('0.0.0.0:0', this.credentials);
-    const address = new Address('localhost', port.toString())
-    this.server.start();
-    this.localPeerInfo.connect(address)
+    this.server.bindAsync('0.0.0.0:0', this.credentials, (err, boundPort) => {
```
CMCDragonkai: BTW, this should come from the CLI parameters or enviroment variable. I would suggest both where the parameters overrides the env variable.

webpack.config.ts
```diff
@@ -0,0 +1,189 @@
...
+    new DeclarationBundlePlugin({
+      ignoreDeclarations: true
+    })
+  ],
+  watchOptions: {
+    ignored: /node_modules/
+  }
+}
+
+const agentConfig = {
```
CMCDragonkai: I'm wondering why there's so many separate builds here. I thought there would be:

1. a library build
2. a cli build

Especially since the cli can use subcommands to launch daemon or whatever.

README.md
robert-cronin: This is a work in progress since `grpc_tools_node_protoc` doesn't seem to work on nixos, might need your help on that one, it should install with `grpc-tools` but `grpc_tools_node_protoc` doesn't seem to be added to the path or installed properly

CMCDragonkai: Does it require native compilation? Is the failure occurring during `npm install`? If so, there are ways around it...

CMCDragonkai: Some npm installs require compilation. This can be done in a nix-shell if the nix-shell has all the tooling required. There's quite a bit of stuff like `node-pre-gyp` or whatever. And python2. I did this before when I was trying to put `dat` and other binary npm stuff into Nixpkgs: https://github.com/svanderburg/node2nix/issues/71 and https://github.com/NixOS/nixpkgs/pull/32940.

However that's for using the packages with CLI executables in Nixpkgs itself. That may be the right idea to do if we can use the protoc compilers from `shell.nix` and that means pushing those packages to upstream (but first in our overlay).

Of course bringing all the tools needed to do compilation for Node is a bit of a mess.

CMCDragonkai: BTW have you seen this: https://github.com/agreatfool/grpc_tools_node_protoc_ts? It builds on top of the grpc_tools_node_protoc to generate typescript code instead of just JS code.

CMCDragonkai: There you see: https://github.com/grpc/grpc-node/blob/master/packages/grpc-tools/package.json#L27. The `grpc-tools` package requires using `node-pre-gyp`. 

As far as I remember, native Node dependencies has 3 kinds of buildchains:

* node-gyp-build
* node-gyp
* node-pre-gyp

So grpc-tools is using `node-pre-gyp`.

I wrote the section on native compilation here: https://github.com/MatrixAI/nixpkgs/blob/master/doc/languages-frameworks/node.section.md

So packaging that into Nix might be the right idea. You can start by putting all the build tools into your shell.nix.

src/lib/peers/PeerManager.ts
CMCDragonkai: We can have things like `PK_X` as our environment variable.

webpack.config.ts
robert-cronin: That is a good point, the agent code can be merged into the lib code. The daemon script is a little more challenging to separate because it is a static script to be called by `fork(..)`

.env.example
robert-cronin: sounds good, I will leave it in as an option and tell the user in README that if they want to use step-cli for testing, they will needs these environment variables

webpack.config.ts
robert-cronin: I managed to combine the agent and lib code, so now we are down to three separate builds. I could probably get it down to 2 if I just use webpacks copy plugin

robert-cronin: actually that probably wouldn't be easy since it has relative imports that need to change with webpack externals.

README.md
robert-cronin: I have added in `nodePackages.node-pre-gyp` to `nativeBuildInputs` within `shell.nix`. I am getting a property missing error. In the section on native compilation, the addition is to `buildInputs` so maybe that is the issue?

robert-cronin: Funnily enough, npm seems to download `node-pre-gyp` and compile just fine until it gets to actually using the command, in which this error appears:
```
➜  js-polykey git:(ipc) ✗ grpc_tools_node_protoc                                                                             ~/Documents/github/js-polykey
/home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/protoc.js:41
    throw error;
    ^

Error: spawn /home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/protoc ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn /home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/protoc',
  path: '/home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/protoc',
  spawnargs: [
    '--plugin=protoc-gen-grpc=/home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/grpc_node_plugin'
  ],
  cmd: '/home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/protoc --plugin=protoc-gen-grpc=/home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/grpc_node_plugin'
}
```

robert-cronin: I think this will be really useful, `grpc_tools_node_protoc` can compile the protos into js code that imports `grpc-js` directly and so we would have no need for loading protos anymore. So its worth getting this right.

src/lib/peers/PeerManager.ts
robert-cronin: oh yes, the port was an environment variable before, I've just forgotten to keep it that way when I changed to `grpc-js`. It is fixed now :+1: `process.env.PK_PORT ?? 0`

webpack.config.ts
robert-cronin: I might leave this as 3 unless we can come up with a way to compile scripts that will be `fork`ed with webpack.

README.md
CMCDragonkai: The node-pre-gyp would be best used as part of creating a new nodePackage in nixpkgs upstream. See the PR I created for dat. It's a similar process.

CMCDragonkai: ENOENT means something doesn't exist. Perhaps the tool is trying to execute something that doesn't exist. This can happen due to absolute paths or paths that don't expect a NixOS environment. I had a similar problem occur when trying to package dvc for Nixpkgs: https://github.com/iterative/dvc/issues/1509

You might want to chase down that error, and find out exaxtly what the path is failing.

Does this path exist? `/home/robbie/Documents/github/js-polykey/node_modules/grpc-tools/bin/protoc`.

CMCDragonkai: Recommend reading: https://github.com/svanderburg/node2nix

CMCDragonkai: And: https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/node.section.md

webpack.config.ts
CMCDragonkai: It is possible to "fork" with 1 cli script, by doing "self-forking". Basically the script calls itself. I've seen this pattern before, it's pretty ingenious. It's like recursion at the process level. But when you call yourself again, you provide parameters that understands that it is going to be a daemon.

src/lib/peers/PeerManager.ts
CMCDragonkai: Not just the port, but the address to. So you would have `PK_HOST`.

CMCDragonkai:
I'm guessing there's only 1 thing blocking this merge atm:

1. The protoc issue and native compilation of the dependency

This problem is really about understanding how to use nix.

Once this is done, we merge and then start reviewing each part in detail... in preparation for the UI work and the demo integration into Ouroboros and/or Matrix Relay VPN. These 2 will require you to talk to @nzhang-zh and @DrFacepalm

CMCDragonkai:
In the mean time, can we also get a demo of the `pk` command line.

If you can record that, with https://asciinema.org/ and show that to the team for the all the common use-cases: generating secrets, passing secrets around, renaming stuff, sharing stuff... etc etc.

README.md
CMCDragonkai: I suggest putting up an issue in the grpc-node repo. I did this before for the `dvc` repo regarding "Packaging for NixOS". And put down your woes there. Someone there might know something about it and help. The `dvc` guys were helpful before too.

.env.example
CMCDragonkai: Cool

Issue #55, opened 2020-07-10
Exposing different interfaces for use of secrets within polykey


robert-cronin:
We need to figure out how to expose the secrets contained within polykey for use in difference contexts.

Some potential interfaces are:
-  clipboard
- stdout,
- a file descriptor
- output to file (or just pipe to file/redirect to file)
- "env variable" injection 
- http

CMCDragonkai:
Should keep track of this issue too: https://github.com/NixOS/rfcs/pull/59#issuecomment-657252560

CMCDragonkai:
## Environment Variables

So to give an example of environment variables.

In any Unix shell you can set environment variables:

```
x=3
echo "$x"
```

You can expose these variables to subprograms:

```
export x=3
echo "$x"
bash -c 'echo "$x"'
```

However subprograms CANNOT set environment variables in the parent process. Which is the current shell.

So how would polykey help in the process of setting environment variables?

There are 2 ways:

1. Is for the `pk` command to spit out instructions to set environment variables in the shell. This requires the user to copy the command. This I've seen in other programs, but it's not really a nice UX.
2. Is to do what `env` does. The `env` command basically is a "fork+exec" at the shell-level but with the additional ability to set environment variables.

Let's explore 2.

To do 2. it is possible to not even use `env` at all.

```
x=3 bash -c 'echo "$x"'
```

This ensures that the environment variable is set for only the subprocess. And it does not export it to every other subprocess. This is good because of principle of least privilege.

So why use `env`? It is because it has lot of extra options to deal with environment variable manipulation and also signal manipulation and also dealing multi-parameter shebangs.

So how does this apply to Polykey?

The second way for Polykey to interface via environment variables is to expose an `env` like interface.

Suppose I wanted to pass some secret into some program that uses environment variables.

I could run it with polykey like this.

```
pk secrets get K --run 'bash -c "echo $K"'
```

The above basically "chains" the secret coming out of polykey directly injecting as an environment variable into the command in `--run`.

See `nix-shell` for examples of this. The `nix-shell` has `[--command cmd] [--run cmd]`. The first is an expression of a shell command. The second is the ACTUAL program being executed. So `--run` should be equivalent to doing a `execvp`. Whereas `--command` should be equivalent to a `execv`. See: https://stackoverflow.com/a/20823413. So the main difference is that `--command` runs in the "shell", whereas `--run` runs the actual program.

[...]

PR #64, opened 2020-07-22
Add OAuth2

### LINKED ISSUE #70: Network Interface
## HTTP or GRPC or take your pick of any networked protocol

This interfacing method makes Polykey equivalent to any other password/secret service like Vault or AWS Secret manager.

Basically this requires a third party networked client to send a request message with an authentation token that PK checks, and then returns a response with the relevant secret.

This has a little of complication here, because now you have these authentication tokens that PK needs to keep track of. It can do this statelessly with HMAC tokens, or it can do it statefully. But essentially this makes PK a "web service" like any other secret management system.

So if somebody wanted to use polykey, but wanted to use it IN THE EXACT SAME WAY they already use AWS Secret Manager or Vault. Then they could. But it does add some extra work necessary to provide the token generation/distribution ability.

To see how to do this, just look at Vault. They have an existing token system for this particular usecase.

And also OAuth2 is basically this. @DrFacepalm 

To simplify this implementation, you could just take an OAuth2 server javascript impl and use that as the token system.

It seems like we can just use this: https://github.com/oauthjs/node-oauth2-server

Once you do this you have these concepts to model:

* Users
* Tokens
* The relationship between users, tokens and vaults

So basically 1 user, may be given N tokens, and each token has the "authority" a.k.a. "scope" to access a subset of all vaults.

See the OAuth2 spec here: https://tools.ietf.org/html/rfc6749#section-1.2

And talk to @DrFacepalm since he's already gone through it.

This thing should be left as low priority for now.

### LINKED ISSUE #83: Separate CLI commands into separate files.
They are currently in single domain specific files like `Secrets.ts` or `Peers.ts` but we should probably separate them into folders and commands like this perhaps:
```
secrets/
├── index.ts
├── AddNewSecret.ts
├── GetSecret.ts
```

### LINKED ISSUE #109: Default port for TCP and UDP
We have decided on a default port of `1314` for polykey but there is an issue of port conflicts if there are multiple pk instances serving on the same ip address. We need some way to check first if a port is taken. There might be a way to do this in gRPC but if not we can take some ideas from here: https://stackoverflow.com/questions/19129570/how-can-i-check-if-port-is-busy-in-nodejs/35251815#:~:text=Check%20out%20the%20amazing%20tcp,port%20is%20accepting%20connections%20tcpPortUsed.

### LINKED ISSUE #115: CLI unclear help message for `polykey vaults new`
Haven't checked other subcommands but the help message for `pk vaults new` does not mention required argument.

```sh
[nix-shell:~/matrix.ai/js-polykey]$ ./bin/polykey vaults new --node-path ~/.polykey
```

This does not return error but also doesn't do anything. Maybe nicer to get an error message?

```sh
[nix-shell:~/matrix.ai/js-polykey]$ ./bin/polykey vaults ls --node-path ~/.polykey
no vaults found
```

Realized I need to supply a vault name but help message does not indicate this argument.

```sh
[nix-shell:~/matrix.ai/js-polykey]$ ./bin/polykey vaults new -h --node-path ~/.polykey
Usage: polykey vaults new [options]

create new vault(s)

Options:
  --node-path <nodePath>  node path
  -h, --help              display help for command


[nix-shell:~/matrix.ai/js-polykey]$ ./bin/polykey vaults new myvault --node-path ~/.polykey
vault created at '/home/ning/.polykey/myvault'
```

robert-cronin:
This may be extended to add other interfaces in accordance with issue #55 
Fixes #70 
Fixes #83  
Fixes #109
Fixes #115


robert-cronin:
Below are a 2 demos from the new OAuth2 HTTP API (both are using bearer tokens issued by the CLI). Note we are not using any kind of login system as polykey does not need to be used to login to any other service nor is any other service going to be integrated into polykey, the OAuth2 code is purely for HTTP based clients that cannot use the pk cli.

Via Swagger Docs:
Polykey OAuth2 Bearer Token Flow Demo gif

Via PK CLI and curl:
https://asciinema.org/a/cq11TZKa81OfWUW5iqD5xyDYq

CMCDragonkai:
Nice only 43 file changes this time!

CMCDragonkai:
A few comments from our discussion:

* The primary reason for OAuth is so that applications that are HTTP-based can integrate into Polykey as an alternative to GRPC (which is used internally between Polykey keynodes and Matrix OS)
* Therefore graphical choosing of scopes is probably an unnecessary feature. This feature is more designed for the cases whether one might "login to Polykey from Facebook/Google" or "login into X with Polykey". The former doesn't make sense for Polykey's usecase as it is decentralised app and not a web service that requires some social information ahead of time. That being said, there might be some usecase of doing this in the same way keybase federated digital identities cryptographically, but is that really an OAuth thing? Maybe it is... but we can leave this feature for after the MVP and attach it to #33. The latter is very unlikely to happen.
* Why not `pk oauth` and not `pk api` as `pk api` is a bit confusing.

CMCDragonkai:
But that swagger system is REALLY cool.

package.json
```diff
@@ -54,31 +54,73 @@
     "@types/commander": "^2.12.2",
     "@types/configstore": "^4.0.0",
     "async-mutex": "^0.2.4",
+    "axios": "^0.19.0",
```
CMCDragonkai: Please use normal `fetch`, there's no need to use `axios` at all. I know I used it before, but if we can avoid bringing in lots of dependencies, that's good.

package.json
```diff
@@ -54,31 +54,73 @@
...
+    "body-parser": "^1.19.0",
+    "bunyan": "^1.8.14",
+    "chai": "^4.2.0",
```
CMCDragonkai: Isn't this for testing?

package.json
```diff
@@ -54,31 +54,73 @@
...
+    "oauth2orize": "^1.11.0",
+    "ono": "^5.0.1",
+    "openapi-sampler": "^1.0.0-beta.15",
     "pako": "^1.0.11",
+    "passport": "^0.4.1",
+    "passport-http": "^0.3.0",
+    "passport-http-bearer": "^1.0.1",
+    "passport-local": "^1.0.0",
+    "passport-oauth2": "^1.5.0",
+    "passport-oauth2-client-password": "^0.1.2",
```
CMCDragonkai, line 74: Are all these passport libraries needed? Especially for the limited usecase of OAuth that we are using?

package.json
CMCDragonkai: If so... then `chai` should be in `devDependencies` and other related chai libraries too.

package.json
```diff
@@ -54,31 +54,73 @@
...
+    "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
     "chalk": "^4.0.0",
     "commander": "^5.1.0",
     "configstore": "^5.0.1",
-    "connect": "^3.7.0",
+    "connect-ensure-login": "^0.1.1",
+    "cookie-parser": "^1.4.5",
+    "cors": "^2.8.5",
+    "ejs": "^3.1.5",
```
CMCDragonkai: What are you using this for?

package.json
```diff
@@ -54,31 +54,73 @@
...
+    "bunyan": "^1.8.14",
+    "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
     "chalk": "^4.0.0",
     "commander": "^5.1.0",
     "configstore": "^5.0.1",
-    "connect": "^3.7.0",
+    "connect-ensure-login": "^0.1.1",
+    "cookie-parser": "^1.4.5",
+    "cors": "^2.8.5",
```
CMCDragonkai: Because this is not a web app, we can just set our access control to be `*` for everything. Therefore a specialized `cors` library should not be necessary. Check the MDN docs for this.

CMCDragonkai:
This PR seems to be still in-progress, I see a lot of `console.log` still. I think it's more efficient if we simplify this according to the comments above, and then merge, and start on the cleanup PR, because my review from the previous PR still stands.

package.json
robert-cronin: I shall take all these packages out, they creeped in from following an online tutorial for swagger. Haven't had a chance to remove them yet.

package.json
robert-cronin: nope this is just dependency creep, I will take it out

package.json
robert-cronin: since were now only using bearer tokens and client password flows (I left this one in since it is just another way to get a token and is nothing like the other ways of doing oauth like facebook or github logins.) I will remove the others

package.json
robert-cronin: this was for providing a login page for the authentication code flow, but that is no longer relevant and these will dissappear :+1:

package.json
robert-cronin: same here, this will be removed

robert-cronin:
control on the cli is now done via `pk oauth...`

robert-cronin:
OAuth2 tokens are now using JWT and API is on HTTPS:
Polykey with JWT and HTTPS

robert-cronin:
All 80 tests updated and successful on local build :+1:  ready for review/merge

Issue #66, opened 2020-08-03
npm release in gitlab-ci doesn't work


robert-cronin:
I am getting the following error when trying to deploy automatically with `npm publish`:
```
npm ERR! code E404
npm ERR! 404 Not Found - PUT https://registry.npmjs.org/js-polykey - Not found
npm ERR! 404 
npm ERR! 404  'js-polykey@0.0.10' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

robert-cronin:
might have some luck here: https://docs.gitlab.com/ee/user/packages/npm_registry/#uploading-packages

CMCDragonkai:
Pretty sure it should just be `polykey`, not `js-polykey` right? The `js-` prefix was only intended to disambiguate for our libraries with the same name on Github.

However, we may be using our vendor prefix `@matrixai` for many other things.

I don't think `Polykey` main app can be distributed on npm.

robert-cronin:
there was an issue with having the pname the same as one of the dependencies in nixos. it's not case sensitive either so it's not a simple matter of having polykey for the lib via npm and PolyKey for the GUI via nixpkgs

robert-cronin:
@matrixai/polykey should disambiguate. Shall I rename it to that then and deprecate [js-polykey](https://www.npmjs.com/package/js-polykey)?

CMCDragonkai:
Yes

On 12 August 2020 03:05:24 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@matrixai/polykey should disambiguate. Shall I rename it to that then
>and deprecate [js-polykey](https://www.npmjs.com/package/js-polykey)?
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/MatrixAI/js-polykey/issues/66#issuecomment-672098469

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

robert-cronin:
Closing on account of migration to gitlab

PR #73, opened 2020-08-11
Add Environment Variable Interface

### LINKED ISSUE #67: Environment variables interface
## Environment Variables

So to give an example of environment variables.

In any Unix shell you can set environment variables:

```
x=3
echo "$x"
```

You can expose these variables to subprograms:

```
export x=3
echo "$x"
bash -c 'echo "$x"'
```

However subprograms CANNOT set environment variables in the parent process. Which is the current shell.

So how would polykey help in the process of setting environment variables?

There are 2 ways:

1. Is for the `pk` command to spit out instructions to set environment variables in the shell. This requires the user to copy the command. This I've seen in other programs, but it's not really a nice UX.
2. Is to do what `env` does. The `env` command basically is a "fork+exec" at the shell-level but with the additional ability to set environment variables.

Let's explore 2.

To do 2. it is possible to not even use `env` at all.

```
x=3 bash -c 'echo "$x"'
```

This ensures that the environment variable is set for only the subprocess. And it does not export it to every other subprocess. This is good because of principle of least privilege.

So why use `env`? It is because it has lot of extra options to deal with environment variable manipulation and also signal manipulation and also dealing multi-parameter shebangs.

So how does this apply to Polykey?

The second way for Polykey to interface via environment variables is to expose an `env` like interface.

Suppose I wanted to pass some secret into some program that uses environment variables.

I could run it with polykey like this.

```
pk secrets get K --run 'bash -c "echo $K"'
```

The above basically "chains" the secret coming out of polykey directly injecting as an environment variable into the command in `--run`.

See `nix-shell` for examples of this. The `nix-shell` has `[--command cmd] [--run cmd]`. The first is an expression of a shell command. The second is the ACTUAL program being executed. So `--run` should be equivalent to doing a `execvp`. Whereas `--command` should be equivalent to a `execv`. See: https://stackoverflow.com/a/20823413. So the main difference is that `--command` runs in the "shell", whereas `--run` runs the actual program.

So this would mean that we are embedding the ability to do `env`-like functionality inside the pk commands. However it isn't entirely necessary. It is possible to allow the user to do command substitution.

```
K="$(pk secrets get K)" bash -c 'echo "$K"'
```

However... we should probably care about UX, so let's add the ability to "run" something with a given context. But it should probably not be done in the trivial way I just explained above.

To deal with more complex cases, it may be better to have some "chording" interface. This means almost a small kind of APL-like language in the `pk` commandline. So you can do things like `pk secrets get K, get Y, get C... etc`.

There are so many possible ways to organize/format the output. The best way to do so, is think of it as an language or "formatting" language. So you could spit out JSON, you could spit environment variable setters:

```
K='3' Y='4' C='5'
```

And many possible others.

The point is, this is the "env" variable interface.

robert-cronin:
This PR will add a new functionality to the CLI and nothing to the agent or library. Particularly a new subcomand to the secrets:

```
pk secrets env some-vault:some-secret=SECRET_1
# enters an interactive shell with the secret 'SECRET_1' available as an env varaible
```
By default, the interactive shell is run via process.env.SHELL which I think is a little better than purely $SHELL because it is node specific and comes with those platform considerations already. Users can also customize the command that is executed with the `--command`. Also, the `--run` command can be used to execute a command and then exit.

Also among changes for this PR is a new environment variable setter when `get`ing secrets that the user can copy:
```
pk secrets get --env some-vault:some-secret
# export SOME_SECRET=<secretContent>
```
Fixes #67


robert-cronin:
Pipeline successful. This PR is now ready for review and merge.

CMCDragonkai:
Why is there so many docs changed? I wonder if docs should just be a CI/CD artifact that gets deployed into gh-pages or something.

robert-cronin:
> Why is there so many docs changed? I wonder if docs should just be a CI/CD artifact that gets deployed into gh-pages or something.

Docs have to be regenerated with every commit to point to the commit hash. This is obviously redundant for non-release commits. I'd like to put doc generation into CI/CD. Perhaps I do that now then? it would simplify the review process. The way I've seen it done before is push to a new branch called something like `gh-pages` and only push to that branch on new tags. Then github pages can reference that branch

CMCDragonkai:
Created a new issue for the docs problem #75. And yes I've used gh-pages to great effect. @nzhang-zh can help with configuring the gitlab CI/CD settings to have the right keys to commit/push to gh-pages branch. Not sure if that means a deploy key has to be used and whether that would be a Github API token.

robert-cronin:
Nice, setting environment variables for CI/CD is fairly straightforward, we just need to use a nix image with git installed for the building and pushing so our new custom nix image will come in handy. Just need a script for the building/ pushing process.

CMCDragonkai:
We wouldn't want to embed the secret in the image. It needs to be injected from the CI/CD settings.

robert-cronin:
I would add in a deploy phase for building web pack and publishing to npm but there is a road block in #66 that has to be solved before that happens.

robert-cronin:
With the new PR #79 created for the separate docs issue, I think there is nothing holding this PR back from being merged.

src/cli/Secrets.ts
```diff
@@ -1,34 +1,47 @@
+import process from 'process';
 import commander from 'commander';
-import { actionRunner, pkLogger, PKMessageType, determineNodePath } from '.';
+import { spawn, ChildProcess, exec } from 'child_process';
 import { PolykeyAgent } from '../lib/Polykey';
+import { actionRunner, pkLogger, PKMessageType, determineNodePath } from '.';
+
+
+const pathRegex = /^([a-zA-Z0-9_-]+)(?::)([a-zA-Z0-9_:-]+)(?:=)?([a-zA-Z_][a-zA-Z0-9_]+)?$/
```
CMCDragonkai: Does this regex parse all kinds of unix paths? Remember that unix paths allow spaces and newlines as well.

src/cli/Secrets.ts
```diff
@@ -1,34 +1,47 @@
...
-        // List secrets
-        if (secretNames.length == 0) {
-          pkLogger(`no secrets found for vault '${vaultName}'`, PKMessageType.INFO);
-        } else {
-          if (isVerbose) {
-            pkLogger(`secrets contained within the ${vaultName} vault:`, PKMessageType.INFO);
+        const vaultNames: string[] = Array.from(options.args.values());
+
+        if (!vaultNames.length) {
+          throw Error('no vault names provided')
```
CMCDragonkai, line 26: We need to create a new exception hierarchy for the library and CLI commands. This will allow structured error handling for Polykey. Should be a new issue.

src/cli/Secrets.ts
```diff
@@ -128,37 +156,129 @@ function makeRemoveSecretCommand() {
...
   return new commander.Command('secrets')
     .description('manipulate secrets for a given vault')
     .addCommand(makeListSecretsCommand())
     .addCommand(makeNewSecretCommand())
     .addCommand(makeUpdateSecretCommand())
-    .addCommand(makeRemoveSecretCommand())
-    .addCommand(makeGetSecretCommand());
+    .addCommand(makeDeleteSecretCommand())
+    .addCommand(makeGetSecretCommand())
+    .addCommand(makeSecretEnvCommand());
```
CMCDragonkai, line 281: I think we should abstract alot of the CLI commands into separate files and modularize them. These files like `cli/Vaults` or `cli/Secrets` can act like top level sub commands with each handler being something lower level. I'm a big fan of MVC style, at least for the controllers themselves that can be instantiated with DI & IoC container. That way commands can share state too. And potentially we can compose commands later in the future.

The Haskell onion layering may be a good idea. https://www.parsonsmatt.org/2018/03/22/three_layer_haskell_cake.html and https://news.ycombinator.com/item?id=12604767

Note that these are architectural issues to be resolved later in a separate issue.

CMCDragonkai (commented):
Becareful with adding any files to the git. Make sure the `.gitignore` is taken from our template. Although I haven't been maintaining the templates for TS work, but instead Python and Haskell.

CMCDragonkai:
Ok we merge this first, and then we merge #60. I think #60 should be merged sooner than later, so we can have a proper test harness for subsequent extension.

All the comments above should be turned into new issues to resolve so it doesn't block this PR.

CMCDragonkai:
Conflicts must be resolved. Then merge away. The number of PRs is increasing!

CMCDragonkai:
And also comments need to be turned into issues.

src/cli/Secrets.ts
robert-cronin: I agree, this needs some work. I've changed it a little to also be more compatible with multiplatform scenario (windows doesn't like `:` and I've added in spaces to the regex) 
This probably has more work to go before it captures all scenarios though, so I will add it in as an issue.

robert-cronin: https://github.com/MatrixAI/js-polykey/issues/81

src/cli/Secrets.ts
robert-cronin: this is also linked to https://github.com/MatrixAI/PolyKey/issues/51

robert-cronin: https://github.com/MatrixAI/js-polykey/issues/82

src/cli/Secrets.ts
robert-cronin: https://github.com/MatrixAI/js-polykey/issues/83

CMCDragonkai:
Nice.

PR #118, opened 2020-10-13
Add Cleanup, Validation and Verification code

### LINKED ISSUE #77: Polykey Agent Design
I've always disliked it when programs create their own subshell/repl, it makes it difficult to integrate. This is OK, if the subshell is presenting a totally different world with different semantics, but with Polykey CLI it should just present a FS interface, and standard Unix tooling can work on it easily.

So if we are not creating a sub-shell within the normal shell, we need to instead integrate our "session" into the normal shell.

So that means that during our installation we also end up installing an agent. This is only needed during our CLI phase. However later I want re-use that agent for the GUI as well. So then it acts as a like taskbar agent in Windows and Mac, for Linux it would a user-level service. I would expect that the client and agent both integrate the js-polykey library.

So what does that mean? Well, it is possible for a single desktop to be running multiple keynodes. You can have a keynode in `~/.polykey1` and another keynode in `~/.polykey2`. No problem with this.

Then the agent for now can keep the "login" session for each keynode that it knows about. So a user-level agent keeps track of keynodes for that user on Unix system (or Windows). But a root level agent would keep track of keynodes on the entire machine. But one could also have an agent within a nix-shell keeping track of just keynodes within its domain. So basically multiple agents may exist at the same time (this will make it easier to debug the agent and use it in infrastructure).

So from this point onwards, every polykey operation requires the client which is the `polykey` app to be authenticated. This means either asking for the password to unlock the stored private key, or asking for the private key if that is not stored.

However like the `sudo` command, as soon as we have authenticated once for any operation, we can then save that authentication to the polykey agent such that subsequent operations do not need to be reauthenticated.

We can do the above automatically or on the basis of some switch. But for prototyping purposes, let's do this automatically.

The session can have a time-expiry, and this can be refreshed on every operation. We can this configurable, but for now let's just fix to 1 hr.

The way in which polykey clients connect to the agent can be done similar to how ssh-agent or gpg agent is utilised. A standard environment variable pointing to some sort of socket that exists somewhere. See `SSH_AUTH_SOCK`.

Note that developing this agent also leads to interesting ideas about agent forwarding similar to how SSH works. However I want to leave this out for now, but it does pertain to another integration point, where in this the integration point is with SSH.

### LINKED ISSUE #90: Optimizing IPC communication between agent and client (CLI/GUI)
At the moment, the IPC communication is "bulky" (and not efficient?) and is not using a standard RPC system.

If we change the communication between Agent and Client to gRPC, we have to beware of the security implications: https://github.com/MatrixAI/js-polykey/issues/50#issuecomment-677600761 (basically we have to use mTLS to secure it).

The advantage is that clients can then be operating over the network as well.

### LINKED ISSUE #96: xstate model for communication transitions
We can use xstate to model the communication transitions in polykey.

linked comment: https://github.com/MatrixAI/js-polykey/pull/84#discussion_r486280990

### LINKED ISSUE #97: Add a DHT for Content Discovery
The DHT will mostly be useful for discovery of nodes that do not have social discovery like automated child nodes of a real person's primary keynode.

It also probably won't be useful until we have a large number of nodes like a large federated hierarchy which becomes unruly to manage if machine nodes are progenitors to other machine nodes. Also, as far as I can tell, DHT is mostly used for content discovery and since we are focused primarily on a tight security model, there is probably no way to make content discovery work well unless it is all in a trusted network with lots of transitive permission.

### LINKED ISSUE #98: improved domain modelling
We have a reasonable separation of concerns with good domain modelling, but dependency visualisation with arkit has shown up some overly dependent connections to `PeerManager`:
dependencies

Alot of these connections are to classes that should be managed by `peerManager` directly. This can be achieved with inversion of control. We can provide functions in the child classes that peer manager can manipulate in order to achieve the same result.

### LINKED ISSUE #100: Support all NAT types for hole punching
At the moment we have a UDP hole punching implementation that covers full-cone NAT as all it's doing is mapping an internal host:port address to an external host:port address. This is a weak implementation as any actual use of polykey is likely to find itself behind either a port-restricted cone NAT or symmetric NAT. 

Any hole punching solution probably won't work with symmetric NAT so the relay can be used in that case. The main thing missing from our implementation to support port-restricted cone NAT is to have the intermediary node send a connection request to the NATted node with the other private node's public address. Section 3.2 on [this](https://bford.info/pub/net/p2pnat/) guide is useful for understanding.

### LINKED ISSUE #102: Break up PolykeyAgent and PolykeyClient classes based on domain models
Right now there are 2 large files; one for the agent and one for the client. This is not very cohesive and needs to be split into separate files based on some good domain modelling.

### LINKED ISSUE #103: Abstract all NAT traversal away from the user
At the moment we have a few methods to control setting up the relay and udp hole punch (i.e. requesting it from a sepcific peer) but this should be done automatically without the users help.

Also, the `relayPeerPublicKey` should be removed from the `PeerInfo`

### LINKED ISSUE #107: Quality of life improvements
There are 4 main points to this as outlined in https://github.com/MatrixAI/js-polykey/pull/84#issuecomment-691671347 :
1. Quality of Life improvements and abstraction/modelling improvements and design improvements and standardisation and Quality Assurance stuff.
2. Proper model based testing and property testing using xstate-test and fast-check.
3. Spec sheet that can derive UML diagrams... and so on. Need to see how it all fits together.
4. Nix based integration testing.

### LINKED ISSUE #108: Kademlia DHT for multi-hop relay/NAT traversal
This is kind of a cleanup as well as an extension of existing features. At the moment we have a NAT traversal based on public peers but a polykey node can have only one relay peer set. We want this to be completely abstracted away from the user.

Kademlia will provide this ability. Any node can act as a relay peer. Kademlia is usually used for content routing but in our situation, the content will be public keys of polykey nodes.

### LINKED ISSUE #111: Add a proper logging system to the CLI (handle verbosity levels)
We need to handle verbosity levels properly, at the moment the verbosity flags are not being used.

We can take a look at the [winston](https://www.npmjs.com/package/winston) library for logging.

### LINKED ISSUE #114: Add json schema for request/response components in openapi spec
In order to aid validation, we can replace the existing component definitions in `openapi.yaml` with json schemas, for example:
```
openapi: 3.0.1
info:
  title: ...
  description: ...
  version: 0.0.1
servers:
- url: /
tags:
- name: root
  description: Default namespace
paths:
  /ping:
    get:
      tags:
      - root
      responses:
        200:
          description: Success
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pong'
components:
  schemas:
    Pong:
      $ref: 'components/schemas/Pong.json'
```
and for `Pong.json`:
```
{
  "$schema": "http://json-schema.org/draft-00/schema#",
  "type": "object",
  "properties": {},
  "required": []
}
```
Note we have to use draft-00 as that is what openapi supports.

### LINKED ISSUE #119: Windows Default Keynode Path should be AppData/Roaming
The AppData/Roaming is the official location for application data that is moved with the user profile of the OS.

It makes sense to default it to AppData/Roaming when on Windows.

But on Linux, you should also check for the `$XDG_...` environment variables. Because there is a specific location for linux desktop now: https://wiki.archlinux.org/index.php/XDG_Base_Directory

That only applies for linux with desktop environment.

I don't what the standard is for mac, so just look around.

robert-cronin:
- [x] Fixes #77 
- [x] Fixes #90 
- [x] Fixes #96
- [x] Fixes #98
- [x] Fixes #102  
- [x] Fixes #103 
- [x] Fixes #107 
- [x] Fixes #108 
- [x] Fixes #111 
- [x] Fixes #114
- [x] Fixes #119


CMCDragonkai:
Fixing and having proper testing for the NAT traversal is important here for our 31st October target. That way by November we should all be dogfooding Polykey and using it to share secrets. And with @nzhang-zh integrating it into the Matrix OS, that means we can use it to handle our secrets for the Ouroboros experiment!!!!

robert-cronin:
going ahead and merging :+1:

Issue #128, opened 2020-11-11
Backup Key File In Case User Forgets Keynode Passphrase


robert-cronin:
make a key file for backup for unlocking the keynode - currently we only have root keypair (i.e. passphrase ). However the user can forget his passphrase and its better that we have a backup keyfile that could still unlock the keynode
@gideonairex

robert-cronin:
I am actually not sure that kbpgp does anything like this natively but we can achieve similar functionality by storing a mirrored version of the vault keys in a separate file that is somehow protected by that backup key file

robert-cronin:
I will add this into the next PR as I think it will be relatively straight forward to implement.

CMCDragonkai:
We have a recovery code system similar to bip39 mnemonics. That should be sufficient to solve this problem. See the screenshots that @faithveloro has worked on.

faithveloro:
Attaching screenshots here: 
![✳️ Desktop - Step 3 1_ Backup or Recovery Code](https://user-images.githubusercontent.com/46587577/101643995-a9d4ab80-3a6f-11eb-863f-597d7b8160b1.jpg)
![✳️ Desktop - Step 3 2_ Backup or Recovery](https://user-images.githubusercontent.com/46587577/101644029-b22ce680-3a6f-11eb-93c3-090da4f5b677.jpg)
![✳️ Desktop - Step 3 3 1_ Backup or Recovery](https://user-images.githubusercontent.com/46587577/101644057-b953f480-3a6f-11eb-997f-8eab5fb44f13.jpg)
![✳️ Desktop - Step 3 3 2_ Backup or Recovery](https://user-images.githubusercontent.com/46587577/101644071-bc4ee500-3a6f-11eb-8032-dc8f376efcd0.jpg)
![Desktop - Step 3 3 3_ Backup or Recovery Success](https://user-images.githubusercontent.com/46587577/101644089-bf49d580-3a6f-11eb-8d55-c709364fdf3f.jpg)

robert-cronin:
[...]

Issue #113, opened 2020-09-30
Create the link between root keypair (kbpgp) and CA keypair (node-forge)


robert-cronin:
As per https://github.com/MatrixAI/js-polykey/pull/94#issuecomment-700626997, there are a few options for doing this:

1. Start with the root keypair and then use a KDF to generate the root x509 certificate pair.
2. Start with the root x509 certificate pair, and then use a KDF to generate the root keypair
3. Use the public-key of the root keypair as the actual public key of the x509 certificate
4. Use the x509 certificate and private key AS the root keypair

robert-cronin:
within node-forge, we can do something like this to make the x.509 certificate the one and only piece of sharing needed for peer to peer contact. For example the kbpgp public key and peerAddress can be encoded in the public cert via attributes:
```
const attrs = [
  {
    name: 'commonName',
    value: 'matrixai',
  },
  {
    // kbpgp public key
    type: '1.3.1.4.1',
    value: '<publicKey>'
  },
  {
    // peerAddress
    type: '1.3.1.4.2',
    value: '127.0.0.1:1314'
  }
];
certificate.setSubject(attrs);
```
where `1.3.1.4.1` and `1.3.1.4.2` are custom OIDs that we can use specifically for encoding polykey specific attributes and this works just fine in node-forge

robert-cronin:
Closing on account of migration to gitlab

PR #130, opened 2020-11-12
Fix: UDP direct hole punching, high CPU usage and misc.

### LINKED ISSUE #120: keypair generation command/pregenerating a keynode
We might need to pre-generate a keynode and move it somewhere else (e.g. generating a keypair on a local machine and putting it in metadata to start an EC2 instance with). Currently there is only `pk agent init` command that also starts polykey and thus also automatically generates all the other folders needed to run polykey (.peers, .pki etc). Instead we just want to generate the keynode as that is all that is requried by polykey for initialization.
Another nice feature would be to automatically link the running keynode to the keypair just generated.

### LINKED ISSUE #121: unlock keynode upon start command
the `pk agent start` command only starts the agent, but it should also unlock the node when starting. This requires an extra parameter to the start command of `--private-passphrase`

### LINKED ISSUE #123: files in hidden folders should not be hidden as well

### LINKED ISSUE #126: addPeer should return the peerId derived from the publicKey
at the moment the addPeer command only returns a boolean but it should instead return the peerId derived from the publicKey for later identification purposes.

### LINKED ISSUE #127: NAT traversal for the case of only 2 nodes and 1 is private
At the moment the NAT traversal is only designed for the case of one public node and 2 private nodes but we should also consider the case where we have one public and one private node.

### LINKED ISSUE #129: CLI - process.exit(0) interferes with spawn command for `pk secrets env`
We have a process.exit(0) command in the `actionRunner` for the CLI and this was to prevent the CLI from hanging from any reason when the commands were completed but this is now causing the `pk secrets env` to stop working

### LINKED ISSUE #132: Set an alias for a particular peer
at the moment, the only way to identify a peer in the store is via a hash string of the public key (peerId), but this is not very user friendly. A great idea from @gideonairex is to extract the userId from the public key and set that at the top of each PeerInfo to be browsed by the user or even add in the ability to set a local alias for a particular peer.
This might be accomplished through the use of an extra map in PeerManager. If we wan't to keep the alias local, then it shouldn't be added to the PeerInfo as that might be shared to others.
So our `aliasStore` might be a mapping of peerIds to alias strings.

### LINKED ISSUE #133: Split client and agent tls credentials into separate files
currently they are in the one config.json file but Agent (and probably other clients) require them to be in separate files.
@nzhang-zh

### LINKED ISSUE #134: BooleanMessage should be replaced with just thrown errors in Agent.proto
The BooleanMessage is being used for success flags in the agent/client gRPC communication channels but this is redundant since if there is any error that stops the completion of a command, it is thrown and picked up by the client side error handler. 
We will replace these BooleanMessage's with EmptyMessage

### LINKED ISSUE #135: High CPU Usage While Idling
So something noticed by both @nzhang-zh and myself is a high CPU usage while the polykey-agent is in idle or even while locked. I think I have traced most of this to interval network requests in the multicast broadcasting and peer DHT querying. As suggested in https://github.com/MatrixAI/js-polykey/pull/130#issuecomment-732703877, we could do expontential backoff for peerDHT querying if peers being contacted aren't responding. For the multicast broadcast, we can either increase the interval, but this one was less impactful. We also have a stealth mode if the user wants to completely stop these intervals for a time.

robert-cronin:
This PR is specifically intended to fix a bug with UDP hole punching so integration with Matrix-Agent can continue but will also fix some other important bugs as well:

- [x] Fixes #120 
- [x] Fixes #121 
- [x] Fixes #123 
- [x] Fixes #126 
- [ ] Fixes #127 
- [x] Fixes #129  
- [x] Fixes #132 
- [x] Fixes #133 
- [x] Fixes #134 
- [x] Fixes #135


robert-cronin:
After some testing with flamegraph, it turned out swagger-ui and a few loops were the cause of the idle CPU usage. The loops include multicast broadcasting and the peerDHT trying to connect to unconnected peers. The latter looks to be the biggest source of inefficiency so I think it should initially try to connect to those peers and if its not able to after maybe 3 tries, it stops trying unless the user tries to connect to it. The inspiration for this was libp2p's 'random walk' which allows a peer to maintain its position in the DHT by randomly querying peers every now and again.

Another optimisation might be to increase the interval time

robert-cronin:
This is with swagger-ui:
image
This is without:
image

CMCDragonkai:
Hmm is there a way to optimise the connection establishment itself? Does it involve alot of cryptography atm?

If it is an interval problem, we could do exponential backoff.

But another question is why are we trying to connect to unconnected peers atm?

On 24 November 2020 5:44:29 pm AEDT, Robbie Cronin <notifications@github.com> wrote:
>After some testing with flamegraph, it turned out swagger-ui and a few
>loops were the cause of the idle CPU usage. The loops include multicast
>broadcasting and the peerDHT trying to connect to unconnected peers.
>The latter looks to be the biggest source of inefficiency so I think it
>should initially try to connect to those peers and if its not able to
>after maybe 3 tries, it stops trying unless the user tries to connect
>to it. The inspiration for this was libp2p's 'random walk' which allows
>a peer to maintain its position in the DHT by randomly querying peers
>every now and again.
>
>Another optimisation might be to increase the interval time
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/130#issuecomment-732640680

-- 

CMCDragonkai:
Will this PR also fix the high CPU usage? I didn't see an issue for that.

CMCDragonkai:
Can you update the title of this PR since it seems to address alot more than UDP hole punching.

CMCDragonkai:
BTW, high cpu usage when idling is critical to mobile deployment. @gideonairex 

We need ways of hooking in events that occur on mobile like 4g vs wifi, battery vs powered... etc.

These would impact any background processing or discovery processes.

CPU usage for js-polykey is really important to optimize!

CMCDragonkai:
If @gideonairex has experience here, maybe he can help?

robert-cronin:
Updated title to the two major things being addressed by this PR.
As for the events on mobile, there are some plugins we can use in nativescript:

- powerinfo (i.e. battery vs charging): https://market.nativescript.org/plugins/nativescript-powerinfo/
- connectivity info (4g vs wifi): https://docs.nativescript.org/ns-framework-modules/connectivity

CMCDragonkai:
Cool so the plugins will be useful to trigger those events, but then js-polykey needs the hooks as "governor" to govern all of its loops so that the user which is Electron Polykey or Native Script Polykey can know how to optimize the usage of js-polykey.

By "governor" I mean this:

> https://en.wikipedia.org/wiki/Governor_(device)

CMCDragonkai:
@robert-cronin ETA for this PR?

CMCDragonkai:
@gideonairex you should track this PR as well, looks like some things coming here will result in features in the frontend.

robert-cronin:
> @robert-cronin ETA for this PR?

I am hoping to be finished by Friday, still need to schedule a meeting with @nzhang-zh to ensure the fixes to the hole punching are working correctly

robert-cronin:
Went a bit over schedule, the UDP hole punching was a little more complicated than anticipated but is finally working over the public internet. To keep the code review more manageable, I will be breaking this PR up so that the already solved issues can be reviewed and merged. Issues #113, #128 and #131 will be split into their own PR for social discoverey and other UI/UX issues

CMCDragonkai:
@gideonairex Can you cross review this with @robert-cronin.

I will need you guys to review each others code in accordance with coding guidelines in the orientation repository as well and make discussions available here too.

I won't always have time to review all the code.

CMCDragonkai:
@robert-cronin Make sure to get `lint` and `lintfix` working as well to ensure that all code is formatted.

See: https://gitlab.com/MatrixAI/open-source/js-polykey/-/pipelines/229949699

It's currently failing the build.

Also squash the commits into semantic commits.

CMCDragonkai:
BTW are the tests intended to work via the CI? Seems like it times out.

robert-cronin:
There is a discrepancy between the gitlab-runner on my local machine and the one actually on gitlab and I haven't managed to figure out why yet. I can have a look soon in the next PR hopefully.

robert-cronin:
So in the cleanup for this PR, I remembered a bug in the connection logic that I forgot to deal with; if the direct connection tries and fails (in the case of public to private connections) then the rest of the methods cannot be performed, I was getting around the bug before by commenting out the direct connection method so it would go directly to the hole punching.

The bug seems to be something to do with my custom promiseAny utility, it almost seems to be blocking the event loop for 10 seconds which is the default timeout set for connectDirectly method. I have a couple of things to try; first is the try to increase the timeout for the connectFirstChannel method so all options can be tried or reduce the timeout for the connectDirectly method so the other methods can have their try. Even though Node.js is async/await, the call stack is single threaded so it still has to empty before the event loop can pop the next task onto it (unless we use webworkers, but that seems a little overkill).

CMCDragonkai:
@robert-cronin is this still relevant?

CMCDragonkai:
Development occurs on gitlab now.

PR #210, opened 2021-07-13
Secret env command

### LINKED ISSUE #205: The `pk secrets env` command for meeting Development Environment Usecase
### Specification

1. The `pk secrets env` command should operate similar to the unix `env` command
2. We can use `pk secrets env` to do double duty: inject environment variables into a new subcommand, or allow sourcing of environment variables into an existing subshell
3. The task for the development environment usecase is specified in https://github.com/MatrixAI/Polykey-Desktop/issues/77
4. The `pk secrets env someprogram arg1 arg2` should do a proper process replacement when possible

### Additional context

Copied excerpt:

> Now when we want to start development on the project, we have 2 options:
> 
> 1. Open up a shell with Polykey sourcing the environment variables
> 2. Use Polykey to inject environment variables into a sub-shell
> 
> Either result is the same, because you'll be running a shell with environment variables set so that your test runs of your software can inherit this environment configuration. Let's demonstrate both.
> 
> In the first case, we will reuse the `source` command and ask `pk` to constuct the equivalent of the `.env` file but entirely in-memory and only for this specific usage. We must also use process substitution `<(...)` so that the `pk` command output can be redirected into a temporary file descriptor that is read by the `source` command.
> 
> ```sh
> # one secret
> source <(pk secrets env my-software-project:AWS_ACCESS_KEY_ID)
> 
> # multiple secrets
> source <(pk secrets env my-software-project:AWS_ACCESS_KEY_ID my-software-project:GOOGLE_MAPS_API_KEY)
> 
> # globbing style (you can use globstar as well, this will only export immediate files)
> source <(pk secrets env my-software-project:*)
> 
> # use the -e flag to export all variables
> source <(pk secrets env -e my-software-project:*)
> # use the -- to separate so you can export just one out of many
> source <(pk secrets env -- my-software-project:AWS_ACCESS_KEY_ID -e my-software-project:GOOGLE_MAPS_API_KEY)
> source <(pk secrets env -e -- my-software-project:AWS_ACCESS_KEY_ID my-software-project:GOOGLE_MAPS_API_KEY)
> ```
> 
> Now your shell has the relevant environment variables set by Polykey, and they will exist for as long as this current shell is alive.
> 
> In the second case, we ill use `pk secrets env` command to run a subshell, it can run any subprogram, but in the context of a development environment, you usually want a shell.
> 
> ```sh
> # one secret
> pk secrets env my-software-project:AWS_ACCESS_KEY_ID bash
> 
> # multiple secrets
> pk secrets env my-software-project:AWS_ACCESS_KEY_ID my-software-project:GOOGLE_MAPS_API_KEY bash
> 
> # globbing style
> pk secrets env my-software-project:* bash
> 
> # the -e flag has no effect when using it this way
> # this is because the subprogram determines whether to export variable or not
> # it usually exports the variable
> pk secrets env -e my-software-project:* bash
> ```
> 
> Now you have a subshell that has the environment variables configured, it will also automatically export it to child programs. When you have finished your development, you can just `exit` the shell, and the environment variables are gone!

### Tasks

1. Ensure that `pk secrets env` can be used to inject environment variables and run a subprocess
2. Ensure that `pk secrets env` can be used to source environment variables and output something that can be used as a file to be sourced by a shell
3. Deal with the fact that nodejs doesn't have an exec that can replace the current node process: https://github.com/nodejs/node/issues/21664
   * You can deal with this the way babel deals with it

> * Babel [attempts to use `kexec`](https://github.com/babel/babel/blob/master/packages/babel-node/src/babel-node.js#L87-L88) where available, which [is a POSIX-only module that replaces the process literally](https://www.npmjs.com/package/kexec). Absent that, it falls back to [its own implementation](https://github.com/babel/babel/blob/master/packages/babel-node/src/babel-node.js#L90-L109) that works like the other two examples.

scottmmorris:
### Description
This PR aims to refactor the existing `pk secret env` command according to the specifications outlined in the Development Environment Usecase. Essentially the command should be able inject environment variable or allow sourcing variables into an existing subshell.

### Issues Fixed
* Fixes MatrixAI/Polykey-CLI#31 

### Tasks
1. [x] Ensure that `pk secrets env` can be used to inject environment variables and run a subprocess
2. [x] Ensure that `pk secrets env` can be used to source environment variables and output something that can be used as a file to be sourced by a shell
3. [x] Handle export flag
4. [x] Handle globbing options specified in the usecase
5. [x] Handle command options with spaces (may extend to other secret and vault CLI commands)
8. [x] Extending globbing to match documentation here (https://git-scm.com/docs/gitignore)
6. [ ] Testing for `secret env` command executing a subshell
7. [ ] Deal with the fact that nodejs doesn't have an exec that can replace the current node process

### Final checklist

* [ ] Domain specific test (`npm test -- tests/bin/secret.test.ts`)
* [ ] Full tests with `npm test`
* [ ] Lint fixed
* [ ] Squash and rebased


scottmmorris:
Ive also changed the naming for `pk secrets create` to be rpimarily `pk secrets put` with an alias of `create`. The main functionality of this command has been added now as well

scottmmorris:
I've updated the description of this PR to include testing for the launching subshells with injected secrets. At the moment the standard jest testing that I have been writing wouldn't be able to deal with it but there should be a way to test this behaviour with jest. Roger has linked these packages https://github.com/jprichardson/node-suppose & https://github.com/nodejitsu/nexpect as references. I have validated the correct operation by running the command but it will still be better to include automated tests as well.

The other task left for this is optimisation using `kexec` which I haven't figured out yet. But this will be left until the end because it is an optimization feature.

Also the globbing currently only handles the single globstar `*`. So cases like `vault:*` or `vault:secretdir/*` will both work (getting all files in the base and secretdir directories respectively). Eventually this should be extended to mimic the documentation here (https://git-scm.com/docs/gitignore)

scottmmorris:
Globbing has now been extended to include both `*` and `**` functionality

CMCDragonkai:
Which globbing library did you use? Make sure that it is well supported 
and ideally in typescript.

scottmmorris:
At the moment I wrote my own code to handle globbing. A lot of the globbing libraries I looked at use a cwd to return output which is difficult with the EncryptedFS. [This](https://www.npmjs.com/package/fast-glob) one allows you to supply an fs and is up-to-date but that would require a bit more work in creating a grpc call to handle this and returning the files. Alternatively, [this](https://www.npmjs.com/package/glob) library was another popular one but doesn't accept any other filesystems as far as I am aware of

scottmmorris:
After trying the first library it seems that EFS is incompatible with the fs implementation specified by `fast-glob`

```
No overload matches this call.
  Overload 1 of 2, '(source: string | string[], options: Options & EntryObjectPredicate): Promise<Entry[]>', gave the following error.
    Type 'EncryptedFS' is not assignable to type 'Partial<FileSystemAdapter>'.
      Types of property 'readdirSync' are incompatible.
        Type '(path: PathLike, options?: { encoding: BufferEncoding; withFileTypes?: false; }) => string[]' is not assignable to type 'ReaddirSynchronousMethod'.
          Types of parameters 'options' and 'options' are incompatible.
            Property 'encoding' is missing in type '{ withFileTypes: true; }' but required in type '{ encoding: BufferEncoding; withFileTypes?: false; }'.
```

CMCDragonkai:
The fs that can be supplied would be the VFS instance, not the EFS 
instance. Would fast-glob work in that case? I'm working on EFS 
extending VFS, so then EFS instance will be a VFS instance anyway.

scottmmorris:
After some testing we have opted to introduce the `tiny-glob` source code into polykey. Have got the source code to work with vfs and have moved it into a utility function. I am creating the new grpc command now which will use this function.

CMCDragonkai:
We expect some edge case bugs with the globbing since the upstream repo has bugs already in it that hasn't been fixed (https://github.com/terkelg/tiny-glob/issues). But it's not critical so we move ahead.

As for bringing in the globbing code, make sure you're rewriting important parts with TypeScript, as the upstream `tiny-glob` is written in JS.

CMCDragonkai:
This requires a rebase on top of client-refactoring given that much has changed.

CMCDragonkai:
Sorry this was closed, it should have had its target changed to master.

Can you try to edit this @scottmmorris? If not, it should just be reopened again...

https://github.com/isaacs/github/issues/1557#issuecomment-511415762

Ok I think this PR must be "recreated" then.

Issue #200, opened 2021-07-05
Wrapper methods in GRPCClientClient to get parameter types of the wrapped GRPC call


tegefaulkes:
### Specification
Here is an example of a function from src/client/GRPCClientClient.ts:39:
```ts
  public vaultsCreate(...args) {
    if (!this._started) throw new clientErrors.ErrorClientClientNotStarted();
    return grpcUtils.promisifyUnaryCall<clientPB.StatusMessage>(
      this.client,
      this.client.vaultsCreate,
    )(...args);
  }
```
While the return type is explicit here as `grpcUtils.promisifyUnaryCall<clientPB.StatusMessage>`.
The type information for the parameters are completely missing. What do I pass to this? is it one parameter or many?
To find out I need to go look at the definition of `this.client.vaultsCreate`
```ts
public vaultsCreate(request: Client_pb.VaultMessage, metadata: grpc.Metadata, options: Partial<grpc.CallOptions>, callback: (error: grpc.ServiceError | null, response: Client_pb.StatusMessage) => void): grpc.ClientUnaryCall;
```

We should explicitly state the parameters and their types like so:
```ts
  public vaultsCreate(vaultMessage: clientPB.VaultMessage):Promise<clientPB.StatusMessage> {
    if (!this._started) throw new clientErrors.ErrorClientClientNotStarted();
    return grpcUtils.promisifyUnaryCall<clientPB.StatusMessage>(
      this.client,
      this.client.vaultsCreate,
    )(vaultMessage);
  }
```

### Additional context
This is used by the CLI and the Polykey GUI to handle comunication to the agent. it will be used extensively.  adding type information will make it:
- easier to write code for.
- proper compile time type checking by the typescript compiler.

As it curretly is error will only come up during runtime when calling the function.

### Tasks
1. [ ] Replace `...args` with the respective parameter and type for each function.

CMCDragonkai:
@tegefaulkes can you rewrite this issue with the new issue templates? Is this a design issue or development issue? It cannot be both.

tegefaulkes:
There still are no templates when creating a new issue. I took this one from the .github development template.
The table at the top doesn't seem to be working. Also, why is there a name and title in it?

CMCDragonkai:
That doesn't happen for me. Go to the GitHub issues and click new issue. You'll see it.

On 6 July 2021 11:21:56 am AEST, Brian Botha ***@***.***> wrote:
>There still are no templates when creating a new issue. I took this one
[...]

PR #275, opened 2021-10-27
GRPC API Review Changes

### LINKED ISSUE #249: Transport Agnostic RPC
We have reviewed the gRPC API, and worked with it extensively. It's time to work out a better RPC layer for PK. There are several problems to to consider here:

* Cohesion
  - https://developers.google.com/protocol-buffers/docs/proto3
* Pagination
  - #237 
  - https://gitlab.com/MatrixAI/Engineering/Polykey/polykey-design/-/issues/22
* Streaming and unification with pagination
  - Client streaming can support some client usecases where only the client process has the data
* Authentication & Sessions - usage of metadata and bearer & basic tokens
  - #240
* Client and Server Interception (Middleware)
  - How client interception works: https://github.com/MatrixAI/js-polykey/pull/296#issuecomment-989740159
  - https://github.com/grpc/grpc-node/issues/419
  - Client interceptors are available and are used in `src/client`
  - Server interceptors are not yet available, so instead we use DIed functions like `authenticator`
* CQRS in order to maintain consistency between different clients
  - https://gitlab.com/MatrixAI/Engineering/Polykey/polykey-design/-/issues/23
* Naming Standardisation
  - #218
* HTTP API portability and Web Gateway Portability
  - #166
  - #235
  - https://github.com/deeplay-io/nice-grpc/issues/25
* Error handling
  - https://www.grpc.io/docs/guides/error/
  - We serialize our PK exceptions and then pass them to the client where it deserializes exceptions
  - Do we have to consider backwards compatibility here? What happens if the agent side serializes an exception that doesn't exist on the client side code, the client side should gracefully handle this by throwing a generic exception that indicates the client doesn't understand the exception (but provide exception details)
  - Resolve trailing metadata usage and sensitive information being exposed in exceptions https://github.com/MatrixAI/js-polykey/issues/249#issuecomment-969850137
* Typescript Compatibility
  - #200 
* API versioning
  - How should PK API be versioned? Should we always try to maintain 100% backwards compatibility, or have some way of versioning incompatible changes?
  - How is a version signaled in gRPC? In REST this can be done in the URL or in the header metadata
  - Guide from Microsoft https://docs.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-5.0
  - GRPC service version would be independent of the js-polykey version which is independent from the state version, which would give us 3 versions to manage:
    * pk source version - updated when source code changes, minor is new features, major is any new incompatibility from anywhere: cli, service, node state... etc, e.g. `1.0.0 -> 1.1.0 -> 2.0.0 -> 2.0.1` 
    * grpc service version - only major version and only changes on breaking changes, e.g. `1 -> 2 -> 3`
    * node state version - only major version and only changes on breaking changes to the node state, e.g. `1 -> 2 -> 3`

We have 2 main proto files:

1. `proto/schemas/Client.proto`
2. `proto/schemas/Agent.proto`

And a `Test.proto` as well, this will need to be used to generate the marshaling code.

### Additional context

* #155 - compatibility with Mobile operating systems
* #166 - having a transport agnostic RPC will make it easier for third party integration as it's possible to extend the different transport options to communicate with the PK agent - such as TCP, and UDP... etc
* #235 - would not be a problem anymore if a web-based transport is enabled, either that or if electron directly bridges into the nodejs runtime (rather than having the FE call into a BE proxy) - this is also relevant to any browser-extensions
* #248 - some RPC functionality is intended to support asynchronous API, should consider these design requirements for notifications
* #234 - our RPC should be compatible on the transport layer for P2P communication, P2P hole punching occurs underneath the RPC layer, so the P2P side has to bootstrap from a lower level communication protocol which is purely message oriented
* #400 - the RPC should be agnostic to IPv4 or IPv6
* #297 - RPC mechanism must be cancellable ideally using cancellable promises, which includes cancelling the underlying side effect
* #243 - it's a good idea for the RPC to have timeouts on its calls and to handle such timeouts, in particular we would want to be able to create "custom" protocols on top of any streams, or better would be to "lift" such protocols into the the underlying RPC system such as the nodes claiming process
* https://exploringjs.com/nodejs-shell-scripting/ch_web-streams.html - Explanation of how to use Web Streams

### Tasks

1. [ ] - Review the proto3 guide, naming standard with reference to the proto files, look for ways to abstract or clean up the data
   * #279
   * [x] - Version the `Client.proto` and `Agent.proto` with version names and test out multiple-version services, and find out what the client does when the version wanted is not available
   * [ ] - Test out grpc reflection for versioning if necessary
   * [ ] - Update all the primitive types, and any usage of 64 bit int should have the string hint
   * [ ] - Add some benchmarks to the grpc, and check if `MAX_CONCURRENT_CONNECTIONS` is used
   * [ ] - Clean up types #200 or gRPC abstractions based on new grpc libraries
   * [ ] - Consider how to shutdown the grpc server and terminate all client connections as well
2. [ ] - Figure out CQRS, pagination and streaming
   * [ ] - Clean up js-pagination client and server side utilities
   * [ ] - Pagination on unary calls, and streaming calls that can be initialised with a pagination parameters (which would enable some sort of CQRS and event sourcing concepts)
3. [ ] - Review in reference to authentication and sessions
   * [ ] - Fix Exceptions with Sensitive Data and Trailing Metadata Usage https://github.com/MatrixAI/js-polykey/issues/249#issuecomment-969850137
4. [ ] - Check the HTTP API and web gateway compatibility
5. [ ] - Deal with the fact that the new grpc-js `1.4.1` version has some incompatibility: `TypeError: http2Server.on is not a function`.

tegefaulkes:
### Description
This PR is for changes relating to the API review discussion.

### Issues Fixed
Fixes #249 

### Tasks
1. [ ] - Review the proto3 guide, naming standard with reference to the proto files, look for ways to abstract or clean up the data
   * [ ] - Version the `Client.proto` and `Agent.proto` with version names and test out multiple-version services, and find out what the client does when the version wanted is not available
     * [ ] - create a new version for a service in the protobuf definitions.
     * [ ] - run both services at the same time.
     * [ ] - test communicating with a V1 service.
     * [ ] - test communicating with a v2 service.
     * [ ] - test communicating with a "removed" service.
   * [ ] - Test out grpc reflection for versioning if necessary
   * [ ] - Update all the primitive types, and any usage of 64 bit int should have the string hint
   * [ ] - Add some benchmarks to the grpc, and check if `MAX_CONCURRENT_CONNECTIONS` is used
   * [ ] - Clean up types #200 or gRPC abstractions based on new grpc libraries
   * [ ] - Consider how to shutdown the grpc server and terminate all client connections as well
2. [ ] - Figure out CQRS, pagination and streaming
   * [ ] - Clean up js-pagination client and server side utilities
   * [ ] - Pagination on unary calls, and streaming calls that can be initialised with a pagination parameters (which would enable some sort of CQRS and event sourcing concepts)
3. [ ] - Review in reference to authentication and sessions
4. [ ] - Check the HTTP API and web gateway compatibility
5. [ ] - Deal with the fact that the new grpc-js `1.4.1` version has some incompatibility: `TypeError: http2Server.on is not a function`.

### Final checklist
* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased


CMCDragonkai:
@tegefaulkes please copy the tasks from the #249 into this and flesh them out as well.

tegefaulkes:
OK, so the way we can include the `google.protobuf` types is pretty simple. the definitions are bundled with the `protoc` releases. for now i'm copying the google folder from the `protoc` release in `src/proto/schema/google`. from here we can just import in in a proto file with `import "google/proto/message.proto".

tegefaulkes:
With the `Any` message type we can embed any message in that field.  to create an Any message we can use the `.pack()` method. we provided it serialized binary form of a message `echoMessage.seralizeBinary()` and name to identify the message. `'common.EchoMessage'`.  we can retrieve the message using the `.unpack()` method. we provide it the `DeseralizeBinary()` for the expected message and the identifier name for the message. If the name matches the message name then the message is returned otherwise null is returned.

```ts
  const anyMessage = new Any();
  const echoMessage = new messages.common.EchoMessage();
  anyMessage.pack(echoMessage.serializeBinary(), 'common.EchoMessage');

  const test2 = anyMessage.unpack(
    messages.common.EchoMessage.deserializeBinary,
    'common.EchoMessage',
  );
  console.log(test2); // Is a EchoMessage
  const test3 = anyMessage.unpack(
    messages.common.StatusMessage.deserializeBinary,
    'common.StatusMessage',
  );
  console.log(test3); // Is null
  ```

The only use for this that comes to mind is a wrapper message for the NAT relay.

**Comment by tegefaulkes** [comment]
`Timestamp` message is pretty simple. we can use it to send date info without converting to nanoseconds first.
```ts
  const timeStampMessage = new Timestamp();
  timeStampMessage.fromDate(new Date());

  console.log(timeStampMessage.getNanos());
  console.log(timeStampMessage.getSeconds());
  console.log(timeStampMessage.toDate());
```

CMCDragonkai:
Our JS dates don't have nanoseconds, only up to milliseconds.

tegefaulkes:
The `Struct` type allows us to send primitive POJOs. this might be pretty useful.
```ts
  const testObject = {
    1: 1,
    "two": 2,
    "three": "Three",
  }
  const structMessage = Struct.fromJavaScript(testObject);

  console.log(structMessage.toJavaScript());
  // { '1': 1, three: 'Three', two: 2 }
```

tegefaulkes:
for the service versioning I need to..
1. create a new version for a service in the protobuf definitions.
2. run both services at the same time.
3. test communicating with a V1 service.
4. test communicating with a v2 service.
5. test communicating with a "removed" service.

I occurs to me that we only really need to version the Agent service. The client is bundled with the agent so the service the use to  communicate should match. However different versions of nodes should be able to communicate with each other.

CMCDragonkai:
I think a single service version is enough for both client and agent. Both client and service should be using the same version integer.

On 29 October 2021 10:36:06 am AEDT, Brian Botha ***@***.***> wrote:
>for the service versioning I need to..
>1. create a new version for a service in the protobuf definitions.
>2. run both services at the same time.
>3. test communicating with a V1 service.
>4. test communicating with a v2 service.
>5. test communicating with a "removed" service.
>
>I occurs to me that we only really need to version the Agent service. The client is bundled with the agent so the service the use to  communicate should match. However different versions of nodes should be able to communicate with each other. 
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/275#issuecomment-954299564
-- 

CMCDragonkai:
No the version number is only incremented on breaking changes. Meaning most of the time we will ensure backwards compatibility. This means when the version mismatch it means there is no possible communication.

On 29 October 2021 10:36:06 am AEDT, Brian Botha ***@***.***> wrote:
>for the service versioning I need to..
>1. create a new version for a service in the protobuf definitions.
>2. run both services at the same time.
>3. test communicating with a V1 service.
>4. test communicating with a v2 service.
>5. test communicating with a "removed" service.
>
>I occurs to me that we only really need to version the Agent service. The client is bundled with the agent so the service the use to  communicate should match. However different versions of nodes should be able to communicate with each other. 
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/275#issuecomment-954299564
-- 

tegefaulkes:
Re-based on master.

tegefaulkes:
I split off part of the tasks into it's own issue at #279 and PR at #280.

CMCDragonkai:
PR description to be updated since #280 got merged. Should also clear this PR and start off from master if there are no useful changes beyond what was done for #280.

CMCDragonkai:
@tegefaulkes before we forget what this PR was doing, I'm guessing the changes here were all prototyping and can be squashed into one to rebase?

tegefaulkes:
I think the only changes this has was prototyping using the google protobuf messages and versioning. AFAIK those changes already exist in the master branch. It can be squashed and re-based. Keep in mind there are a lot of conflicts with importing messages.

CMCDragonkai:
Going to close this for now, as the aim of #249 has changed to focus on replacing GRPC itself with something more generic. Plus this code is wildly outdated now.

Issue #274, opened 2021-10-26
Investigate and fix random `NodeGraph` test failures


joshuakarp:
### Describe the bug

The tests in `NodeGraph` can intermittently fail across runs. We'd experienced some similar failures in the past, and I thought I'd resolved them. The past failures were to do with some test utility functions that generate node IDs for a specific bucket index, such that they can be used for dynamic testing. They were previously causing failures by generating node IDs that overflowed into a different bucket index.

If this is still the cause of failure, now that we can use the 32 byte array as a node ID, I can potentially make these test utilities a bit more robust. Previously, they were just incrementing a character in the node ID to force an expected XOR value.

### To Reproduce

The test failures only occur randomly. It took me 5 runs of `NodeGraph.test.ts` to experience a failure in the tests. Some instances of these:

<details>

```
[nix-shell:~/Documents/MatrixAI/js-polykey]$ npm test -- tests/nodes/NodeGraph.test.ts 

> @matrixai/polykey@1.0.0 test /home/josh/Documents/MatrixAI/js-polykey
> jest "tests/nodes/NodeGraph.test.ts"

Determining test suites to run...
GLOBAL SETUP
 FAIL  tests/nodes/NodeGraph.test.ts (5.69 s)
  NodeGraph
    ✓ finds correct node address (11 ms)
    ✓ unable to find node address (6 ms)
    ✓ adds a single node into a bucket (5 ms)
    ✕ adds multiple nodes into the same bucket (16 ms)
    ✓ adds a single node into different buckets (6 ms)
    ✓ deletes a single node (and removes bucket) (6 ms)
    ✕ deletes a single node (and retains remainder of bucket) (14 ms)
    ✓ enforces k-bucket size, removing least active node when a new node is discovered (83 ms)
    ✓ retrieves all buckets (in expected lexicographic order) (13 ms)
    ✓ refreshes buckets (2050 ms)
    ✓ finds a single closest node (3 ms)
    ✓ finds 3 closest nodes (8 ms)
    ✓ finds the 20 closest nodes (92 ms)
    ✓ updates node (6 ms)

  ● NodeGraph › adds multiple nodes into the same bucket

    expect(received).toEqual(expected) // deep equality

    Expected: {"address": {"ip": "6.6.6.6", "port": 6666}, "lastUpdated": Any<Date>}
    Received: undefined

      222 |         lastUpdated: expect.any(Date),
      223 |       });
[...]

PR #266, opened 2021-10-25
Introducing vault sharing and the restrictions of pulling and cloning with vault permissions

CMCDragonkai:
### Description

This pull request aims to allow vaults to be shared and unshared with other Node Ids. Specification and discussion can be found inside the Tasklist issues.

### Secret Commands

There are some changes to secrets commands here. These should be extracted out of this PR and moved to another PR as they have different specification requirements.

### Issues Fixed

* Fixes MatrixAI/Polykey#259 
* Fixes MatrixAI/Polykey#253 
* Fixes MatrixAI/Polykey#305
* Fixes MatrixAI/Polykey#299 
* Fixes MatrixAI/Polykey#309
* Fixes MatrixAI/Polykey#335 
* Fixes MatrixAI/Polykey#336
* Fixes MatrixAI/Polykey#337 
* Fixes MatrixAI/Polykey#338
* Fixes MatrixAI/Polykey#339
* Fixes MatrixAI/Polykey#340
* Fixes MatrixAI/Polykey#342 
* Related to MatrixAI/Polykey#294 - in terms of introducing withF and withG to be used
* Related to MatrixAI/Polykey-Docs#3 - review sharing diagrams when changes are done

### Tasks

Reference for residual TODOs is here https://github.com/MatrixAI/js-polykey/pull/266#issuecomment-994242267 .

- [x] 6. MatrixAI/Polykey#337 
- [x] 7. MatrixAI/Polykey#338
- [x] 9. MatrixAI/Polykey#340
- [x] 2. MatrixAI/Polykey#342 
- [x] 8. MatrixAI/Polykey#339
- [x] 1. MatrixAI/Polykey#305  
- [x] 3. MatrixAI/Polykey#259 
- [x] 4. MatrixAI/Polykey#335 
- [x] 5. MatrixAI/Polykey#336

### Final checklist
* [x] Domain specific tests
* [x] Full tests
* [x] Updated inline-comment documentation
* [x] Lint fixed
* [x] Squash and rebased
* [x] Sanity check the final build

PR #280, opened 2021-10-29
Protobuf domains

### LINKED ISSUE #279: Creating `.proto` definition subdomains.
### Specification
We need to update the protobuf definitions to split up messages for each domain into its own domain `.proto` file. we can use package names and imports to achieve this.

### Additional context
Relates to #249 

### Tasks
* [x] Prototype by abstracting into subdirectories under src/proto/schemas, and integrate google's protobuf library by first referencing it and then just copying verbatim
* [x] Create common domain types and common message types and share between Client.proto and Agent.proto

tegefaulkes:
### Description
This PR addresses changes relating to adding subdomains to the proto definitions and importing useful `google.protobuf` types.

The new directory structure looks like this now. when using a message from a domain in a rpc you import it with `import "domains/vaults.proto"` and refer to the messages with `domains.vaults.Vault`. 
```bash
schemas/
├── agent.proto
├── client.proto
├── domains
│   ├── common.proto
│   ├── gestalts.proto
│   ├── identities.proto
│   ├── keys.proto
│   ├── nodes.proto
│   ├── notifications.proto
│   ├── permissions.proto
│   ├── secrets.proto
│   ├── sessions.proto
│   └── vaults.proto
└── test.proto
```

### Issues Fixed
* Fixes #279 
* Relates to #249 

### Tasks
   * [x] - Prototype by abstracting into subdirectories under `src/proto/schemas`, and integrate google's protobuf library by first referencing it and then just copying verbatim
     * [x] - work out how to import google.protobuf messages.
     * [x] - examine google.protobuf.Any usage.
     * [x] - examine timestamp usage.
     * [x] - examine struct usage.
   * [x] - Create common domain types and common message types and share between `Client.proto` and `Agent.proto`

### Final checklist
* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
Right now I have the message subdomains in the same directory  as the `Agent.proto` and `Client.proto` in `proto/schema`. Do we want to move them into their own sub directory like `proto/schema/domains`

Do we want to keep the agent and client specific messages separate. group them like `vaultsAgent.proto`, `vaultsCommon.proto` and `vaultsClient.proto`?

so we want to split the services into the domains as well? Currently the clientService is defined in `Client.proto` it could be split into the domain proto files such as the vaults rpc definitions in the `vaults.proto`.

scripts/proto-generate.sh
```diff
@@ -14,4 +14,5 @@ exec protoc \
   --js_out=import_style=commonjs,binary:src/proto/js \
   --ts_out="grpc_js:${script_dir}/../src/proto/js" \
   --grpc_out="grpc_js:${script_dir}/../src/proto/js" \
-  "${script_dir}/../src/proto/schemas/"*.proto
+  "${script_dir}/../src/proto/schemas/"**/*.proto \
```
CMCDragonkai: Does this already handle globbing all the proto files? Then the second line of `"${script_dir}/../src/proto/schemas/"*.proto \` should be unnecessary.

src/agent/GRPCClientAgent.ts
```diff
@@ -1,7 +1,7 @@
 import type { TLSConfig } from '../network/types';
 
 import { GRPCClient, utils as grpcUtils } from '../grpc';
-import * as agentPB from '../proto/js/Agent_pb';
+import { messages } from '.';
```
CMCDragonkai: Don't import from the current directory's index. We only do that from outside the directory (and even then, it's usually intended for exports). In our own code, we should be always as specific as possible. And furthermore we should be using `agentPB`. Because `messages` is too vague.

src/agent/agentPB.ts
```diff
@@ -0,0 +1,4 @@
+export * as common from '../proto/js/domains/Common_pb';
```
CMCDragonkai: This seems strange, why export things from proto js if you're inside `agent`. I don't think this should be done... Again we want to avoid too much recursive imports/exports, it complicates the structure.

CMCDragonkai:
Can you only copy in the google protos that we are actually using? For the google protos that we are not using, please delete them.

Also on this MR in the description, you should describe the new structure. Use `tree src/proto/schemas` and copy the output to the MR description. Makes it easier to show context.

Also give some annotations explaining why you did what you did. As in put them where.

src/proto/schemas/domains/Permissions.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+import "domains/Nodes.proto";
```
CMCDragonkai: So the root of imports is always from `src/proto/schemas` then?

src/proto/schemas/domains/Permissions.proto
```diff
@@ -0,0 +1,18 @@
...
+import "domains/Identities.proto";
```
CMCDragonkai: How are we deciding whether a proto file should be capitals or lowercase? Originally they were capitals because they were `service Agent` and part of `package agentInterface`.

Can you check the grpc/protobuf spec, and explain which files should be capitals, which should be lowercase based on naming standards of services vs packages.

CMCDragonkai: See: https://developers.google.com/protocol-buffers/docs/style

src/proto/schemas/domains/Permissions.proto
```diff
@@ -0,0 +1,18 @@
...
+
+package permission;
```
CMCDragonkai: According to the style guide.

> Package name should be in lowercase, and should correspond to the directory hierarchy. e.g., if a file is in my/package/, then the package name should be my.package.

scripts/proto-generate.sh
tegefaulkes: If I do just the line with the glob then it it only generates for the sub-directory.

CMCDragonkai: Try: https://unix.stackexchange.com/a/49917/56970

CMCDragonkai: Update the bash script accordingly.

src/agent/agentPB.ts
tegefaulkes: It's just to make it clearer in the code when using the messages. I wanted to replicate the 'name spacing' so in the proto if the message was defined in vaults EG `vaults.Vault` then we'd refer to it in the code as `messages.vaults.Vault`. 

It also allows us to import the messages from one place.

src/proto/schemas/domains/Permissions.proto
tegefaulkes: Yes, the root is set by the `  --proto_path="${script_dir}/../src/proto/schemas" \` line in the proto-generate.sh

src/proto/schemas/domains/Permissions.proto
tegefaulkes: Ok, I can fix that up.

CMCDragonkai:
Yea that's fine, but I wouldn't export this from `src/agent/index.ts`. 
The `src/agent/index.ts` and `src/client/index.ts` should only export 
what's in the local directory.

tegefaulkes:
Oh, so you're saying I should import the agentPb directly instead of through the index.ts?

src/agent/GRPCClientAgent.ts
tegefaulkes: Fixed.

src/agent/agentPB.ts
CMCDragonkai: Yea that's fine, but I wouldn't export this from `src/agent/index.ts`. The `src/agent/index.ts` and `src/client/index.ts` should only export what's in the local directory.

And yes it should be imported from JS directly. Just don't export it from `src/agent/index.ts`. These things are not in the `agent` directory. They are generate JS files from `src/proto/js`.

tegefaulkes: Fixed.

src/proto/schemas/domains/Permissions.proto
tegefaulkes: The style guide doesn't mention what should be done for file names. so far I've just had them as capital like `Vaults.proto` but I think we can decide on convention here.

CMCDragonkai: They say it should always be lower case:

> Files should be named lower_snake_case.proto

https://developers.google.com/protocol-buffers/docs/style#file_structure

scripts/proto-generate.sh
tegefaulkes: That fixed it. it was enabled in the console but disabled when running the script. adding `shopt -qs globstar` fixed it.

CMCDragonkai:
Ok so I see what you did here...

```
├── google
│   └── protobuf
│       ├── any.proto
│       ├── api.proto
│       ├── descriptor.proto
│       ├── duration.proto
│       ├── empty.proto
│       ├── field_mask.proto
│       ├── struct.proto
│       ├── timestamp.proto
│       ├── type.proto
│       └── wrappers.proto
```

When we start using these, we can copy just the ones we want to use. 

Which are all copied from: https://github.com/protocolbuffers/protobuf/tree/master/src/google/protobuf

So atm, none of them are being used, so we can push that to later increment of the #275.

CMCDragonkai:
I like the new structure. Now we can clearly see that `src/proto/schemas/agent.proto` and `src/proto/schemas/client.proto` are used for the services. However I would want to change the service name to `AgentService` instead of just `Agent`. This follows the style guide and why is given here: https://docs.buf.build/lint/rules/#service_suffix mainly to ensure no overlap in naming between message types and service types.

CMCDragonkai:
I'm going to change it to `agent_service.proto` and `AgentService` correspondingly.

The `package` is supposed to have the version specifiers as per the #249. Let me check what they should be.

CMCDragonkai:
I noticed that `domains.common.EchoMessage` can be used by `TestService`.

Also because we are regenerating the code, we should be deleting the `src/proto/js` so that the old files aren't left there.

CMCDragonkai:
From https://github.com/MatrixAI/js-polykey/issues/249#issuecomment-946307094

Because the package name is meant to follow the directory structure. It would mean that `schemas` should have a sub directory. Maybe something like:

```
agent/v1/agent_service.proto
client/v1/client_service.proto
```

Of course that would imply that we have 2 different versions. So instead we might do something like:

```
polykey/v1/agent_service.proto
polykey/v1/client_service.proto
polykey/v1/test_service.proto
```

And instead of `domains`, I reckon we can use `types` as the directory since they are just different types.
That would ensures that our package name is `polykey.v1`

CMCDragonkai:
Also the package name should be the directory, but not contain the name of the file. This follows golang packaging/file hierarchy.

CMCDragonkai:
This means all proto files part of the same package is part of the same namespace. So when referring to them, you just use `types.EchoMessage` and `types.VaultMessage` instead of `types.vaults.VaultMessage` or `types.common.EchoMessage`. Again this is standard in golang where protobuf is part of the same ecosystem.

CMCDragonkai:
Note that imports are all like this now:

```
import "polykey/v1/types/nodes.proto";
import "polykey/v1/types/identities.proto";
```

Which means one has to refer to them using the full package name: `polykey.v1.types.Node`.

CMCDragonkai:
Oh it seems you can use the same types within the same `package`. That means you don't really need to import right?

CMCDragonkai:
I see that the `List` was reused twice as a message name. That's why you created separate packages for each file. But according to the style guide, the packages are meant to be different directories. I reckon here it's better to differentiate the 2 list types and not conflict in the same namespace. So all proto files within the same package are in the same namespace and package corresponds to directory name here.

CMCDragonkai:
From a naming standard pov, the messages are more freeform. However specific kinds of messages should have `Request` and `Response` suffix. See: https://docs.buf.build/lint/rules/#rpc_request_standard_name-rpc_response_standard_name-rpc_request_response_unique But I'll leave this out for now.

Right now I need to ensure that all our types are not conflicting with each other.

CMCDragonkai:
So we'll drop the `Message` suffix. But we will need to add wrapper types for `Request` and `Response` suffixes.

CMCDragonkai:
Ok to preserve what @tegefaulkes did, I will instead do something like:

```
schemas/polykey/v1/agent_service.proto
schemas/polykey/v1/client_service.proto
schemas/polykey/v1/test_service.proto
schemas/polykey/v1/vaults/vaults.proto
schemas/polykey/v1/secrets/secrets.proto
```

This means we have a package like `polykey.v1.secrets.Rename`. Then the directory here means the protofiles.

That way it expands more cleanly and less need to to write the prefix/suffix within each file.

CMCDragonkai:
Updated to:

```
»» ~/Projects/js-polykey/src/proto
 ♖ tree .              ⚡(protobuf_domains) pts/2 23:05:40
.
├── js
│   └── polykey
│       └── v1
│           ├── agent_service_grpc_pb.d.ts
│           ├── agent_service_grpc_pb.js
│           ├── agent_service_pb.d.ts
│           ├── agent_service_pb.js
│           ├── client_service_grpc_pb.d.ts
│           ├── client_service_grpc_pb.js
│           ├── client_service_pb.d.ts
│           ├── client_service_pb.js
│           ├── gestalts
│           │   ├── gestalts_grpc_pb.js
│           │   ├── gestalts_pb.d.ts
│           │   └── gestalts_pb.js
│           ├── identities
│           │   ├── identities_grpc_pb.js
│           │   ├── identities_pb.d.ts
│           │   └── identities_pb.js
│           ├── keys
│           │   ├── keys_grpc_pb.js
│           │   ├── keys_pb.d.ts
│           │   └── keys_pb.js
│           ├── nodes
│           │   ├── nodes_grpc_pb.js
│           │   ├── nodes_pb.d.ts
│           │   └── nodes_pb.js
│           ├── notifications
│           │   ├── notifications_grpc_pb.js
│           │   ├── notifications_pb.d.ts
│           │   └── notifications_pb.js
│           ├── permissions
│           │   ├── permissions_grpc_pb.js
│           │   ├── permissions_pb.d.ts
│           │   └── permissions_pb.js
│           ├── secrets
│           │   ├── secrets_grpc_pb.js
│           │   ├── secrets_pb.d.ts
│           │   └── secrets_pb.js
│           ├── sessions
│           │   ├── sessions_grpc_pb.js
│           │   ├── sessions_pb.d.ts
│           │   └── sessions_pb.js
│           ├── test_service_grpc_pb.d.ts
│           ├── test_service_grpc_pb.js
│           ├── test_service_pb.d.ts
│           ├── test_service_pb.js
│           ├── utils
│           │   ├── utils_grpc_pb.js
│           │   ├── utils_pb.d.ts
│           │   └── utils_pb.js
│           └── vaults
│               ├── vaults_grpc_pb.js
│               ├── vaults_pb.d.ts
│               └── vaults_pb.js
└── schemas
    └── polykey
        └── v1
            ├── agent_service.proto
            ├── client_service.proto
            ├── gestalts
            │   └── gestalts.proto
            ├── identities
            │   └── identities.proto
            ├── keys
            │   └── keys.proto
            ├── nodes
            │   └── nodes.proto
            ├── notifications
            │   └── notifications.proto
            ├── permissions
            │   └── permissions.proto
            ├── secrets
            │   └── secrets.proto
            ├── sessions
            │   └── sessions.proto
            ├── test_service.proto
            ├── utils
            │   └── utils.proto
            └── vaults
                └── vaults.proto

26 directories, 55 files
```

CMCDragonkai:
The usage of `polykey/v1` means that if a breaking change occurs, we will create `polykey/v2` and always the new proto files  there. Only changes need to be created there, but the service files will always need to be copied over. But files that aren't changed, they can be kept the same.

For example if `permissions.proto` is breaking change, then we would have:

```
src/schemas/polykey/v2/permissions/permissions.proto
src/schemas/polykey/v2/agent_service.proto
src/schemas/polykey/v2/client_service.proto
src/schemas/polykey/v2/test_service.proto
```

But of course we should limit such changes, and try to make things as backwards compatible as possible.

CMCDragonkai:
If I copy back in the google proto. There's actually some missing proto files. The original globbing wasn't working. So with the new globstar it is now working, so there's some extra google proto to bring in.

By removing the `types.proto` and `api.proto` this solves the problems such that it works now. So for now I'm leaving:

```
»» ~/Projects/js-polykey/src/proto/schemas/google/protobuf
 ♜ tree .                                   ⚡(protobuf_domains) pts/2 23:16:00
.
├── any.proto
├── descriptor.proto
├── duration.proto
├── empty.proto
├── field_mask.proto
├── struct.proto
├── timestamp.proto
└── wrappers.proto

0 directories, 8 files
```

In case we need it for the future completion of #249.

CMCDragonkai:
Now to update all the src code with the proper imports.

CMCDragonkai:
All code swapped over. Ran `lintfix`.

The last thing to do is to run tests.

If the relevant tests pass for this, then merging time.

All other MRs need to rebase on top. @emmacasolin can you help rebase the CLI retry MR https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/213 cause this makes some big changes to the import paths.

@scottmmorris you'll have to rebase for #266.

CMCDragonkai:
All tests pass. Except of course for nodes being dealt with here #274.

PR #310, opened 2021-12-22
Extracting Node Connection Management out of `NodeManager` to  `NodeConnectionManager`

joshuakarp:
### Description

The refactoring effort to make our nodes domain more robust.

```
                              ┌──────────┐
                              │ Sigchain │
                              └────▲─────┘
                                   │
                                   │
                                   │
                            ┌──────┴──────┐
                            │ NodeManager ├──────────────────┐
                            └──────┬──────┘                  │
                                   │                         │
                                   │                         │
                                   │                         │
                       ┌───────────▼───────────┐             │
┌──────────────┐       │                       │             │
│ ForwardProxy ◄───────┤ NodeConnectionManager │             │
└──────────────┘       │                       │       ┌─────▼─────┐
                       │   ┌──────────────┐    ├───────► NodeGraph │
┌──────────────┐       │   │NodeConnection│    │       └───────────┘
│ ReverseProxy ◄───────┤   └──────────────┘    │
└──────────────┘       │                       │
                       └─────▲─────▲────▲──────┘
                             │     │    │
                             │     │    │
                  ┌──────────┘     │    └─────────────┐
                  │                │                  │
             ┌────┴────┐    ┌──────┴─────┐  ┌─────────┴──────────┐
             │Discovery│    │VaultManager│  │NotificationsManager│
             └─────────┘    └────────────┘  └────────────────────┘
```

### Issues Fixed

* Fixes #225
* Fixes #224
* Fixes #254 
* Fixes #274
* Fixes #216

### Tasks

1. `NodeConnectionManager`:
   - draft API plan:
   ```ts
   class NodeConnectionManager {
      // creates/retrieves a connection, and uses to perform passed f
      public withConnection<T>(targetNodeId, f: (conn: NodeConnection => Promise<T>)): Promise<T>;
      // simply creates a connection, but doesn't use it for anything yet
      public createConnection(targetNodeId): Promise<void>;
      public destroyConnection(targetNodeId): Promise<void>;
      // Migrated from NodeManager:
      // helper function to create and set a NodeConnection (called twice in createConnection)
      protected establishNodeConnection(targetNodeId, lock): Promise<void>;
      public findNode(targetNodeId): Promise<NodeAddress>;
      public openConnection(egressHost, egressPort, timer): Promise<void>;
      // Migrated from NodeGraph:
      public getClosestLocalNodes(targetNodeId): Promise<Array<NodeData>>;
      public getClosestGlobalNodes(targetNodeId): Promise<NodeAddress | undefined>;
[...]

Issue #308, opened 2021-12-20
Split CLI and GRPC API tests


CMCDragonkai:
### Specification

It's time to split up the test modules for these areas, this will our test performance, and give us understanding of the test coverage of all our API behaviour:

* `tests/bin`
* `tests/agent`
* `tests/client`

For `tests/bin`, each command can have its own test. Look at `tests/bin/agent` as a template for what to do. Right now all the other commands are single test modules in their own directory like `tests/bin/identities/identities.test.ts`. These should all be split up into individual commands. Note that most of the tests should be using the shared global agent. However some tests may be independent. This is where you can use the nested describe pattern as demonstrated in `tests/bin/agent/status.test.ts`.

Some tips:

* Use the global shared agent
* Use module mocking for prompting
* Use the correct `pkExec`, `pkStdio`, `pkSpawn`, `pkExpect` utility
* Use 1 nested describe at the bottom in case there is a need to use a the global shared agent and also have independent tests
* Make sure to revert all side effects in the global shared agent! In every test!

We may optimise it further when realising that some tests can just use their own agent instead of sharing the agent. This is because by using the global keypair, starting agents is really fast now. This will probably need to be used in the nodes commands anyway.

For `tests/agent` and `tests/client`, this is about the GRPC API. We have now split up each RPC handler into their own file, this allows much more isolated and specific unit testing of each handler. Refer to `tests/client/service/agentStop.test.ts` as the template for each handler test. All `rpc*.test.ts` should be converted appropriately. Refer to #249 when you have questions or discussions about API design. This is just about splitting the tests up, however we may find we can fix things up regarding the API as this issue is tackled.

### Additional context

* #264 - concurrent testing scalability should improve as we separate our tests out
* #292 - PR for CI/CD that brought in changes that allows isolated GRPC API tests in for client service and agent service
* #249 - Overall GRPC API review
* #306 - Node claims tests are still broken and need to be fixed

### Tasks

1. Split up `tests/bin`
2. Split up `tests/client/service` and refer to `src/client/service` and `tests/client/service/agentStop.test.ts`
3. Split up `tests/agent/service` and refer to `src/agent/service`
4. Check for other API conformance/design issues, such as error handling.

CMCDragonkai:
Should also revert any cases where `beforeEach` and `afterEach` is being used to control the lifecycle of the unit under test. The lifecycle of the unit under test should always be in each individual test.

CMCDragonkai:
[...]

Issue #299, opened 2021-11-29
Review js-id usage for NodeId, VaultId, NotificationId, PermId, ClaimId, Gestalts and `GenericIdTypes.ts`


joshuakarp:
### Specification

The current `GenericIdTypes` module is slightly out of place within the larger architecture of the `js-polykey` codebase. We should find some time to review this, and refactor its required implementation to more suitable areas of the codebase. It was originally introduced in order to facilitate usage of the `js-id` module within Polykey.

### Additional context

* relevant comment on this here https://github.com/MatrixAI/js-polykey/pull/266#issuecomment-984254501 (more generally talking about `js-id` too) - discussion originally from here https://vimeo.com/manage/videos/652339635 5:00 - 17:30
* originally introduced here https://github.com/MatrixAI/js-polykey/commit/c5f22e50398b0bc85fdcd3de5da36ac04cef6ed5
* #253 - vault id usage in EFS and when cloning

### Tasks

1. Add some of the encoding wrappers into the `IdInternal` class, this requires a new PR in `js-id`. Include `toMultibase` and `toUUID` and `toBuffer` as instance methods, and the `fromString`, `fromMultibase`, `fromUUID`, and `fromBuffer` as static methods (these represent alternatives of creating the `IdInternal`). Make sure that these functions return `Id` as the type and not `IdInternal`.
2. Create a new PR for addressing the NodeId changes that fixes #254 and relates to #299 (this requires more work to be done, not just node id).
3. This PR will focus on applying the new `NodeId` which is an opaque alias of `Id` (which itself is an alias of `IdInternal & number`).
4. All the areas which is currently using `NodeId` is expecting an encoded string. These will need to be changed to expect a sort of `type NodeIdEncoded = string;`, you still need a `nodesUtils.encodeNodeId`.
5. The encoding functions and decoding functions of `NodeId` should be inside `nodes/utils.ts`. This should be similar to how I'm doing it for `VaultId`. An alternative might be to create class extension or generics, but this seems complicated.
6. Raw usage of the `NodeId` will mostly occur inside the nodes domain, in particular the `NodeGraph` can use the `Uint8Array` directly. Note that the `js-db` currently doesn't yet allow direct usage of the `Id`, because it requires a `Buffer` type. However there is an issue https://github.com/MatrixAI/js-db/issues/3 to allow it to eventually take `Id`, so for now, just use `idUtils.toBuffer` and `idUtils.fromBuffer` when interacting with the DB.
[...]

PR #311, opened 2021-12-28
CLI and Client & Agent Service test splitting

emmacasolin:
### Description
This PR will split our CLI and GRPC API tests into individual files for each command/handler, as well as refactor the tests to improve performance and coverage.

### Issues Fixed
* Fixes https://github.com/MatrixAI/js-polykey/issues/308
* Fixes https://github.com/MatrixAI/js-polykey/issues/306
* Fixes https://github.com/MatrixAI/js-polykey/issues/312

### Tasks
1. [x] For all `tests/bin`, `tests/client`, and `tests/agent`: 
  - Split tests for each command/handler into their own individual files (except for commands that directly correspond to another command, e.g. `identitiesTokenGet`/`identitiesTokenPut`/`identitiesTokenDelete`)
  - Have all tests use the global shared agent and/or the nested describe pattern used by `tests/bin/agent/status.test.ts` where appropriate
  - Use module mocking for prompting
  - Use the correct `pkExec`, `pkStdio`, `pkSpawn`, `pkExpect` utility
  - Only use nested describe if there is a need to use own agent (rather than global agent)
  - Revert all side effects in the global shared agent and the end of each test
  - Remove all state setup/teardown from `beforeEach` and `afterEach`
  - Make sure `WARN` is the default wherever `LogLevel` is used
9. [x] Remove any unnecessary tests from `GRPCClientClient.test.ts` and `GRPCClientAgent.test.ts`
10. [x] Check for other API conformance/design issues, such as error handling
11. [x] Check order of stopping (in particular proxy stopping in notifications manager)
11. [x] Replace all use of `clearXXX()` with destroy methods in tests
12. [x] Restart method for agent for clearing side effects from the global agent that cannot be otherwise undone
13. [x] Replace `polykeyStartupTimeout` with `defaultTimeout`
14. [x] Fix up node claiming in tests (some tests set them up as singly-signed claims when they should be signed by both nodes) - from MatrixAI/Polykey#306
15. [x] Integrate `EventBus` into PK to be used for asynchronously emitting root keypair changes
16. [x] Fix up imports to only be using selective imports (except for '../utils')
17. [x] Fix identities bin tests and gestalts client service tests

Tests that need splitting:

`tests/bin` (these should be fully simulated wherever possible using the global agent, then own agent using global keypair, then own agent fully created from scratch)
- `agent`
  - [x] lock
  - [x] lockall
  - [x] start
  - [x] status
  - [x] stop
  - [x] unlock
- `identities`
[...]

PR #318, opened 2022-01-20
Change `NodeId` from encoded string to an opaque alias of `Id` which is a typed array from js-id

### LINKED ISSUE #254: Refactor `NodeId` as a proxy/singleton instance
### Specification

The node ID of a Polykey keynode is the public key fingerprint (based on the root keypair of the keynode). As such, on a key renewal/refresh, the node ID of the keynode will also change. Therefore, the node ID can be seen as dynamic state.

Currently we've had to be careful when we retrieve the node ID. We shouldn't be storing it as local, static state, otherwise this state needs to be updated when the root key changes.

The current solution has been to inject the `KeyManager` into any class where we require the node ID to be retrieved, such that we can call `getNodeId` to compute the node ID directly from the root key. This isn't the ideal solution: we shouldn't have to expose all of the `KeyManager` whenever we need just the node ID.

We should look into changing the node ID to be similar to the `Id` class in `js-id`, or making use of the `js-permaproxy` library to do this (https://github.com/matrixai/js-permaproxy). Therefore, only the node ID's proxy instance would need to be injected where we need the node ID.

Would we expose any other functionality other than a getter? If we wanted to do an "update" functionality, we'd actually need to perform an update on the root key itself (given that the node ID is derived from the public key). So perhaps instead, we could look into abstracting the root key as a proxy, such that a key update/renewal can be called without the `KeyManager`... this is potentially completely unnecessary though.

### Additional context

* original discussion of changing the node ID to be a proxy/singleton instance https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/205#note_708488228
* previous thread from !209 (nodes claims) discussing the problems arising in the `nodes` domain when injecting the node ID as a local, static property https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/209#note_675531967
* #312 - Propagate keypair changes to other domains
* #299 - NodeId applied to the `js-id`

### Tasks

1. ...
2. ...
3. ...

### LINKED ISSUE #261: Canonicalise node ID representation
### Specification

We'd like to canonicalise the representation of a keynode's node ID.

Recall that the node ID of a node is based on the public key of a keynode: it is a public key fingerprint. At the moment, we produce a SHA-256 hash of the public key, resulting in an array of 32 bytes. This array of bytes can be currently seen as the primitive canonicalised representation of the node ID.

The representation needs to provide:
* fixed size encodings: the Kademlia system for node ID resolution requires a fixed-size representation of a node ID, such that we can perform bitwise XOR operations between node IDs.
* URL-safe characters

Previously, we were using `base64` as the encoding choice for a human-readable version of the node ID. This produced fixed-size node IDs of 44 characters, but used unsafe characters (e.g. `/`, `+`).

After that, we were then looking at `base58btc`. This has the advantage of no longer producing URL-unsafe characters, as well as some quality-of-life improvements by not including visually-similar characters. However, it can produce different sized node IDs. See below:

<details>

```ts
import * as keysUtils from './src/keys/utils';
import {
  util as forgeUtil,
} from 'node-forge';
import { base58btc } from "multiformats/bases/base58";

async function main() {
  const pemPublicKey43 = `-----BEGIN PUBLIC KEY-----
  MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEA0dTpU0P4AMR3bWs9t4KQ
  WWkQZZsjqMLj4Hr2c3n6bYhqCl4e69P7tFUG6mrIGomUb/JCqZ5cik+IyDoKz/rw
  wB6vJ51ajT4nLgiCgKbiMeU3gdlXKDpJSBxi85sFFjprbmJiO41r2ZezE7bS62Sl
  k8vkPI2IPhH4H/Zun60iO0lRBMd/aWSZ03hNZHGBe1JVTzkoxLneaNOd5XRQ90KO
  TPButem2+5tU91oQsFH4VZONHuJbHEuYeiJUbFtgKNvwDe8w/0BmDoV5r9R7oEjI
  busgF0SQv6DkQILL2PfhJAsYQdY3azd07q3Nhm4YgQJg3r7XbN4+goPUwCr3073d
  rM1Ulc2zh9Oukx7iqzlGjjQWJ7q6bd2ewn/E9dlQDNc8KP4Fq08aMo1bRGSX1Pl8
  h5E5lr/FjKgNmxLwe+gBqo/iMvGYISo6IedEpw2xLcWxQrZ6yu9MzCi+6l6mr/M7
  PQqye8FzMXMxaaqIBi0WUkgg3qTynqAM/jde64GcBdLv4UfkbSkgSRlJb5srmcrc
  jckqpjeP4yRO0ex0ynkMMi78LT0yG3E/XYqFx+XD6EGEpEKMbVPwBmOyEapRbVVI
  DLBEmR6e6WuvdnyiJksCUpZhuDrVyJuVOImL8FJMFtanpPxwZe3FpNaGJQxa0m1d
  VQE/QpKzYZL0pMhBhHr/JBkCAwEAAQ==
  -----END PUBLIC KEY-----`;
  const pemPublicKey44 = `-----BEGIN PUBLIC KEY-----
  MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEA3SABwHq1vJ9ydCl2eW/Q
  r3guKAPdesaxkm85k4EhJ2ywvDcw0wjh39KbcU5ymexxDGzKa3DkhrGzzHbQmS37
  Ee0Nvm6VR2PI+rhmRtJAVztAPPTMj3YHJXaqpoffILXt508jAnvQ38477tZ8hxSR
  /NfZ2OYCLJrCENXvIy6NOPWBwcXPJUL02fDPpaZsvcnoLpIptOnIEuQXL3YJHjKd
  w6yYT4g0sy0MSqp8NFPzzlcY4k5gKO1KR/G2/o2DFErd0n2a7k+SY/ChuUaZpk6U
  q88zfLSRdUCM4pe469ZRrlz7lRoQUemBrmDx0g00MS9GsSEgOZFz3xqt+/3jE+vg
  vt9Ul4929z+9onK5NvjwVR3gRDAl9IFW6P3YVdnyPUo3tNkgQ3Q9Pew/cmulGVf3
  js/DiS9QVjivscfWGbVnkRPuNlUQ0WX4mNEDtYUEGsuYIRkgvBwHYFO1TGQWn3P3
  vEfUg61LzP8WHoYxb/nmYcd14t8zt9reQnCOk+/ON+Btp1Q1ZhIiy7uxG0IGE7m2
  Nawvql9YGFheFC56TYnikm3jfw4boyaqwCeVmqRBAe/NCXcIzDK76AmEvfjzitOl
  2f8rtFMcRE9LFzAYf0bGMbtLEWmE3tMbxDeO3B+DXKtUjeRun6TWca1biyK8niI0
  xsGTzpaAygxgj5DHF8OjP9ECAwEAAQ==
  -----END PUBLIC KEY-----`;
  const publicKey43 = keysUtils.publicKeyFromPem(pemPublicKey43);
  const publicKey44 = keysUtils.publicKeyFromPem(pemPublicKey44);

  const bytes43 = keysUtils.publicKeyToFingerprintBytes(publicKey43);
  const bytes44 = keysUtils.publicKeyToFingerprintBytes(publicKey44);
  console.log('publicKeyToFingerprintBytes length 43:', bytes43.length);
  console.log('publicKeyToFingerprintBytes length 44:', bytes44.length);
  const fTypedArray43 = forgeUtil.binary.raw.decode(bytes43);
  const fTypedArray44 = forgeUtil.binary.raw.decode(bytes44);
  console.log('fTypedArray 43:', fTypedArray43);
  console.log('fTypedArray 44:', fTypedArray44);
  console.log('Hex string 43:', Buffer.from(fTypedArray43).toString('hex'));
  console.log(base58btc.baseEncode(fTypedArray43));
  console.log(keysUtils.publicKeyToFingerprint(publicKey43));

}

main();
```

```
[nix-shell:~/Documents/MatrixAI/js-polykey]$ npm run ts-node -- nodeIds.ts 

> @matrixai/polykey@1.0.0 ts-node /home/josh/Documents/MatrixAI/js-polykey
> ts-node -r tsconfig-paths/register "nodeIds.ts"

publicKeyToFingerprintBytes length 43: 32
publicKeyToFingerprintBytes length 44: 32
fTypedArray 43: Uint8Array(32) [
  10,  45,  47,  45, 125, 109,  40,  40,
  20, 160,  60, 163, 101, 100, 189, 222,
  38,  35, 246, 172,  73, 251, 232, 221,
  13,  97, 139, 151, 240, 251, 129, 201
]
fTypedArray 44: Uint8Array(32) [
  233, 254, 155, 220,   4, 146, 225,  43,
  130, 201,  87,  50,  52, 141, 221, 156,
   15, 162, 131,  58, 120, 191, 146,  36,
   49, 133,  50,  92, 203,  95, 232,  13
]
Hex string 43: 0a2d2f2d7d6d282814a03ca36564bdde2623f6ac49fbe8dd0d618b97f0fb81c9
gj3DUVjzAGHvmP9ZRFB5geWLVRwXUSdEUtMcVvxozLx
gj3DUVjzAGHvmP9ZRFB5geWLVRwXUSdEUtMcVvxozLx

```

</details>

As a temporary solution, we've adopted a `base32hex` encoding, producing a fixed fixed 53 character node ID, such as `v6n7m9vuf44pfqq133re8v91ju7a8968h4nkirjt8r7r9pbcoqqq0` (with a prepended `v` character from `multibase`): https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/205#note_711237922

### Additional context

* https://github.com/MatrixAI/js-polykey/issues/254 - will need to look at this issue concurrently
* discussions on `base58` and `base58btc`
  * https://github.com/dougal/base58/issues/8
  * https://en.bitcoin.it/wiki/Base58Check_encoding
  * https://learnmeabitcoin.com/technical/base58
  * https://bitcoin.stackexchange.com/questions/22636/what-is-the-correct-base58-encoding-for-this-uncompressed-public-key-ripemd-hash

### Tasks

1. Choose a standard means of representing the node ID
2. Change all current usages of node ID to be this new standard (e.g. within `NodeGraph` (kademlia), across GRPC, etc)

tegefaulkes:
### Description
This PR will focus on applying the new `NodeId` which is an opaque alias of `Id` (which itself is an alias of `IdInternal & number`).

1. Add some of the encoding wrappers into the `IdInternal` class, this requires a new PR in `js-id`. Include `toMultibase` and `toUUID` and `toBuffer` as instance methods, and the `fromString`, `fromMultibase`, `fromUUID`, and `fromBuffer` as static methods (these represent alternatives of creating the `IdInternal`). Make sure that these functions return `Id` as the type and not `IdInternal`.
3. This PR will focus on applying the new `NodeId` which is an opaque alias of `Id` (which itself is an alias of `IdInternal & number`).
4. All the areas which is currently using `NodeId` is expecting an encoded string. These will need to be changed to expect a sort of `type NodeIdEncoded = string;`, you still need a `nodesUtils.encodeNodeId`.
5. The encoding functions and decoding functions of `NodeId` should be inside `nodes/utils.ts`. This should be similar to how I'm doing it for `VaultId`. An alternative might be to create class extension or generics, but this seems complicated.
6. Raw usage of the `NodeId` will mostly occur inside the nodes domain, in particular the `NodeGraph` can use the `Uint8Array` directly. Note that the `js-db` currently doesn't yet allow direct usage of the `Id`, because it requires a `Buffer` type. However there is an issue https://github.com/MatrixAI/js-db/issues/3 to allow it to eventually take `Id`, so for now, just use `idUtils.toBuffer` and `idUtils.fromBuffer` when interacting with the DB.
7. Once this PR is merged into master, rebase #266 and #310 on top to benefit from it.

### Issues Fixed

- Fixes #254 
- Fixes #261 
- Related #299 

### Tasks

1. [x] Add some of the encoding wrappers into the `IdInternal` class, this requires a new PR in `js-id`. Include `toMultibase` and `toUUID` and `toBuffer` as instance methods, and the `fromString`, `fromMultibase`, `fromUUID`, and `fromBuffer` as static methods (these represent alternatives of creating the `IdInternal`). Make sure that these functions return `Id` as the type and not `IdInternal`. - Update to 3.2.0 of js-id
2. [x] Update `NodeId` type to use an opaque alias of `Id`.
3. [x] Update all uses of `NodeId` to use the new type where possible.
4. [x] Where a encoded string of `NodeId` is required, use a `NodeIdEncoded` instead.
5. [x] Don't get the NodeId from the Certificate. get the Public key from the `KeyManager` and convert it from that. 
6. ~[ ] `PolykeyAgent` should store the `NodeId`. It should be updated via the event bus. ~ - stay with using `this.keyManager.getNodeId()`.~
7. ~[ ] Anything that needs the NodeId should take a lambda that fetches it from the PolykeyAgent.~ - stick with injecting `KeyManager` where it is needed
8. [x] Add `NodeId` validation to RPC services and possibly bin commands.
9. [x] review types and utils imports.

example of usage here.
https://github.com/MatrixAI/js-polykey/issues/254#issuecomment-1016205861

Areas that need changes:

1. ~Status needs the string encoded to put into the status JSON file~ - Status file does indeed to use the encoded string, but `StatusInfo` should still use the `NodeId` and not `NodeIdEncoded` because we want to do any validation on the boundary and ensure that our internal types in the context of the program uses types that imply that the validation occurred already, and therefore the `nodeId` is indeed a valid node id.
2. GRPCClient needs the string encoded to pass as a URL parameter for the proxy
3. keymanager certificate needs to have string encoded node id
4. sigchain claims needs to have string encoded node id
5. `GestaltGraph.ts` needs to use the encoded string of the NodeId - its just because it is designed like this it's due to the `GestaltKey`
6. The `NodeInfo` for now has to use `NodeIdEncoded` because it is being put into the database as a value as part of a POJO which gets JSON encoded. Plus that `GestaltGraph` and `Discovery` uses it as a string, but in the future it may be better done as `NodeId` which follows the same reason as `1.` about validation being done on the boundary.

@tegefaulkes please flesh the above out.

Guidelines:

1. Use NodeId binary form where possible
2. Use NodeId encoded as multibase when it goes to a file, to outside network, to the console/terminal
3. Use NodeId binary string if it's put into a private POJO... - do this only when you're really sure that this will never be seen by anything else
4. Use NodeId buffers when using DB.
5. when comparing two ids for equality convert them to strings with `.toString()`

Unresolved tests addressed in other PRs:

These were failing in the commit I branched off of. May be fix in #310 or #308 
- ping.test.ts
- find.test.ts
- notificationsManager.test.ts

relevant to the vaults domain.
- vaultManager.test.ts

Part of nodes and #310 
- NodeConnection.test.ts
- nodeManager.test.ts

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/acl/ACL.ts
+++ b/src/acl/ACL.ts
@@ -7,7 +7,7 @@ import type { Ref } from '../types';
 
 import { Mutex } from 'async-mutex';
 import Logger from '@matrixai/logger';
-import { utils as idUtils } from '@matrixai/id';
+import { IdInternal, utils as idUtils } from '@matrixai/id';
 import {
   CreateDestroyStartStop,
   ready,
@@ -130,8 +130,16 @@ class ACL {
     nodeId2: NodeId,
   ): Promise<boolean> {
     return await this._transaction(async () => {
-      const permId1 = await this.db.get(this.aclNodesDbDomain, nodeId1, true);
-      const permId2 = await this.db.get(this.aclNodesDbDomain, nodeId2, true);
+      const permId1 = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId1.toBuffer(),
+        true,
+      );
+      const permId2 = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId2.toBuffer(),
+        true,
+      );
       if (permId1 != null && permId2 != null && permId1 === permId2) {
         return true;
       }
@@ -147,7 +155,7 @@ class ACL {
         Record<NodeId, Permission>
       > = {};
       for await (const o of this.aclNodesDb.createReadStream()) {
-        const nodeId = (o as any).key as NodeId;
+        const nodeId = IdInternal.create<NodeId>((o as any).key);
         const data = (o as any).value as Buffer;
         const permId = makePermissionId(
           await this.db.deserializeDecrypt(data, true),
@@ -196,15 +204,16 @@ class ACL {
         );
         const nodePerm: Record<NodeId, Permission> = {};
         const nodeIdsGc: Set<NodeId> = new Set();
-        for (const nodeId in nodeIds) {
+        for (const nodeIdString in nodeIds) {
+          const nodeId: NodeId = IdInternal.fromString(nodeIdString);
           const permId = await this.db.get(
             this.aclNodesDbDomain,
-            nodeId as NodeId,
+            nodeId.toBuffer(),
             true,
           );
           if (permId == null) {
             // Invalid node id
-            nodeIdsGc.add(nodeId as NodeId);
+            nodeIdsGc.add(nodeId);
             continue;
           }
           const permRef = (await this.db.get(
@@ -213,7 +222,7 @@ class ACL {
           )) as Ref<Permission>;
           if (!(vaultId in permRef.object.vaults)) {
             // Vault id is missing from the perm
-            nodeIdsGc.add(nodeId as NodeId);
+            nodeIdsGc.add(nodeId);
             continue;
           }
           nodePerm[nodeId] = permRef.object;
@@ -244,7 +253,11 @@ class ACL {
   @ready(new aclErrors.ErrorACLNotRunning())
   public async getNodePerm(nodeId: NodeId): Promise<Permission | undefined> {
     return await this._transaction(async () => {
-      const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+      const permId = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId.toBuffer(),
+        true,
+      );
       if (permId == null) {
         return;
       }
@@ -275,15 +288,16 @@ class ACL {
       }
       const perms: Record<NodeId, Permission> = {};
       const nodeIdsGc: Set<NodeId> = new Set();
-      for (const nodeId in nodeIds) {
+      for (const nodeIdString in nodeIds) {
+        const nodeId: NodeId = IdInternal.fromString(nodeIdString);
         const permId = await this.db.get(
           this.aclNodesDbDomain,
-          nodeId as NodeId,
+          nodeId.toBuffer(),
           true,
         );
         if (permId == null) {
           // Invalid node id
-          nodeIdsGc.add(nodeId as NodeId);
+          nodeIdsGc.add(nodeId);
           continue;
         }
         const permRef = (await this.db.get(
@@ -292,7 +306,7 @@ class ACL {
         )) as Ref<Permission>;
         if (!(vaultId in permRef.object.vaults)) {
           // Vault id is missing from the perm
-          nodeIdsGc.add(nodeId as NodeId);
+          nodeIdsGc.add(nodeId);
           continue;
         }
         perms[nodeId] = permRef.object;
@@ -318,7 +332,11 @@ class ACL {
     action: GestaltAction,
   ): Promise<void> {
     return await this._transaction(async () => {
-      const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+      const permId = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId.toBuffer(),
+        true,
+      );
       const ops: Array<DBOp> = [];
       if (permId == null) {
         const permId = await aclUtils.generatePermId();
@@ -341,7 +359,7 @@ class ACL {
           {
             type: 'put',
             domain: this.aclNodesDbDomain,
-            key: nodeId,
+            key: nodeId.toBuffer(),
             value: idUtils.toBuffer(permId),
             raw: true,
           },
@@ -369,7 +387,11 @@ class ACL {
     action: GestaltAction,
   ): Promise<void> {
     return await this._transaction(async () => {
-      const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+      const permId = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId.toBuffer(),
+        true,
+      );
       if (permId == null) {
         return;
       }
@@ -394,7 +416,11 @@ class ACL {
           this.aclVaultsDbDomain,
           idUtils.toBuffer(vaultId),
         )) ?? {};
-      const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+      const permId = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId.toBuffer(),
+        true,
+      );
       if (permId == null) {
         throw new aclErrors.ErrorACLNodeIdMissing();
       }
@@ -419,7 +445,7 @@ class ACL {
         {
           type: 'put',
           domain: this.aclNodesDbDomain,
-          key: nodeId,
+          key: nodeId.toBuffer(),
           value: permId,
           raw: true,
         },
@@ -448,7 +474,11 @@ class ACL {
       if (nodeIds == null || !(nodeId in nodeIds)) {
         return;
       }
-      const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+      const permId = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId.toBuffer(),
+        true,
+      );
       if (permId == null) {
         return;
       }
@@ -491,7 +521,7 @@ class ACL {
     for (const nodeId of nodeIds) {
       const permIdBuffer = await this.db.get(
         this.aclNodesDbDomain,
-        nodeId,
+        nodeId.toBuffer(),
         true,
       );
       if (permIdBuffer == null) {
@@ -537,7 +567,7 @@ class ACL {
       ops.push({
         domain: this.aclNodesDbDomain,
         type: 'put',
-        key: nodeId,
+        key: nodeId.toBuffer(),
         value: idUtils.toBuffer(permId),
         raw: true,
       });
@@ -558,7 +588,11 @@ class ACL {
     nodeId: NodeId,
     perm: Permission,
   ): Promise<Array<DBOp>> {
-    const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+    const permId = await this.db.get(
+      this.aclNodesDbDomain,
+      nodeId.toBuffer(),
+      true,
+    );
     const ops: Array<DBOp> = [];
     if (permId == null) {
       const permId = await aclUtils.generatePermId();
@@ -576,7 +610,7 @@ class ACL {
         {
           type: 'put',
           domain: this.aclNodesDbDomain,
-          key: nodeId,
+          key: nodeId.toBuffer(),
           value: idUtils.toBuffer(permId),
           raw: true,
         },
@@ -608,7 +642,11 @@ class ACL {
 
   @ready(new aclErrors.ErrorACLNotRunning())
   public async unsetNodePermOps(nodeId: NodeId): Promise<Array<DBOp>> {
-    const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+    const permId = await this.db.get(
+      this.aclNodesDbDomain,
+      nodeId.toBuffer(),
+      true,
+    );
     if (permId == null) {
       return [];
     }
@@ -635,7 +673,7 @@ class ACL {
     ops.push({
       type: 'del',
       domain: this.aclNodesDbDomain,
-      key: nodeId,
+      key: nodeId.toBuffer(),
     });
     // We do not remove the node id from the vaults
     // they can be removed later upon inspection
@@ -653,10 +691,11 @@ class ACL {
         return;
       }
       const ops: Array<DBOp> = [];
-      for (const nodeId in nodeIds) {
+      for (const nodeIdString in nodeIds) {
+        const nodeId: NodeId = IdInternal.fromString(nodeIdString);
         const permId = await this.db.get(
           this.aclNodesDbDomain,
-          nodeId as NodeId,
+          nodeId.toBuffer(),
           true,
         );
         // Skip if the nodeId doesn't exist
@@ -703,7 +742,11 @@ class ACL {
     nodeIdsJoin: Array<NodeId>,
     perm?: Permission,
   ): Promise<Array<DBOp>> {
-    const permId = await this.db.get(this.aclNodesDbDomain, nodeId, true);
+    const permId = await this.db.get(
+      this.aclNodesDbDomain,
+      nodeId.toBuffer(),
+      true,
+    );
     if (permId == null) {
       throw new aclErrors.ErrorACLNodeIdMissing();
     }
@@ -719,7 +762,7 @@ class ACL {
     for (const nodeIdJoin of nodeIdsJoin) {
       const permIdJoin = await this.db.get(
         this.aclNodesDbDomain,
-        nodeIdJoin,
+        nodeIdJoin.toBuffer(),
         true,
       );
       if (permIdJoin === permId) {
@@ -750,7 +793,7 @@ class ACL {
       ops.push({
         type: 'put',
         domain: this.aclNodesDbDomain,
-        key: nodeIdJoin,
+        key: nodeIdJoin.toBuffer(),
         value: permId,
         raw: true,
       });
@@ -789,15 +832,16 @@ class ACL {
     }
     const ops: Array<DBOp> = [];
     const nodeIdsGc: Set<NodeId> = new Set();
-    for (const nodeId in nodeIds) {
+    for (const nodeIdString in nodeIds) {
+      const nodeId: NodeId = IdInternal.fromString(nodeIdString);
       const permId = await this.db.get(
         this.aclNodesDbDomain,
-        nodeId as NodeId,
+        nodeId.toBuffer(),
         true,
       );
       if (permId == null) {
         // Invalid node id
-        nodeIdsGc.add(nodeId as NodeId);
+        nodeIdsGc.add(nodeId);
         continue;
       }
       const permRef = (await this.db.get(
@@ -806,7 +850,7 @@ class ACL {
       )) as Ref<Permission>;
       if (!(vaultId in permRef.object.vaults)) {
         // Vault id is missing from the perm
-        nodeIdsGc.add(nodeId as NodeId);
+        nodeIdsGc.add(nodeId);
         continue;
       }
       const vaultActions: VaultActions | undefined =
--- a/src/agent/service/nodesCrossSignClaim.ts
+++ b/src/agent/service/nodesCrossSignClaim.ts
@@ -6,6 +6,7 @@ import type { KeyManager } from '../../keys';
 import type * as nodesPB from '../../proto/js/polykey/v1/nodes/nodes_pb';
 import { utils as grpcUtils } from '../../grpc';
 import { utils as claimsUtils, errors as claimsErrors } from '../../claims';
+import { utils as nodesUtils } from '../../nodes';
 
 function nodesCrossSignClaim({
   keyManager,
@@ -66,7 +67,7 @@ function nodesCrossSignClaim({
         }
         // Verify the claim
         const senderPublicKey = await nodeManager.getPublicKey(
-          payloadData.node1,
+          nodesUtils.decodeNodeId(payloadData.node1),
         );
         const verified = await claimsUtils.verifyClaimSignature(
           constructedEncodedClaim,
@@ -79,12 +80,12 @@ function nodesCrossSignClaim({
         const doublySignedClaim = await claimsUtils.signIntermediaryClaim({
           claim: constructedIntermediaryClaim,
           privateKey: keyManager.getRootKeyPairPem().privateKey,
-          signeeNodeId: nodeManager.getNodeId(),
+          signeeNodeId: nodesUtils.encodeNodeId(nodeManager.getNodeId()),
         });
         // Then create your own intermediary node claim (from X -> Y)
         const singlySignedClaim = await sigchain.createIntermediaryClaim({
           type: 'node',
-          node1: nodeManager.getNodeId(),
+          node1: nodesUtils.encodeNodeId(nodeManager.getNodeId()),
           node2: payloadData.node1,
         });
         // Should never be reached, but just for type safety
--- a/src/agent/service/nodesHolePunchMessageSend.ts
+++ b/src/agent/service/nodesHolePunchMessageSend.ts
@@ -22,7 +22,7 @@ function nodesHolePunchMessageSend({
       // back to the source node.
       if (
         nodeManager.getNodeId() ===
-        nodesUtils.makeNodeId(call.request.getTargetId())
+        nodesUtils.decodeNodeId(call.request.getTargetId())
       ) {
         const [host, port] = networkUtils.parseAddress(
           call.request.getEgressAddress(),
@@ -32,7 +32,7 @@ function nodesHolePunchMessageSend({
         // If so, ask the nodeManager to relay to the node
       } else if (
         await nodeManager.knowsNode(
-          nodesUtils.makeNodeId(call.request.getSrcId()),
+          nodesUtils.decodeNodeId(call.request.getSrcId()),
         )
       ) {
         await nodeManager.relayHolePunchMessage(call.request);
--- a/src/bin/utils/processors.ts
+++ b/src/bin/utils/processors.ts
@@ -1,6 +1,6 @@
 import type { FileSystem } from '../../types';
 import type { RecoveryCode } from '../../keys/types';
-import type { NodeId } from '../../nodes/types';
+import type { NodeId, NodeIdEncoded } from '../../nodes/types';
 import type { Host, Port } from '../../network/types';
 import type {
   StatusStarting,
@@ -17,6 +17,7 @@ import * as binErrors from '../errors';
 import * as clientUtils from '../../client/utils';
 import { Status } from '../../status';
 import config from '../../config';
+import { utils as nodesUtils } from '../../nodes';
 
 /**
  * Prompts for existing password
@@ -191,7 +192,7 @@ async function processRecoveryCode(
  */
 async function processClientOptions(
   nodePath: string,
-  nodeId?: NodeId,
+  nodeId?: NodeIdEncoded,
   clientHost?: Host,
   clientPort?: Port,
   fs = require('fs'),
@@ -214,12 +215,13 @@ async function processClientOptions(
     if (statusInfo === undefined || statusInfo.status !== 'LIVE') {
       throw new binErrors.ErrorCLIStatusNotLive();
     }
-    if (nodeId == null) nodeId = statusInfo.data.nodeId;
+    if (nodeId == null)
+      nodeId = nodesUtils.encodeNodeId(statusInfo.data.nodeId);
     if (clientHost == null) clientHost = statusInfo.data.clientHost;
     if (clientPort == null) clientPort = statusInfo.data.clientPort;
   }
   return {
-    nodeId,
+    nodeId: nodesUtils.decodeNodeId(nodeId),
     clientHost: clientHost,
     clientPort: clientPort,
   };
@@ -233,7 +235,7 @@ async function processClientOptions(
  */
 async function processClientStatus(
   nodePath: string,
-  nodeId?: NodeId,
+  nodeId?: NodeIdEncoded,
   clientHost?: Host,
   clientPort?: Port,
   fs = require('fs'),
@@ -261,12 +263,14 @@ async function processClientStatus(
       clientPort: Port;
     }
 > {
+  let nodeIdDecoded: NodeId | undefined =
+    nodeId != null ? nodesUtils.decodeNodeId(nodeId) : undefined;
   // If all parameters are set, no status and no statusInfo is used
-  if (nodeId != null && clientHost != null && clientPort != null) {
+  if (nodeIdDecoded != null && clientHost != null && clientPort != null) {
     return {
       statusInfo: undefined,
       status: undefined,
-      nodeId,
+      nodeId: nodeIdDecoded,
       clientHost,
       clientPort,
     };
@@ -286,13 +290,13 @@ async function processClientStatus(
     throw new binErrors.ErrorCLIStatusMissing();
   }
   if (statusInfo.status === 'LIVE') {
-    if (nodeId == null) nodeId = statusInfo.data.nodeId;
+    if (nodeIdDecoded == null) nodeIdDecoded = statusInfo.data.nodeId;
     if (clientHost == null) clientHost = statusInfo.data.clientHost;
     if (clientPort == null) clientPort = statusInfo.data.clientPort;
     return {
       statusInfo,
       status,
-      nodeId,
+      nodeId: nodeIdDecoded,
       clientHost,
       clientPort,
     };
@@ -300,7 +304,7 @@ async function processClientStatus(
   return {
     statusInfo,
     status,
-    nodeId,
+    nodeId: nodeIdDecoded,
     clientHost,
     clientPort,
   };
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -18,6 +18,7 @@ import { CreateDestroy, ready } from '@matrixai/async-init/dist/CreateDestroy';
 import * as discoveryErrors from './errors';
 import * as gestaltsUtils from '../gestalts/utils';
 import * as claimsUtils from '../claims/utils';
+import { utils as nodesUtils } from '../nodes';
 
 interface Discovery extends CreateDestroy {}
 @CreateDestroy()
@@ -112,7 +113,8 @@ class Discovery {
         // The sigchain data of the vertex (containing all cryptolinks)
         let vertexChainData: ChainData = {};
         // If the vertex we've found is our own node, we simply get our own chain
-        if (vertexGId.nodeId === this.nodeManager.getNodeId()) {
+        const nodeId = nodesUtils.decodeNodeId(vertexGId.nodeId);
+        if (nodeId.equals(this.nodeManager.getNodeId())) {
           const vertexChainDataEncoded = await this.nodeManager.getChainData();
           // Decode all our claims - no need to verify (on our own sigchain)
           for (const c in vertexChainDataEncoded) {
@@ -123,9 +125,7 @@ class Discovery {
           }
           // Otherwise, request the verified chain data from the node
         } else {
-          vertexChainData = await this.nodeManager.requestChainData(
-            vertexGId.nodeId,
-          );
+          vertexChainData = await this.nodeManager.requestChainData(nodeId);
         }
 
         // TODO: for now, the chain data is treated as a 'disjoint' set of
@@ -136,7 +136,7 @@ class Discovery {
 
         // Now have the NodeInfo of this vertex
         const vertexNodeInfo: NodeInfo = {
-          id: vertexGId.nodeId,
+          id: nodesUtils.encodeNodeId(nodeId),
           chain: vertexChainData,
         };
 
@@ -150,12 +150,14 @@ class Discovery {
           // If the claim is to a node
           if (claim.payload.data.type === 'node') {
             // Get the chain data of the linked node
-            const linkedVertexNodeId = claim.payload.data.node2;
+            const linkedVertexNodeId = nodesUtils.decodeNodeId(
+              claim.payload.data.node2,
+            );
             const linkedVertexChainData =
               await this.nodeManager.requestChainData(linkedVertexNodeId);
             // With this verified chain, we can link
             const linkedVertexNodeInfo: NodeInfo = {
-              id: linkedVertexNodeId,
+              id: nodesUtils.encodeNodeId(linkedVertexNodeId),
               chain: linkedVertexChainData,
             };
             await this.gestaltGraph.linkNodeAndNode(
@@ -200,7 +202,7 @@ class Discovery {
           }
         }
         // Add this node vertex to the visited
-        visitedVertices.add(gestaltsUtils.keyFromNode(vertexGId.nodeId));
+        visitedVertices.add(gestaltsUtils.keyFromNode(nodeId));
       } else if (vertexGId.type === 'identity') {
         // If the next vertex is an identity, perform a social discovery
         // Firstly get the identity info of this identity
@@ -222,14 +224,14 @@ class Discovery {
           // Claims on an identity provider will always be node -> identity
           // So just cast payload data as such
           const data = claim.payload.data as ClaimLinkIdentity;
-          const linkedVertexNodeId = data.node;
+          const linkedVertexNodeId = nodesUtils.decodeNodeId(data.node);
           // Get the chain data of this claimed node (so that we can link in GG)
           const linkedVertexChainData = await this.nodeManager.requestChainData(
             linkedVertexNodeId,
           );
           // With this verified chain, we can link
           const linkedVertexNodeInfo: NodeInfo = {
-            id: linkedVertexNodeId,
+            id: nodesUtils.encodeNodeId(linkedVertexNodeId),
             chain: linkedVertexChainData,
           };
           await this.gestaltGraph.linkNodeAndIdentity(
@@ -314,7 +316,7 @@ class Discovery {
       // Verify the claim with the public key of the node
       const verified = await claimsUtils.verifyClaimSignature(
         encoded,
-        await this.nodeManager.getPublicKey(data.node),
+        await this.nodeManager.getPublicKey(nodesUtils.decodeNodeId(data.node)),
       );
       // If verified, add to the record
       if (verified) {
--- a/src/gestalts/GestaltGraph.ts
+++ b/src/gestalts/GestaltGraph.ts
@@ -1,15 +1,15 @@
 import type { Buffer } from 'buffer';
 import type {
+  Gestalt,
   GestaltAction,
   GestaltActions,
-  GestaltKey,
-  GestaltNodeKey,
   GestaltIdentityKey,
+  GestaltKey,
   GestaltKeySet,
-  Gestalt,
+  GestaltNodeKey,
 } from './types';
 import type { NodeId, NodeInfo } from '../nodes/types';
-import type { ProviderId, IdentityId, IdentityInfo } from '../identities/types';
+import type { IdentityId, IdentityInfo, ProviderId } from '../identities/types';
 import type { Permission } from '../acl/types';
 import type { DB, DBLevel, DBOp } from '@matrixai/db';
 import type { ACL } from '../acl';
@@ -24,6 +24,7 @@ import * as gestaltsUtils from './utils';
 import * as gestaltsErrors from './errors';
 import { utils as aclUtils } from '../acl';
 import * as utils from '../utils';
+import { utils as nodesUtils } from '../nodes';
 
 interface GestaltGraph extends CreateDestroyStartStop {}
 @CreateDestroyStartStop(
@@ -282,7 +283,7 @@ class GestaltGraph {
       if (gId.type === 'node') {
         ops.push(
           ...(await this.unlinkNodeAndIdentityOps(
-            gId.nodeId,
+            nodesUtils.decodeNodeId(gId.nodeId),
             providerId,
             identityId,
           )),
@@ -316,7 +317,9 @@ class GestaltGraph {
 
   @ready(new gestaltsErrors.ErrorGestaltsGraphNotRunning())
   public async setNodeOps(nodeInfo: NodeInfo): Promise<Array<DBOp>> {
-    const nodeKey = gestaltsUtils.keyFromNode(nodeInfo.id);
+    const nodeKey = gestaltsUtils.keyFromNode(
+      nodesUtils.decodeNodeId(nodeInfo.id),
+    );
     const ops: Array<DBOp> = [];
     let nodeKeyKeys = await this.db.get<GestaltKeySet>(
       this.graphMatrixDbDomain,
@@ -326,10 +329,13 @@ class GestaltGraph {
       nodeKeyKeys = {};
       // Sets the gestalt in the acl
       ops.push(
-        ...(await this.acl.setNodePermOps(nodeInfo.id, {
-          gestalt: {},
-          vaults: {},
-        })),
+        ...(await this.acl.setNodePermOps(
+          nodesUtils.decodeNodeId(nodeInfo.id),
+          {
+            gestalt: {},
+            vaults: {},
+          },
+        )),
       );
     }
     ops.push(
@@ -383,7 +389,12 @@ class GestaltGraph {
     for (const key of Object.keys(nodeKeyKeys) as Array<GestaltKey>) {
       const gId = gestaltsUtils.ungestaltKey(key);
       if (gId.type === 'node') {
-        ops.push(...(await this.unlinkNodeAndNodeOps(nodeId, gId.nodeId)));
+        ops.push(
+          ...(await this.unlinkNodeAndNodeOps(
+            nodeId,
+            nodesUtils.decodeNodeId(gId.nodeId),
+          )),
+        );
       } else if (gId.type === 'identity') {
         ops.push(
           ...(await this.unlinkNodeAndIdentityOps(
@@ -425,7 +436,9 @@ class GestaltGraph {
     identityInfo: IdentityInfo,
   ): Promise<Array<DBOp>> {
     const ops: Array<DBOp> = [];
-    const nodeKey = gestaltsUtils.keyFromNode(nodeInfo.id);
+    const nodeKey = gestaltsUtils.keyFromNode(
+      nodesUtils.decodeNodeId(nodeInfo.id),
+    );
     const identityKey = gestaltsUtils.keyFromIdentity(
       identityInfo.providerId,
       identityInfo.identityId,
@@ -475,10 +488,13 @@ class GestaltGraph {
     //     join node gestalt's permission to the identity gestalt
     if (nodeNew && identityNew) {
       ops.push(
-        ...(await this.acl.setNodePermOps(nodeInfo.id, {
-          gestalt: {},
-          vaults: {},
-        })),
+        ...(await this.acl.setNodePermOps(
+          nodesUtils.decodeNodeId(nodeInfo.id),
+          {
+            gestalt: {},
+            vaults: {},
+          },
+        )),
       );
     } else if (
       !nodeNew &&
@@ -489,12 +505,13 @@ class GestaltGraph {
         Object.keys(identityKeyKeys) as Array<GestaltKey>,
         [identityKey],
       );
-      const identityNodeIds = Array.from(
-        identityNodeKeys,
-        (key) => gestaltsUtils.ungestaltKey(key).nodeId,
+      const identityNodeIds = Array.from(identityNodeKeys, (key) =>
+        gestaltsUtils.nodeFromKey(key),
       );
       // These must exist
-      const nodePerm = (await this.acl.getNodePerm(nodeInfo.id)) as Permission;
+      const nodePerm = (await this.acl.getNodePerm(
+        nodesUtils.decodeNodeId(nodeInfo.id),
+      )) as Permission;
       const identityPerm = (await this.acl.getNodePerm(
         identityNodeIds[0],
       )) as Permission;
@@ -505,7 +522,7 @@ class GestaltGraph {
       // and the perm record update
       ops.push(
         ...(await this.acl.joinNodePermOps(
-          nodeInfo.id,
+          nodesUtils.decodeNodeId(nodeInfo.id),
           identityNodeIds,
           permNew,
         )),
@@ -513,10 +530,13 @@ class GestaltGraph {
     } else if (nodeNew && !identityNew) {
       if (utils.isEmptyObject(identityKeyKeys)) {
         ops.push(
-          ...(await this.acl.setNodePermOps(nodeInfo.id, {
-            gestalt: {},
-            vaults: {},
-          })),
+          ...(await this.acl.setNodePermOps(
+            nodesUtils.decodeNodeId(nodeInfo.id),
+            {
+              gestalt: {},
+              vaults: {},
+            },
+          )),
         );
       } else {
         let identityNodeKey: GestaltNodeKey;
@@ -524,11 +544,11 @@ class GestaltGraph {
           identityNodeKey = gK as GestaltNodeKey;
           break;
         }
-        const identityNodeId = gestaltsUtils.ungestaltKey(
-          identityNodeKey!,
-        ).nodeId;
+        const identityNodeId = gestaltsUtils.nodeFromKey(identityNodeKey!);
         ops.push(
-          ...(await this.acl.joinNodePermOps(identityNodeId, [nodeInfo.id])),
+          ...(await this.acl.joinNodePermOps(identityNodeId, [
+            nodesUtils.decodeNodeId(nodeInfo.id),
+          ])),
         );
       }
     }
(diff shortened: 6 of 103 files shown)


tegefaulkes:
I think this PR will fix #261 as well.

CMCDragonkai:
Are you going to create the PR on js-id too to include the additional static methods and instance methods for `IdInternal` there?

CMCDragonkai:
Are we still going to use a lambda and pass that around instead of passing the key manager to get the node id? I have a feeling that this isn't providing enough benefit to bother doing it. So I'm keeping how we are passing `KeyManager` everywhere.

CMCDragonkai:
I suggest these encoding and decoding utilities to be put into `nodes/utils.ts`:

```ts
function encodeNodeId(nodeId: NodeId): string {
  return idUtils.toMultibase(nodeId, 'base32hex');
}

function decodeNodeId(nodeIdString: string): NodeId | undefined {
  return idUtils.fromMultibase(nodeIdString) as NodeId | undefined;
}
```

Where `type NodeId = Opaque<'NodeId', Id>;`.

CMCDragonkai:
Will need this merged for me to finish off #266. I'm already put the above functions in, but overall NodeId changes will be significant across the codebase.

tegefaulkes:
One annoying quirk of using `NodeId` as a key to a map or object. it is seen as a string when iterated over.
```ts 
const nodeIds: Record<NodeId, null>
for (const nodeId in nodeIds) {
    // nodeId is a string here.
    const newNodeId: NodeId = IdInternal.fromString(nodeId)!;
}
```

CMCDragonkai:
That's to be expected. POJO keys are **always** strings.

CMCDragonkai:
Is there any other database of nodes other than NodeGraph? Remember the node id should used as a buffer key during lookup.

The gestalt graph should also be using the encoded NodeIds because they are likely to be shown to the end user in the GUI.

tegefaulkes:
I used `toBuffer()` in a few places where it was being used in the db. so far that has only been the ACL.

I've run into the gestalt graph already, I'll have a deeper look at it at a later stage.

tegefaulkes:
Most of the changes are done. 
- I need to take a closer look a `GestaltGraph` and make sure it's using the encoded form of `NodeId`.
- A lot of test may be failing due to comparing `NodeId`s with `toEqual()`. Should be using `toStrictEqual()`.
- Something weird is going on with `GRPCClientAgent` failing to extend `GRPCClient<AgentServiceClient>` in some tests with `TypeError: Class extends value undefined is not a constructor or null`. 

The error is very peculiar
```
    TypeError: Class extends value undefined is not a constructor or null

      19 | interface GRPCClientAgent extends CreateDestroy {}
      20 | @CreateDestroy()
    > 21 | class GRPCClientAgent extends GRPCClient<AgentServiceClient> {
         |                               ^
      22 |   /**
      23 |    * Creates GRPCClientAgent
      24 |    * This connects to the agent service

      at Object.<anonymous> (src/agent/GRPCClientAgent.ts:21:31)
      at Object.<anonymous> (src/agent/index.ts:2:1)

```

CMCDragonkai:
The `toEqual` should work the same as `toStrictEqual`. Maybe you mean `toBe`.

Also is this PR rebased on master? It seems there are conflicting files.

tegefaulkes:
My mistake. Anything comparing two objects such as the `NodeId` with `toBe` should use `ToEqual` or `toStringEqual` instead.
`toBe` will fail with;
```
Error: expect(received).toBe(expected) // Object.is equality

If it should pass with deep equality, replace "toBe" with "toStrictEqual"

Expected: [188, 56, 23, 20, 225, 85, 252, 109, 252, 39, 99, 174, 160, 84, 204, 189, 95, 29, 30, 165, 143, 100, 165, 65, 157, 237, 190, 55, 12, 47, 227, 72]
Received: serializes to the same string

```

tegefaulkes:
Some problems with the sigChain that need sorting out. One small problem is that it expected strings for the `NodeId` in the `ClaimData` and did some shortcuts with that POJO internally. This makes it a little tricky to neatly convert the `NodeId`s into strings internally. so for simplicity sake I'm going to have it use `NodeInEncoded` for the `NodeIds` instead.

CMCDragonkai:
Sigchain claims are JSON, so it will make sense to store them as encoded strings and not the original buffer. Unless of course there's an on-demand conversion when you read it out of the DB. But if the claims are stored as POJO values in the DB then storing it as an encoded string is fine.

On 24 January 2022 11:40:36 am AEDT, Brian Botha ***@***.***> wrote:
>Some problems with the sigChain that need sorting out. One small problem is that it expected strings for the `NodeId` in the `ClaimData` and did some shortcuts with that POJO internally. This makes it a little tricky to neatly convert the `NodeId`s into strings internally. so for simplicity sake I'm going to have it use `NodeInEncoded` for the `NodeIds` instead.
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/318#issuecomment-1019610473
>
>Message ID: ***@***.***>
-- 

tegefaulkes:
Most domains have been fixed up. currently failing test files are;
- NotificationsManager.test.ts
- bin/agent/status.test.ts
- notifications/utils.test.ts
- VaultOps.test.ts - Problem with mocking?
- VaultInternal.test.ts - Problem with mocking?
- vaultManager.test.ts
- notifications.test.ts
- bin/itentitites/identities.test.ts
- bin/nodes/claim.test.ts
- bin/keys/renew.test.ts
- bin/keys/reset
- client/service/keysKeyPairRenew.test.ts
- bin/nodes/add.test.ts
- client/service/keysKeyPairReset.test.ts
- client/rpcGestalts.test.ts
- client/service/notificationsSend.test.ts
- nodes/NodeGraph.test.ts - Random failures. fixed in nodes PR

CMCDragonkai:
Just a reminder, if there are places where you need to decode an encoded node ID, and it may return undefined, but you know for sure that it won't return undefined because you created it or took it from a trusted source, like for example the database. Then you can just use `!` to assert that it will be defined when you decode it.

We should however check for `undefined` whenever we are getting a NodeId from an untrusted source, and exceptions should be thrown on the boundary.

src/PolykeyAgent.ts
```diff
@@ -494,7 +495,7 @@ class PolykeyAgent {
         async (keyChangeData: RootKeyPairChangeData) => {
           this.logger.info('Propagating root keypair change');
           await this.status.updateStatusLive({
-            nodeId: keyChangeData.nodeId,
+            nodeIdEncoded: nodesUtils.encodeNodeId(keyChangeData.nodeId),
```
CMCDragonkai: Can you keep the key as `nodeId` here? The `status.json` keys should not change.

src/PolykeyAgent.ts
```diff
@@ -583,7 +584,7 @@ class PolykeyAgent {
       await this.sessionManager.start({ fresh });
       await this.status.finishStart({
         pid: process.pid,
-        nodeId: this.keyManager.getNodeId(),
+        nodeIdEncoded: nodesUtils.encodeNodeId(this.keyManager.getNodeId()),
```
CMCDragonkai: Same here, we should still be calling it `nodeId` on the outside world. The fact that it is encoded as a value is an implementation detail.

To the outside world, node ids are encoded. Internally node ids are buffers. That's how it should be understood.

src/acl/ACL.ts
```diff
@@ -130,8 +130,16 @@ class ACL {
     nodeId2: NodeId,
   ): Promise<boolean> {
     return await this._transaction(async () => {
-      const permId1 = await this.db.get(this.aclNodesDbDomain, nodeId1, true);
-      const permId2 = await this.db.get(this.aclNodesDbDomain, nodeId2, true);
+      const permId1 = await this.db.get(
+        this.aclNodesDbDomain,
+        nodeId1.toBuffer(),
```
CMCDragonkai, line 135: The `aclNodesDb` is `NodeId -> PermissionId`. You should ensure that both the key and value are raw buffers, and the `raw` option is set to true here so that the DB doesn't do any conversion/transcoding of the buffers. The `raw` parameter is the third parameter. It ensures that you return a `Buffer` directly.

Also this assumes that `PermId` is also transitioned to js-id as well, so it is also a buffer, however if that isn't the case, disregard this comment. But this is relevant to #299.

src/acl/ACL.ts
```diff
@@ -147,7 +155,7 @@ class ACL {
         Record<NodeId, Permission>
       > = {};
       for await (const o of this.aclNodesDb.createReadStream()) {
-        const nodeId = (o as any).key as NodeId;
+        const nodeId = IdInternal.fromBuffer<NodeId>((o as any).key)!;
```
CMCDragonkai: @tegefaulkes I believe you can shorten this to just `Id.create<NodeId>((o as any).key)`.

The `!` here asserts that it works. But if you know it works, you can also just pass the buffer directly into `create` method, rather than `fromBuffer`.

CMCDragonkai:
Did you remove the encoding/decoding functions from `NodeGraph`, node that you have the ability to use `NodeId` as a buffer directly? This includes:

* `nodeIdtoU8` in `src/nodes/utils.ts`

src/PolykeyAgent.ts
tegefaulkes: This has been changed back.

src/PolykeyAgent.ts
tegefaulkes: This has also been changed back.

tegefaulkes:
Yes, they have been removed.

tegefaulkes:
There are two very weird errors that are causing me problems right now.

 mocking for `keysUtils` is failing in some cases. causing an error such as `Error ... is not a function` Usually the first use of a `keysUtils` method. In some it seems to be caused by importing `keysutils` with `import * as keysUtils from '@/keys/utils';` in stead of `import { utils as keysUtils} from '@/keys';`.

A whole bunch of tests are failing to even run with the error
```

  ● Test suite failed to run

    TypeError: Class extends value undefined is not a constructor or null

      19 | interface GRPCClientAgent extends CreateDestroy {}
      20 | @CreateDestroy()
    > 21 | class GRPCClientAgent extends GRPCClient<AgentServiceClient> {
         |                               ^
      22 |   /**
      23 |    * Creates GRPCClientAgent
      24 |    * This connects to the agent service

      at Object.<anonymous> (src/agent/GRPCClientAgent.ts:21:31)
      at Object.<anonymous> (src/agent/index.ts:2:1)
```
 `src/agent/GRPCClientAgent.ts`. I have no idea why this is happening. The crazy thing is that the tests for `GRPCClientAgent` run without problem. even weirder is that the tests that have this error don't directly depend on `GRPCClientAgent`. it seems to be happening to mostly keys and identities tests?

tegefaulkes:
for the error `TypeError: Class extends value undefined is not a constructor or null`, googling the problem suggests that there is a circular import somewhere.

tegefaulkes:
Used a tool for checking cyclic dependencies. Looks like we have a few.
```
1) errors.ts > acl/errors.ts
2) errors.ts > agent/errors.ts
3) errors.ts > bootstrap/errors.ts
4) errors.ts > claims/errors.ts
5) errors.ts > client/errors.ts
6) errors.ts > gestalts/errors.ts
7) errors.ts > git/errors.ts
8) errors.ts > grpc/errors.ts
9) errors.ts > identities/errors.ts
10) errors.ts > keys/errors.ts
11) errors.ts > network/errors.ts
12) errors.ts > nodes/errors.ts
13) errors.ts > notifications/errors.ts
14) errors.ts > schema/errors.ts
15) errors.ts > sessions/errors.ts
16) errors.ts > sigchain/errors.ts
17) errors.ts > status/errors.ts
18) errors.ts > vaults/errors.ts
19) network/types.ts > keys/types.ts
20) claims/types.ts > identities/types.ts
21) nodes/types.ts > claims/types.ts
22) network/types.ts > keys/types.ts > nodes/types.ts
23) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > claims/utils.ts
24) nodes/types.ts > nodes/NodeConnection.ts > claims/utils.ts
25) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts
26) network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts
27) nodes/types.ts > nodes/NodeConnection.ts
28) nodes/types.ts > nodes/NodeConnection.ts > nodes/utils.ts
29) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts
30) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts
31) vaults/types.ts > vaults/VaultInternal.ts
32) config.ts > network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
33) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
34) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
35) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts
36) vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts
37) nodes/NodeGraph.ts > nodes/NodeManager.ts
```

These 4 cycles are new compared to master and may be causing the problem.
```
nodes/types.ts > nodes/NodeConnection.ts > nodes/utils.ts
config.ts > network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
```

CMCDragonkai:
Mutual recursion in `error.ts` is intended.

Maybe @emmacasolin can shed some light here? Did something get introduced during the test splitting?

CMCDragonkai:
Can you check if master has those errors in the tests? If so, we can use bisect to find the commit where it came in.

tegefaulkes:
It happened from my changes in this branch since the master branch doesn't have these errors. Likely from importing `nodesUtils` in a bunch of places.

tegefaulkes:
I think I fixed it? I had to split up an import.
```Diff
// In src/agent/GRPCClientAgent.ts
- import { GRPCClient, utils as grpcUtils } from '../grpc';
+ import * as grpcUtils  from '../grpc/utils';
+ import GRPCClient from '../grpc/GRPCClient';
```

tegefaulkes:
All of the tests that were failing to run are running now.

CMCDragonkai:
Yea generally speaking our utilities shouldn't have mutual imports. Can we break those up by ensuring that domain utils only imports root utils? And higher level domains can import lower level domains but not vice versa. And if 2 things need a common thing we can lift it up to a third thing or move it to root utils.

CMCDragonkai:
> Used a tool for checking cyclic dependencies. Looks like we have a few.
> ```
> 1) errors.ts > acl/errors.ts
> 2) errors.ts > agent/errors.ts
> 3) errors.ts > bootstrap/errors.ts
> 4) errors.ts > claims/errors.ts
> 5) errors.ts > client/errors.ts
> 6) errors.ts > gestalts/errors.ts
> 7) errors.ts > git/errors.ts
> 8) errors.ts > grpc/errors.ts
> 9) errors.ts > identities/errors.ts
> 10) errors.ts > keys/errors.ts
> 11) errors.ts > network/errors.ts
> 12) errors.ts > nodes/errors.ts
> 13) errors.ts > notifications/errors.ts
> 14) errors.ts > schema/errors.ts
> 15) errors.ts > sessions/errors.ts
> 16) errors.ts > sigchain/errors.ts
> 17) errors.ts > status/errors.ts
> 18) errors.ts > vaults/errors.ts
> 19) network/types.ts > keys/types.ts
> 20) claims/types.ts > identities/types.ts
> 21) nodes/types.ts > claims/types.ts
> 22) network/types.ts > keys/types.ts > nodes/types.ts
> 23) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > claims/utils.ts
> 24) nodes/types.ts > nodes/NodeConnection.ts > claims/utils.ts
> 25) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts
> 26) network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts
> 27) nodes/types.ts > nodes/NodeConnection.ts
> 28) nodes/types.ts > nodes/NodeConnection.ts > nodes/utils.ts
> 29) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts
> 30) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts
> 31) vaults/types.ts > vaults/VaultInternal.ts
> 32) config.ts > network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
> 33) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
> 34) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
> 35) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts
> 36) vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts
> 37) nodes/NodeGraph.ts > nodes/NodeManager.ts
> ```
> 
> These 4 cycles are new compared to master and may be causing the problem.
> ```
> nodes/types.ts > nodes/NodeConnection.ts > nodes/utils.ts
> config.ts > network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
> keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
> nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
> ```

@tegefaulkes 

Make `node/types.ts` **not** import NodeConnection. The types are imported by the classes but not the other way around.

Same goes for any other domains.

tegefaulkes:
That exists because of the object map.
```ts
type NodeConnectionMap = Map<
  NodeId,
  {
    connection?: NodeConnection;
    lock: MutexInterface;
  }
>;
```

I can remove that and move it to directly where it's needed.

tegefaulkes:
How do we want to handle validation of NodeIds? it used to be handled via the genericIdType conversion functions. do we want a similar mechanism here?

tegefaulkes:
Failing tests are;
- bin/nodes/find.test.ts
- nodes/NodeManager.test.ts - failing in master commit
- notifications/NotificationsManager.test.ts - failing in master commit
- nodes/NodeConnection.test.ts - failing in master commit
- discovery/Discovery.test.ts 
- bin/identities/identities.test.ts - discovery test failing, related to above.
- bin/nodes/claim.test.ts - Failure related to notifications
- bin/notifications/notifications.test.ts - invalid vault id - related to vaults and genericIdTypes
- vaults/VaultOps.test.ts - weird keysUtils mocking bug. cyclic imports related?
- agent/GRPCClientAgent.test.ts 
- bin/nodes/add.test.ts - Doesn't fail on invalid node, need some NodeId verification.
- bin/nodes/ping.test.ts - failing in master commit
- vaults/VaultInternal.test.ts - weird keysUtils mocking bug. cyclic imports related?
- vaults/VaultManager.test.ts - invalid vault id - related to vaults and genericIdTypes

tegefaulkes:
- bin/nodes/find.test.ts
- discovery/Discovery.test.ts
- bin/identities/identities.test.ts - discovery test failing, related to above.
- agent/GRPCClientAgent.test.ts
- vaults domain keysUtils mocking problem.

Are worth looking into first. The rest should be reviewed to see if they fit in this PR or fit into other in progress PRs.

emmacasolin:
Just a note that `notifications/NotificationsManager.test.ts` has been fixed in https://github.com/MatrixAI/js-polykey/pull/311

CMCDragonkai:
> That exists because of the object map.
> ```ts
> type NodeConnectionMap = Map<
>   NodeId,
>   {
>     connection?: NodeConnection;
>     lock: MutexInterface;
>   }
> >;
> ```
> 
> I can remove that and move it to directly where it's needed. 

Yea that should be in NodeConnectionManager.

CMCDragonkai:
> How do we want to handle validation of NodeIds? it used to be handled via the genericIdType conversion functions. do we want a similar mechanism here?

Syntactically should be done on the boundary any other validation depends on the domain.

tegefaulkes:
Ok, so I need to add `nodeId` validation on the boundary of the RPC and possibly the bin commands. Validation should be pretty simple, just check the number of bytes the `NodeId` has after decoding it.

CMCDragonkai:
Regarding the node ID validation, the multibase decoder will check if it is base32hex first, then if not, return undefined, if yes, check if the length is 32 bytes, if yes, then return NodeId, if not, return undefined. This means that any function using this `nodesUtils.decodeNodeId` may need to convert `undefined` into an exception. There can an exception in `nodes/errors.ts` called `ErrorNodeIdInvalid` (which might already exist) that can be thrown.

CMCDragonkai:
Regarding these cycles:

```
19) network/types.ts > keys/types.ts
20) claims/types.ts > identities/types.ts
21) nodes/types.ts > claims/types.ts
22) network/types.ts > keys/types.ts > nodes/types.ts
23) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > claims/utils.ts
24) nodes/types.ts > nodes/NodeConnection.ts > claims/utils.ts
25) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts
26) network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts
27) nodes/types.ts > nodes/NodeConnection.ts
28) nodes/types.ts > nodes/NodeConnection.ts > nodes/utils.ts
29) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts
30) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts
31) vaults/types.ts > vaults/VaultInternal.ts
32) config.ts > network/types.ts > keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
33) keys/types.ts > nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
34) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts > keys/utils.ts
35) nodes/types.ts > nodes/NodeConnection.ts > notifications/types.ts > vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts
36) vaults/types.ts > vaults/VaultInternal.ts > vaults/utils.ts
37) nodes/NodeGraph.ts > nodes/NodeManager.ts
```

The cycles involving nodes should be fully removed by #310.

The cycles involving vaults should be fully removed by #266.

All other cycles should be removed in this PR as well (perhaps as a separate commit).

---

@tegefaulkes what was the tool you used?

tegefaulkes:
The tool is called `madge` https://www.npmjs.com/package/madge

tegefaulkes:
Might be a small problem but we can't test two `NodeId`s for equality directly. EG
```ts
const node1: NodeId == ...;
const node2: NodeId == ...;

node1 === node2; // false
node1 == node2; // false
node1.toString() === node2.toString(); // true
```

One problem with this is that while we can use a `NodeId` as a key to a `Map` we can't actually use the key to retrieve from the map properly.

CMCDragonkai:
The node IDs are Uint8array objects. Testing for equality actually tests for object identity. As in are the 2 pointers pointing to the same object.

For equality of value, you can check the string form yes, and use Uint8array API given that it's an array and you check for array equality. Copy it from chai or another library.

As for putting them in the Map, this is expected. You either out it in as an object key or as a string form. You will need to remember which you did.

tegefaulkes:
This should be ready for review now. the only tests left are related to other PRs.

src/agent/service/nodesCrossSignClaim.ts
```diff
@@ -66,7 +67,7 @@ function nodesCrossSignClaim({
         }
         // Verify the claim
         const senderPublicKey = await nodeManager.getPublicKey(
-          payloadData.node1,
+          nodesUtils.decodeNodeId(payloadData.node1)!,
```
CMCDragonkai: Is it guaranteed that `payloadData.node1` is a valid and filled node id?

src/gestalts/GestaltGraph.ts
```diff
@@ -172,11 +174,11 @@ class GestaltGraph {
           const vertexKeys = gKs;
           gestalt.matrix[vertex] = vertexKeys;
           if (gId.type === 'node') {
-            const nodeInfo = await this.db.get<NodeInfo>(
+            const gestaltNodeInfo = await this.db.get<GestaltNodeInfo>(
```
CMCDragonkai: What was the reason for changing this type?

src/gestalts/types.ts
```diff
@@ -22,13 +23,18 @@ type GestaltIdentityKey = Opaque<'GestaltIdentityKey', string>;
...
 type Gestalt = {
   matrix: GestaltMatrix;
   nodes: GestaltNodes;
   identities: GestaltIdentities;
 };
+// Need a type that plays nice with the DB. Used internally only.
+type GestaltNodeInfo = {
+  id: NodeIdEncoded;
+  chain: ChainData;
+};
```
CMCDragonkai: The `GestaltNodeInfo` is new. What was the `NodeInfo` before, and why was this moved into `gestalts/types.ts`?

src/grpc/GRPCClient.ts
```diff
@@ -99,13 +101,13 @@ abstract class GRPCClient<T extends Client> {
       );
       // Encode as a URI in order to preserve the '+' characters when retrieving
       // in ForwardProxy from the http_connect_target URL
-      const encodedNodeId = encodeURIComponent(nodeId);
+      const encodedURINodeId = encodeURIComponent(nodeIdEncoded);
```
CMCDragonkai: I suggest this should be `nodeIdEncodedForURI`.

src/keys/utils.ts
```diff
@@ -176,11 +177,10 @@ function publicKeyToFingerprintBytes(
   return fString;
 }
 
-function publicKeyToFingerprint(publicKey: PublicKey): PublicKeyFingerprint {
+function publicKeyToNodeId(publicKey: PublicKey): NodeId {
```
CMCDragonkai: I suggest that since `NodeId` is defined in `nodes`, this function should instead be in `nodes/utils`.

Then the `KeyManager` can import `nodesUtils` and call that function.

Do we need `publicKeyToFingerprintBytes` function anymore too?

src/gestalts/GestaltGraph.ts
tegefaulkes: The Id doesn't convert back properly if it's if it's part of an object. so I had to convert it to a string before putting it into the DB.

src/gestalts/types.ts
tegefaulkes: This is just for use in the DB, NodeInfo is used elsewhere but due to a problem with putting and getting it from the DB i needed a version I could use internally in gestalts.

src/network/ForwardProxy.ts
```diff
@@ -299,7 +300,9 @@ class ForwardProxy {
       return;
     }
     const url = new URL(`pk://${request.url}`);
-    const nodeId = url.searchParams.get('nodeId') as NodeId | null;
+    const rawNodeId = url.searchParams.get('nodeId');
+    const nodeId =
+      rawNodeId != null ? nodesUtils.decodeNodeId(rawNodeId) : undefined;
```
CMCDragonkai: Replace this with just:

```ts 
const nodeIdEncodedForURL = url.searchParams.get('nodeId');
const nodeId = nodeIdEncodedForURL != null ? nodesUtils.decodeNodeId(nodeIdEncodedForURL) : undefined;
```

Use the same name as when we encoded as well. Keeps it tidy and searchable.

~You just need to change `decodeNodeId` to take `any` type and return `undefined` if it isn't a string.~

src/keys/utils.ts
CMCDragonkai: Actually disregard this comment, mainly because it uses alot of utilities inside `keysUtils` already. So conversion to `NodeId` stays here.

However still `publicKeyToFingerprintBytes`, if it is not used elsewhere, it could also be just embedded into this function.

src/acl/ACL.ts
tegefaulkes: `PermId` is an js-id and in that example is already being used as you described. same for all other times we get the PermId from the db.

src/agent/service/nodesCrossSignClaim.ts
tegefaulkes: It should be valid, yes.

src/gestalts/types.ts
CMCDragonkai: Ok I reproduced this.

The `NodeInfo` is using `NodeId`, and `Uint8Array` requires marshalling when going to and from JSON, because the DB will encode the data prior.

```
// '{"b":{"0":97,"1":98,"2":99}}'
s = JSON.stringify({ b: Uint8Array.from([97, 98, 99]) });
// { b: { '0': 97, '1': 98, '2': 99 } }
JSON.parse(s);
```

Note that this problem has been addressed in Node's buffer: https://nodejs.org/api/buffer.html#buftojson. The native json stringify will automatically encode the buffer into a special representation, however during JSON.parse, you have to then unmarshal it manually back to a buffer object.

The `JSON.parse` has a second parameter called `reviver` that can be used to unmarshal JSON strings in a custom way, and convert these back into ID. Same with `JSON.stringify` called `replacer` which can be used to marshal IDs appropriately.

So the choices are:

1. Use `NodeIdEncoded` when storing these `NodeId` into DB as values (as keys they can be stored as raw buffers as designed).
2. Create custom marshalling/unmarshalling to deal with this.

I think we should add a `toJSON` for `IdInternal` class, since we can use this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior.

At the very least this allows us to potentially decode from a custom JSON encoding as well. This seems to have general implications across the code wherever we use `NodeId` as values in the DB.

@joshuakarp thoughts?

src/agent/service/nodesCrossSignClaim.ts
CMCDragonkai: @joshuakarp is going to review all of these uses of assertions to catch any bugs in logic.

src/gestalts/GestaltGraph.ts
CMCDragonkai: Discussed here: https://github.com/MatrixAI/js-polykey/pull/318#discussion_r793264040

src/acl/ACL.ts
CMCDragonkai: Is `PermId` being encoded as JSON or as raw buffer when stored as a value in the DB?

src/network/ForwardProxy.ts
CMCDragonkai: Disregard:

~You just need to change `decodeNodeId` to take `any` type and return `undefined` if it isn't a string.~

src/keys/utils.ts
tegefaulkes: Removed `publicKeyToFingerprintBytes` and in-lined it into `publicKeyToNodeId`

src/network/ForwardProxy.ts
tegefaulkes: Fixed.

tegefaulkes: Ah, i'll roll that one change back.

CMCDragonkai: Yep, just change these 3 lines accordingly.

src/acl/ACL.ts
tegefaulkes: Looks like it's getting stored as a buffer. When it is retrieved it's used as a key for another `db.get` so it remains a buffer.

src/gestalts/types.ts
joshuakarp: Remember that `NodeInfo` is essentially a serialised map of sigchain data for some particular node, and this is stored within the `GestaltGraph` (found during discovery, etc).

I'd expect that the `GestaltGraph` is eventually going to be displayed in a human readable fashion. Therefore, the marshalling back and forth from the `UInt8Array` `NodeId` seems to be unnecessary, as in the end we'd want to display the encoded version of the `NodeId` anyway.

Would you agree with the above @CMCDragonkai ?

CMCDragonkai: I'm going to add `toJSON` method into the `IdInternal`. This should return something like:

```ts
{type: "IdInternal", data: [97, 98, 99]}
```

This will make it similar to Node buffers in this case, this is at least one step.

In addition, I'll add the `equals` method to allow easier more efficient equality comparisons between `Id` objects, and overloading `==` is not possible atm unless one is a primitive first.

CMCDragonkai: @joshuakarp currently `NodeInfo` is using `NodeId` and not `NodeIdEncoded`. If the `NodeInfo` will end up always showing the encoded version, then `NodeInfo` should still be used, but have its `id` become `NodeIdEncoded`, like:

```
type NodeInfo = {
  id: NodeIdEncoded;
  chain: ChainData;
};
```

Right now when I use `ag` to find all references of `NodeInfo`, I see it used in `GestaltGraph`, `NodeConnection` and `Discovery`. It seems `Discovery` and `GestaltGraph` are both limited to using the encoded node id for now. And `NodeConnection` is only using it as a comment.

In that case, for now it makes sense to continue using `NodeInfo`, but just update the type.

We should then understand that `NodeInfo` represents a human readable POJO of node information...

We may need to consider how `NodeInfo` would be used in the future to see if it is worth it to keep it in its raw buffer.

tegefaulkes: Using the Reviver and  Replacer seems like a good solution here. but we will have to add the `toJSON` method to `IdInternal` in `js-id`.

joshuakarp: > In that case, for now it makes sense to continue using NodeInfo, but just update the type.
> 
> We should then understand that NodeInfo represents a human readable POJO of node information...

I agree with this, and so agree that `NodeIdEncoded` is the correct type to be using here.

CMCDragonkai: @tegefaulkes can we switch back to `NodeInfo` in that case and just change to `NodeIdEncoded` atm? I just checked `IdentityInfo`, and that also uses string forms of `ProviderId` and `IdentityId` as well.

Seems like `NodeInfo` is primarily used by gestalts and sigchain related work, and we will refactor this when we get there if need be.

tegefaulkes: Did some quick testing.
```ts
// Added to IdInternal
  public toJSON(): string {
    return this.toString();
  }
```

example of usage.
```ts
const id = IdInternal.fromString('testString');
    const testObj = {
      test: 'test',
      id
    }

    const jsonId = JSON.stringify(id)
    const jsonObj = JSON.stringify(testObj);

    const revivedId = JSON.parse(jsonId, (_, value) => IdInternal.fromString(value));
    const revivedObj = JSON.parse(jsonObj, (key, value) => {
      if (key === 'id') return IdInternal.fromString(value)
      return value
    });

    expect(revivedId).toEqual(id);
    expect(revivedObj).toEqual(testObj);
  })
```

So it seems to work fine, but the only annoyance is that we have to tailor the reviver for each use.

src/agent/service/nodesCrossSignClaim.ts
joshuakarp: Given that `payloadData` is coming from a decoded claim from some other keynode, I'm not sure that we can assert that the node ID is valid.

The flow is as follows, where keynode `B` is verifying a claim from keynode `A`:
1. `B` receives the claim from `A`
2. `B` attempts to decode the claim (remember, internally it's a JWS with base64 fields) - this also does some structural validation with `ClaimNodeSinglySigned.json`, but does not validate the values of fields
3. `B` attempts to then use the node ID from the decoded claim

So here we actually don't have a guarantee that the node ID is valid (if we assume that a keynode can send a malformed claim).

src/gestalts/types.ts
tegefaulkes: I'll change it back to nodeInfo and switch that to using NodeIdEncoded. 

ignore my two comments, they're out of place in this discussion since the comments here didn't update while I was writing them.

src/agent/service/nodesCrossSignClaim.ts
joshuakarp: The remainder of the usage of node ID in `nodesCrossSignClaim.ts` looks fine to me though.

src/agent/service/nodesHolePunchMessageSend.ts
```diff
@@ -32,7 +32,7 @@ function nodesHolePunchMessageSend({
         // If so, ask the nodeManager to relay to the node
       } else if (
         await nodeManager.knowsNode(
-          nodesUtils.makeNodeId(call.request.getSrcId()),
+          nodesUtils.decodeNodeId(call.request.getSrcId())!,
```
joshuakarp: Similar to https://github.com/MatrixAI/js-polykey/pull/318/files#r793286473, I don't think we can make this assertion. This is on the service boundary, so we've performed no validation on the node ID to ensure it's correct.

src/bin/utils/processors.ts
```diff
@@ -219,7 +220,7 @@ async function processClientOptions(
     if (clientPort == null) clientPort = statusInfo.data.clientPort;
   }
   return {
-    nodeId,
+    nodeId: nodesUtils.decodeNodeId(nodeId)!,
```
joshuakarp: This makes the assumption that the node ID from `Status` will be valid. Can we ever have a `Status` file that contains invalid data?

src/bin/utils/processors.ts
```diff
@@ -286,21 +289,22 @@ async function processClientStatus(
     throw new binErrors.ErrorCLIStatusMissing();
   }
   if (statusInfo.status === 'LIVE') {
-    if (nodeId == null) nodeId = statusInfo.data.nodeId;
+    if (nodeIdDecoded == null)
+      nodeIdDecoded = nodesUtils.decodeNodeId(statusInfo.data.nodeId)!;
```
joshuakarp: Similar assumption here with `Status` (https://github.com/MatrixAI/js-polykey/pull/318/files#r793293939).

src/discovery/Discovery.ts
```diff
@@ -112,7 +113,8 @@ class Discovery {
         // The sigchain data of the vertex (containing all cryptolinks)
         let vertexChainData: ChainData = {};
         // If the vertex we've found is our own node, we simply get our own chain
-        if (vertexGId.nodeId === this.nodeManager.getNodeId()) {
+        const nodeId = nodesUtils.decodeNodeId(vertexGId.nodeId)!;
```
joshuakarp: ~~We make the assumption here that the `GestaltKey` stored in our local `GestaltGraph` will be containing a valid node ID. This seems fine.~~

Actually, I'm incorrect.

The `vertexGId` comes from other claims that we've discovered during the discovery process. That is, claims coming from other nodes. So again, I'm not sure we can make this assertion. Otherwise we assume that any claim from any node must have a valid node ID. 

This goes for all usages of `decodeNodeId` in `Discovery.ts`.

joshuakarp:
I need to continue to review the usage of `decodeNodeId` and the assertions made with respect to the broader context of nodes, gestalts, etc.

src/agent/service/nodesHolePunchMessageSend.ts
tegefaulkes: I must've missed that one.

src/bin/utils/processors.ts
tegefaulkes: It's possible. but not under normal operation.

tegefaulkes:
Barring any major that comes up in review. This should be ready to merge early tomorrow.

CMCDragonkai:
https://github.com/MatrixAI/js-id/pull/12 got merged and released as `3.3.0`. There are some important changes there:

1. The ability to use `toJSON` and `fromJSON` when dealing with JSON marshalling https://github.com/MatrixAI/js-polykey/pull/318#pullrequestreview-864380012 - note that there we don't need it for now, but it may be useful in the future if `NodeInfo` changes or other places require it.
2. The `equals` method is faster than doing string comparison. So any usage of id string comparison should use `id.equals` instead. Map will still use pointer comparison, POJOs will still use binary string comparison. The `equals` does not overload the `==` operator.
3. The `fromBuffer` and `fromString` both do not return `undefined`, so any usage of `!` assertions should not be needed.

src/gestalts/types.ts
CMCDragonkai: Going to resolve this for now then since we are using `NodeInfo` with `NodeIdEncoded`. But 3.3.0 allows for the json marshalling if we need it.

src/acl/ACL.ts
CMCDragonkai: Ok great. We bear in the mind that if is ever wrapped in JSON we have to use 3.3.0's ability of `fromJSON` and the `reviver` parameter.

CMCDragonkai:
What test failures are left to other PRs? Can you specify them in the PR description as well so we don't lose them @tegefaulkes.

src/nodes/NodeManager.ts
```diff
@@ -298,10 +315,10 @@ class NodeManager {
       const claimId = c as ClaimIdString;
       const payload = verifiedChainData[claimId].payload;
       if (payload.data.type === 'node') {
-        const endNodeId = payload.data.node2;
+        const endNodeId = nodesUtils.decodeNodeId(payload.data.node2)!;
```
joshuakarp: Same as https://github.com/MatrixAI/js-polykey/pull/318#discussion_r793286473. No validation on node ID yet, so this makes an assumption that node2 of the node to node claim is a valid node ID.

src/nodes/NodeManager.ts
```diff
@@ -385,11 +402,11 @@ class NodeManager {
...
   public async relayHolePunchMessage(message: nodesPB.Relay): Promise<void> {
     const conn = await this.getConnectionToNode(
-      message.getTargetId() as NodeId,
+      nodesUtils.decodeNodeId(message.getTargetId())!,
     );
     await conn.sendHolePunchMessage(
-      message.getSrcId() as NodeId,
-      message.getTargetId() as NodeId,
+      nodesUtils.decodeNodeId(message.getSrcId())!,
+      nodesUtils.decodeNodeId(message.getTargetId())!,
```
joshuakarp: Again, on a service boundary. Do we expect every hole punch message to have valid source and target node IDs? Or can these be malformed during transit? I'd assume the latter to be safer.

src/bin/utils/processors.ts
CMCDragonkai: We would like to have `StatusInfo` contain a legitimate `NodeId`. To do this, we would need to ensure that during the `readStatus` call, it should returning a proper `NodeId` within `StatusInfo`.

So `StatusLive` should be using `NodeId` and `NodeIdEncoded`, and the `readStatus` and `writeStatus` should be decoding and encoding the node Id respectively.

To ensure that we have a proper `NodeId`, you want to have the `nodesUtils.decodeNodeId` function check if the resulting length is 32 bytes. If that's incorrect, then it should return `undefined`. This gives us a light weight validation upon decoding the node id.

In terms of validation, it should always occur at the boundaries of the app. In this line inside `processClientOptions`, the correct boundary is actually the `readStatus` method of `Status`, not this particular function. So if the node id is invalid, it will throw `ErrorStatusParse`.

Now there's a separate problem to deal with which the tests have dummy node ids that don't fit the spec for what a node id is. Right now we have to keep with a set size for node ids, which is 32 bytes. This is because the node graph cannot work with arbitrary sized node ids. These tests should be updated to use a dummy node id generator which exist in `tests/utils.ts`.

```ts
const nodeId: NodeId = IdInternal.create(await keysUtils.getRandomBytes(32));
```

So tests should be updated to use legitimate node ids like above, and we can see that `StatusInfo` is a place where `NodeId` has to be used instead of `NodeIdEncoded`.

src/agent/service/nodesHolePunchMessageSend.ts
CMCDragonkai: Service handlers are boundaries of the PK app on the network level. So validating the node ID by decoding it first should be done early in the service handler. If it fails to work, an exception should be thrown here. These validation exceptions can be thrown from the `src/agent/errors.ts`.

There can be a common set of exceptions that happen for both client service and agent service, and even in the CLI as well when there is some validation being done on input data before they get "decoded" into proper types. Where should these exceptions go? It's not ideal to spread them out being `src/bin/errors.ts` and `src/agent/errors.ts` and `src/client/errors.ts`. Some exceptions are put into the domains themselves `ErrorInvalidNodeId` is in `src/nodes/errors.ts` but is this actually used by the nodes domain?

For now, it can be kept in `src/nodes/errors.ts` in this situation, but in the future where there are other common exceptions, we may need to create a separate validation domain with `validation/errors.ts` to keep these errors together.

So `validateNodeId` should just be rolled into `decodeNodeId`, and try to keep the errors consistent like `ErrorXInvalid` vs `ErrorInvalidX`.

tegefaulkes:
`NodeId` validation is now done inside of `nodesUtils.decodeNodeId()`. I've fixed up all of the `NodeId`s in tests to use a random `NodeId` generator. just need to check the tests are still working.

src/agent/service/nodesHolePunchMessageSend.ts
tegefaulkes: `decodeNodeId` now validates the `NodeId` all parts of the code using this have been updated.

src/bin/utils/processors.ts
tegefaulkes: All relevant tests have been updated to use `generateRandomNodeId()`. I haven't updated  Status to do the verification directly yet.

CMCDragonkai:
Fleshed out spec about how `NodeId` is used regarding `StatusInfo` and `NodeInfo` (and there being an exception for `NodeInfo` atm).

src/bin/utils/processors.ts
tegefaulkes: Fixed Status to use `NodeId`. had to add it's own reviver and replacer to make it work.

tegefaulkes:
Tests that are not getting resolved here are;

These were failing in the commit I branched off of.
- ping.test.ts
- find.test.ts
- notificationsManager.test.ts

relevant to the vaults domain. 
- vaultManager.test.ts

Part of nodes and #310 
- NodeConnection.test.ts
- nodeManager.test.ts

tests/nodes/utils.ts
```diff
@@ -99,4 +97,18 @@ async function nodesConnect(localNode: PolykeyAgent, remoteNode: PolykeyAgent) {
...
 
-export { generateNodeIdForBucket, incrementNodeId, nodesConnect };
+function generateRandomNodeId(): NodeId {
+  // Make it easy to read with an identifier
+  const base = 'Node-' + Math.trunc(Math.random() * 10000).toString(16) + '-';
+  const random = keysUtils.getRandomBytesSync(32 - base.length);
+  return IdInternal.fromBuffer<NodeId>(
+    Buffer.concat([Buffer.from(base), random]),
+  );
+}
```
CMCDragonkai: I think `base` doesn't require the extra random number. You already getting random bytes.

Also this should be in `tests/utils.ts`.

CMCDragonkai: Here's a better example:

```ts
function generateRandomNodeId(): NodeId {
  // Make it easy to read by using hex
  const random = keysUtils.getRandomBytesSync(16).toString('hex');
  return IdInternal.fromBuffer<NodeId>(
    Buffer.from(random),
  );
}
```

tegefaulkes: It has been updated.

PR #325, opened 2022-02-07
CLI Retry Options due to Intermittent Network

CMCDragonkai:
### Description

This introduces retry options to `CommandFind` and `CommandPing`.

This was extracted out of #266 because it wasn't related to vaults functionality.

Still trying to find what the reason for these options are for, and perhaps this should be considered in light of the CLI over-all design.

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/src/bin/errors.ts
+++ b/src/bin/errors.ts
@@ -51,28 +51,12 @@ class ErrorCLIFileRead extends ErrorCLI {
   exitCode = sysexits.NOINPUT;
 }
 
-class ErrorSecretPathFormat extends ErrorCLI {
-  description = "Secret name needs to be of format: '<vaultName>:<secretPath>'";
-  exitCode = 64;
-}
-
-class ErrorVaultNameAmbiguous extends ErrorCLI {
-  description =
-    'There is more than 1 Vault with this name. Please specify a Vault ID';
-  exitCode = 1;
-}
-
-class ErrorSecretsUndefined extends ErrorCLI {
-  description = 'At least one secret must be specified as an argument';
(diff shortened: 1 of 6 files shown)


CMCDragonkai:
@joshuakarp do you have any comments about this? What would be the usecase of these retry options?

CMCDragonkai:
Remember, that sometimes retry can be done by the end user as well. They can do things like `watch -n 1` command. Shell scripting can be used as well. But my question is what is the purpose of retrying for `CommandFind` and `CommandPing`?

Note that `ping` command itself repeatedly pings. Is the `CommandPing` meant to do this as well?

Also this PR currently has the incorrect short parameters in the tests. Always use the long parameters when testing or scripting CLI calls.

CMCDragonkai:
This requires re-review due to big changes in our nodes ping and connection logic in #326.

CMCDragonkai:
So primarily the idea here is that one could do this for a few commands where it supports it. The `--retry-count` will retry something up to a limit, while `--retry-interval` will be the delay between retries.

At the same time, one must define what it means to retry something. What exceptions are considered retryable... maybe it depends on the command itself. It's not exactly generic.

At any case if this is needed, this needs to rebased and redone due to big changes in PK.

CMCDragonkai:
Closed because work is done on Polykey-CLI now.

CMCDragonkai:
New issue tracking this https://github.com/MatrixAI/Polykey-CLI/issues/16

PR #321, opened 2022-01-28
General Data Validation - Boundary IO locations should be doing data validation and marshalling (and the decommissioning of GenericIdTypes for all IDs)

CMCDragonkai:
### Description

Boundary validation seems like a general problem that people are hitting over and over. It's best to fix this once and for all across all domains to ensure that we always have a consistent way of doing boundary validation.

#318 was merged with problems involving general data validation.

It's come to attention that data validation is not consistently applied across all of the IO boundaries that PK does.

Data validation is a very crucial task that we've missed, I originally thought that the combination of GRPC protobuf and the domain types would be enough. But it appears it is not, and our IO boundaries require a general validation system.

I've used general validation systems/libraries in the past for many different applications, mostly web SaaS, and I've never liked any of them.

After working with `JSON.parse`, I got some inspiration that combines the ideas from `JSON.parse`'s `reviver` parameter and the usage of recursion schemes used in functional languages, and the boundary issue mentioned in the Yesod framework https://www.yesodweb.com/book/introduction to create a `parse` function that do triple duty:

1. Be capable of validating nested data structure without any static constraints
2. Be capable of doing IO during validation, in order to validate complex dynamic data relationships
3. Be capable of sanitisation and type-marshalling and any other transformations that could occur

The lack of tuple structures, tuple-based pattern matching in JS means the composition of graph-operating function fragments is less syntactically nice, however the general principle of recursion schemes is followed.

Example prototype:

<details>

```ts
import { CustomError } from 'ts-custom-error';

class ErrorParse extends CustomError {
  public readonly errors: Array<ErrorRule>;
  constructor(errors: Array<ErrorRule>) {
    const message = errors.map((e) => e.message).join('; ');
    super(message);
    this.errors = errors;
  }
}

class ErrorRule extends CustomError {
  public keyPath: Array<string>;
  public value: any;
  public context: object;
  constructor(message?: string) {
    super(message);
  }
}

/**
 * Functional data-validation, inspired by recursion schemes and `JSON.parse` reviver
 * Post-fix depth first traversal on the data structure tree
 * Capable of modifying the data structure during traversal
 * Works like `JSON.parse` but on POJOs rather than JSON strings
 * Compose validation rules into the `rules` function
 */
async function parse(
  rules: (keyPath: Array<string>, value: any) => Promise<any>,
  data: any,
  options: { mode: 'greedy' | 'lazy' } = { mode: 'lazy' }
): Promise<any> {
  const errors: Array<ErrorRule> = [];
[...]

PR #331, opened 2022-02-14
Fixing the GRPC `toError()` bug with `ErrorValidation`

emmacasolin:
### Description
There is currently a bug with the new `ErrorValidation`, which was brought in as part of the new validation system in #321. Since `ErrorValidation` extends `ErrorPolykey`, but also uses a different constructor, `toError()` is unable to properly construct `ErrorValidation`s when throwing errors. Since we want `ErrorValidation` to be able to be used as both a standard `ErrorPolykey`, but also as a way of joining together multiple `ErrorParse`s into a single error, we need to do something like this:
```
class ErrorValidation extends ErrorPolykey {
  description = '...';
  exitCode = sysexits.DATAERR;
  public errors: Array<ErrorParse>;

  constructor(message, data) {

    if (data.errors != null) {
      let errors = [];
      for (const eData of data.errors) {
        const errorParse = new ErrorParse(eData.message);
        errorParse.keyPath = ...eData.keyPath;
        errors.push(errorParse);
      }
      this.errors = errors;
    }


  }

  static createFromErrors(errors: Array<ErrorParse>): ErrorValidation {
    const message = errors.map((e) => e.message).join('; ');
    const data = {
      errors: errors.map((e) => ({
        message: e.message,
        keyPath: e.keyPath,
        value: e.value.valueOf(),
      })),
    };
    const e = new ErrorValidation(message, data);
    e.errors = errors;
    return e;
    // super(message, data);
    // this.description = 'Input data failed validation';
    // this.exitCode = sysexits.DATAERR;
    // this.errors = errors;
  }
}
```

### Issues Fixed
* Fixes https://github.com/MatrixAI/js-polykey/pull/321#issuecomment-1033215131

### Tasks
1. [x] Fix the constructor inside `ErrorValidation` to match `ErrorPolykey`
2. [x] Add a static method to `ErrorValidation` to deconstruct an array of `ErrorParse`s
3. [x] Tests for both usages of `ErrorValidation`, including with `toError()`

### Final checklist
* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/validation/errors.ts
+++ b/src/validation/errors.ts
@@ -2,24 +2,45 @@ import { CustomError } from 'ts-custom-error';
 import { ErrorPolykey, sysexits } from '../errors';
 
 /**
- * This packages the `ErrorParse` array into the `data` property
- * This is to allow encoding to and decoding from GRPC errors
+ * Generic error containing all parsing errors that occurred during
+ * execution.
  */
 class ErrorValidation extends ErrorPolykey {
-  public readonly errors: Array<ErrorParse>;
-  constructor(errors: Array<ErrorParse>) {
+  description = 'Input data failed validation';
+  exitCode = sysexits.DATAERR;
+  public errors: Array<ErrorParse>;
+  constructor(message, data) {
+    super(message, data);
+    if (data.errors != null) {
+      const errors: Array<ErrorParse> = [];
+      for (const eData of data.errors) {
+        const errorParse = new ErrorParse(eData.message);
+        errorParse.keyPath = eData.keyPath;
+        errorParse.value = eData.value;
+        errorParse.context = eData.context;
+        errors.push(errorParse);
+      }
+      this.errors = errors;
+    }
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


emmacasolin:
The bug has been fixed, however there aren't tests for the majority of GRPC handlers that validation errors are handled correctly. This could be added either in this PR or in #311.

emmacasolin:
Lintfixed + all tests passing. I added some tests to check that `toError()` is working correctly with some of the handlers that already have split tests on master. The rest can wait until #311. This is now ready for review/merge.

CMCDragonkai:
This looks good, the only thing is to make the commit message more descriptive. State what you are actually fixing in this case:

```
Fixing `ErrorValidation` serialisation over GRPC by putting `ErrorParse` data into `ErrorValidation.data.errors`
```

Would be more valid..

CMCDragonkai:
After you change the message, proceed to merge.

emmacasolin:
Changed the commit name - merging now.

PR #351, opened 2022-02-25
Adding `pk agent restart` CLI command and GRPC handler

emmacasolin:
### Description
There are some circumstances where we may want to restart a Polykey agent, for example:
- If functionality is added in the future that requires a restart to take effect
- If we want to delete data that cannot otherwise be removed (e.g. sigchain data, registered providers) using the `fresh` parameter
- We want to change certain configurations that are set during the `PolykeyAgent.start()` command (e.g. network config)

In these cases it may be useful to have an `agent restart` command that is accessible to the CLI. This has been implemented here and can be brought in if deemed useful.


CMCDragonkai:
This command will likely be useful for when PK is being updated, or config changes, or extension changes. But in doing so, we should also consider how to do this entire start/stop operation gracefully. That is, we could imagine rather than fully killing the process, we just "reload" the process image, or just re-start the domains without restarting the agent itself, since agent state are all in the domains. It will require a fully agent-lock though, and if we need to consider graceful restart, we need to consider putting a lock on all network operations.

CMCDragonkai:
Needs rebase and redesign due to significant changes in PK.

CMCDragonkai:
Closed because work is done on Polykey-CLI now.

CMCDragonkai:
New issue tracking this https://github.com/MatrixAI/Polykey-CLI/issues/15

PR #370, opened 2022-04-06
fix: bug with GPRC client call through proxy

### LINKED ISSUE #369: GRPC call failure when multiple nodes connecting to 1 seed node.
### Specification

There seems to be a problem with the GRPC client calls in a very specific circumstance. In the test `tests/nat/noNAT.test.ts:239` `'agents in different namespaces can ping each other via seed node'` we are trying to connect to the same seed node from two agents. the sequence is 
1. seed1 starts
2. seed2 starts
3. agent1 starts
4. agent2 fails to start.

The problem here is during the polykey.start() sequence in agent2 we are calling `nodeConnectionManager.syncNodeGraph()`. this is making a agent-agent GRPC call. It seems that any client GRPC call to the seed1 causes a error `ErrorGRPCClientCall: Generic call error - 13 INTERNAL: Received RST_STREAM with code 2 triggered by internal client error: Protocol error` to be thrown.

This only happens within this test. trying the same procedure manually causes no problem. Replicating the circumstances using  polykey agents and `NodeConnectionManager`s to make the GRPC calls works fine as well.

The error implies that the the stream is being closed because of a protocol error during communication. it is hard to tell where this is coming from.

We need to work out why this problem is happening.

Some things to verify;
1. If we see `WARN:ConnectionForward 127.0.0.1:53659:Client Error: Error: write EPIPE` is it because the `clientSocket` is closed and the `tlsSocket` is trying to write data to it and thus emitting an `EPIPE`.
6. If `GRPCClient` is triggering the `clientSocket` to be closed on the `ConnectionForward`
7. Why is `GRPCClient` closing, and what is causing the subsequent protocol error `ErrorGRPCClientCall: Generic call error - 13 INTERNAL: Received RST_STREAM with code 2 triggered by internal client error: Protocol error`.

### Additional context

- Related #357 
- Related comments
  - https://github.com/MatrixAI/js-polykey/pull/357#issuecomment-1085257576
  - https://github.com/MatrixAI/js-polykey/pull/357#issuecomment-1085416097
  - https://github.com/MatrixAI/js-polykey/pull/357#issuecomment-1085435593
  - https://github.com/MatrixAI/js-polykey/pull/357#issuecomment-1085463939

### Tasks

1. Explore and discuss the problem further.
2. Fix up logger naming and handle connections to `0.0.0.0` to prevent accidental usage: https://github.com/MatrixAI/js-polykey/issues/369#issuecomment-1087043771
3. Refactor how `createConnection` works in NCM so it's not outputting that it's creating a connection multiple times, even though it fetches it from the cache https://github.com/MatrixAI/js-polykey/issues/369#issuecomment-1087055754
4. Change the `withConnF` error handling to use the `ResourceRelease` and pass the error in there, and reduce all the relevant error instances to a limited set of exceptions, preferably only those that relate to connection failures. Consider how this may be delegated by connectivity state changes though (let's avoid duplicate logic between connectivity state and exceptions there). https://github.com/MatrixAI/js-polykey/issues/369#issuecomment-1087056950

tegefaulkes:
### Description
This PR is to address the issue with #369 and other small fixes.

This is basically the fix:

```
      this.tlsSocket.once('data', () => {
        this.serverSocket.pipe(this.tlsSocket as TLSSocket, { end: false });
      });
```

We couldn't get `pause` or `cork` to work. So now we only start piping the server socket to the tls socket on the reverse side upon first receiving data.

### Issues Fixed

* Fixes #369 

### Tasks
1. [X] fix the `RST_stream`  GRPC bug. Do this by delaying any response through the proxy till after we receive the MAGIC frame.
1. [X] Add in a check for `handleConnectForward` for the wild card addresses since they should not be connectable. - this could be done in `Proxy.establishConnectionForward` because it is used by both `handleConnectForward` and `openConnectionForward`, and it would throw a specific exception for this. Have a look at `src/network/errors.ts` to see what is acceptable.
2. [X] Add additional tests for this in `Proxy.test.ts`
3. [X] Ensure that `NodeConnection.name` is embedded in the `NodeConnection` logger.
4. [X] update logging in `NodeConnectionManager` to be less confusing when creating and using connections.
6. [x] test for if the proxy connections are established first and then client connects.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/network/utils.ts
+++ b/src/network/utils.ts
@@ -29,6 +29,10 @@ function isHost(host: any): host is Host {
   return isIPv4 || isIPv6;
 }
 
+function isHostWildcard(host: Host): boolean {
+  return host === '0.0.0.0' || host === '::';
+}
+
 /**
  * Validates hostname as per RFC 1123
  */
@@ -353,6 +357,7 @@ export {
   pingBuffer,
   pongBuffer,
   isHost,
+  isHostWildcard,
   isHostname,
   isPort,
   toAuthToken,
--- a/tests/nodes/NodeConnectionManager.termination.test.ts
+++ b/tests/nodes/NodeConnectionManager.termination.test.ts
@@ -355,6 +355,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
+        },
         logger: logger,
       });
 
@@ -408,6 +411,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
+        },
         logger: logger,
       });
       const agentNodeId = polykeyAgent.keyManager.getNodeId();
@@ -483,6 +489,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
+        },
         logger: logger,
       });
 
@@ -551,6 +560,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
+        },
         logger: logger,
       });
 
@@ -624,6 +636,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
+        },
         logger: logger,
       });
 
@@ -697,6 +712,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
+        },
         logger: logger,
       });
 
--- a/src/network/ConnectionReverse.ts
+++ b/src/network/ConnectionReverse.ts
@@ -295,7 +295,9 @@ class ConnectionReverse extends Connection {
         await this.stop();
       });
       this.tlsSocket.pipe(this.serverSocket, { end: false });
-      this.serverSocket.pipe(this.tlsSocket, { end: false });
+      this.tlsSocket.once('data', () => {
+        this.serverSocket.pipe(this.tlsSocket as TLSSocket, { end: false });
+      });
       this.clientCertChain = clientCertChain;
       this.logger.info('Composed Connection Reverse');
     } catch (e) {
--- a/src/network/Proxy.ts
+++ b/src/network/Proxy.ts
@@ -434,6 +434,13 @@ class Proxy {
           timer,
         );
       } catch (e) {
+        if (e instanceof networkErrors.ErrorProxyConnectInvalidUrl) {
+          if (!clientSocket.destroyed) {
+            await clientSocketEnd('HTTP/1.1 400 Bad Request\r\n' + '\r\n');
+            clientSocket.destroy(e);
+          }
+          return;
+        }
         if (e instanceof networkErrors.ErrorConnectionStartTimeout) {
           if (!clientSocket.destroyed) {
             await clientSocketEnd('HTTP/1.1 504 Gateway Timeout\r\n' + '\r\n');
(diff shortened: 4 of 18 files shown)


CMCDragonkai:
It turns out that `networkUtils.resolvesZeroIP` is used by `network/Connection.ts` to resolve zero IPs to localhost. This works, but all our logging gets confusing.

So to avoid future confusion, then we should just be preventing any usage of zero IP by using a `isHostWildcard` check can be done.

The question is where should this be done? During the construction of `Connection` or done also in the `Proxy` where connections are being opened forward and reverse and as well as via the `CONNECT` handler.

src/network/utils.ts
```diff
@@ -29,6 +29,13 @@ function isHost(host: any): host is Host {
   return isIPv4 || isIPv6;
 }
 
+function isHostWildcard(host: Host): boolean {
+  const [isIPv4] = Validator.isValidIPv4String(host);
+  if (isIPv4 && host === '0.0.0.0') return true;
+  const [isIPv6] = Validator.isValidIPv6String(host);
+  return isIPv6 && host === '::';
+}
+
```
CMCDragonkai: Why not just do `host === '0.0.0.0' || host === '::'`?

tests/nodes/NodeConnectionManager.termination.test.ts
```diff
@@ -408,6 +411,9 @@ describe(`${NodeConnectionManager.name} termination test`, () => {
       polykeyAgent = await PolykeyAgent.createPolykeyAgent({
         password,
         nodePath: nodePath,
+        networkConfig: {
+          proxyHost: '127.0.0.1' as Host,
```
CMCDragonkai, line 415: The `proxyHost` is the external facing host of the proxy connections now.

The default I believe is `::` or `0.0.0.0` which means bind to all interfaces. But during testing you are forcing it to be localhost to avoid polluting the network.

Is that your intention with this?

CMCDragonkai:
@emmacasolin if you do the above simple fix, does it solve your problem from before?

emmacasolin:
> @emmacasolin if you do the above simple fix, does it solve your problem from before?

I haven't tested it using namespaces yet, however @tegefaulkes has said that it fixes the case that we were using for testing before inside `noNAT.test.ts` https://github.com/MatrixAI/js-polykey/issues/369#issuecomment-1089850961

emmacasolin:
> > @emmacasolin if you do the above simple fix, does it solve your problem from before?
> 
> I haven't tested it using namespaces yet, however @tegefaulkes has said that it fixes the case that we were using for testing before inside `noNAT.test.ts` [#369 (comment)](https://github.com/MatrixAI/js-polykey/issues/369#issuecomment-1089850961)

Confirming that this fix also works for agents/seed nodes that are in different namespaces.

tests/nodes/NodeConnectionManager.termination.test.ts
emmacasolin: I think in this case it's because later on the proxy host/port is retrieved using

```ts
await nodeGraph.setNode(agentNodeId, {
  host: polykeyAgent.proxy.getProxyHost(),
  port: polykeyAgent.proxy.getProxyPort(),
});
```

Which would return `0.0.0.0` (the default value) if not set explicitly to `127.0.0.1`. Alternatively instead of using `polykeyAgent.proxy.getProxyHost()` to get the host you would need to set it to `127.0.0.1` there instead, otherwise you'll be trying to connect to `0.0.0.0`.

src/network/utils.ts
tegefaulkes: That works too. I'll fix it up.

tests/nodes/NodeConnectionManager.termination.test.ts
tegefaulkes: What @emmacasolin said was the idea. It's the easiest way to ensure that everything has the correct IP, including the logging.

CMCDragonkai: If `getProxyHost()` returns `0.0.0.0`. This is technically correct. Because it is infact hosted on all IP addresses in the OS.

This is why if you want to target a server, that may have been bound on `0.0.0.0`, you must then use `networkUtils.resolvesZeroIP()` to ensure that `0.0.0.0` or `::` gets "resolved" to `127.0.0.1` or `::1`. You cannot just use the host parameter as is.

I provisioned for this before inside `Connection` constructor which uses that `resolvesZeroIP`, but I didn't expect the confusion that would arise from logging, which would still use the original address of `0.0.0.0` or `::` when the `Connection` was constructed. This conversion still exists in `Connection` in this PR.

What we changed here, is to actually disallow targeting wildcard addresses when attempting to connect to another host in the whole `network/Proxy.ts`, even though internally it can handle it. I think this is good because it ensures explicit checking and prevents confusion with our logging.

In the future, we could use the type system to help prevent this. If we had "negation" types. That is a function could accept `Host` type EXCEPT where the `Host` is a wildcard address. https://catchts.com/type-negation

src/network/utils.ts
tegefaulkes: Pushed up the change.

tests/nodes/NodeConnectionManager.termination.test.ts
tegefaulkes: Anything here I need to address in this PR? or is this some notes on what needs to be done at some point?

CMCDragonkai: Nothing to change. It's good.

PR #366, opened 2022-03-28
WIP: Upgrading @matrixai/async-init, @matrixai/async-locks, @matrixai/db, @matrixai/errors, @matrixai/workers, Node.js and integrating @matrixai/resources

CMCDragonkai:
### Description

Several core libraries have been updated and so they need to be integrated into PK. Refer to https://github.com/MatrixAI/js-encryptedfs/pull/63 to see how this was handled for EFS and repeat here.

1. `@matrixai/async-init` - no major changes
2. `@matrixai/async-locks` - all instances of `Mutex` should change to use `Lock` or `RWLockWriter`
3. `@matrixai/db` - the DB now supports proper DB transactions and has introduced `iterator`, however no more sublevels (use `KeyPath`/`LevelPath` instead)
4. `@matrixai/errors` - we make all of our errors extend `AbstractError<T>` and also provide static descriptions to all of them, as well as use the `cause` chain
5. `@matrixai/workers` - no major changes here
6. `node.js` - we should upgrade to Node 16 in order to integrate promise cancellation, as well as using `Promise.any` for ICE protocol
7. `@matrixai/resources` - since the `@matrixai/db` no longer does any locking, the acquisition of the `DBTransaction` and `Lock` has to be done together with `withF` or `withG`

Along the way, we can explore how to deal with indexing #188 and #257, which should be easier now that DB has root levels of `data`, `transactions`.

#### Locking changes

<details>

There were three important types of locks brought in by `js-async-locks`:

1. Lock (this is the equivalent of `Mutex` which we were using previously)
2. RWLockWriter (a write-preferring read/write lock)
3. LockBox (a collection of related locks)

In most cases, we are removing locks in favour of using optimistic concurrency control (OCC). This means that most domain locks should be removed with a few exceptions:

[...]

PR #326, opened 2022-02-08
Testnet Deployment

joshuakarp:
### Description

Once MatrixAI/Polykey#310 has been merged, we'll finally be able to move onto the deployment of our testnet into AWS.

I foresee this to be achievable in a few stages:
1. Deploy Polykey onto AWS and ensure we have a working remote keynode (#194)
2. Create the testnet seed nodes and store these securely somewhere (#285)
3. Use these seed nodes for our domain name (#177)

* Relates to MatrixAI/Polykey#159 
* Relates to MatrixAI/Polykey#161 

### Issues Fixed

* ~~Fixes MatrixAI/Polykey#324~~ being completed in MatrixAI/Polykey#310

Testnet deployment:

* Fixes MatrixAI/Polykey#194
* Fixes MatrixAI/Polykey#285 
* Fixes MatrixAI/Polykey#177 

NodeGraph Structure:

* Relates MatrixAI/Polykey#244 
* Fixes MatrixAI/Polykey#317 
* Fixes MatrixAI/Polykey#158
* Fixes MatrixAI/Polykey#359 

Node Adding Policies:

* Fixes MatrixAI/Polykey#212 
* Fixes MatrixAI/Polykey#322
* Fixes MatrixAI/Polykey#344 
* Fixes MatrixAI/Polykey#363 

Node Removal Policies:

* Fixes MatrixAI/Polykey#150 
* Fixes MatrixAI/Polykey#226 
* Fixes MatrixAI/Polykey#345

### Tasks

- ~~[ ] 2. Complete MatrixAI/Polykey#324~~ - being completed in MatrixAI/Polykey#310
- [x] 1. Read through and revisit all previous issues
- [ ] 3.  Complete MatrixAI/Polykey#194
  * Ultimately provides us with a deployment of Polykey on AWS!
  * Task list there should be sufficient to follow process.
- [ ] 4.  Complete MatrixAI/Polykey#285 
- [ ] 5. Complete MatrixAI/Polykey#177 
- ~~[ ] 6. Created automated testing that utilises the testnet~~ - to be done in MatrixAI/Polykey#357 
  - ~~[ ] - MatrixAI/Polykey#159~~
  - ~~[ ] - MatrixAI/Polykey#161~~
- [x] 7. Add agent service `nodes` tests https://github.com/MatrixAI/js-polykey/issues/194#issuecomment-1032173720
  * [x] - `nodesChainDataGet`
  * [x] - `nodesClosestLocalNode`
[...]

Issue #290, opened 2021-11-15
Create IPC locking library `@matrixai/js-file-locks` to introduce RWlocks in IPC


joshuakarp:
### Specification

We'd like to investigate creating our own locking library, based off our current usage of `fd-lock`.

Previously we were using both [`proper-lockfile`](https://github.com/moxystudio/node-proper-lockfile) and [`fd-lock`](https://github.com/mafintosh/fd-lock) in a more conglomerated fashion. Now, we are purely using `fd-lock` for the `Status` (our refactored agent lockfile https://github.com/MatrixAI/js-polykey/pull/283) and `Session` usage (implemented [here](https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/213)).

Note the distinction between "locking" here. Here, we're referring to "file locking" (as per the `fd` prefix). For example, for the `Status`, we lock this status file to assert that an agent process is currently executing. Conversely, we're using `async-mutex` for intra-process locking (locking some resource from within a process, as opposed to between processes). Remember that we also have an existing [read-write locking implementation](https://gist.github.com/CMCDragonkai/4de5c1526fc58dac259e321db8cf5331) that utilises `async-mutex`.

Our own library should support:
* OS-specific read-write locks
* Native promise API
* Explore the addition of POSIX locks

OS-specific read-write locks should be the initial priority, and should be seen as a basic extension of functionality from `fd-lock` (we can use this as a source of inspiration to copy from).

### Additional context
* Original issue: MatrixAI/Polykey#251
* MatrixAI/Polykey-CLI#18 - Should also allow an API to check if the lock is already locked so we can enable concurrent coalescing

### Tasks

1. ...
2. ...
3. ...

CMCDragonkai:
Note that `async-mutex` is an in-process locking, while file locking is for inter-process locking. We would want to expose rwlocks directly from flock and equivalent in windows for this to work.

CMCDragonkai:
The `src/utils.ts` is now `src/utils`. It's now a directory with `src/utils/locks.ts` which contains all locking utility functions.

Here is where the `RWLock` based on `async-mutex` is. Can be used for other locking utilities that use file locking, or more abstractions are built on top of IPC file locking or in-memory async-mutex (promise-based) locking.

CMCDragonkai:
[...]

Issue #257, opened 2021-10-21
Integrate autoindexing into the `vaults` domain


joshuakarp:
### Specification

Once autoindexing has been implemented into `js-db`, this should be brought into `js-polykey` and integrated into the `vaults` domain. 

### Additional context

* autoindexing issue in `js-db` https://github.com/MatrixAI/js-db/issues/1 and the corresponding PR https://github.com/MatrixAI/js-db/pull/2
* usage of the vault name for creation of a vault https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/205#note_700936060 and when enforcing uniqueness on creation https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/205#note_692109270

### Tasks

Provide support for:
1. Using the index to look up by vault name (currently, it performs a linear search over all vaults)
2. Using the index to look up by node ID from the `remote` field (i.e. a cloned vault that originally came from the specified node ID) - this would mean the node ID needs to be part of the metadata
3. Using the index to enforce the uniqueness constraint on vault name and to prevent duplicate vault names

CMCDragonkai:
We're not going to do autoindexing anymore. See reasons in https://github.com/MatrixAI/js-db/issues/1

PR #374, opened 2022-05-20
Upgrading lib dependencies and node.js version

### LINKED ISSUE #294: Fine Grained Concurrency Control Standardisation
### Specification

Locking in js-polykey has gone through alot of iterations. The most recent iteration is in `js-db` usage of locks where fine grained locks provided by `async-mutex` is used, as well as the `RWLock` abstraction in `src/utils.ts`.

For most domains such as `ACL`, the locks are too coarse-grained, causing one to lock the entire domain itself. Many of these locks should be replaced with the usage of DB transactions as it is done in `js-encryptedfs`.

In some places, fine grained locks can replace the existing coarse grained locking or even replace the usage of condition variables. Currently the `src/network/Connection.ts` and derived classes makes use of a `_composed` boolean which did double duty in terms of indicating when composition was done but as a way to prevented repeated concurrent calls of `compose`. The first duty is fine, but the second duty should done with a fine grained lock shared between the all the calls that should be blocked when composition operation is occurring. This means all the methods that currently check `_composed` and throw exceptions when it is not true.

Transactions has been introduced to js-db. With this we can replace a lot of the existing locking with the use of the new db transactions. The general changes that need to be implemented are as follows.

1. Updating any transact wrappers by removing them or using `withF`, `withG` locking directly.
2. In all cases where there is a conflict just throw it up the stack. We will expect to handle them within the handlers or look deeper into it later.
3. `ErrorDBTransactionConflict` Error should never be seen by the user. We should catch and override it with a more descriptive error for the context. 
4. Transactions should be started within the handlers and passed all the way down to where they are needed. The idea is to make each handler attomic. 
5. Concurrency testing should be introduced but only after SI transactions has be implemented.
6. All usage of DB should be updated to use the new API. This means removing sublevels and utilising `LevelPath` and `KeyPath`s instead.
7. All usage of db streams should be replaced with the db iterator.
8. All instances of `db.put`, `db.get` and `db.del` should be using transactions via `tran.put/get/del`

This applies to all domains that make use of DB OR domains that depend on others that make use of DB. The goal here is to make any even starting from the handlers atomic. 

There are limitations to this however. Since a transaction can fail if there is overlapping edits between transactions. We can't really include changes to the db that will commonly or guarantee conflict. Example of this are counters or commonly updated fields. So far this has been seen in;

1. `NotificationsManager`. Makes use of a counter so any transactions that include Adding or removing a notification WILL conflict. Reads also update metadata so concurrently reading the same message WILL conflict.
2. More to follow?

Some cases we will need to make use of locking along with a transaction. A good example of this is in the `NotificationManager` where we are locking the counter update. When this is the case we need to take extra care with the locking. Unless the lock wraps the whole transaction it is still possible to conflict on the transaction. we can't compose operations that rely on this locking with larger transactions. 

An example of this problem is.
```
start tran1
start tran2

start lock
	tran1 update counter
end lock

start lock
	tran2 update counter
end lock

end tran1
end tran2 // conflict!

To avoid this tran2 has to start after tran1 ends. 
We need to force serialisation of the transactions here.
As a result we can't compose with a transaction outside of
the lock.
```

This means that some operations or domains can't be composed with larger transactions. It has yet to be seen if this will cause an issue since more testing is required to confirm any problem. I suppose this means we can't mix pessimistic and optimistic transactions. So far it seems it will be a problem with the following domains.

1. Vaults domain - Lifecycle and editing of vaults relies heavily on locks. 
2. NotificationsManager - Locking is needed for the counter and preventing other conflicts. 

**Note that this has nothing to do with IPC locking as in #290.**

### Additional Context

* https://github.com/MatrixAI/js-db/issues/4 - true snapshot isolation
* https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/213#note_740902479 - Issue involving the testing of `NodeConnection` and network connections and having a race condition with `this._composed` being true, but the properties that should only be used when composition finishes being undefined.
* https://github.com/MatrixAI/js-db/pull/10 - Enabling `withF` and `withG` usage in DB

### Tasks
1. Updating any transact wrappers by removing them or using `withF`, `withG` locking directly.
2. In all cases where there is a conflict just throw it up the stack. We will expect to handle them within the handlers or look deeper into it later.
3. `ErrorDBTransactionConflict` Error should never be seen by the user. We should catch and override it with a more descriptive error for the context. 
4. Transactions should be started within the handlers and passed all the way down to where they are needed. The idea is to make each handler attomic. 
5. Concurrency testing should be introduced but only after SI transactions has be implemented.
6. All usage of DB should be updated to use the new API. This means removing sublevels and utilising `LevelPath` and `KeyPath`s instead.
7. All usage of db streams should be replaced with the db iterator.
8. All instances of `db.put`, `db.get` and `db.del` should be using transactions via `tran.put/get/del`

### LINKED ISSUE #304: Integrate Error Chaining in a js-errors or @matrixai/errors package
### Specification

Now that `@matrixai/errors` is ready, we can use it to incorporate error chaining into PK. This will involve a series of steps:

#### Chaining

In all places where we are catching an error and then throwing a new error in its place, we need to be including the original error as the `cause` of the new error, e.g.
```ts
} catch (e) {
  throw new SomePolykeyError(e.message, { cause: e });
}
```

Note that the `cause` should always be an exception/error. If not specified it defaults to `unknown`. When a error is thrown, the top-level error should contain the full instances of the errors in its cause chain, as one big nested object. This data structure can be serialised and deserialised recursively, where every error in the chain has a single cause, and that cause is contained as a property within it.

It is important to realise that the `replacer` will remove any entry if the returned value is `undefined`. Unless it is an array, it will replace that value with `null`. This means that because our `cause` may be `undefined`, it may not exist during deserialisation. Any usage of `fromJSON` must be aware that the `cause` property may not exist.

#### Client Server Error Architecture

```
toError: ServiceError -> ErrorPolykeyRemote                         fromError: Error -> ServerStatusResponse

                                                                         ┌─────────────────────────────┐
      ┌───────────────────────────────┐                                  │ NodeId 1 - Server           │
      │ Client                        │             Request              │                             │
      │                               │                                  │ ┌─────────────────────────┐ │
      │ Receives toError ServiceError ├──────────────────────────────────┼─► Service Handlers        │ │
      │ Throws ErrorPolykeyRemote     │                                  │ │                         │ │
      │ cause: ErrorX | ErrorUnknown  ◄──────────────────────────────────┼─┤ Sends fromError ErrorX  │ │
      │ data: { nodeID: 1, ... }      │                                  │ └─────────────────────────┘ │
      └───────────────────────────────┘      Response with error         │                             │
                                             JSON string in metadata     └─────────────────────────────┘
                                          error: { type: ErrorX, ... }
```

It's important to realise:

* Only the client constructs `ErrorPolykeyRemote`
* The `ErrorPolykeyRemote` should contain `nodeId: NodeId`, `host`, `port` and other connection information in its own `data` which can aid debugging which server responded with this error
* The `ErrorPolykeyRemote` should also have information about which call triggered this, and perhaps even request information, it can have a lot of information which can add later.
  - Additional client/connection metadata should be set in the `data` property of `ErrorPolykeyRemote`
  - Give `data` a more specific type like `data: ClientMetadata & POJO` so that it forces that `ClientMetadata` must be available
  - We can use `type ClientMetadata = { nodeId: NodeId; host: Host; port: Port }`. And additional information can be provided later. Like what the call was.
  - The `ClientMetadata` type should be in `src/types.ts` to avoid import loops, since it has to use the `NodeId` type from `nodes` and `Host` and `Port` from the the `network` domains.
  - This keeps the type along with `ErrorPolykeyRemote` together at the top level `src`.
* The server only serialises the `ErrorX` that `ErrorX` may be any kind of error. The server must filter the information it sends to the client to ensure it is not leaking sensitive information. This can include the `stack` since the `stack` is rarely required by the client.
* Even when the server sends the error back to the client, it must also log out this error, but only if the error is considered a "server error". That is if the is a "client error" such as 4xx errors in HTTP, then it should not be logged out. This includes:
  - Authentication
  - Validation
  - Precondition
  - Concurrency Conflict
  - Anything else coming from HTTP 4xx
  - Example: missing object/resource
* If the client encounters unknown data during deserialisation, it only returns `ErrorUnknown` at the root of the JSON data. Any other unknown data should be returned as-is.
* This means all remote errors are **always** `ErrorPolykeyRemote`, but the `cause` may be `ErrorUnknown` or `ErrorX`. Where `ErrorX` may contain anything after that (as long as the runtime schema checking during `fromJSON` calls work)
* The GRPC metadata key for the serialised error JSON should just be `error`, in the future if we switch to using JSON RPC, this error JSON will probably be encoded as part of the JSON RPC protocol - see https://www.jsonrpc.org/specification#error_object and https://eth.wiki/json-rpc/json-rpc-error-codes-improvement-proposal
* GRPC itself only supports "HTTP2" kind of errors, but these codes aren't really being used, all errors from the remote node is considered one kind of HTTP error, that being of `UNKNOWN`. We use that to represent an "application" error. This basically means we don't use most of GRPC/HTTP2's standard error codes, as our "application layer errors" are on top. This is because GRPC/HTTP2's standard error codes were designed more for the HTTP layer. Although some of the HTTP error codes are still used internally by GRPC.
* This is the same philosophy taken by JSON RPC over HTTP: http://www.simple-is-better.org/json-rpc/transport_http.html, where success or error at the application layer is still 200 response code at the HTTP layer. HTTP errors are specific to the HTTP layer.
  ```
  200 OK
  for responses (both for Response and Error objects)
  204 No Response / 202 Accepted
  for empty responses, i.e. as response to a Notification
  307 Temporary Redirect / 308 Permanent Redirect
  for HTTP-redirections (note that the request may not be automatically retransmitted)
  405 Method Not Allowed
  if HTTP GET is not supported: for all HTTP GET requests
  if HTTP GET is supported: if the client tries to call a non-safe/non-indempotent method via HTTP GET
  415 Unsupported Media Type
  if the Content-Type is not application/json
  others
  for HTTP-errors or HTTP-features outside of the scope of JSON-RPC
  ```

#### JSON serialisation/deserialisation

Our gRPC `toError` and `fromError` utilities will need to be modified to be able to serialise and deserialise the error chain when sending errors between clients/agents or agents/agents.

- These can utilise JSON's replacer/reviver properties of stringify/parse - this is non-recursive and processes objects in fragments, with everything being orchestrated by stringify/parse
- The replacer needs to serialise errors into a standardised structure `{ type: 'ClassName', data: { ... } }` and also filter out sensitive data (the stack) when errors are being sent to an agent (as opposed to a client) - it works top to bottom, creating a structure first and filtering out unwanted fields afterwards
- The reviver needs to covert this structure back into its original data type - it works bottom to top, converting fields to their correct types and then finally constructing the finished data type
  - If the received data can't be deserialised into a known error type it should be converted to an `ErrorPolykeyUnknown`, with the unknown data in the `data` field
- We also no longer need to separate errors in the gRPC metadata into `name`/`message`/`data` and this is now just one data structure - this can just be one `error` field
- Need to filter out sensitive information when sent from the service handlers, this includes `stack` and other special data. This would a case by case basis for specific exception classes. That is, the `replacer` will need to execute a filter by checking against specific exception classes and filtering them out. Filter rules should be acquired from `src/client` and `src/agent`.

#### Logging of Errors at the Service Handlers

- Loggers should be added to the service handlers in order to log out errors, since these do not shut down the agent process
  - Not all errors should be reported though - only validation/authentication/other precondition errors
  - This logger is separate from the logger used for grpc internals
  - How to determine if an error is for the client or for the server? This is a case by case basis, unlike HTTP, we do not have 4xx and 5xx codes emitted at the beginning, all exceptions assume programmatic usage. Therefore there is a smaller amount of client errors then there are server errors. The filter should just check that if part of "client errors" set, then don't emit. Create set of "client errors", this set can be defined in both `src/client` and `src/server` (code duplication is fine here), since each service can define their own set of what is considered a "client error".

#### Reporting the Error to the user at the Root Exception Handler

We have 2 root exception handlers, one at the client and one at the agent. This addresses the client side, however unless things are different, the same applies to the agent side.

On the client side, this is done in `src/bin/polykey.ts`.

When an exception is received, we must interpret 3 things:

* The `cause` chain
* The `exitCode`
* The metadata in `data`

According to issue #323, the `--format json` doesn't currently affect the STDERR. This needs to be done now because the errors being reported are quite complex, and during testing, we expectation utilities should be parsing JSON to make it easier to test.

Therefore `binUtils.outputFormatter` will need to ensure that `error` type is a human formatted JSON, while `json` type can now be the JSON output for exceptions. This is doable now due to the `toJSON` utility function inherited from `AbstractError`.

Furthermore in order to acquire the options passed into the command, use `rootCommand.opts()`, this method was added into to `PolykeyCommand` to enable us to acquire the options.

```ts
const opts = rootCommand.opts();
if (opts.format === 'json') {
  process.stderr.write(
    binUtils.outputFormatter({
      type: 'json',
      data: e.toJSON(),
    });
  );
} else {
  // use `error` format
  process.stderr.write(
    binUtils.outputFormatter({
      type: 'error',
      // ...
    });
  );
}
```

The desired format for human format should be something like:

```
ErrorName: description - message
  K\tV
  K\tV
  cause ErrorName: description - message
    K\tV
    K\tV
```

Note the usage of `\t` for separation while spaces are used for indentation. We can change this format later.

The `exitCode` is currently a bit ambiguous, it was originally intended to mean that if this exception was the last exception caught by the process, this code should be used for the process exit code.

It seems more ideal to allow the process to decide what the exit code should be based on the family of exceptions. But since we have built our exit code this way, we need to continue to use it like this.

However now that we have a `cause` chain, we have to decide how to deal with the exit code when we are getting `ErrorPolykeyRemote`. Because of this scenario:

```
C -> A1 -> A2
```

We must differentiate exceptions that originate from the client, or the first agent or the second agent. If an exception comes from A2, we may see `ErrorPolykeyRemote` wrapping another `ErrorPolykeyRemote`.

So right now a policy can be made:

1. Find the first `cause` property that is not `ErrorPolykeyRemote`, and use that `exitCode`
2. Otherwise get the `exitCode` of the first exception.

This only works because `ErrorPolykeyRemote`'s cause type is limited to `ErrorPolykey`, it cannot be undefined or anything else.

### Additional context

* https://github.com/MatrixAI/js-db/commit/5172d01e694e7f0f58023fb41c76844da0a6284f
* https://github.com/mathew-kurian/TraceError.js - can use this stack style for reporting
* https://github.com/MatrixAI/js-logger/issues/3 - Needs design integration with js-logger so these 2 libraries would work together
* https://github.com/MatrixAI/js-polykey/pull/321#issuecomment-1028563774 - discussion on error handling between local node and remote node for agent & client service, realising the need for additional metadata indicating where the exception comes from, as well as handling non-ErrorPolykey errors
* https://github.com/MatrixAI/js-polykey/pull/311#issuecomment-1049484685 - discussion of how error chaining will make debugging of exceptions that originate from outside the domain being tested easier
* https://github.com/MatrixAI/js-errors/pull/1#issuecomment-1086832650 - Initial ideas on error serialisation and deserialisation in this thread

### Tasks

- [x] 1. Convert all errors from `CustomError` to new `AbstractError<T>` from `@matrixai/errors`
- [x] 2. Give all errors a static description and `errCode`
- [x] 3. In all places where one error is caught and a different one is thrown, include the original error as the `cause`
- [x] 4. Modify the validation utils to use new error chaining
- [x] 5. Modify gRPC `fromError` and `toError` functions to properly handle chained errors
- [x] 6. Include a timestamp when any error is thrown
- [x] 7. Add error logging to service handlers
- [x] 8. Update all packages relying on `js-errors` to `1.1.0` to use the `fromJSON` and `toJSON`
- [x] 9. PK should adapt its replacer and reviver functionality and bring in "rules" from both `src/client` and `src/agent`. Can use DI in the client service vs the agent service to change the serialisation/deserialisation.
- [x] 10. Logging should also use separate rules, one for client service, one for agent service.
- [x] 11. Handle the error chain at the root exception handler for `src/bin/polykey.ts` and `src/bin/polykey-agent.ts`
- [x] 12. Ensure that `AbstractError.fromJSON` can handle non-existent `cause` property

### LINKED ISSUE #323: Format option should affect the output formatting of exceptions on STDERR for the CLI
### Specification

Now that we centralised all of our error functionality into https://github.com/MatrixAI/js-errors, we have a more structured way of creating and propagating way. Due to this, we need our output for these errors to become more formalised as well in order to display this information. We need to be able to communicate the information from our errors to both humans and machines, so we need to modify the `--format` option to apply to the STDERR as well as the STDOUT.

Readable details included in both styles of output would need to include:

* Structured error information coming from https://github.com/MatrixAI/js-logger/issues/3
* Chained error information coming from MatrixAI/Polykey#304 
* Extra data that is put into the `data` property
* Whether the exception came from local or from a remote PK node https://github.com/MatrixAI/js-polykey/issues/304#issuecomment-1028561787
* Contextual information about parsing/validation errors

The json output is the most simple to achieve, since we now have `toJSON` and `fromJSON` methods for our errors that allow them to be serialised and deserialised easily. `toJSON` is called automatically whenever `JSON.stringify()` is called, so nothing needs to change about how we already handle the output of json data (except for handling non-polykey errors, which have no `toJSON` method).

A bit more thought needs to go into the human-readable output style, for example:
* Indentation: we want to indent every error in the cause chain, and we want to indent error properties underneath the name of the error
* Formatting: we want to use tabs between keys and values for readability, and keys should be displayed in camelCase
* Data: we want to include important information without crowding the output - we want to minimise repetition (e.g. don't print the exitCode for remote errors since these are inherited from the cause, and don't print the stack for any errors)

There is one point of ambiguity, which is that of logging for the PK agent. When starting the PK agent, we can use the `--format` option, but this would affect the output of the starting command and the output of logs going to STDERR for the agent process. Should this be 2 separate options? I believe this has been discussed here https://github.com/MatrixAI/js-logger/issues/3#issuecomment-994229643 and the conclusion was that no, it should be 1 option, and it affects STDOUT, STDERR of the CLI command as well as the STDERR logging of the agent process if any.

This means that the format option needs to be propagated to the exit handlers, as well as the error loggers in `polykey-agent.ts`. This requires dynamic setting of the format property after it has been passed in as an option.

### Additional context

* https://github.com/MatrixAI/js-logger/issues/3 - We need structured logging in order to completely address this issue
* MatrixAI/Polykey#304 - Issue for error chaining including implementation details

### Tasks

1. Modify the output formatter to display more detailed error information (including the error chain)
2. Propagate the `--format` option to `CommandPolykey` (and all subcommands), the exit handlers, and `polykey-agent`

### LINKED ISSUE #356: Align `VaultInternal` `with` context API functions with `NodeConnectionManager`
### Specification
When performing operations with a vault, expect to pass a callback that receives the vault. I would want to align the API between `NodeConnectionManager` and `VaultManager`. This requires further specification.

Refer to #335 

### Additional context
* Relates #335 
* Relates #266 

### Tasks

1. ...
2. ...
3. ...

tegefaulkes:
### Description
This PR has been copied from #366 under a new branch to allow for branch and CI/CD protections.

Several core libraries have been updated and so they need to be integrated into PK. Refer to https://github.com/MatrixAI/js-encryptedfs/pull/63 to see how this was handled for EFS and repeat here.

1. `@matrixai/async-init` - no major changes
2. `@matrixai/async-locks` - all instances of `Mutex` should change to use `Lock` or `RWLockWriter`
3. `@matrixai/db` - the DB now supports proper DB transactions and has introduced `iterator`, however no more sublevels (use `KeyPath`/`LevelPath` instead)
4. `@matrixai/errors` - we make all of our errors extend `AbstractError<T>` and also provide static descriptions to all of them, as well as use the `cause` chain
5. `@matrixai/workers` - no major changes here
6. `node.js` - we should upgrade to Node 16 in order to integrate promise cancellation, as well as using `Promise.any` for ICE protocol
7. `@matrixai/resources` - since the `@matrixai/db` no longer does any locking, the acquisition of the `DBTransaction` and `Lock` has to be done together with `withF` or `withG`

Along the way, we can explore how to deal with indexing #188 and #257, which should be easier now that DB has root levels of `data`, `transactions`.

#### Locking changes

<details>

There were three important types of locks brought in by `js-async-locks`:

1. Lock (this is the equivalent of `Mutex` which we were using previously)
2. RWLockWriter (a write-preferring read/write lock)
3. LockBox (a collection of related locks)

In most cases, we are removing locks in favour of using optimistic concurrency control (OCC). This means that most domain locks should be removed with a few exceptions:

- Counters/serialised data - places where we have strict serialisation, such as counters and lexicographically ordered keys, require locking the DB in order to maintain this order. This applies to the keys used by the Discovery Queue in `Discovery.ts`, the `NotificationId`s and message count used by `NotificationsManager.ts`, the `ClaimId`s and sequence numbers used by `Sigchain.ts`, and if the set/ping node or refresh buckets queues introduced in #326 become persistent then we would need locking there as well. This could be achieved for these cases by introducing a `LockBox` to the affected domains and locking the relevant keys when we access the db, e.g. `withF([this.db.transaction(), this.locks.lock(...lockRequests)], async ([tran]) => {})` where `this.locks` is a `LockBox` and `...lockRequests` is an array of `[KeyPath, Lock]` (see EFS `INodeManager.ts`).
- Object maps - in `NodeConnectionManager` and `VaultManager` we need to lock groups of objects, this can be done using a LockBox where we are referencing the same ID each time we want to lock a specific object.

Everywhere else we expect that conflicts will be rare, so we don't use locks in order to simplify our codebase. In the case of a conflict, we can either retry (if safe) or bubble up the error to the user.

</details>

#### Errors changes

<details>

The new `js-errors` allows us to bring in error chaining, along with more standardised JSON serialisation/deserialisation (for sending errors across gRPC). With this error chaining ability, there are now three ways that we can handle/propagate errors:

1. Error re-raise/re-throw: This re-raises the same error, in this case it just means you couldn't handle the error.

```ts
try {
  throw new ErrorPolykey('something happened');
} catch (e) {
  throw e;
}
```

2. Error override: This overrides the error with a whole new error. Use this when you are wrapping a library of internal errors and presenting transformed external errors to the user.

```ts
try {
  throw new ErrorPolykey('something happened');
} catch (e) {
  throw new ErrorPolykey2(e.message);
}
```

3. Error chain: This raises a new error and chains the new error with the old error. Use this when your code represents an onion of different layers each capable of handling different errors.

```ts
try {
  throw new ErrorPolykey('something happened');
} catch (e) {
  throw new ErrorPolykey2(e.message, { cause: e });
}
```

In all places where we are catching one error and throwing a different error in its place, we should be using approach 3 (error chain). If we just want to bubble the original exception upwards then use approach 1 (re-raise/re-throw). Finally, if we want to hide the original error from the user (perhaps it contains irrelevant implementation details or could be confusing and thus requires additional context) we can use approach 2 (error override). There is a fourth approach that exists in Python for errors that occur as a direct result of handling another error, however, this does not exist in TypeScript (in such a case we would use approach 3). When using approach 2 (and in some cases approach 3) you may want to log out the original error in addition to throwing the new error.

##### JSON serialisation/deserialisation

When sending errors between agents/from client to agent we need to serialise errors (including the error chain if this exists). Then, on the receiving side, we need to be able to deserialise back into the original error types.

We are able to do this using `JSON.stringify()` (serialisation) and `JSON.parse()` (deserialisation). These methods allow us to pass in a replacer/reviver to aid with converting our error data structure, as well as being combined with `toJSON()` and `fromJSON()` utility methods on the error class itself. These are implemented on `AbstractError` from `js-errors`, however, we need to extend these to work with `ErrorPolykey` in order to handle the additional `exitCode` property. While `toJSON()` can simply call `super.toJSON()` and add in the extra field, `fromJSON()` needs to be completely reimplemented (although this can be copied from `AbstractError` for the most part). Similarly, the replacer and reviver can be based on the replacer and reviver used in the `js-errors` tests.

##### ErrorPolykeyRemote

Errors are propagated between agents and clients as follows:

```
toError: ServiceError -> ErrorPolykeyRemote                         fromError: Error -> ServerStatusResponse

                                                                         ┌─────────────────────────────┐
      ┌───────────────────────────────┐                                  │ NodeId 1 - Server           │
      │ Client                        │             Request              │                             │
      │                               │                                  │ ┌─────────────────────────┐ │
      │ Receives toError ServiceError ├──────────────────────────────────┼─► Service Handlers        │ │
      │ Throws ErrorPolykeyRemote     │                                  │ │                         │ │
      │ cause: ErrorX | ErrorUnknown  ◄──────────────────────────────────┼─┤ Sends fromError ErrorX  │ │
      │ data: { nodeID: 1, ... }      │                                  │ └─────────────────────────┘ │
      └───────────────────────────────┘      Response with error         │                             │
                                             JSON string in metadata     └─────────────────────────────┘
                                          error: { type: ErrorX, ... }
```

`ErrorPolykeyRemote` is constructed on the client side (not the server side) inside our gRPC `toError()` utility. After the received error is deserialised, it is wrapped as the cause property of a new `ErrorPolykeyRemote`, which should also contain the `nodeId`, `host`, and `port` of the agent that originally sent the error in its data property. In order to access this information, it needs to be passed through from wherever the client/agent method is called (this would be bin commands for the client service and domain methods for the agent service). The data can then be passed through to the `promisify...()` methods, which in turn call `toError()`.

##### Testing

Now that we have an error chain, we need to adjust our tests to be able to perform checks on these. In many cases where we were originally expecting some specific `ErrorPolykey` we will now be receiving an `ErrorPolykeyRemote` with the original error in its cause property. For simple cases like this it is simple enough to just perform the existing checks on the cause property of the received error rather than the top-level error, however this approach becomes more complicated for longer error chains. Additionally, we may want to perform checks on the top-level `ErrorPolykeyRemote` (such as checking the metadata for the sending agent).

In this case, it would be useful to create an expectation utility that allows one to perform checks on the entire error chain, from the top-level error to the final error in the chain. This could look something like this:

```ts
function expectErrorChain(thrownError: any, expectedErrors: Array<any>) {
  let checkingError = thrownError;
  for (error in errors) {
    // Check expected type first
    expect(checkingError).toBeInstanceOf(...);
    // Match against every expected property
    expect(checkingError.message).toBe(error.message);
    if (error instanceof ErrorPolykey) {
      expect(checkingError.data).toEqual(error.data);
      // ...
    }
    // Move along to the next in the cause chain
    checkingError = checkingError.cause;
  }
}
```

We could also think about using the `toJSON()` method on each error to allow us to use jest's `expect().toMatchObject()` matcher rather than having to check every error property individually. Also potentially including parameters to specify which properties of the error you do and don't want to check against.

##### Additional context
- https://github.com/MatrixAI/js-polykey/issues/304 - issue for error chaining/JSON serialisation and de-serialisation with additional implementation details in the description
- https://github.com/MatrixAI/js-errors/pull/2 - bringing JSON serialisation/de-serialisation into `js-errors`

</details>

#### Database changes

<details>

Changes include but are not limited to

1. Updating any transact wrappers by removing them or using `withF`, `withG` locking directly.
2. In all cases where there is a conflict just throw it up the stack. We will expect to handle them within the handlers or look deeper into it later.
3. `ErrorDBTransactionConflict` Error should never be seen by the user. We should catch and override it with a more descriptive error for the context. 
4. Transactions should be started within the handlers and passed all the way down to where they are needed. The idea is to make each handler attomic. 
5. Concurrency testing should be introduced but only after SI transactions has be implemented.
6. All usage of DB should be updated to use the new API. This means removing sublevels and utilising `LevelPath` and `KeyPath`s instead.
7. All usage of db streams should be replaced with the db iterator.
8. All instances of `db.put`, `db.get` and `db.del` should be using transactions via `tran.put/get/del`

This applies to all domains that make use of DB OR domains that depend on others that make use of DB. The goal here is to make any even starting from the handlers atomic. 

There are limitations to this however. Since a transaction can fail if there is overlapping edits between transactions. We can't really include changes to the db that will commonly or guarantee conflict. Example of this are counters or commonly updated fields. So far this has been seen in;

1. `NotificationsManager`. Makes use of a counter so any transactions that include Adding or removing a notification WILL conflict. Reads also update metadata so concurrently reading the same message WILL conflict.
2. More to follow?

Some cases we will need to make use of locking along with a transaction. A good example of this is in the `NotificationManager` where we are locking the counter update. When this is the case we need to take extra care with the locking. Unless the lock wraps the whole transaction it is still possible to conflict on the transaction. we can't compose operations that rely on this locking with larger transactions. 

An example of this problem is.
```
start tran1
start tran2

start lock
	tran1 update counter
end lock

start lock
	tran2 update counter
end lock

end tran1
end tran2 // conflict!

To avoid this tran2 has to start after tran1 ends. 
We need to force serialisation of the transactions here.
As a result we can't compose with a transaction outside of
the lock.
```

This means that some operations or domains can't be composed with larger transactions. It has yet to be seen if this will cause an issue since more testing is required to confirm any problem. I suppose this means we can't mix pessimistic and optimistic transactions. So far it seems it will be a problem with the following domains.

1. Vaults domain - Lifecycle and editing of vaults relies heavily on locks. 
2. NotificationsManager - Locking is needed for the counter and preventing other conflicts. 

</details>

#### Node.js changes

<details>

After upgrading to Node v16 we will be able to bring in some new features.

1. `Promise.any` - we are currently using `Promise.all` in our `pingNode()` method in `NodeConnectionManager` (this is being done in #326) but this should change to using `Promise.any`. This is because `Promise.all` waits for every promise to resolve/reject, however we only care about whichever finishes first and we want to cancel the rest. This change can be made in #326 after this PR is merged.
2. `AggregateError` - this error is emitted by `Promise.any` if all of the given promises are rejected. In our case this would mean that we were unable to ping the desired node via direct connection or signalling (and eventually relaying once this is implemented), so we may want to catch this and re-throw some other error to represent this. We will also need to add this into our error serialisation/deserialisation.
4. `AbortController` - there are a number of places that have been identified in #297 where we could use the new `AbortController`/`AbortSignal`. This can be done in a separate PR when rebasing #326.

</details>

### Issues Fixed

* Fixes #294 
* Fixes #304 
* Fixes #356
* Fixes #323 
* Related https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/37
* Related #244 
* Related #188 
* Related #257 
* Related https://github.com/MatrixAI/js-db/issues/1
* Related #290 - this will only be important for file locking related operations for Inter-Process communication

### Tasks

- [x] 1. Upgrade all exceptions to use `AbstractError` and use `cause` chain and static descriptions
- [x] 2. Remove all sublevel object maintenance
- [x] 3. Update all usages of `db.get` and `db.put` and `db.del` to use `KeyPath`
- [x] 4. Remove all uses of `batch` and domain locking and instead migrate to using `DBTransaction` and the `withF` and `withG` from `@matrixai/resources`
- [x] 5. Replace all uses of `createReadStream`, `createKeyStream` and `createValueStream` with `db.iterator`
- [x] 6. All domain methods that involve DB state should now take a `tran?: DBTransaction` optional parameter in the last parameter to allow one to compose a transaction context
- 7. ~If possible, get EFS to use the new DB as well, which can be done by making `DB` take a "prefix", thus recreating sublevels, but only for this usecase. This would mean both EFS and PK use the same DB, with EFS completely controlling a lower level.~ Put on hold, Requires an update to `js-db` and is low priority. Make a new issue for this?
- [x] 8. See where `RWLock` can be used instead when doing concurrency control on the `DBTransaction` to raise the isolation level to avoid non-repeatable reads, or phantom reads or lost-updates. - SI means that non-repeatable reads and phantom reads are not possible, so that waits on SI merge, while lost-updates is where the counter race condition occurred and that was resolved with just the `Lock`.
- [x] 9. Upgrade to Node 16 (follow procedure from https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/37)
- [x] 10. Integrate `@matrixai/async-locks` to use the `Lock` class instead of `Mutex` from `async-mutex`
- [x] 11. Use `@matrixai/resources` `withF` and `withG` to replace transactions
- [x] 12. Clean up left over commentary from migration
- [x] 13. Update nodeConnectionManager and VaultManager to use lockbox for the object maps.
- [x] 14. Update NodeConnectionManager withConnF to properly use new resources API features. This includes the `ResourceAquire` making use of error handling.
- [x] 15. Update usage of `DB.iterator`/`tran.iterator` as well as `keyAsBuffer`/`valueAsBuffer`
- [x] 16. Update to latest version of EFS with fixed `decrypt` and latest DB
- [x] 17. Filter error logging on the agent and client service handlers to only log out server errors
- [x] 18. Fix formatting of CLI errors to display as much (unmodified) info as possible
- [x] 19. Clean up commits
- [x] 20. Move `ClientMetadata` and `ErrorPolykeyRemote` under the `grpc` domain
- [x] 21. Concurrency checks and DB conflict retry in service handlers (pending https://github.com/MatrixAI/js-db/pull/19)
- [x] 22. Replace the `utils.promisify` here with https://github.com/MatrixAI/js-db/pull/19#issuecomment-1143273485

Tasks are divided by domain for conversion. Assignments are as follows. Domains labelled 'last' depend on a lot of other domains and will be divided later.
- [X] acl - Emma
- [x] agent - last
- [x] bootstrap - last
- [x] client - last
- [x] discovery - Brain
- [X] gestalts - Emma
- [X] identities - Emma
- [X] nodes - Emma , depends on sigchain
- [x] notifications - Brian
- [x] sessions - Emma
- [x] sigchain - Brain
- [X] vaults - Brian
- [x] polykeyAgent - last

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

TBD spec:

<details>

Error changes:

Testing for an error with a cause.

1. Pending changes from JSON serialisation/deserialisation from the new AbstractError
2. ErrorPolykeyRemote needs to have remote node metadata
3. Work out how ErrorPolykeyRemote is propagated from agent to agent to client.
6. Create new expectation utilities that can standardise how we expect things to fail and the checks on those exceptions, as well as the cause property
7. Bring in the standards of exception override vs exception chain into the spec

DB changes:

domains - Each domain used to have a `transact` or `withTransaction` wrapper. Unless this wrapper actually needs some additional context, and it is just calling `withF` and `withG` directly, then avoid this and just replace it with direct usage of `withF` and `withG` utilities. We should identify domains that require additional context, in those cases we continue to use a with transaction wrapper.

transaction, and its utilisation of SI transactions

```
withF([db.transaction()], async ([tran]) => {
    await tran.put(k1, ...);
    await tran.put(k2, ...);
});

withF([db.transaction()], async ([tran]) => {
    await tran.put(k2, ...);
    await tran.put(k3, ...);
});
```

```
ErrorDBTransactionConflict
  .data => [k1, k2, k3]
```

In all cases, you just throw this exception up, and propagate it all the way to the client.

The pk client however will change the message to be more specific to the user's action.

Retrying this can be identified and sprinkled into the codebase afterwards.

1. Work out a way to write a user-friendly message when a ErrorDBTransactionConflict occurs in the PolykeyClient (at the CLI)
2. Identify where automatic retries can occur, and make those catch the `ErrorDBTransactionConflict` and resubmit the transaction

1. Identify write-skews, this is where we must use solutions for dealing with write skews => locking and materialising the write conflict
2. Identify serialisation requirements - counter updates, and where PCC is demanded as part of the API, in those situations use the LockBox or lock

Handlers need to now start the transaction. For agent and client.

We have `tran?: DBTransaction` where it allows public methods to create their own transaction context.

But this is only for debugging and convenience.

During production, the transaction context is meant to be setup at the handler level.

This means that each handler is its own atomic operation.

By default they start out at the handler level, this will make it easier for us to identify our atomic operations and to compare them.

Tests - introduce concurrency tests, consider some new cocurrency expectation combinators that allow us to more easily say one of these results has to be X, and the rest has to be Y.

Concurrency tests can be introduced domain by domain. - Use `Promise.allSettled`

We can however focus our concurrency testing at the handler level, because we expect that to be the unit of atomic operations.

All ops removed, changed to using keypath and levelpath.

In the iterator:

* For `tran.iterator` you will always have the `key` no matter what because it is being used for matching between the overlay and underlying db data.
* The `keyAsBuffer` and `valueAsBuffer` are by default `true`, if you change them to false, it will decode the 2, and return as decoded values. If you need both, use `dbUtils.deserialize()`.
* More docs for tran.iterator
* Eliminate all subleveldown, we are no longer using that (can be brought back later, would only be useful to share db betweeen EFS and PK DB)
* Eliminate all streams

```
withF([db.transaction()], async ([tran]) => {
    await tran.put('k1', 'adfdsf);
    const i = tran.iterator();
    const r1 = await tran.get('k1')
    i.seek('k1')
    const r2 = i.next()
    // r1 !== r2 (possible)
    // with SI r1 === r2
});
```

Just think that the iterator has the same snapshot as the original transaction.

Don't start writing concurrency tests until we have SI transactions. Serial tests are still fine.

Nodev16 Changes:

1. Promise.any - for ICE protocol for the DC and signalling
2. AggregateError - its in our serialisation and deserialisation, but it's only emitted by Promise.any
3. AbortController - separate PR (when rebasing testnet deployment) - identify where these may be changed polling

Locking Changes:

All async locks and async mutex is to be replaced with `js-async-locks`.

In the `js-async-locks`:

1. Lock - mutex
2. RWLockWriter - read/write
3. LockBox - locking collection

Convert object-map pattern to using LockBox by referencing the same ID.

Use INodeManager in EFS as a guide.

</details>


tegefaulkes:
This PR needs to target the staging branch when that gets created.

emmacasolin:
The vaults client service tests are the only ones that aren't split like the others into separate test files. They also set up the client service differently from the other tests and are missing some parameters due to the errors changes. Going to split and align them with the other tests to try and get them running.

tegefaulkes:
I'm looking into merging the vault's EFS db with the main DB. At first blush it seems easy. We can just provide the DB to the EFS constructor. However I don't think that the DB supports prefix/sublevel anymore. Do We need to re implement that on the DB before we can continue with this? @CMCDragonkai

CMCDragonkai:
It would require the DB to maintain knowledge of a levelpath to be prepended to any queries.

But other questions remain, what about the root paths? Does the subDB maintain their own transactions and indexes? Or are they shared with the top DB?

Anyway for now not important we can address it later.

tegefaulkes:
Tests in `tests/notifications/NotificationsManager.test.ts` is currently failing randomly. I've narrowed it down to the function `getNotificationIds` returning a malformed `NotificationId` in the list. For example normally the `NotificationId` should be 16 bytes. but we are getting undersized ones.
```
    IdInternal(9) [Uint8Array] [
        0, 187,  28,   5,
        0, 119, 190, 113,
      109
    ]
```
If I had to guess there is a problem with the iterator reading the keys out properly. It could be that the a malformed ID is used when adding the message to the DB. But also possible there is a bug with the DB's handling of the path delimiter. This could happen if the ID itself contains the delimiter and that wasn't handled properly. 

I'd start with checking if the ID is malformed before it's put into the database. If it's the delimiter problem then the IDs that fail should always contain the same character.

CMCDragonkai:
> Tests in `tests/notifications/NotificationsManager.test.ts` is currently failing randomly. I've narrowed it down to the function `getNotificationIds` returning a malformed `NotificationId` in the list. For example normally the `NotificationId` should be 16 bytes. but we are getting undersized ones.
> 
> ```
>     IdInternal(9) [Uint8Array] [
>         0, 187,  28,   5,
>         0, 119, 190, 113,
>       109
>     ]
> ```
> 
> If I had to guess there is a problem with the iterator reading the keys out properly. It could be that the a malformed ID is used when adding the message to the DB. But also possible there is a bug with the DB's handling of the path delimiter. This could happen if the ID itself contains the delimiter and that wasn't handled properly.
> 
> I'd start with checking if the ID is malformed before it's put into the database. If it's the delimiter problem then the IDs that fail should always contain the same character.

The delimiter in the database is a the `\x00` null byte. There are tests in js-db that cover this usecase to ensure that it's properly escaping keys that have the `\x00` null byte.

I'm not sure if the js-id case is covered well enough, but @emmacasolin you can check the js-db tests.

Just make sure that this branch is actually using the latest js-db check (`package-lock.json` and `node_modules/@matrixai/db/package.json`).

emmacasolin:
> The vaults client service tests are the only ones that aren't split like the others into separate test files. They also set up the client service differently from the other tests and are missing some parameters due to the errors changes. Going to split and align them with the other tests to try and get them running.

Spec for this:

I've currently split all of the vaults client service tests (not secrets), however there were no existing tests for some of the handlers so those are currently todos.

Tests that have been split into separate files and are passing:
- `vaultsCreate`, `vaultsDelete`, and `vaultsList` (all combined into a single file since they're interdependent)
- `vaultsLog`
- `vaultsPermissionSet`, `vaultsPermissionUnset`, and `vaultsPermissionGet` (again, combined due to interdependence)
- `vaultsRename`
- `vaultsVersion`

Tests that have been split into separate files but are still todo:
- `vaultsClone`
- `vaultsPull`
- `vaultsScan`

Tests that are written but are yet to be split:
- [x] `vaultsSecretsNew`/`vaultsSecretsDelete`/`vaultsSecretsGet`
- [x] `vaultsSecretsNewDir`/`vaultsSecretsList`
- [x] `vaultsSecretsMkdir`
- [x] `vaultsSecretsEdit`
- [x] `vaultsSecretsRename`
- [x] `vaultsSecretsStat`

emmacasolin:
I'm noticing a pattern with the malformed ids.

If you look at the valid ids, they all look something like this:
```
   6,  40, 197, 143, 103, 104, 112,   0, 187, 239, 222, 246, 239, 105, 136, 223
   6,  40, 197, 143, 103, 194, 112,   0, 153, 154,  25,  17,  55,  81,   7, 147
   6,  40, 197, 143, 105,  71, 112,   0, 153,  51, 165, 184,  95, 177,  94, 116
   6,  40, 197, 143, 108, 192, 112,   0, 148,  42, 166, 251,  64, 233, 161,  23
   6,  40, 197, 143, 108, 208, 112,   0, 158, 229,  40,  54, 129, 244,  55, 177
   6,  40, 197, 143, 108, 221, 112,   0, 131, 130,  32,  78, 126, 110,  85, 120
```

Note that they all only contain one 0.

Looking at the malformed ids, these all have a second 0 somewhere in the array:
```
                                      0, 188, 217, 193,  78,   0, 120, 201, 211
                       0,  12, 112,   0, 134, 242, 214, 168, 251, 149, 166, 127
                                      0, 164,  67, 242,   2,  10, 126,   0,  44
                            0, 112,   0, 139,  57,  29,  46,  18, 182, 188, 110
                                      0, 179, 100, 199,   0, 123,  31,  30, 180
                                      0, 142,   9,   0, 136, 186, 123, 100,  47
```

Since 0 is the delimiter, it's likely that these extra zeros are being misinterpreted. Notification Ids are just randomly generated lexicographic integers, and these can contain zeros. This is the same case for Discovery Queue Ids, so this is likely why I was seeing occasional random failures in those tests in the past as well. I'll have to take a look at the ids before they ever get put into the db in order to confirm this theory.

---

Support for this idea is that the ids are indeed not malformed when they are created and put into the db. Here's an example of one of them:
```
# Before going into the db
   6,  40, 199,   7, 225, 174, 112,   0, 178,  28,   0, 153, 149, 157,  24, 131

# The key buffer read out of the db for this key
<Buffer 00 b2 1c 00 99 95 9d 18 83>

# The id after being read from the db
                                      0, 178,  28,   0, 153, 149, 157,  24, 131
```

tegefaulkes:
I had a look at `js-db` and I'm guessing the problem is in `parseLevel`, `src/utils.ts:155` in `js-db`. I can't confirm but I feel that it should be handling the `if (b === esc[0])` case before the `if (b === sep[0])` case. That is, it should be checking for escaped separators before handling the separators. 

Referring to this part of the code.
```ts
  for (let i = 0; i < buf.byteLength; i++) {
    const b = buf[i];
    if (b === sep[0]) {
      // Note that `buf` is a subarray offset from the input
      // therefore the `sepEnd` must be offset by the same length
      sepEnd = i + (sepStart + 1);
      break;
    } else if (b === esc[0]) {
      const n = buf[i + 1];
      // Even if undefined
      if (n !== esc[0] && n !== sep[0]) {
        throw new errors.ErrorDBParseKey('Invalid escape sequence');
      }
      // Push the n
      levelBytes.push(b, n);
      // Skip the n
      i++;
    } else {
      levelBytes.push(b);
    }
  }
  ```

**Comment by CMCDragonkai** [comment]
@emmacasolin you should be able to go to js-db's `tests/utils.test.ts`. Look for the test `parse key paths`.

In there, there are several tests for different key possiblities.

You can take generated IDs, and stick them there. Put one of the failing IDs in.

The test checks if enocding it to the a key then decoding it back to a keypath is equvalent.

Do something like:

```ts
const keyPaths: Array<KeyPath> = [
  // ...existing keys being checked
  [generatedId1],
  [generatedId2],
  [generatedId3, generatedId4],
  [failingGeneratedId]
];
// ... continue
```

You can generate the ids directly in that test, or just copy base the buffers.

tegefaulkes:
Out of the 5 failing proxy tests. 4 of them pass if we comment out the failing `expect` that is checking for an error. The Error itself is a write after end error. I'm not sure if the error is a problem if we ignore it. I haven't had any luck finding the write that is the source of the error but I haven't looked too deeply at the forward/reverse connections.

The last error is puzzling. I'm getting a type error from code internal to node.js.
```
Cannot read properties of null (reading 'finishShutdown')
TypeError: Cannot read properties of null (reading 'finishShutdown')
    at JSStreamSocket.finishShutdown (node:internal/js_stream_socket:160:12)
    at node:internal/js_stream_socket:147:14
    at processTicksAndRejections (node:internal/process/task_queues:78:11)
    at runNextTicks (node:internal/process/task_queues:65:3)
    at processImmediate (node:internal/timers:437:9)
    ```
    I can't find anything that could be causing this. It seems to only be a problem with that one test. The test itself is checking if the connection times out properly. It seems that the test is passing fine and the error itself only happens after the `test()` and `afterEach()` completes. 
    
    I can't see any problem with the timeout mechanism in the `ReverseConnection`. I couldn't tell you where this is coming from except for the fact that `js_stream_socket` is possibly related to the node.js `net` internal library.

Running theory is that We're not handling the shutdown of some network resource properly in this specific test case. I need to think on it some more.

**Comment by emmacasolin** [comment]
> Out of the 5 failing proxy tests. 4 of them pass if we comment out the failing `expect` that is checking for an error. The Error itself is a write after end error. I'm not sure if the error is a problem if we ignore it. I haven't had any luck finding the write that is the source of the error but I haven't looked too deeply at the forward/reverse connections.
> 
> The last error is puzzling. I'm getting a type error from code internal to node.js.
> 
> ```
> Cannot read properties of null (reading 'finishShutdown')
> TypeError: Cannot read properties of null (reading 'finishShutdown')
>     at JSStreamSocket.finishShutdown (node:internal/js_stream_socket:160:12)
>     at node:internal/js_stream_socket:147:14
>     at processTicksAndRejections (node:internal/process/task_queues:78:11)
>     at runNextTicks (node:internal/process/task_queues:65:3)
>     at processImmediate (node:internal/timers:437:9)
>     ```
>     I can't find anything that could be causing this. It seems to only be a problem with that one test. The test itself is checking if the connection times out properly. It seems that the test is passing fine and the error itself only happens after the `test()` and `afterEach()` completes. 
>     
>     I can't see any problem with the timeout mechanism in the `ReverseConnection`. I couldn't tell you where this is coming from except for the fact that `js_stream_socket` is possibly related to the node.js `net` internal library.
> 
> Running theory is that We're not handling the shutdown of some network resource properly in this specific test case. I need to think on it some more.
> ```

I'm not sure if this is related to this issue, but when running `npm install` I get the following errors (these errors only show up in the terminal, they aren't caught by my IDE)

```
src/http/utils.ts:23:19 - error TS2345: Argument of type 'Duplex' is not assignable to parameter of type 'Socket'.
  Type 'Duplex' is missing the following properties from type 'Socket': connect, setTimeout, setNoDelay, setKeepAlive, and 8 more.

23       sockets.add(socket);
                     ~~~~~~

src/http/utils.ts:25:24 - error TS2345: Argument of type 'Duplex' is not assignable to parameter of type 'Socket'.

25         sockets.delete(socket);
                          ~~~~~~


Found 2 errors in the same file, starting at: src/http/utils.ts:23
```

**Comment by emmacasolin** [comment]
> @emmacasolin you should be able to go to js-db's `tests/utils.test.ts`. Look for the test `parse key paths`.
> 
> In there, there are several tests for different key possiblities.
> 
> You can take generated IDs, and stick them there. Put one of the failing IDs in.
> 
> The test checks if enocding it to the a key then decoding it back to a keypath is equvalent.
> 
> Do something like:
> 
> ```ts
> const keyPaths: Array<KeyPath> = [
>   // ...existing keys being checked
>   [generatedId1],
>   [generatedId2],
>   [generatedId3, generatedId4],
>   [failingGeneratedId]
> ];
> // ... continue
> ```
> 
> You can generate the ids directly in that test, or just copy base the buffers.

The failing ids are indeed failing in this test. The problem occurs when you have two separators in the key part (this doesn't happen for the level part or when using escapes) that are both in front of other data. For example,
```ts
[Buffer.concat([utils.sep, Buffer.from('foobar'), utils.sep])] // passes - separators are not inbetween data
[Buffer.concat([Buffer.from('foobar'), utils.sep, Buffer.from('foobar')])] // passes - only one separator
[Buffer.concat([Buffer.from('foobar'), utils.sep, Buffer.from('foobar'), utils.sep])] // passes - only one of the separators is in front of data
[Buffer.concat([utils.sep, Buffer.from('foobar'), utils.sep, Buffer.from('foobar')])] // FAILS
```

With that last one, the original keypath is `['foobarfoobar']` but after being encoded and decoded it's `['foobar', 'foobar']`

tegefaulkes:
I've have stop gap fixes for the problematic network tests. As I've said before 4 of the 5 tests are fixed by just ignoring the error. It doesn't seem to cause a problem with the rest of the code base and may just be an issue specific to the tests. 
The other tests that with the weird error mentioned [here](https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1135537879) was fixed by setting the `proxyHost` when starting the proxy.

As it stands all of the proxy tests are passing now. I can look deeper into the cause of the problems but I feel a this point it is low priority. If we want to explore them deeper then I can make a new issue for it to be address later.

CMCDragonkai:
Need to verify:

```
    "@grpc/grpc-js": "1.3.7",
```

Change to internals

```
    "@grpc/grpc-js": "1.6.7",
```

CMCDragonkai:
There are 2 issues here that I need to have a look at:

1. For the DB key parsing issue https://github.com/MatrixAI/js-db/pull/22 - the existence of the separator in the terminal part of the key path causes problems during `parseKey`, it gets parsed as 2 separate parts. This is because the `parseKey` doesn't know when to stop parsing. It needs to know when it's the last part and stop parsing any separators. Escaping the last part is not a good solution as this affects all iteration usage including outside the DB.
2. The network tests are still an issue (https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1136709893), I'm in the middle of copying back from `master` branch what the existing network tests were and finding out what's the exact problem here. Furthmore the network tests are still not using direct imports, it needs to be changed there to avoid cross-cutting concerns during testing.

CMCDragonkai:
The adding of the `ClientMetadata` in `src/types.ts` ends up causing `src/types.ts` to depend on the internals. This is a loop. We need to prevent loops. Internal domains depend on higher levels, but high level types cannot depend on internal domain types. (Which means we have to change where `ClientMetadata` goes.

CMCDragonkai:
I'm getting type errors on expected number of arguments for the `tests/grpc/utils`. @emmacasolin can you update them according to your changes to GRPC:

<details>

```
 FAIL  tests/network/index.test.ts
  ● Test suite failed to run

    tests/grpc/utils/GRPCClientTest.ts:76:22 - error TS2554: Expected 3 arguments, but got 2.

    76     return grpcUtils.promisifyUnaryCall<utilsPB.EchoMessage>(
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    77       this.client,
       ~~~~~~~~~~~~~~~~~~
    78       this.client.unary,
       ~~~~~~~~~~~~~~~~~~~~~~~~
    79     )(...args);
       ~~~~~

      src/grpc/utils/utils.ts:369:3
        369   f: (...args: any[]) => ClientUnaryCall,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
    tests/grpc/utils/GRPCClientTest.ts:84:22 - error TS2554: Expected 3 arguments, but got 2.

    84     return grpcUtils.promisifyReadableStreamCall<utilsPB.EchoMessage>(
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    85       this.client,
       ~~~~~~~~~~~~~~~~~~
    86       this.client.serverStream,
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    87     )(...args);
       ~~~~~

      src/grpc/utils/utils.ts:470:3
        470   f: (...args: any[]) => ClientReadableStream<TRead>,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
    tests/grpc/utils/GRPCClientTest.ts:92:22 - error TS2554: Expected 3 arguments, but got 2.

     92     return grpcUtils.promisifyWritableStreamCall<
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     93       utilsPB.EchoMessage,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~
    ... 
     97       this.client.clientStream,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     98     )(...args);
        ~~~~~

      src/grpc/utils/utils.ts:542:3
        542   f: (...args: any[]) => ClientWritableStream<TWrite>,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
    tests/grpc/utils/GRPCClientTest.ts:103:22 - error TS2554: Expected 3 arguments, but got 2.

    103     return grpcUtils.promisifyDuplexStreamCall<
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~
    104       utilsPB.EchoMessage,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~
    ... 
    108       this.client.duplexStream,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    109     )(...args);
        ~~~~~

      src/grpc/utils/utils.ts:653:3
        653   f: (...args: any[]) => ClientDuplexStream<TWrite, TRead>,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
    tests/grpc/utils/GRPCClientTest.ts:114:22 - error TS2554: Expected 3 arguments, but got 2.

    114     return grpcUtils.promisifyUnaryCall<utilsPB.EchoMessage>(
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    115       this.client,
        ~~~~~~~~~~~~~~~~~~
    116       this.client.unaryAuthenticated,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    117     )(...args);
        ~~~~~

      src/grpc/utils/utils.ts:369:3
        369   f: (...args: any[]) => ClientUnaryCall,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
    tests/grpc/utils/GRPCClientTest.ts:127:22 - error TS2554: Expected 3 arguments, but got 2.

    127     return grpcUtils.promisifyReadableStreamCall<utilsPB.EchoMessage>(
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    128       this.client,
        ~~~~~~~~~~~~~~~~~~
    129       this.client.serverStreamFail,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    130     )(...args);
        ~~~~~

      src/grpc/utils/utils.ts:470:3
        470   f: (...args: any[]) => ClientReadableStream<TRead>,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
    tests/grpc/utils/GRPCClientTest.ts:135:22 - error TS2554: Expected 3 arguments, but got 2.

    135     return grpcUtils.promisifyUnaryCall<utilsPB.EchoMessage>(
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    136       this.client,
        ~~~~~~~~~~~~~~~~~~
    137       this.client.unaryFail,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    138     )(...args);
        ~~~~~

      src/grpc/utils/utils.ts:369:3
        369   f: (...args: any[]) => ClientUnaryCall,
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        An argument for 'f' was not provided.
```

</details>

I think you didn't update these when the errors metadata was meant to be put in.

Alternatively to make them backwards compatible we can default the parameter to be an empty object.

CMCDragonkai:
Experiment with network fixes is here: https://github.com/MatrixAI/js-polykey/pull/376, I'll get to this after js-db issues fixed. In the mean time @tegefaulkes after all the other things are fixed, best to have a look at it.

tegefaulkes:
All except for 2 tests are passing. Almost done.

CMCDragonkai:
@emmacasolin when you're fixing up the vaults or other test domains please replace imports to be more specific where you see fit.

These 2 will still need to be fixed:

* https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1136803336
* https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1136823839

---

Quick update regarding the js-db and the key-parsing issue. To fix the ability to have separators in the terminal part it will involve an API change to how iterators work. Details are now in https://github.com/MatrixAI/js-db/pull/19. I'll have to cherry pick what I have there into staging since the changes are also combined with the ability to use `keyAsBuffer` and `valueAsBuffer`. This will end up affecting EFS and PK where you are using iterators.

emmacasolin:
@tegefaulkes I've replaced the places in the tests where you had used `{} as ClientMetadata` with actual/mocked data, as this information is required in order for ErrorPolykeyRemotes to be constructed. Therefore, in those tests the NodeId, Host, Port, and Command name are now either taken from constants already used in the tests or mocked. Note that I didn't fix this for `rpcVaults` since I'm splitting those anyway, and part of doing that is constructing the client in the same way as the other tests (which does not require the metadata to be set manually since it uses an actual client).

Also there were some failing tests in `grpc/utils.test.ts` due to more recent errors changes. Specifically the descriptions for polykey errors were not being serialised/deserialised so that is fixed now, and there was a very brittle test for expecting error messages to be empty for ErrorPolykeyUnknown, however these are constructed in our reviver (and could change) so I removed the tests for the message and just added a note above the test for how the unknown data is encoded (because this is also implemented in the reviver).

tegefaulkes:
Cool. I think with that last commit all of the tests are passing. except for the random failures due the the DB bug. 

I think @CMCDragonkai mentioned that we could make the `ClientMetadata` optional but since everything is fixed to be using it properly now I'm not sure it's needed.

 I should mention that I changed how the metadata was converted in the `ErrorPolykeyRemote` toJSON and fromJSON methods. Some tests were breaking because metadata wasn't revived properly when it was coming from a remote ageng. EG a chain on `ErrorPolykeyRemote`>`ErrorPolykeyRemote`>someError.
 
 I'm going to go look at the proxy network errors for now.

emmacasolin:
The splitting of the vaults/secrets client service tests is done now. Most of the tests I left as they were and just changed the setting up to align with the rest of the client service tests. This means that there isn't really any mocking done so the tests are quite inefficient (the actual underlying behaviour is being tested elsewhere so the client service tests can mock this and just expect actions to have been performed without actually checking that the expected action occurred correctly).

The clearest example of this is in `vaultsLog.test.ts` where I needed to extend the timeout for the `beforeEach` in order to accommodate needing to create some commits so that the command has something to check for. When this test runs on its own it's fine (the entire suite containing three tests takes 30 seconds, which is slow but doesn't time out), however when run with the rest of the vaults/secrets client service tests it takes closer to 100 seconds. Using mocking here would solve the issue, but it's not really a high priority fix.

Also of note is that there was a mistake in `vaultsSecretsMkdir` where the response type of the call was stated to be an EmptyMessage when it's actually supposed to be a StatusMessage so the test was failing when I did a check on the type of the response message. Looking at some of the other vaults client service handlers there seems to be the same issue for other handlers, however Mkdir was the only one that was caught through these tests. Not sure if the others need to be changed over too but it applies to vaultsCreate, vaultsScretsGet, and vaultsSecretsNewDir.

Last thing is that there actually weren't any existing tests for vaultsScan, vaultsPull, or vaultsClone so these have been left as todo. Not sure if writing these tests is a priority for this PR or not.

CMCDragonkai:
4.0.1 of js-db has been published. To integrate `4.0.1` you'll need to make changes to the `iterator`.

Mostly this comes in a couple ways:

1. The `iterator` is now `DBIterator`, this means it returns `KeyPath` as part of it's `next()` call. This means
   ```ts
   for await (const [kP, v] of db.iterator()) {
     // if you want the key
     dbUtils.keyPathToKey(kP);
     // however you should be able to do everything you need to do purely with just `kP`
   }
   ```
2. The ability to use `keyAsBuffer` and `valueAsBuffer` options. Note that `keyAsBuffer` doesn't change the `KeyPath` type. It just means that the `KeyPath` is an array of utf-8 encoded strings rather than an array of `Buffer`.

One way to quickly identify what to change is to a full search for `db.iterator` or `tran.iterator`. Then change the `k` to `kP` or `key` to `keyPath` to be precise that will make it clearer.

The change has to start on EFS first, then back to this PK. @tegefaulkes and @emmacasolin please plan out how to divide the work to update EFS and PK appropriately.

tegefaulkes:
Pipeline passing with some caveats. Had to re-do some tests that failed due to the DB bug.

CMCDragonkai:
The DB will need to be updated again pending this update: https://github.com/MatrixAI/js-db/pull/28. We're using base encoding to eliminate ambiguities in the keys. There's no more escape byte needed anymore. By using base 128, keys are enlarged by about 15%, the time complexity will be the same as our escaping algorithm before at `O(n)`. Anything larger than base128 would need to use a more complex 2-byte symbol alphabet, so it's not worth it right now as we can re-use the rfc4648 algorithm and just change some constants to make it work.

Right now some new fuzz tests are being added to js-db by @emmacasolin to ensure that it will be correct. The key is to ensure that our base128 algo is using an lexicographically ordered alphabet.

Then EFS and this PR's update of the DB can proceed. Note that `db.dump` and `tran.dump` are getting they key types updated to just be `KeyPath`.

CMCDragonkai:
Try out https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md to find out import cycles. My initial attempt is here: https://github.com/MatrixAI/TypeScript-Demo-Lib-Native/pull/30.

Seems like some additional packages are required: https://github.com/alexgorbatchev/eslint-import-resolver-typescript/issues/106

---

Many places we have legal cycles like in errors. But pretty much everywhere else import cycles should be disallowed.

Thus `ClientMetadata` needs to be moved. Where did this go?

tegefaulkes:
#376 has been merged into this.

CMCDragonkai:
1. @tegefaulkes make sure to check that the usage of iterator and `keyAsBuffer` is done once https://github.com/MatrixAI/js-db/pull/28 is merged and also update the DB here accordingly.
2. @emmacasolin will be working on EFS https://github.com/MatrixAI/js-encryptedfs/pull/66 to get the DB updated there, and PK will need to be updated with the latest EFS when that finishes too.
3. Also @emmacasolin has some final error logging tasks for the service handlers #304 as well as a minor format change from #323.
4. Other than that, @tegefaulkes you just need to start cleaning up the commits for this PR.
5. @emmacasolin please move `ClientMetadata` and `ErrorPolykeyRemote` under the `grpc` domain. Currently `ClientMetadata` ends up making `src/types.ts` importing `src/nodes/types.ts` and `src/network/types.ts` and that is a big scope-creep here. Having it in the `grpc` domain isn't the best, but the only quick solution here.
6. Pending my own work on SI snapshot: https://github.com/MatrixAI/js-db/pull/19

emmacasolin:
The DB has been updated and released as 4.0.2, and EFS has been updated with this change and is now released as 3.5.2, so both of those can be updated here.

tegefaulkes:
Cool. I'll make the changes now.

tegefaulkes:
I get the following error when trying to use `valueAsBuffer: false` option in the iterator

```ts
    for await (const [keyPath, permId] of tran.iterator<PermissionId>({ valueAsBuffer: false }, [
      ...this.aclNodesDbPath,
    ])) {
      // Looping here
    }
```

```
ErrorDBParseValue: Unexpected token � in JSON at position 0
    at Object.deserialize (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/@matrixai/db/src/utils.ts:240:13)
    at constructor_.deserializeDecrypt (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/@matrixai/db/src/DB.ts:684:41)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.next (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/@matrixai/db/src/DB.ts:521:21)
    at Object.next (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/@matrixai/db/src/DBTransaction.ts:276:29)
    at Object.[Symbol.asyncIterator] (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/@matrixai/db/src/DBTransaction.ts:362:24)
    at constructor_.getNodePerms (/home/faulkes/matrixCode/polykey/js-polykey/src/acl/ACL.ts:138:38)
    at withF (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)
    at Object.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/tests/acl/ACL.test.ts:170:23)

```

Unless I'm using this wrong. This needs to be fixed in DB. @CMCDragonkai @emmacasolin

tegefaulkes:
I have cleaned up the commits and fixed up the Iterator usage. There seems to be a bug with the iterator's `valueAsBuffer: false` option. the iterator seems to be failing to deserialize the the value. Once that is fixed and the dependencies are updated again then we can merge this PR.

tegefaulkes:
Found the problem I was having with the DB. I was being brain dead.I'll push up the fix in a sec.
Besides that, before we merge we need to check for any `TODO` and `FIXME` comments that were created during this PR and resolve them.

tegefaulkes:
At `src/vaults/VaultManager.ts:54`
```ts
// FIXME, this is a HACK since this type isn't exported
type LockRequest<L extends Lockable> = [
  key: ToString,
  lockConstructor: new () => L,
  ...lockingParams: Parameters<L['lock']>,
];
```

This is copied from the LockBox types. It was needed to generate the Lock requests from mapping vaultIds. It would be useful to be able to import this type but it was not exported. Might be something we need to add to `@matrixai/async-locks`.

CMCDragonkai:
Ok can you patch `@matrixai/async-locks` to export that type. Do a PR.

tegefaulkes:
If i recall there was a possible update to js-db and efs that was still pending. I think that's all that is left for this PR. I'm going to check what still needs to be done for them.

CMCDragonkai:
@emmacasolin please add https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1141612328 to the PR description and tick them off in the tasks section.

ETA on this being ready for review?

CMCDragonkai:
@emmacasolin 

> I'm noticing that the vaults service handlers aren't using validateSync to do validation, so they would be throwing ErrorParse rather than ErrorValidation. Will fix those up to use validateSync.

Are there any other domains where this applies to?

CMCDragonkai:
Can also replace the `utils.promisify` here with https://github.com/MatrixAI/js-db/pull/19#issuecomment-1143273485. It preserves the types of the callback function. I'm pretty sure it works but it's worth checking.

emmacasolin:
> @emmacasolin please add [#374 (comment)](https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1141612328) to the PR description and tick them off in the tasks section.
> 
> ETA on this being ready for review?

Should be ready by tomorrow

> @emmacasolin
> 
> > I'm noticing that the vaults service handlers aren't using validateSync to do validation, so they would be throwing ErrorParse rather than ErrorValidation. Will fix those up to use validateSync.
> 
> Are there any other domains where this applies to?

Just secrets, which is technically part of vaults in the client service anyway.

CMCDragonkai:
Some test failures were still happening on CI/CD, is this expected?

emmacasolin:
> Some test failures were still happening on CI/CD, is this expected?

No, there seem to be issues with cloning and pulling now but the errors aren't very helpful, for example

```
● VaultManager › With remote agents › clone and pull vaults using a vault id

    ErrorPolykeyRemote: Cannot read properties of undefined (reading 'valueOf')

      201 |       if (key === 'UNKNOWN' && errorData != null) {
      202 |         const error: Error = JSON.parse(errorData, reviver);
    > 203 |         const remoteError = new grpcErrors.ErrorPolykeyRemote(
          |                             ^
      204 |           metadata,
      205 |           error.message,
      206 |           {

      at toError (src/grpc/utils/utils.ts:203:29)
      at gf (src/grpc/utils/utils.ts:453:13)
      at Object.request (src/vaults/VaultInternal.ts:826:31)
      at Function.connect (node_modules/isomorphic-git/index.cjs:6757:17)
      at _fetch (node_modules/isomorphic-git/index.cjs:7453:15)
      at _clone (node_modules/isomorphic-git/index.cjs:7726:42)
      at Object.clone (node_modules/isomorphic-git/index.cjs:7849:12)
      at src/vaults/VaultInternal.ts:169:11
      at src/nodes/NodeConnectionManager.ts:182:16
```

I've put console.logs above the only two places in the PK codebase that have a `valueOf`, but it's not coming from either place. I've also tried logging out the vaultIds inside the service handlers to see if those are what are undefined, but so far I haven't seen anything unusual.

CMCDragonkai:
This might have been brought in by 8237452. Your changes to the `vaultsGitInfoGet` and `vaultsGitInfoPack`.

Can you check reverting those changes and then try again? And then investigate those 2 in particular and see why it's happening. Since the git cloning occurs in an HTTP tunnel via a GRPC duplex stream we wouldn't necessarily use any validation inside the tunnel, but I haven't gone into it in detail.

Issue #234, opened 2021-08-30
Using QUIC/HTTP3 to replace utp-native for the Data Transfer Layer in the networking domain


CMCDragonkai:
### Specification

We can separate the networking to 3 layers:

* P2P application layer - e.g. kademlia, automerge, nodes domain
* RPC layer - e.g. grpc (and once considered jsonrpc but it's too late now)
* Data Transfer layer - UTP, Wireguard, QUIC

The Data Transfer Layer is particularly special since it is the lowest part of our stack and is always fundamentally built on top of UDP. These are its characteristics:

* Must be built on top of UDP (due to hole-punching requirements)
* Capable of NAT-traversal via hole punch packets
* Capable of being proxied for traversing symmetric NAT
* End to end encrypted - based on MTLS or otherwise
* Low latency and high throughput (and ideally low power for maintaining the connection liveness)
* Compatible with Linux, Mac, Windows, iOS, Android

Currently we use the TLS on top of UTP protocol as the Data Transfer Layer. This is implemented with a forward and reverse proxy termination points bridging TCP to the UTP protocol.

The proxy termination bridges was necessary due to us using GRPC as the RPC layer and its implementation fixing on HTTP2 stack, but with an escape hatch via a connect proxy. This is more of an implementation detail than anything else. But it does give us a fairly generic system that can turn any TCP-protocol into something NAT-traversable.

The usage of MTLS also enables a seamless usage of the same X.509 certificate that we already use for the rest of PK.

The underlying library being used here is `utp-native` https://github.com/mafintosh/utp-native, it is C++ module that incorporates the utp C++ library https://github.com/bittorrent/libutp and wraps it into a NodeJS module. It works fine for Linux, Window, and Mac. However there is no clear path for usage on Android or iOS. See https://github.com/mafintosh/utp-native/issues/30. This could be done by using https://github.com/janeasystems/nodejs-mobile (but this is not as popular), or by compiling libutp natively for iOS and Android, and then wrapping it out as native code on NativeScript/React Native.

The `utp-native` is also OLD. It has some issues like:

* https://github.com/mafintosh/utp-native/issues/40
* Poor documentation that we had to work out how it worked at a low level by going into the source code
* Doesn't support IPv6 https://github.com/mafintosh/utp-native/issues/15

All of this means that continuing down with trying to use `utp-native` might just mean flogging a dead horse.

[...]

PR #376, opened 2022-05-25
WIP: Fixing network issues in the new feature-dep-upgrades

CMCDragonkai:
Experiment in fixing the network

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2655,7 +2655,7 @@
     "node_modules/@protobufjs/aspromise": {
       "version": "1.1.2",
       "resolved": "https://registry.npmjs.org/@protobufjs/aspromise/-/aspromise-1.1.2.tgz",
-      "integrity": "sha1-m4sMxmPWaafY9vXQiToU00jzD78="
+      "integrity": "sha512-j+gKExEuLmKwvz3OgROXtrJ2UG2x8Ch2YZUxahh+s1F2HZ+wAceUNLkvy6zKCPVRkU++ZWQrdxsUeQXmcg4uoQ=="
     },
     "node_modules/@protobufjs/base64": {
       "version": "1.1.2",
@@ -2670,12 +2670,12 @@
     "node_modules/@protobufjs/eventemitter": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@protobufjs/eventemitter/-/eventemitter-1.1.0.tgz",
-      "integrity": "sha1-NVy8mLr61ZePntCV85diHx0Ga3A="
+      "integrity": "sha512-j9ednRT81vYJ9OfVuXG6ERSTdEL1xVsNgqpkxMsbIabzSo3goCjDIveeGv5d03om39ML71RdmrGNjG5SReBP/Q=="
     },
     "node_modules/@protobufjs/fetch": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@protobufjs/fetch/-/fetch-1.1.0.tgz",
-      "integrity": "sha1-upn7WYYUr2VwDBYZ/wbUVLDYTEU=",
+      "integrity": "sha512-lljVXpqXebpsijW71PZaCYeIcE5on1w5DlQy5WH6GLbFryLUrBD4932W/E2BSpfRJWseIL4v/KPgBFxDOIdKpQ==",
       "dependencies": {
         "@protobufjs/aspromise": "^1.1.1",
         "@protobufjs/inquire": "^1.1.0"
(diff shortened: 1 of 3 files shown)


CMCDragonkai:
@tegefaulkes this is where I'm investigating the network.

CMCDragonkai:
Based off: https://github.com/MatrixAI/js-polykey/pull/374#issuecomment-1136709893

I don't think we should be cutting corners here regarding the networking, we should always ensure this is working well enough since it's the basis of all the other functionality of nodes and vaults.

tegefaulkes:
Attempting to revert the version of `@grpc/grpc-js` from `1.6.7` to `1.3.7`. This is resulting in the following build errors.

<details>

```
 npm run build

> @matrixai/polykey@1.0.0 build
> rm -r ./dist || true; tsc -p ./tsconfig.build.json

node_modules/@grpc/grpc-js/build/src/object-stream.d.ts:11:18 - error TS2430: Interface 'IntermediateObjectWritable<T>' incorrectly extends interface 'Writable'.
  The types returned by 'end(...)' are incompatible between these types.
    Type 'void' is not assignable to type 'this'.
      'this' could be instantiated with an arbitrary type which could be unrelated to 'void'.

11 export interface IntermediateObjectWritable<T> extends Writable {
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@grpc/grpc-js/build/src/server-call.d.ts:64:5 - error TS2416: Property 'end' in type 'ServerWritableStreamImpl<RequestType, ResponseType>' is not assignable to the same property in base type 'Writable'.
  Type '(metadata?: any) => void' is not assignable to type '{ (cb?: (() => void) | undefined): this; (chunk: any, cb?: (() => void) | undefined): this; (chunk: any, encoding: BufferEncoding, cb?: (() => void) | undefined): this; }'.
    Type 'void' is not assignable to type 'this'.
      'this' could be instantiated with an arbitrary type which could be unrelated to 'void'.

64     end(metadata?: any): void;
       ~~~

node_modules/@grpc/grpc-js/build/src/server-call.d.ts:77:5 - error TS2416: Property 'end' in type 'ServerDuplexStreamImpl<RequestType, ResponseType>' is not assignable to the same property in base type 'Duplex'.
  Type '(metadata?: any) => void' is not assignable to type '{ (cb?: (() => void) | undefined): this; (chunk: any, cb?: (() => void) | undefined): this; (chunk: any, encoding?: BufferEncoding | undefined, cb?: (() => void) | undefined): this; }'.
    Type 'void' is not assignable to type 'this'.
      'this' could be instantiated with an arbitrary type which could be unrelated to 'void'.

77     end(metadata?: any): void;
       ~~~


Found 3 errors in 2 files.

Errors  Files
     1  node_modules/@grpc/grpc-js/build/src/object-stream.d.ts:11
     2  node_modules/@grpc/grpc-js/build/src/server-call.d.ts:64
```

</details>

The problem seems to be that GRPC extends some types provided by node. and given the new version of node GRPC isn't extending the type properly anymore. Specifically the `end` method was changed to return `this` instead  of `void. 

This must've been the install/build errors that @emmacasolin saw.

I don't really have any evidence for or against this version change being the problem however.

tegefaulkes:
Found one of the problems. For the 4 tests that have the write after end error. We can fix this by changing the following.
```ts
      tlsSocket.on('end', () => {
        logger.debug('Reverse: receives tlsSocket ending');
        tlsSocketEnd();
        if (utpConn.destroyed) {
          logger.debug('Reverse: destroys tlsSocket');
          tlsSocket.destroy();
        } else {
          logger.debug('Reverse: responds tlsSocket ending');
          // Why are we ending the tlsSocket during it's end event?
          // tlsSocket.end(); // removing this removes the error. 
          utpConn.end(); // Seems like we should be ending the utpConn when the TLS ends.
          tlsSocket.destroy();
        }
      });
```

tegefaulkes:
The problem with the last test also seems to be a problem with the test itself. It seems to be caused by trying to connect to `0.0.0.0` with the UTP socket. That's why it was fixed when setting the `proxyHost` when starting the proxy. Otherwise it was taking the default host of '0.0.0.0'.

The fix for this is still the same. Just need to add the following to starting the proxy;
```ts
    await proxy.start({
      serverHost: serverHost(),
      serverPort: serverPort(),
      forwardHost: localHost, // Add this
      proxyHost: localHost, // Add this
      tlsConfig: {
        keyPrivatePem: keyPairPem.privateKey,
        certChainPem: certPem,
      },
    });
```

Theses should be added to all of the other tests just in case. We only want to bind the the localhost for testing anyway.

tegefaulkes:
@CMCDragonkai What were you doing in the WIP commit? so I know how to rename or squash it out.

CMCDragonkai:
My original goal was to revert your changes to `Proxy.test.ts`, in doing so I copied over from master branch.

Then slowly reintroduce your changes as I can see incrementally what the effects are.

Some of your changes was a bit confusing, especially removing lines of code that close/end sockets or other resources. I didn't finish it.

CMCDragonkai:
> Found one of the problems. For the 4 tests that have the write after end error. We can fix this by changing the following.
> 
> ```ts
>       tlsSocket.on('end', () => {
>         logger.debug('Reverse: receives tlsSocket ending');
>         tlsSocketEnd();
>         if (utpConn.destroyed) {
>           logger.debug('Reverse: destroys tlsSocket');
>           tlsSocket.destroy();
>         } else {
>           logger.debug('Reverse: responds tlsSocket ending');
>           // Why are we ending the tlsSocket during it's end event?
>           // tlsSocket.end(); // removing this removes the error. 
>           utpConn.end(); // Seems like we should be ending the utpConn when the TLS ends.
>           tlsSocket.destroy();
>         }
>       });
> ```

Receiving an `end` event is not the same as sending an `end` event. The former is about receiving an FIN package, the latter is about sending a FIN packet.

If it fails to send a FIN packet, then that's because the utp connection is destroyed. But as you can see I already checked for that.

This is why I said it sounds familiar, since it's like a problem we have dealt with before.

tegefaulkes:
Should be close to merging. There is some things I need to check first.

tegefaulkes:
Looking over how Proxy handles the `utpConn`. It seems that we just don't handle any errors coming out of the `utpConn`. Since this problem only happens when the socket is ending anyway I think we can just safely ignore the error in the tests. Or do what was suggested and log out a warning during the test if an error happens.

tegefaulkes:
This is ready to merge now.

CMCDragonkai:
I think you should be logging out a warning here. If you're happy to proceed, just merge it into #374. I'll come back around to reviewing the networking issue soon. Main important thing is to code to the protocol, and add adjustments when the implementation doesn't meet the protocol. The utp-native library seems flaky atm, and assuming SI transaction works well enough, we may be in a position of integrating a different network stack like QUIC #234 (as I'll have the understanding and tooling for integrating C/C++ code into our JS libraries).

Issue #334, opened 2022-02-15
Review the Identities CLI and RPC Handlers


emmacasolin:
### Specification
Our Identities CLI seems incomplete and needs to be reviewed. In particular:
- The `identitiesToken*` rpc handlers have no corresponding CLI commands - something like `pk identities token ...` should be added
  - We need to consider what kind of access we want the user to have to this data (authenticating a provider already "puts" a token - need to consider the implications of allowing the user to put/get/delete their own tokens)
  - Our 'tokens' are data of the following type:
```ts
      type TokenData = {
        accessToken: string;
        refreshToken?: string;
        accessTokenExpiresIn?: number;
        refreshTokenExpiresIn?: number;
      };
```
- The `identitiesProvidersList` rpc handler has no corresponding CLI command - possible command could be `pk identities providers`
  - As it's currently implemented, this just returns a list of the providers (provider ids) currently registered to the IdentitiesManager (and thus supported by Polykey)
- The `identitiesAuthenticatedGet` rpc handler (which returns the user's own authenticated identities) currently uses the command `pk identities authenticated` - this name is very similar to `pk identities authenticate` and could get confusing
  - The naming of this command, as well as its relationship with the token manipulation commands (particularly `identitiesTokenGet`) should be reviewed
- At the moment, the gestalts and identities commands are both grouped under the `pk identities` subcommand
  - We should consider separating these, especially as the functionality of our CLI grows and we need to be able to reuse command names (e.g. we may need both a `gestalts list` and an `identities list` command)
- Many of the identities CLI commands receive data from the server in the form of a stream, however this data is outputted as a single array
  - This should be modified such that the data returned from the stream is streamed to stdout as we iterate through it (as opposed to after)

### Additional context
- General CLI review issue: https://github.com/MatrixAI/js-polykey/issues/268
- Beginning of a thread discussing some of the issues with the identities CLI: https://github.com/MatrixAI/js-polykey/pull/320#discussion_r804223154

### Tasks
1. Review what functionality should be exposed to the CLI and what should not
2. Create CLI commands for functionality that is not yet exposed to the CLI
3. Consider separating identities and gestalts commands in the CLI
4. Ensure that data returned as a stream to the CLI is outputted to STDOUT as a stream

CMCDragonkai:
There's a problem with `pk identities trust` command.

From: https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1041312635.

[...]

PR #378, opened 2022-06-01
Preparing for Testnet Deployment

### LINKED ISSUE #150: Remove "revoked" node ID -> node address mappings in `NodeGraph`
The buckets database used to store node ID -> node address (host:port) mappings is a fixed-size structure, of `n` buckets containing `k` mappings each.

Because a public key can be revoked and refreshed, and the node ID is the public key fingerprint, there is the potential for a keynode to require a new node ID (and thus, new certificate/s as well).

This introduces a trade-off:

* We introduce a redundant key-value pair in our buckets database, where two node IDs are mapping to the same host and port. After learning of the updated node ID, do we remove this node ID from the database altogether? Furthermore, we can have malicious actors that repeatedly perform this action, and congest the network's node databases, by introducing a multitude of keys that map to the same host and port.
* But, other systems may require a legitimate earlier node ID -> node address mapping in order to connect (as they may not know about the newest node ID)
  * NOTE: do we actually want this node ID to be considered valid anymore though? That is, If the public key has been refreshed, would we not want to remove this node ID -> node address mapping entirely? Essentially, when a public key is refreshed, we would therefore consider it as a "new node" in the network, where all other nodes are unaware of it.
  * Perhaps we still do. While the node ID may be considered as compromised, the host and port that it maps to would likely remain unchanged?

### LINKED ISSUE #158: Tree-based Kademlia Data Structure - k-bucket library
This was part of our package.json - https://github.com/tristanls/k-bucket.

It may not be used due our kademlia being persistent. But it may be a valuable source to understand how to make our own kademlia implementation more performant.

For now I'm removing k-bucket from package.json.

### LINKED ISSUE #177: Update testnet.polykey.io to point to the list of IPs running seed keynodes
Done on cloudflare.

image

This will depend on us deploying the js-polykey keynode containers onto AWS.

Also change `bootstrap.polykey.io` to `seed.polykey.io`.

### LINKED ISSUE #194: Testnet Node Deployment (testnet.polykey.io)
### Specification

It is now time for our second attempt at testnet deployment.

image

We had previously already done a PK deployment on AWS using ECS back when PK was `0.0.41`. While the AWS deployment worked, we hit a lot of problems which meant we had to go through an 8 month long refactoring process over the entire codebase.

Now that the codebase is finally refactored, we're ready for the second attempt.

The AWS architecture is basically the same as before, but our configuration should be a lot more simpler. There are some changes though.

* Before we had to deal with Node root certificates, now root certificates are no longer relevant to the testnet/mainnet deployment
* We are now separating into 2 clusters of PK seed nodes: `mainnet.polykey.io` and `testnet.polykey.io`. The `mainnet` is intended for production use, and we will first prototype our testnet deployment and testnet will be where new versions of PK are tested before being released on production.
* Both mainnet and testnet seed nodes will be trusted by default, but the PK releases should default to use the mainnet and have a switch to use the testnet.
* We don't know yet whether we should be using NLB or not, we may decide not to use a NLB at all. But there shouldn't be any sort of session state that is required for P2P functionality
* NLBs cannot be used with PK clients that are debugging the testnet/mainnet nodes, because they would resolve any possible node, and in this case there is in fact network session state. Instead PK client debugging has to be done with the container IPs.
* We know that IPv6 isn't supported yet so we will have IPv4 and DNS support.
* We should be using well known ports here of `1314` UDP and `1315` TCP for the ingress port and the client port respectively.
* The PK nodes are not stateless, they do require node state. However this node state is not important to us to persist. So any EBS volume mounted into the ECS container should work. Basically we just need a mutable temporary directory. What kind of mutations are there? Well the kademlia node graph is persisted atm and is not in-memory.

### Additional context

* https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/issues/197 - old issue detailing how to configure the AWS infrastructure (we're doing it manually right now) **refer to this when starting on this issue**
* https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/issues/237 - old issue regarding old task definition envrionment variables

### Tasks

1. [x] - Upload image to ECR "elastic container registry"
   * check https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/issues/197#note_496723119
   * make sure you have the right authentication details
2. [x] - Create ECS "elastic container service" Task Definition for the new image uploaded to ECR
   * the Task Definition describes how to execute the container
   * just like how we executed in `docker run`, we will need the same parameters
   * the PK agent will need a writable directory to be the node state, if we don't specify anything this will just be a temporary scratch layer by ECS, so we should be using a volume mount of some sort, the data inside this PK agent is not important, therefore any AWS volume should be ok, however an NFS/EFS volume might help us in case we want to debug things
   * additional environment variables for unattended bootstrapping and port/host binding
     - MatrixAI/Polykey#202 - bootstrapping
     - MatrixAI/Polykey#286 - ports
3. [x] - Start the ECS service, just cluster of 1, test that it is working by using the PK CLI and directly contacting the ECS IP address and port for `PK_PORT`.
4. [x] - Integrate firewall (security group), NLB and elastic IP to the NLB and then attach the `testnet.polykey.io` domain to the NLB
   * NLB won't maintain a session between connections in order to point to the same agent, ideally we won't need to have a common for NAT-busting purposes, this shouldn't be a problem for NAT-busting for either hole-punching relay or actual relay, as far I know
   * if it is a problem, an alternative to NLB, is domain-level load balancing, where multiple EIPs are presented by `testnet.polykey.io` and they are randomised, cloudflare supports this https://www.cloudflare.com/en-au/learning/performance/what-is-dns-load-balancing/ resolve this in MatrixAI/Polykey#177 
~5. [ ] - Update reference documentation with testnet architecture including AWS, use a component diagram with relevant AWS resources~ - waiting on Pulumi... intermediate diagram available https://github.com/MatrixAI/js-polykey/issues/194#issuecomment-1179923272

### LINKED ISSUE #212: Improve node search over k-buckets (getClosestLocalNode)
### Specification
To optimise reading out the closes nodes to a target node we need to apply some improvements.

The `getClosestNode` function needs to take a `nodeId` and `limit` as parameters. The nodeId is the node we're calculating the distance relative to. The limit is how many nodes we wish to return. The limit defaults to the `nodeBucketLimit` as per the Kademlia spec.

We need to avoid reading out all of the buckets and iterating over empty buckets. This can be achieved by using a `readStream` over the `nodeGraphBucketsDb` level. This level contains sub levels for each bucket. Each sub level contains the nodeId:nodeInfo key:value pairs. Using the `nodeGraphBucketsDb` level we can iterate over each stored node in bucket order all at once.  Note when setting the gt or lt on the stream we need to start from the desired bucket. In this case the starting point is the bucket 'above' the desired starting bucket. the key we want to start from takes the form of  a `Buffer` with `<prefix><higherBucketId><prefix>`. Iterating less than this gives us the target bucket plus all lower buckets. Above this is all of the higher buckets.

When we run out of lower buckets we need to iterate over the higher buckets from where we started.  If we run into limit while iterating over the nodes we need to get the whole of the last bucket we read. since nodes are out of order within a bucket we need whole buckets to ensure we obtain the closest nodes.

The resulting list is sorted by distance using `nodesUtils.bucketSortByDistance` and the list is truncated down to the provided limit.

##### As implemented
We iterate over the nodes directly across the buckets. the nodes are read out in the following order. 
1. all nodes within the target bucket `N`
2. nodes in order of bucket 0 to bucket N-1
3. nodes in order of bucket N+1 to 255.

When we reach our specified limit we read the whole of the last bucket we've read and add that to the list. we then sort all of the nodes and truncate the list back down to the limit and return that.

### Additional context

* discussion thread https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/195#note_612883452 (text above copied from a previous comment of mine)
* https://github.com/MatrixAI/js-polykey/issues/212#issuecomment-1069810881

### Tasks
* [x] 1. Update `getClosestNodes` implementation to use a `readStream` to iterate over each node sequentially across all of the buckets.
* [x] 2. if we run out of 'closer' buckets we iterate over the `higher` buckets.
* [x] 3. buckets are not ordered via distance, so if we read any node from a bucket into the list we need to get the whole bucket.
* [x] 4. the returned list needs to be sorted by distance and truncated down to the limit.

### LINKED ISSUE #226: Expunge any added "invalid" nodes from k-buckets
### Specification

> Any node that is added that turns out to be invalid, should be expunged from the any state in the nodes.

We need a means of handling any nodes that are found to be invalid.

I make the initial assumption that an "invalid" keynode can only be introduced by a manual insertion into the buckets database. Therefore, an "invalid" node can only be inserted from the `nodes add` CLI call, or from the specified seed nodes when we instantiate a keynode. I'm of the opinion that nodes added from discovery (through Kademlia) should be expected to be correct. However, do note that any nodes added by `nodes add` can be discovered through Kademlia. Thus, we're currently at a state where "invalid" nodes could be discovered.

From my perspective, there's a couple of ways a node can be deemed as "invalid":
* invalid host/port: any malformed IP is caught at the CLI level, but an invalid well-formed IP + port can't be ascertained easily (a timeout of connection can signify that the node is simply offline).
* invalid node ID: malformed node IDs are caught at CLI level, and we also check the node ID against the certificate on connection.

As such, we don't currently have an easy way of determining an "invalid" node.

**I believe the best approach to this is making appropriate restrictions on how/what we add to the buckets database.** See additional commentary in the below section regarding this.

### Additional context

This was originally brought up in review of the nodes CLI refactoring. 

When discussing the `nodes add` command, I made the initial assumption that we would need to create a connection to the other node before we could successfully add to the database. This would ensure that the node details must be correct.

However, this is not the case https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/198#note_641488589:

> I'm not sure about this constraint. PK nodes should be able to be offline at random times, and it's a fairly lazy network.
> I reckon you should be able to add nodes into the graph without initial connection.
> The problem with kademlia you mentioned is known in literature. It's called kademlia poisoning. Variants are like sybil attacks as well.
> We haven't investigated mitigations against it right now, and the kademlia system can be suspect to DDOS attack as well.
> The problem you say can also occur by just someone subverting its own kademlia database, so the mitigation for this can not be on the assumption that all nodes are friendly. It requires a different approach. But we can address this later after the release.
> For now, you just want to ensure that when you do the discovery you might want to avoid adding nodes into your search if that node is offline or cannot be accessed (you may try to discard them early) so you don't keep around dead nodes in your node graph.
> Of course you don't want delete dead nodes if they part of your gestalt though.
> We will need to go through a proper design overview to work through these issues.

### Tasks

1. ...
2. ...
3. ...

### LINKED ISSUE #285: Testnet securely maintain a pool of recovery codes
Once we have the ability to use `PK_RECOVERY_CODE` to automatically bootstrap the PK keynodes, we need to create at least 1 recovery code and 1 root key to be used.

The recovery code must be kept secret. I'll maintain this right now. The root key will be inside AWS's block device mounted into the ECS container and this will be kept safe inside AWS.

The recovery code will need to be used as an environment variable for ECS for the testnet.

Eventually we can store the recovery code inside a running Polykey node, and make use of AWS integrations, like our wiki page: "Service Deployment Secrets with AWS ECS".

Doing this should ensure that we don't need to maintain the volume state mounted in to the ECS container, it just has to be mutable, but it can be deleted, since everything can be recovered.

### Tasks

1. [x] - Use `pk bootstrap` locally to generate a recovery code and root key.
2. [x] - Save the recovery code securely.
3. [x] - Try using `pk bootstrap` on a different directory and see if the same root key is used. Compare them.
4. [x] - Delete the root key.
5. [x] - Use the recovery code for ECS Task Definition

This will be done for 1 single testnet node. We can scale this up later.

### LINKED ISSUE #317: Test communications when root key pair is renewed and NodeID is changed
### Specification
When the root keypair is renewed, we expect the old nodeId and client certificate to still be valid (in constrast to when it is reset).

From https://github.com/MatrixAI/js-polykey/issues/312#issuecomment-1016137891:

Changing the node id will impact communications between:
- Agent to Agent - this is mTLS
- CLI to Agent - normal TLS (client does not provide cert)
- GUI to Agent - normal TLS (client does not provide cert)

Under Agent to Agent, this is going through the forward and reverse proxies, and also the GRPCClientAgent. The GRPC client does not use TLS but the TLS is being done between forward and reverse proxies. This means that when we perform a keyManager.renewRootKeyPair, the old NodeId is still valid until the certificate expiration date. This means nodes contacting the changing agent with the old NodeId should still work, but receive an automatic update to the NodeGraph when the node id changes. I believe the logic to check the cert chain is in network/utils.ts verifyServerCertificateChain, but the logic to update the node id in the nodegraph is not verified yet.

Under CLI to Agent and GUI to Agent, they are using direct TLS on the GRPCClient. Existing GRPCClient connections should continue to work and not be interrupted by the TLS change. I believe I had tests for this under tests/grpc/GRPCServer.test.ts as changing the private key and certificate on the fly. However new connections will need to be made with a new target Node ID. This can occur by reading the status each time a new connection is made. However I think a similar TLS verification logic could take place where the old node ID can still be valid.

So we need tests to ensure this is the case. One for tests/grpc/GRPCClient.test.ts to ensure that the GRPCClient node id can still work. Another for tests/network as well.

Then we would extend these tests into tests/nodes for NodeConnection, and also in tests/bin to test CLI situation.

### Additional context
- Previous issue for propagating root keypair changes https://github.com/MatrixAI/js-polykey/issues/312
- https://github.com/MatrixAI/js-polykey/pull/311#discussion_r787389824
- https://github.com/MatrixAI/js-polykey/issues/312#issuecomment-1014188218

### Tasks
1. Check logic for updating the certchain and node id
2. TLS verification to allow old node ids to still be used
3. Tests for GRPCClient, network domain, nodes, and CLI

### LINKED ISSUE #322: Investigate means of authenticating a node ID when adding to `NodeGraph`
### Specification
With #344 more or less solved we need a way of authenticating when adding forward connections. this includes authenticating the old connections  when a bucket fills up. In both cases we can so a `pingNode` to check if a node is alive.  However the current implementation of `pingNode` is insufficient for the job as it only checks if a node is alive.

`NodeManager.pingNode` needs to be updated to match the following criteria;
1. It needs to establish a proxy connection to the target. if a connection can be established then the target is considered alive. if the connection is established then we can consider it authenticated as well.
2. It needs to take an optional address information for checking if direct connections without the nat bashing proceedure can be done. This will be useful for ICE

The `NodeManager.setNode` needs to be updated to do the following.
1. when a bucket is full the oldest node needs to be verified with `pingNode`. This is already done, just need to update ping for this. This will need to concurrently check alpha of the oldest connections.
2. Needs an optional force parameter to garentee adding the node.

`NodeConnectionManager` needs to be updated to;
1. takes `NodeManager` as part of start.
2. establishing a connection should trigger `NM.qeueSetNode`

`NodeManager` needs to be updated to;
1. startStop asyinc init pattern.
2. implement a async queue for adding nodes to the nodeGraph. 

After that we need to make sure all instance where we need to add a node just calls `NodeManager.setNode`. Where this is done still needs to be speced out but this may be outside the scope of this issue. Create a new issue for this?

### Additional context

- another very relevant issue: #226
- Related #344 
- Related #326 
- Related #148 

### Tasks
- [x] update `NodeManager.pingNode` to match the above specification.
- [x] update `NodeManager.setNode` to match the above specification.
- [x] implement setnode queue. this is similar to the queue used by discovery. 
- [x] seperate pingNode out of setNode
- [x] concurrently check alpha old nodes if a bucket is full when setting node.
- [x] convert NCM and NM to startstop, they rely on each other now with NCM calling NM.qeueSetNode.
- [x] establishing a NodeConnection should trigger setting a node in the nodeGraph by callong NM.qeueSetNode.
- [x] Creating a new connection should not be blocked by setting node.
- [x] Integrate into `pk nodes set` (with `--no-ping` and `--force`) and `pk nodes ping`

### LINKED ISSUE #324: Support host (IP address) in `parseSeedNodes`
### Specification

`parseSeedNodes` currently assumes that the provided address of the seed node is a hostname, and does not provide support for a seed node with an IP address as a host address:

```ts
// src/bin/utils/parsers.ts - parseSeedNodes()
if (!networkUtils.isHostname(idHostPort[1])) {
  throw new commander.InvalidOptionArgumentError(
    `${idHostPort[1]} is not a valid hostname`,
  );
}
```

Furthermore, it should also provide support for IPv6 addresses - as such, we can no longer rely on a colon/semi-colon `split` usage. It's been suggested to use `new URL` to construct a URL from the `nodeId@host:port` format (will require a prepended `pk://` 'protocol') such that we can acquire the different components of the seed nodes.

### Additional context

- discussed in sprint meeting https://vimeo.com/674273240 approx 23:00
- original comment here https://github.com/MatrixAI/js-polykey/pull/321#issuecomment-1030758342 and  https://github.com/MatrixAI/js-polykey/pull/321#issuecomment-1030979784

### Tasks

1. Utilise `new URL` to generate the `nodeId@host:port` 
2. Include the appropriate exception handling for invalid formats
   - can we specify the particular component of the `nodeId@host:port` that's invalid if an exception is raised (it's currently done this way through the use of `split`)? Look into this.
3. Move function to validation utils
4. Add test for IP address in seed node

### LINKED ISSUE #344: Seed node not adding details of connecting node to its `NodeGraph`
### Specification

Authentication a node on the other side of a connection is done buy the `Proxy` class when a connection is made. This includes obtaining the `NodeId`, certificates and connection info. I assume that the certificates are authenticated at this stage. This is all the information that we need to authenticate and add a node to the node graph with the added benefit of it being obtained at the time a connection is established.

We need to add a callback to the `Proxy` that is triggered with the `NodeId`, `remoteHost` and `remotePort` information whenever a connection is made. For simplicity this will be triggered when a `ForwardConnection` and `ReverseConnection` is made. So `Proxy` needs to take a `(remoteNodeId, remoteHost, remotePort) => void` callback that is triggered whenever a connection is established and authenticated.

A new event needs to be added to the `EventBus` for this connection event. The handler for this event needs to call `NodeManager.setNode()` to add the information to the nodeGraph.

### Additional context
- Proxies was updated in #360 , #361. Changes affect this.
- Uses the same events system from #312 #311  ? I'm not sure where it was implemented.

### Tasks
- [x] 1. Add a `connectionEstablishedCallback` to the `Proxy` that is triggered whenever a connection is authenticated and established.
- [x] 2. Add a new event to the `events` `EventBus` in the `PolykeyAgent` and have this triggered by the callback.
- [x] 3. Add an event handler for the new event that calls the `NodeManager.setNode()` method thus adding the node to the nodeGraph.
- [x] 4. add tests for;
  - [x] Proxy triggering the callback with the correct information.
  - [x] Integration test for the `PolykeyAgent` to check that the event properly propagates and updates the `NodeGraph`.

### LINKED ISSUE #345: Supporting the "refresh" operation for `NodeGraph` buckets
### Specification
Kademlia has this to say about refreshing buckets;

"Buckets will generally be kept constantly fresh, due to the traffic of requests travelling through nodes. To avoid pathological cases when no traffic exists, each node refreshes a bucket in whose range it has not performed a node lookup within an hour. Refreshing means picking a random ID in the bucket’s range and performing a node search for that ID. To join the network, a node u must have a contact to an already participating node w. u inserts w into the appropriate k-bucket. u then performs a node lookup for its own node ID. Finally, u refreshes all k-buckets further away than its closest neighbour. During the refreshes, u both populates its own k-buckets and inserts itself into other nodes’ k-buckets as necessary." [kademlia spec](https://www.scs.stanford.edu/~dm/home/papers/kpos.pdf#page=4)

Given this we need to implement the following

1. A refresh bucket method needs to be added that does the following. This may need to be part of an async queue since refreshing a bucket can be expensive.
    1. Picks a random node inside the range of a given bucket.
    2. preforms a node lookup on the generated `NodeId`.
2. We need to refresh a bucket if it hasn't been used within a certain time. This defaults to 1 hour. The timeout can be implemented as a priority queue and a single timer tracking the shortest time, no need to maintain 255 timers at once.
    1. track timeouts for each bucket
    2. refresh timeout whenever a bucket is used.
    3. when a timeout on a bucket is triggered we need to do a refresh bucket operation on that bucket.
3. When entering the network we need to refresh all k-buckets further away than its closest neighbour.
    1. when entering the network after searching for itself in the network the node then needs to refresh every bucket 'above' it's closest neighbour.

`refreshBucket` is pretty simple. It generates a random `NodeId` within the targeted bucket and preforms a `nodeFind` operation for that nodeId. We generate the desired 'distance' by generating 32 random bytes and then zeroing the bits above the target bucket bit and forcing 1 for the target bucket bit. The random `NodeId` is made by XOR-ing the distance with the base `NodeId`. 

The activity timers for buckets were implemented the following way. We maintain a map of `bucketIndex -> deadline` to track when each bucket needs to be refreshed. We then maintain a single timer that triggers when the closest deadline has passed. When the timer is triggered we add any bucket that has passed it's deadline to the queue and reset the timer to the next closest deadline. When resetting timers for buckets that have been access we just update the map entry for that timer. if the updated bucket is the closest deadline then we update the timer again. When a bucket is is the queue, it's deadline is disabled by setting it to `0` and is not considered when updating the timer or adding to queue. 

The `refreshBucket` queue is implemented using a set of unique bucket indexes so a bucket can only be in the queue once. The queue is asynchronously digested doing a single `refreshBucket` at a time. When a `refreshBucket` operation is completed for a bucket it's deadline is reinstated. if a bucket is updated while in the queue then it's deadline is reinstated and removed from the queue. 

Entering the network has been updated. `NodeConnectionManager.syncNodeGraph` has been updated such that when it completes the initial sync it will add any buckets above the closest node's bucket to the `refreshBucket` queue.

### Additional context

- originally discussed when talking about adding a node to another node's `NodeGraph` in https://github.com/MatrixAI/js-polykey/issues/344#issuecomment-1043941991

### Tasks
1. [x] Implement `refreshBucket` as per the above spec.
3. [x] Implement refresh timeouts for every node bucket as per the above spec
4. [x] Update how the node enters the network as per the above spec.

### LINKED ISSUE #359: Setting a Node involves NodeManager and NodeGraph, where NodeManager handles bucket overflow policy
### Specification
Adding a node to a `NodeGraph`'s bucket needs to be updated. There are 3 cases when adding a node that need to be handled
1. If the node already exists in the bucket the last updated need to be updated.
2. If the node is not already in the bucket AND the bucket has room. The node is added into the bucket.
3. If the node is not in the bucket AND the bucket is full. We need to ping the oldest node in the bucket, if the old node does not respond we drop it for the new node, otherwise we drop the new node and update the last updated for the old node.

We need to add a `force` flag for always add the new node for manually adding nodes.

The core of the functionality needs to be moved into the `NodeManager`. since the `NodeGraph` is purely data structure the checking of the nodes before adding them should be handled inside the NodeManager.

As implemented, the `NodeGraph` just stores the nodes within buckets and tracks the number of nodes in the bucket. The more complex logic is handled by the `NodeManager` now. 

### Additional context

- https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1070138301
- Relates #326 
- Relates #212 
- Relates #322 
- Relates #344 
- Relates #150 
- Relates #226 
- Relates #345 

### Tasks
- [x] 1. Remove bucket limit checking and core adding logic from `NodeGraph` to `NodeManager.addNode` method. `NodeGraph` is purely data-structure anything that requires complex checking or networking needs to be outside it.
- [x] 2. Handle cases when adding a node.
  1. [x] Node already exists.
  2. [x] Node exists, bucket has room.
  3. [x] Node exist, bucket is full.
- [x] 3. Add `force` parameter to force adding a node if bucket is full.
- [x] Update testing.
  - [x] fix any tests broken by logic moving to `NodeManager`
  - [x] add test for the 3 cases when adding nodes.
  - [x] add test for `force` parameter

### LINKED ISSUE #363: Adding timeout overrides to `NodeConnectionManager` methods
### Specification
While we can specify the timeouts when creating a `NodeConnectionManager` there are cases where we want to be able to specify a timeout when creating a connection. 

For this we will need to add an optional `connConnectTime` parameter to any `NodeConnectionManager` method that can create a connection. this ill override the timeout of trying to create a connection if set.

### Additional context
- Relates #322 needs this to avoid blocking for a long time when checking a node.

### Tasks
1. Add optional timeouts to all of the `NodeConnectionManager` methods that could create a new connection.

tegefaulkes:
### Description
Note that this has been migrated from MatrixAI/Polykey#326.

Once MatrixAI/Polykey#310 has been merged, we'll finally be able to move onto the deployment of our testnet into AWS.

I foresee this to be achievable in a few stages:
1. Deploy Polykey onto AWS and ensure we have a working remote keynode (#194)
2. Create the testnet seed nodes and store these securely somewhere (#285)
3. Use these seed nodes for our domain name (#177)

* Relates to MatrixAI/Polykey#159 
* Relates to MatrixAI/Polykey#161 
* Migrated from MatrixAI/Polykey#326 
* Relates to MatrixAI/Polykey#297 

### Issues Fixed

* ~~Fixes MatrixAI/Polykey#324~~ being completed in MatrixAI/Polykey#310

Testnet deployment:

* Relates to MatrixAI/Polykey#194
* Relates to MatrixAI/Polykey#285 
* Relates to MatrixAI/Polykey#177 

NodeGraph Structure:

* Relates MatrixAI/Polykey#244 
* Fixes MatrixAI/Polykey#317 
* Fixes MatrixAI/Polykey#158
* Fixes MatrixAI/Polykey#359 

Node Adding Policies:

* Fixes MatrixAI/Polykey#212 
* Fixes MatrixAI/Polykey#322
* Fixes MatrixAI/Polykey#344 
* Fixes MatrixAI/Polykey#363 

Node Removal Policies:

* Fixes MatrixAI/Polykey#150 
* Fixes MatrixAI/Polykey#226 
* Fixes MatrixAI/Polykey#345

### Tasks

- ~~[ ] 2. Complete MatrixAI/Polykey#324~~ - being completed in MatrixAI/Polykey#310
- [x] 1. Read through and revisit all previous issues
- ~~[ ] 3.  Complete MatrixAI/Polykey#194~~ - Extracted out, should be finished after this PR merges
  * Ultimately provides us with a deployment of Polykey on AWS!
  * Task list there should be sufficient to follow process.
- ~~[ ] 4. Complete MatrixAI/Polykey#285~~ - Extracted out, should be finished after this PR merges
- ~~[ ] 5. Complete MatrixAI/Polykey#177~~ - Extracted out, should be finished after this PR merges
- ~~[ ] 6. Created automated testing that utilises the testnet~~ - to be done in MatrixAI/Polykey#357 
  - ~~[ ] - MatrixAI/Polykey#159~~
  - ~~[ ] - MatrixAI/Polykey#161~~
- [x] 7. Add agent service `nodes` tests https://github.com/MatrixAI/js-polykey/issues/194#issuecomment-1032173720
  * [x] - `nodesChainDataGet`
  * [x] - `nodesClosestLocalNode`
  * [x] - `nodesHolePunchMessage`
- [x] 8. Changed `Cmd` to `EntryPoint` because PK only has 1 executable and makes `docker run ...` easier.
- [x] 9. The `--seed-nodes='<defaults>;...'` will now mean that any specified seed nodes overrides the defaults rather than the other way around
- [x] 10. If seed nodes contains the own `NodeID`, the own `NodeId` will be automatically filtered.
- [x] 11. Make `pk agent start` return status information such as `nodeId` and not just `recoveryCode`
- [x] 12. All PK commands can contact a remote agent (except `agent start` and the like), it needs to be made more obvious whether we are contacting the local agent or remote agent https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1040120023
  - This is done by making the required client connection parameters of node ID, client host and client port all or nothing when set. This means if they are partially set, this results in a new exception `ErrorCLIClientOptions`.
  - If they are all set, that means a connection to an agent not specified by the local status.
  - If none are set, this means relying on the local status.
- [x] 13. The seed node should be adding the details of any node connecting to it to its own `NodeGraph`: see issue MatrixAI/Polykey#344
    - currently the seed node does **not** add an entry in its `NodeGraph` for a connecting node
    - this is essential for the Kademlia implementation to start to function, and for the node's details to eventually propagate out into the wider Polykey network
    - see https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1042518638
- [x] 14. Investigate and fix encountered Polykey errors/problems during testnet testing https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1041190329:
   - ~~[ ] a. No output provided on `pk identities trust`? However, I understand that we've previously shied away from providing "success" output on every single one of our commands. It can be tricky to diagnose problems when this is the case though.~~ - moved this to https://github.com/MatrixAI/js-polykey/issues/334#issuecomment-1043779027
   - ~~[ ] b. `pk identities trust` seemingly not adding a node to the gestalt graph when it doesn't already exist in `GestaltGraph`/`NodeGraph` - see https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1042543612 and https://matrixai.slack.com/archives/CEAUUV5QX/p1645069875382019~~ - moved this to https://github.com/MatrixAI/js-polykey/issues/334#issuecomment-1043779027
   - [x] c. `SyntaxError: Unexpected end of data` on an "invalid" node ID when using `decodeNodeId` - rebased on master has new js-id 3.3.1 that catches syntax errors when decoding multibase encoded strings
- [x] 15. Refactor `NodeGraph` structure
  * [x] The new structure will use 3 nested sublevels to achieve MatrixAI/Polykey#244 
    - `buckets` sublevel will contain each bucket, where each bucket sublevel contains `NodeId` to `NodeData`
    - `meta` sublevel contains each bucket, where each bucket sublevel is a config structure, currently only `count`
    - `lastUpdated` or `index` sublevel that contains each bucket, and each bucket sublevel contains `lexi(lastUpdated)-NodeId` to `NodeId`, the key is a compound index, allowing us to efficiently acquire the most up-to-date or least up-to-date node entry
  * [x] New methods like `getNodes` and `getBuckets` will allow us to interrogate the state of the `NodeGraph` efficiently, by streaming data out of the `NodeGraph`, this will be important for debugging, and later analytics
  * [x] Integration of `DBTransaction` and new iterator feature of sublevels that allow us to maintain a snapshot of the DB at a point in time
  * [x] Reviewing kademlia trie structure and bringing ideas in - reviewed and discovered that trie structure properties are not necessary in our NodeGraph, our design is already efficient and optimal
- [x] 16. Optimise `getClosestNode` method based on new `NodeGraph` structure
- [x] 17. Relationship between `NodeGraph` and `NodeManager` should be tightly coupled due to the adding and removal policies of nodes, maintaining the bucket limit must involving pinging old nodes
- [x] 18. tests for changing root keys and connections.
  - Check how active connections work during a keychange.
  - Check if new connections use new TLS config
  - check if a renew/reset key event propigates to where it is needed.
-  [x] 19. cleaning up and small fixes to nodesUtils parser functions
-  [x] 20. adding force and ping flags to nodes add command

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
This was migrated from MatrixAI/Polykey#326 since we needed to change to a feature branch. Work from this PR/branch from now on.

CMCDragonkai:
@tegefaulkes please remember to tick off anything above in case they are done, or cross out things that are no longer relevant in this PR.

CMCDragonkai:
Change the base branch to target MatrixAI/Polykey#374 branch.

tegefaulkes:
Starting rebase on feature-dep-upgrades.

tegefaulkes:
I've done the first blush rebase but there are some parts that I have to fix. Most of it is just the `NodeGraph` since It had a major refactor in this PR. So I need to go over it and apply the locking and DB changes. On top of that there are likely a bunch of bugs that were introduced that I have to review and fix.

tegefaulkes:
Finished with the re-base + initial fixes. Everything is building and linting fine now. I'm going to start fixing the tests.

src/utils/utils.ts
```diff
@@ -154,14 +154,16 @@ function promisify<T>(f): (...args: any[]) => Promise<T> {
   };
 }
 
-/**
- * Deconstructed promise
- */
-function promise<T>(): {
+export type PromiseType<T> = {
```
CMCDragonkai: 1. The `export` should be at the very end.
2. This is a nominal type for the deconstructed promise, prefer calling this `PromiseDeconstructed`.

src/utils/utils.ts
```diff
@@ -220,6 +222,67 @@ function arrayZipWithPadding<T1, T2>(
...
+  iterable: AsyncIterable<T>,
+): Promise<Array<T>> {
+  const arr: Array<T> = [];
+  for await (const item of iterable) {
+    arr.push(item);
+  }
+  return arr;
+}
+
+function bufferSplit(
```
CMCDragonkai, line 251: This may not be necessary anymore if we are able to use `KeyPath` directly.

Primarily for composite keys where we wanted to have it sorted by a particular index, but still acquire the second part of the key.

But don't remove this function because it may be useful in the future.

src/nodes/utils.ts
```diff
@@ -1,29 +1,75 @@
...
+  NodeBucket,
+  NodeBucketIndex,
+} from './types';
 import { IdInternal } from '@matrixai/id';
-import { bytes2BigInt } from '../utils';
+import lexi from 'lexicographic-integer';
+import { bytes2BigInt, bufferSplit } from '../utils';
+import * as keysUtils from '../keys/utils';
+
+// FIXME:
```
CMCDragonkai: Shoudl look for all the FIXMEs and resolve them. Also remove any spurious commentary.

src/utils/utils.ts
CMCDragonkai: Look at `lexicographic level iteration order` test in js-db to see the rules of iteration that the DB does now, then you can see if it applies to what we want to use for composite keys.

tegefaulkes:
I can't get the iterator to work properly with the `gt`  or `lt` options. I'm not sure if I'm using it wrong or if it's broken. Looking at the js-db tests we don't seem to be testing if they work so it might be broken.

CMCDragonkai:
This PR will need to target staging and rebase on top. However @emmacasolin is still pending some changes to the CI/CD so we can hold off on that for now. There are changes from MatrixAI/Polykey#374 that this should rebase on, because of all the fixes to logging.

CMCDragonkai:
@tegefaulkes is this PR's description updates to point to abort controller changes?

CMCDragonkai:
Also please clean up the task list, and start adding ETAs into it as well.

What is the final task list?

tegefaulkes:
Re-based and targeting staging now.

tegefaulkes:
Notes on MatrixAI/Polykey#378 completion
- MatrixAI/Polykey#244 is mostly done but depends on SI transactions. Is there any API changes/fixes that need to be done when SI is merged?
- MatrixAI/Polykey#317, I don't know if this has been addressed. I'll need to look into it more.
- MatrixAI/Polykey#158, is rulled irelevant. Will be auto closed when merged.
- MatrixAI/Polykey#359, Seems to be done.
- MatrixAI/Polykey#212, is done
- MatrixAI/Polykey#322, has one last task to complete/ look into
- MatrixAI/Polykey#344, seems to be done
- MatrixAI/Polykey#363, Looks done.
- MatrixAI/Polykey#345, Seems to be done.

MatrixAI/Polykey#194, MatrixAI/Polykey#285 and MatrixAI/Polykey#177 depend on this PR but are not directly fixed by it. I think they need to be done after MatrixAI/Polykey#378 is merged so It makes sense to extract them to their own epic.

SO in summary only the following issues possibly needs work still.
- MatrixAI/Polykey#244 - Only if there are any API changes that need to be fixed here. otherwise it's just updating a dependency?
- MatrixAI/Polykey#317 - Wasn't really my donain of work. So I don't know if anything has been done for it and what still needs to be done.
- MatrixAI/Polykey#322 - Need to check if force and no ping flags have been added to a bin command but otherwise done.

MatrixAI/Polykey#150 and MatrixAI/Polykey#226 haven't had any direct work on them but I'd argue that the functionality required for them is already handled. An invalid node will not be added to the NodeGraph. and any nodes that become invalid or inactive are passively filtered out when adding new nodes.

The code as it stands now is building and should be passing tests. it is also reasonably up to date with the staging branch.

tegefaulkes:
`NodeConnectionManager.getClosestGlobalNodes` is using an priority queue for finding nodes. The shortlist is an array and only filters out contacted nodes when adding or contacting. So it's allowing for duplicate nodes. This will be a problem in larger networks.  So i'll need to fix this but i'll come back to it after the other tasks have been addressed. I'll also have to check if there are other priority queues that have the same problem. `SyncNodeGraph` maybe?

CMCDragonkai:
There will be left over work from merging this PR:

1. Fixing up the queues MatrixAI/Polykey#329 so that we have a single consistent queue implementation for all the places where we are doing queuing.
2. Fixing up asynchronous cancellation and timer integration MatrixAI/Polykey#297, we have `AbortController` usage now in a some places under the nodes domain, but it's not consistent with the rest of the PK yet

CMCDragonkai:
@tegefaulkes there's a bunch of stuff that looks like a conflict with staging branch, may require a rebase again? Possible changes from @emmacasolin?

tegefaulkes:
It's due to staging branch being force pushed causing commits here that are duplicates from staging to conflict with themselves. I just need to do a re-base again. the conflicts should be minimal since I think most of the staging branch changes are CI/CD related.

tegefaulkes:
This is ready to merge if CI/CD passes.

CMCDragonkai:
Cool I see all the tests pass. The nix dry run still fails, we need to investigate that later in the staging branch.

There are some WIP commits you have you should squash, once you have cleaned up your commits, we can merge this into staging.

I'm wondering there's 2 issues with the CI/CD commits in staging atm:

1. nix dry run job
2. the usage of `pkg.js` - this may be connected to the nix

We could try to fix them in staging branch first, then merge this on top, or merge this on top first, and then move those CI/CD commits on top of this in order to tackle them. Was there any changes in this branch that depends on the CI/CD changes you rebased on top?

CMCDragonkai:
These commits should be squashed and given description as to what the fixes are:

image

Better description for this? https://github.com/MatrixAI/js-polykey/pull/378/commits/8e04b578d4eccc10dcbd73fb58c934a27e57e932

Better description for this? https://github.com/MatrixAI/js-polykey/pull/378/commits/e3011dac16168c2d710d157edad1b31254cb68bc

emmacasolin:
The wip commits have now been squashed and renamed. All of the tests should still be passing, in which case this can be merged into staging and the CI commits moved on top of this. There shouldn't be anything here that relies on those changes.

CMCDragonkai:
Ok merging into staging, afterwards @emmacasolin you can rebase and move the CI commits above this merge commit.

CMCDragonkai:
Many of the issues didn't autoclose because PK's default branch is still master. I've closed relevant issues manually, but we will switch to staging as the default branch once the CI/CD is all fixed up in the staging branch.

CMCDragonkai:
Some remaining issues that come out of this PR:

* MatrixAI/Polykey#324 - need @tegefaulkes to confirm if this is done, or @emmacasolin can confirm
* MatrixAI/Polykey#317 - this is done here, however @tegefaulkes needs to create new issue for general keypair checks - be as comprehensive as possible
* MatrixAI/Polykey#244 - this is waiting on SI to be done
* MatrixAI/Polykey#194, MatrixAI/Polykey#285, MatrixAI/Polykey#177 are all procedures to be done when staging is ready to be used and docker containers are built (I still need to review the entire codebase in staging).

PR #391, opened 2022-06-28
Packaged executable integration tests - initial integration tests with docker

### LINKED ISSUE #347: Problem with `tests/bin/keys/reset.test.ts` and `tests/bin/keys/renew.test.ts` stopping the PolykeyAgent
### Describe the bug
When `tests/bin/keys/reset.test.ts` and `tests/bin/keys/renew.test.ts` use the shared global agent to run their tests (as opposed to their own separate agents) and are run at the same time, after the first test has finished running it will sometimes shut down the global agent even though there are still other processes using it. It will do this only if the other test is reading from the global data directory at the same time as when the first test is attempting to stop:
```
INFO:KeyManager:Reading /tmp/polykey-test-global-6W6c7T/agent/state/keys/db.key
INFO:KeyManager:Reading /tmp/polykey-test-global-6W6c7T/agent/state/keys/vault.key
INFO:renew test:Stopping PolykeyAgent
```

### To Reproduce
1. Checkout https://github.com/MatrixAI/js-polykey/pull/311/commits/dc53883c951b8a2b1e1978897856509b5962fea3
2. Run `npm test -- ./tests/bin/keys/reset.test.ts ./tests/bin/keys/renew.test.ts` (note that putting `renew` before `reset` seems to not reproduce the same issue)

### Expected behavior
The `Stopping PolykeyAgent` message can only come from the `PolykeyAgent.stop` method, however, this is not being called from inside `globalAgentClose`, as would be expected. We need to trace where this is being called from as it should only be called from `globalAgentClose` when using the global agent, which should only happen after all tests have finished running.

### Additional context
- https://github.com/MatrixAI/js-polykey/pull/311#issuecomment-1043952073
- https://github.com/MatrixAI/js-polykey/pull/311#issuecomment-1046449996

### LINKED ISSUE #389: Integration tests for `packaged executable on target platforms
### Specification

 after creating the packaged executable using `pkg` we need to run a variety of tests on it. We need to do this in all of the main environments such as Linux, windows and MAC. We will also need to test the docker image but I'm nut sure the same tests can apply in that case.
 
 All of the main tests we need to do are already done by the bin tests. So we just need to repurpose all or a subset of these tests and run them on the target platform. Assuming the tests themselves don't need some kind of modification to work on the target platforms. Then the only changes to the tests that need to be made are to exec the packaged executable instead of the ts/js code.
 
 Keep in mind that most of the tests don't actually exec pk to run them but are run in thread using mocks to capture output. With that in mind we will need to review the tests to see if any of them require access to the thread to apply mocks for the tests. We may also apply a subset of the tests to run instead of running all of them. 
 
 I think we can prioritise the tests based on complexity. starting with just running `Polykey` to full integration. I think the stages go as follows.
 1. Running polykey
	 1. bin/agent
	 2. bin/bootstrap.test.ts
	 3. bin/polykey.test.ts
 2. local local
	 1. bin/keys
	 2. bin/sessions.test.ts
	 3. bin/vaults
	 4. bin/secrets
 3. network level
	 1. bin/nodes
	 2. bin/notifications
	 3. bin/identities

We will need a way to specify that we are running the tests targeting the packaged executable. This will be specified as a global however it will be set via and env variable. So the global check the env and provides the parameter to the tests. 

The tests will need to check the target and change it's behaviour based on that. First by changing modifying how the commands are run. The utility functions `pkStdio`, `pkExec` and `pkSpawn` will need to be modified to run the packaged target if that global parameter is set. The functions will need to function with the same API so the the new behaviour will be a drop in replacement. 

Most tests utilise `okStdio` which is running the command 'in thread' with mocking. This could be a problem because if any tests is doing mocking on the command itself then we can't run it in a separate thread using exec. This means the test itself could be incompatible with testing the packaged executable. Either a change to the test needs to be made or the test needs to be filtered out when testing the packaged executable. This is unlikely to be a problem since there seems to be little mocking and any mocking will be done against the polykey instance running 'in thread' anyway.

### Additional context

- Related #379 

### Tasks

1. Allow setting a global variable specifiying the path to the packaged executable under test
2. Allow bin tests to target the specified path when running the tests
3. Fix up or filter out any tests that are incompatable with using the packaged executable for testing.
4. test manually for each target platform to ensure tests are working.
5. configure CI/CD to use integration tests for each target platform and ensure that it works.

### LINKED ISSUE #404: Switch from using the global agent to using global keypairs for tests
### Specification
https://github.com/MatrixAI/js-polykey/issues/264 introduced our global agent which is used in our tests as a way to reduce setup time (since we no longer need to create an agent for every test). However, this solution complicates our tests by requiring all side effects to be reversed (some side effects cannot be undone, e.g. sigchain changes and some identities changes) and preventing tests that use the global agent from running in parallel.

An alternative solution is to have a pre-generated root private key that we can pass into the `PolykeyAgent` creation. This should sidestep the key generation process and speed up agent creation significantly. 

### Additional context
- https://github.com/MatrixAI/js-polykey/issues/264 - Original issue for improving test scalability (where the global agent was introduced)
- https://github.com/MatrixAI/js-polykey/pull/394#issuecomment-1179487869 - Comment discussing the need to switch from using the global agent to global keys

### Tasks

- [x] 1. Add an override to `KeyManager` generation to provide a root private key to skip root key-pair generation.
- [x] 2. Propagate this override up to the `PolykeyAgent` parameters
- [x] 3. create tests confirming this behaviour.
- [x] 4. Propagate this change up to `pk agent start` and `pk bootstrap`. Including new parameter + ENV variable.
- [x] 5. Replace all usages of global agent with root key override.
- [ ] 6. Move global keypair fixtures to a common directory as `tests/fixtures`

### LINKED ISSUE #407: Integration tests for docker image
### Specification

This issue is to track the progress of PKG integration tests for the docker image. 

A selection of bin tests need to be updated to support using the docker image as the target for testing. 

The main tests that need to be updated in order are;
1. `bin/agent/start.test.ts`
2. `bin/bootstrap.test.ts`
3. rest of `bin/agent`
4. `bin/polykey.test.ts`

Depending on the decided scope, the rest of the tests that could be updated are listed in MatrixAI/Polykey-CLI#10

Some things specific to docker that need to be checked refereed from https://github.com/MatrixAI/js-polykey/issues/389#issuecomment-1180048178. 

1. Rather than `--network host`, you'll need an IP address and Port mapping, make sure this works on the CI/CD too
2. Rather than `--pid host` you'll need handling of the signals by the PK agent
3. With `--userns host` and `--user "$(id -u)"` this may still be necessary locally, in order to allow the files written to be readable by the test code which is running as part of our local user. But on the CI/CD this both may be the root user. But experiment with this.

### Tasks

- [x] 1. update to use `testIf` to explicitly run the test if testing docker. Right now we implicitly running them by disabling the tests that are not supported for docker. 
- [x] 2. convert the following test...
    - [x] `bin/agent/start.test.ts`
    - [x] `bin/bootstrap.test.ts`
    - [x] rest of `bin/agent`
    - [x] `bin/polykey.test.ts
    - [x] rest of the bin tests 
- [x] 3. update `.env.example` with any changes to the required env vars.
- [ ] 4. ensure CI/CD job is working.

### LINKED ISSUE #408: Integration tests for windows build

### LINKED ISSUE #409: Integration tests for Mac build
### Specification

This issue is to track the progress of PKG integration tests for the Mac build. 

A selection of bin tests need to be updated to support using the mac executable as the target for testing. 

The main tests that need to be updated in order are;
1. `bin/agent/start.test.ts`
2. `bin/bootstrap.test.ts`
3. rest of `bin/agent`
4. `bin/polykey.test.ts`

### Tasks

1. spec this out some more

### LINKED ISSUE #410: Expand `pkStdio`, `pkExec` and `pkSpawn` utils to target executables
### Specification

We need to expand the `pkStdio`, `pkExec`, and `pkSpawn` utility commands to support testing executable files such as the win, mac builds and docker container. Counterparts to these commands need to be created. These new commands need to mimic their counterparts but execute the target executable.

### Tasks

1. [x] Create counterpart commands that execute the targeted executable
    * [x] `pkExecTarget`
    * [x] `pkStdioTarget` - this can't work with mocking so keep that in mind.
    * [x] `pkSpawnTarget`
3. [x] add switching utility functions that will use a `pk` command or it's counterpart to execute the command.
4. [x] add code docs to each command clearly specifying usage.

### LINKED ISSUE #417: Signal handling for windows
### Specification

Windows handles it's signals differently. According to https://nodejs.org/docs/latest-v16.x/api/process.html#processkillpid-signal. 

"On Windows, where POSIX signals do not exist, the signal argument will be ignored, and the process will be killed forcefully and abruptly (similar to 'SIGKILL'). See [Signal Events](https://nodejs.org/api/process.html#signal-events) for more details.".

However acording to the node signals https://nodejs.org/docs/latest-v16.x/api/process.html#signal-events

- `SIGHUP` is generated on Windows when the console window is closed, and on other platforms under various similar conditions. See [signal(7)](http://man7.org/linux/man-pages/man7/signal.7.html). It can have a listener installed, however Node.js will be unconditionally terminated by Windows about 10 seconds later. On non-Windows platforms, the default behavior of SIGHUP is to terminate Node.js, but once a listener has been installed its default behavior will be removed.
- `SIGBREAK` is delivered on Windows when Ctrl+Break is pressed. On non-Windows platforms, it can be listened on, but there is no way to send or generate it.
- `SIGINT` from the terminal is supported on all platforms, and can usually be generated with Ctrl+C (though this may be configurable). It is not generated when [terminal raw mode](https://nodejs.org/docs/latest-v16.x/api/tty.html#readstreamsetrawmodemode) is enabled and Ctrl+C is used.'SIGINT' from the terminal is supported on all platforms, and can usually be generated with Ctrl+C (though this may be configurable). It is not generated when [terminal raw mode](https://nodejs.org/docs/latest-v16.x/api/tty.html#readstreamsetrawmodemode) is enabled and Ctrl+C is used.
- `SIGTERM` is not supported on Windows, it can be listened on.

So we need to make sure that all of the signals that windows can produce are supported by Polykey. Then we need to create new tests testing that this works.

### Additional context

- Related #408 
- Related #401 

### Tasks

1. manually test and confirm what signals are produced when
    - polykey is killed used `ctrl + C` in the terminal running it.
    - killed by closing the terminal running it.
    - killed by using task manager.
2. Add any missing handlers for the signals produced by windows.
3. add tests confirming the new handlers work.

### LINKED ISSUE #420: Override key generation for all tests
### Specification

With the changes from #404 it is now possible to skip key generation all together. So far this optimisation has been applied to the bin tests but now we need to apply it to the rest of the jest tests.

This can be done by taking every instance of `createPolykeyAgent` and adding the following `keysConfig` parameter to it. Currently `globalRootKeyPems` contains 10 pre-generated private key Pems.
```ts
keysConfig: {
        privateKeyPemOverride: globalRootKeyPems[0],
      },
```

If there are any instances of a `polykey agent start` bin command being used then you can use the following `PK_ROOT_KEY: globalRootKeyPems[0]`  Env variable.

Any Polykey agents that are created inside a `beforeAll` should now be done inside the `beforeEach`. This will give better test isolation.

### Additional context

- Related #404

### Tasks

- [x] 1. convert every instance of `createPolykeyAgent` to use the above config to skip key generation.
- [x] 2. convert any polykey agents that are created inside a `beforeAll` to use `beforeEach`.

tegefaulkes:
### Description

This PR modifies the `tests/bin` tests to target a packaged executable for the tests. This will be done by setting the `pk_pkg_executable_path` as the path to the desired executable we wish to test. This should then apply all bin tests to this executable.

The details are inside of issue MatrixAI/Polykey-CLI#10.

### Issues Fixed

* Fixes MatrixAI/Polykey#410 
* Fixes MatrixAI/Polykey#407 
* Fixes MatrixAI/Polykey#404
* Fixes MatrixAI/Polykey#420 
* Fixes MatrixAI/Polykey#347
* Related MatrixAI/Polykey-CLI#10 
* Related MatrixAI/Polykey-CLI#12
* Related MatrixAI/Polykey-CLI#11 
* Related MatrixAI/Polykey-CLI#13 
* Related MatrixAI/Polykey-CLI#14

### Tasks
Given the scope of this PR, sub-tasks will be tracked with each issue.

- [x] 1. Allow bin tests to target the specified path when running the tests MatrixAI/Polykey#410 
- [x] 2. update tests to support docker image MatrixAI/Polykey#407 
- [x] 6. MatrixAI/Polykey#404 
- [x] 7. test optimisations MatrixAI/Polykey#420
- ~3. MatrixAI/Polykey-CLI#13~ - addressed in new PR
- ~4. update tests to support win build MatrixAI/Polykey-CLI#11~  - addressed in new PR
- ~5. update tests to support mac build MatrixAI/Polykey-CLI#12~  - addressed in new PR

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
Can we keep the env variable simple? Just `PK_PATH` could be used. Then `pkPath` is the variable.

I'm also wondering if there's any self executing logic where in our code and if there is, we may want to out out bundled executable into the `PATH` itself so that it can be done. But any self executing logic should be done more robustly by not relying on the `PATH` and instead finding the path to it's own script.

Note that in comparison to `describeIf` and `testIf`, this env variable is acquired and centralised to the globals in `jest.config.js`. Whereas environmental conditions are done test-module-specific where `describeIf` and `testIf` are used.

CMCDragonkai:
One should note that using `PK_PATH` in our `.env.example` to ensure that such testing can also be done locally. It must by default be unset.

Only if set, is it used.

And you want to have a check that if set, and is empty string or an invalid path or path to non-executable, throw an exception. This could be done in `jest.config.js`.

tegefaulkes:
I've modified pkSpawn to use the packaged executable if the pkgPath is set. I've been testing this on `tests/bin/agent/start.test.ts/ starting in foreground`. It mostly works however it seems that it is not getting terminated properly. The process is getting killed properly. However the `SIGTERM` signal is not getting caught and as such `polykey` is not stopping gracefully. It just immediately terminates. I'll have to check if windows has any quirks causing this.

tegefaulkes:
Tested it under Ubuntu and it works fine. So windows seems to be the problem here.

tegefaulkes:
According to https://nodejs.org/docs/latest-v16.x/api/process.html#processkillpid-signal. 

"On Windows, where POSIX signals do not exist, the signal argument will be ignored, and the process will be killed forcefully and abruptly (similar to 'SIGKILL'). See [Signal Events](https://nodejs.org/api/process.html#signal-events) for more details.".

This might just be a `child_process` problem. I'll check if there is an alternative.

Using `process.kill(agentProcess.pid!, 'SIGTERM');` has the same problem.

CMCDragonkai:
As far I know windows does have signals. But in this case, you're right that windows does not support `SIGTERM`.

This becomes complicated, because we are using SIGTERM to perform graceful exit of the agent. There are equivalent windows events in https://stackoverflow.com/questions/2055753/how-to-gracefully-terminate-a-process#2055810.

So we might have to find a way for nodejs to receive the equivalent of a graceful exit signal, and send that signal instead in our tests if we want it to also work on windows.

tegefaulkes:
According to [this](https://nodejs.org/docs/latest-v16.x/api/process.html#signal-events) we may need to add some new handlers. We will then need to confirm the behaviour on windows by closing the window, ending via task manager and using `kill PID`.
- `SIGHUP` is generated on Windows when the console window is closed, and on other platforms under various similar conditions. See [signal(7)](http://man7.org/linux/man-pages/man7/signal.7.html). It can have a listener installed, however Node.js will be unconditionally terminated by Windows about 10 seconds later. On non-Windows platforms, the default behavior of SIGHUP is to terminate Node.js, but once a listener has been installed its default behavior will be removed.
- `SIGBREAK` is delivered on Windows when Ctrl+Break is pressed. On non-Windows platforms, it can be listened on, but there is no way to send or generate it.

I can't find any info on what exactly happens when you do `kill PID` via CMD or powershell. Hopefully we receive a windows flavoured signal in that case. Worst case we receive no signal and the process exits. In any case I think the tests need to be modified to supply the correct signal depending on platform. Using `process.kill` or `agentProcess.kill` with `SIGTERM` or no signal results in the process abruptly exiting.

CMCDragonkai:
Yea that makes sense. Basically to deal with Windows, we have to also listen for `SIGINT`, `SIGHUP` and `SIGBREAK`.

So in our jest tests, when we want to signal graceful termination, we send it `SIGINT`, that is likely going to be portable. I just coded it to `SIGTERM` since that was considered the standard on unix systems.

I think we could just make our graceful termination trigger upon all of these signals for the polykey agent:

1. `SIGINT`
2. `SIGHUP`
3. `SIGBREAK`
4. `SIGTERM`

Then change all of our tests that are using `SIGTERM` to instead use `SIGINT` which should be portable across all operating systems.

We should also have tests that test each signal for a given system. So on windows, we expect `SIGINT`, `SIGHUP`, `SIGBREAK` all should gracefully shutdown the program.

CMCDragonkai:
Sometimes `SIGINT` is sometimes useful for other things, but I think for daemons generally `SIGINT` means program shutdown too. Interactive shells use `SIGINT` to interrupt what it is currently doing, but not kill the shell. For all other foreground program/daemon, this should should just shutdown the program.

tegefaulkes:
Just some extra notes on what I've tried when testing the signal problem with windows. 
- so far all of the testing was done via the `start in foreground` test under `tests/bin/agent/start.test.ts`. we should separate the test  into it's own file to reduce possible side effects with the test. 
- `SIGINT` should work universally and we already have a handler for it. However under windows we still have the same problem where the process ends abruptly without the signal getting handled.
- `SIGHUP` and `SIGBREAK` causes the `agentProcess.kill` and `process.kill` to throw an error even though they should be valid signals. So we may need an alternative method to trigger these signals via the tests.
- I have not tested what signals happen under windows if you do the following to the running agent under windows;
    - close the console window running the agent. should be a `SIGHUP`?
    - kill the process via task manager. should be a `SIGBREAK`?
    - use `ctrl + c`, `ctrl + d` or other keyboard break signals in the console. should be a `SIGBREAK`?
    - kill using the `kill <PID>` command. Not sure about this one. I just hope we can catch and handle it.
- We need a to work out a way to trigger these signals during tests. using `process.kill` or child_process's `agentProcess.kill` doesn't work the way we expect under windows.

@emmacasolin

tegefaulkes:
I have a working bash script that allows us to run the docker image in a similar way to running the normal polykey command. This runs fine in the terminal. However I can't for the life of me get an output on the stdout and stderr when running the script via `child_process`.

tegefaulkes:
Problem was PBCAK... I wasn't allowing the process any time to output anything before the test ended. it's outputting the streams with no problem.

tegefaulkes:
The script works now. It should mostly be a drop in replacement for running the bin commands. One major difference is that the node path will have to be specified with the `PK_NODE_PATH` env variable. I'm going to look at applying it to the start tests now.

tegefaulkes:
Testing this with the `start in foreground` test has revealed some new quirks. 
1. the PID inside the contain is always 1. more importantly it will not match the PID of the process we are spawning. I think we'll have to remove checks for this in tests or override it depending on what we're testing.
2. We can bind mount a folder and work out of that. but anything the container creates has weird permissions. Essentially the host doesn't have permission to read files created by the docker image. In the case of `start in foreground` test this means we can't read the `status.json` file as part of the test. Annoyingly there doesn't seem to be a quick and easy fix for this.  https://techflare.blog/permission-problems-in-bind-mount-in-docker-volume/ outlines the problem.
3. Sending a signal to the child process doesn't actually propagate to the docker image. so we will have to kill it some other way.

CMCDragonkai:
The container has the root PID... I think you could resolve this by also mapping the host pids in the docker container, just like host network and host IPC?

CMCDragonkai:
For 2., this is because the container is the `root` user. It's possible to have a container run as a separate user, maybe inherit the user of the host system? That may be a compile time thing or a runtime thing, you can have a check.

CMCDragonkai:
For 3., see this https://hynek.me/articles/docker-signals/, we would need to make that this is correct, as our orchestration system will rely on signals. (AWS ECS).

Also the linked article: https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html

tegefaulkes:
1. and 3. is solved by using the `--pid host` flag to share the pid namespace with the host. This avoids the quirks with having a PID of 1 so any signals are properly passed along.
2. Is solved by using `--userns host --user <UID>` to allow the permissions to match up with the host.

With this and cutting out some un-needed expectations,the `start in foreground` test is passing.

tegefaulkes:
Things I need to look into changing. 
1. change the docker image to use entrypoint using the path to the file we run instead of using `cmd`.
2. Make sure that the entrypoint is using the `[ "command" ]` syntax.

CMCDragonkai:
And using `pstree` to confirm that `/bin/sh` is not an intermediate process between docker and polykey.

CMCDragonkai:
Tomorrow separate out what can be merged into staging so we can do the testnet deployment procedures. Focus on `agent start` and `bootstrap` related commands, any fixes in source, or only fixes to tests right now?

tegefaulkes:
All of the agent start tests are functional now. The tests rely on some env variables to be set.
- PK_TEST_CMD=scripts/docker-run.sh - path to the command you want to execute, this case the docker testing script.
- PK_TEST_PLATFORM=DOCKER - This is used to filter what tests to run.
- PK_DOCKER_IMAGE=polykey-1.0.0:lyi2g9qpaq5i05c01hh4gcidm0jv2vh3 - Should be set to the image name after loading the docker image with `docker load --input <path to image>`

Note that I've created a modified `testIf` and `describeIf` called `runTestIf` and `runDescribeIf` that can be used like `runTestIf(condition)('name', ()=>{}, timeout))` so we can do things like `runTestIf(condition).only('name', ()=>{}, timeout))`.

Still todo:
- modify the CI/CD `integration:docker` test job to use the new tests.
- clean up commits and cherry pick to staging.
- convert bootstrap.test.ts to test docker image.
- update entrypoint of the docker image

CMCDragonkai:
> Note that I've created a modified `testIf` and `describeIf` called `runTestIf` and `runDescribeIf` that can be used like `runTestIf(condition)('name', ()=>{}, timeout))` so we can do things like `runTestIf(condition).only('name', ()=>{}, timeout))`.

That seems like an improvement over `testIf` and `describeIf` @emmacasolin. @tegefaulkes why not just replace `testIf` and `describeIf` usage?

> - pk_pkg_path=scripts/docker-run.sh - path to the command you want to execute, this case the docker testing script.

This should be `PK_COMMAND` right? We're using the `PK_` prefix for all of env variables as can be seen in `bin/utils.options.ts`, but because this is a development time env variable, you must put it into `.env.example`, make sure to use `#PK_COMMAND=`. I believe that sets the env variable to empty string. By default users are expected to copy `.env.example` to `.env` which is autoloaded by `nix-shell`. So it should be commented out to indicate you only use this to run tests against a different polykey executable. And later this can expand to an actual command string, like `docker run`... Note it would never be a shell string, but instead a command string just means it would be split by newlines across the child process exec.

> - PK_TEST_PLATFORM=DOCKER - This is used to filter what tests to run.
> - PK_DOCKER_IMAGE="polykey-1.0.0:lyi2g9qpaq5i05c01hh4gcidm0jv2vh3" - Should be set to the image name after loading the docker image with `docker load --input <path to image>`

Are these needed? I was expecting the CI/CD to automatically set the docker image path inside the docker run script that is used by `PK_COMMAND` above?

At the same time, the `PK_TEST_PLATFORM` seems strange. Our existing usage of `describeIf` and `testIf` are based on platform differences that are acquired from in-code, instead of referring to environment variables. The main idea was that `npm test` should just always work, and figure out what to test based on the platform.

In this case, testing a separate pk executable, should just trigger automatically based on the existence of `PK_COMMAND`.

However I imagine that now our integration tests would be changed to run `npm test -- ./tests/bin` to focus only on the `bin` tests.

Kind of like:

```sh
echo 'exec docker run ...' > ./tmp/docker-run.sh
chmod u+x ./tmp/docker-run.sh
PK_COMMAND='./tmp/docker-run.sh' npm test -- ./tests/bin

# In the ideal case, this should be possible
PK_COMMAND='docker run ...' npm test -- ./tests/bin
```

CMCDragonkai:
I am not entirely convinced that some of the bin tests are not able to run via the integration? I think all the bin tests should be runnable for integration. Using `docker run` with all the right namespace options should be no different from running `polykey` directly on the host operating system.

Later we can have "docker" specific tests, but that's separate from `tests/bin` which are meant to be generic tests on the executable of polykey.

Another issue is `tests/nat` which we can address later, since they rely on alot of OS facilities, we don't know if they will run in a docker in side docker. We only know that they run inside docker (but not in dind) - @emmacasolin

CMCDragonkai:
Regarding:

> update entrypoint of the docker image

Before you do this, please verify with `pstree` whether there is in fact a `/bin/sh` intermediate process.

If there isn't, then we don't need to modify this, we can continue using `Cmd`. On the other hand, because polykey container is a single-command container, it may be worth while to replace with `Entrypoint`. But verify it first, and even after using `Entrypoint`, you want to check with `pstree` to see the process tree.

tegefaulkes:
Right now the two tests i'm skipping are 
- 'start in background' - the docker container ends when the main process exits. since this test starts a detached process and exits the main one we end up the the container closing and killing the new process. Since the background process is not really a use case for docker I don't think it's worth converting this test.
- 'start with global agent - theses tests are using `pkStdio`, a global agent and mocking. it's not worth converting pkStdio, global agent and we can't use mocking if we do. So I don't think it's worth the time to convert these tests at this stage.

When using `pstree -pa` I only saw the container -> polykey and some processes under that. I'm not super certain what i'm looking for but I don't think it's running under a `/bin/sh` shell there.

CMCDragonkai:
> When using pstree -pa I only saw the container -> polykey and some processes under that. I'm not super certain what i'm looking for but I don't think it's running under a /bin/sh shell there.

Can you copy paste the output of `pstree` to here on the relevant container running?

tegefaulkes:
> Are these needed? I was expecting the CI/CD to automatically set the docker image path inside the docker run script that is used by PK_COMMAND above?

I think it's much simpler to just set an env variable for the docker image here than trying to meta-generate the script to contain it. Using `PK_TEST_PLATFORM` to filter the tests seemed the simplest since we can just set it when running the job. I wasn't sure of an in code way of checking if we're running against a docker image unless we want to check that `PK_TEST_CMD` is explicitly the path to the `docker-run.sh` script.

tegefaulkes:
pstree output.
```
  |-containerd-shim,485519 -namespace moby -id60ea62c4654cd
  |   |-polykey-agent,485540
  |   |   |-{polykey-agent},485553
  |   |   |-{polykey-agent},485554
  |   |   |-{polykey-agent},485555
  |   |   |-{polykey-agent},485556
  |   |   |-{polykey-agent},485557
  |   |   |-{polykey-agent},485558
  |   |   |-{polykey-agent},485559
  |   |   |-{polykey-agent},485560
  |   |   |-{polykey-agent},485561
  |   |   |-{polykey-agent},485562
  |   |   |-{polykey-agent},485563
  |   |   |-{polykey-agent},485564
  |   |   |-{polykey-agent},485565
  |   |   |-{polykey-agent},485566
  |   |   |-{polykey-agent},485567
  |   |   |-{polykey-agent},485568
  |   |   |-{polykey-agent},485569
  |   |   `-{polykey-agent},485570
  |   |-{containerd-shim},485521
  |   |-{containerd-shim},485522
  |   |-{containerd-shim},485523
  |   |-{containerd-shim},485524
  |   |-{containerd-shim},485525
  |   |-{containerd-shim},485526
  |   |-{containerd-shim},485527
  |   |-{containerd-shim},485528
  |   `-{containerd-shim},485546
```

CMCDragonkai:
> * 'start in background' - the docker container ends when the main process exits. since this test starts a detached process and exits the main one we end up the the container closing and killing the new process. Since the background process is not really a use case for docker I don't think it's worth converting this test.
> -'start with global agent - theses tests are using pkStdio, a global agent and mocking. it's not worth converting pkStdio, global agent and we can't use mocking if we do. So I don't think it's worth the time to convert these tests at this stage.

I see, there doesn't seem to be a way to re-use tests in a different way, so that the only way to group tests that can run inside docker is to use that conditional. Unless we were to create a directory like `tests/bin/docker` specific to `docker` but that would just result in alot of duplication.

Ok so then we need `PK_COMMAND` to refer to the pk command, but a separate env variable as you've done for indicating that this is a docker command. I think this variable needs to be more precise, something like `PK_COMMAND_DOCKER` as a flag. Both of these variables should result in globals being created and referenced in `jest.config.js`. (I'm wondering if it should be `PK_TEST_COMMAND` then, in order to keep it the same?).

* `PK_TEST_COMMAND`
* `PK_TEST_COMMAND_DOCKER`

This can avoid overlaps with env variables used for real in our `src/bin/utils/options.ts`.

> I think it's much simpler to just set an env variable for the docker image here than trying to meta-generate the script to contain it. Using PK_TEST_PLATFORM to filter the tests seemed the simplest since we can just set it when running the job. I wasn't sure of an in code way of checking if we're running against a docker image unless we want to check that PK_TEST_CMD is explicitly the path to the docker-run.sh script.

There are ways to do it, however it's not exactly relevant here. We're not checking whether we are running __inside__ docker (we already are, since we are in gitlab ci), but whether the command is __executing against a docker container image__. That's why I want to be precise here to avoid confusion.

CMCDragonkai:
That `pstree` output looks correct, the shim process always runs. It would be worth seeing an example of it when you just run a client command, instead of starting the agent. But that's more work on the integration side.

CMCDragonkai:
To improve DX, we have to specify what happens if `PK_TEST_COMMAND` or `PK_TEST_COMMAND_DOCKER` is set vs unset vs set with empty value.

That is does `unset` mean the same as `set with empty value`?

My original policy is that they should be different. In fact if set or not set is enough to indicate a boolean, then `set with empty value` is just something that is checked and an error is thrown.

This is kind of like how our existing options are parsed, if empty string, they are still parsed, but there's a check on whether this is allowed or not.

tegefaulkes:
Squashed and cleaned the up the commits. I'll cherry-pick this into staging and start on the bootstrap tests.

tegefaulkes:
Seems to have auto closed when I cherry picked the only commit. I think I can re-open after pushing a new commit.

tegefaulkes:
Seems like running `npm run test` requires nix shell to work properly.  I'll need to fix the `integration:docker` job.

tegefaulkes:
Oh... We may have a problem here. we're using dind to run the docker image inside the gitlab runner. So we don't have node, nix or NPM available to us here. So how do we run any jest tests in this context? I think the same problem will apply to the linux, macos and windows runners.

tegefaulkes:
An alternative is that we can code in an internal sanity self check debug bin command. Just to check if dependencies and basic functionality is working. This way we can sanity check the executables on their respective platforms without depending anything external. It could double as a debug test in the field as well.

CMCDragonkai:
The `integration` jobs needs to acquire the relevant dependencies in platform-specific way.

For example on windows, it has to do:

```yaml
  before_script:
    - choco install nodejs --version=16.14.2 -y
    - choco install python --version=3.9.12 -y
    - refreshenv
  script:
    - npm config set msvs_version 2019
    - npm install --ignore-scripts
    - $env:Path = "$(npm bin);" + $env:Path
    - npm run build --verbose
    - npm test -- --ci
```

I copied that from `build:windows`. Same for the mac runners.

For `dind`, that is an issue simply because the image supplied is `docker:20.10.11`, it only has `docker` available. You will need to change this image to an image that has `docker` AND also has all the available tools we need. I'm thinking you should go into the `docker` image official dockerfile, and see if all we need to do is to add `docker` into our own nix image, and maybe that will work, but maybe that won't be enough, if docker requires a lot of special things.

If `docker` requires a bunch of special things, we may have to either build on top of the `docker` image. See how to do this here: https://nixos.org/manual/nixpkgs/stable/#sec-pkgs-dockerTools.

Try using the `fromImage` to build on top of `docker:20.10.11`...

Later in the future we can acquire all the relevant details and directly embed it into our own nix container.

CMCDragonkai:
According to https://forum.gitlab.com/t/why-services-docker-dind-is-needed-while-already-having-image-docker/43534, it seems to say that

```yaml
  services:
    - docker:20.10.11-dind
```

Is what actually starts the docker daemon... which could mean we just use our own nix container and just push `docker` into our gitlab runner? Maybe that would be simple enough to do?

CMCDragonkai:
Refer to https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#use-docker-in-docker and https://github.com/docker-library/docker/blob/master/docker-entrypoint.sh#L29-L36. It appears the default docker container had some special entrypoint.

CMCDragonkai:
Remember to use `--rm` as well in your `docker run` command.

```sh
docker run -it --rm
```

That should remove the container once it stops running.

CMCDragonkai:
The `integration:*` jobs will require the tooling that `build:*` jobs currently have.

1. Ubuntu requires apt tooling to install nodejs v16 and relevant other tools - another option is to install nix on ubuntu (or use a custom ubuntu image that has nix on it, so that one can use `nix-shell`)
   - using nix should do a single user installation on ubuntu first, then using `nix-shell`... however this is problematic mainly because it doesn't have all the nix configuration that our nix images have for using our caching, so we can setup nix on ubuntu later
3. Windows requires choco - @emmacasolin is also potentially working on chocolatey caching in MatrixAI/Polykey#394
4. Mac requires homebrew - @emmacasolin is also potentially working on improving the homebrew caching in MatrixAI/Polykey#394

CMCDragonkai:
So right now @tegefaulkes you are investigating how to get our `gitlab-runner` image to have `docker` in its package set, and to also setup configuration so it can work similar to docker upstream. Beware of any incompatibility that might arise between `dind` image and our own gitlab runner image. Try to log your progress of your experiments here too.

CMCDragonkai:
@tegefaulkes try to abstract your changes to `pkStdio`, `pkExec`... a little bit. Those commands are becoming quite large. I suggest factoring out the 2 branches and calling them using `pkStdio` so the complexity of the functions are reduced, and we can have more precise behaviour for them.

I have a thought that since `pkStdio` was meant to be used with mocking, and mocking doesn't work for external executables, instead of modifying `pkStdio`, you should just switch those tests to using `pkExec` when the `PK_TEST_COMMAND` is set. This would also be test dependent.

Now since we are all working on the pipeline, to avoid confusion, please create a list above of all the tests you are changing and what you changing them to, so we can review and discuss too.

CMCDragonkai:
The most recent changes that were cherrypicked into staging has introduced test failures in `tests/bin`. Not sure if related again, but we can see `tests/bin` and `tests/nat` failing again. I might have revert all of those commits that were accidentally staged. I'm going to be disabling `integration:merge` to avoid these issues.

CMCDragonkai:
There's an option I wasn't aware of called `--init`: https://stackoverflow.com/questions/43122080/how-to-use-init-parameter-in-docker-run.

Can you see what happens to the docker process tree if it is run with `--init`.

CMCDragonkai:
The `deploy-service.sh` now creates the cluster and also registers a task definition.

While creating the cluster was idempotent, the registration of the task definition is not, it just adds a new task definition all the time. It's sort of a waste to have loads of old task definitions around, especially if nothing actually changed.

For now, as we generate new task definitions, once we perform the updated service, old task definitions can be automatically garbage collected (we may keep at least the last 10 just in case things change).

CMCDragonkai:
Ok I've created `cluster.yaml`, `service.yaml` and `task_definition.yaml`.

The logs are currently not working, and `aws ecs create-service` is not idempotent like `aws ecs create-cluster`.

It gives back:

> An error occurred (InvalidParameterException) when calling the CreateService operation: Creation of service was not idempotent.

These commands are all wildly inconsistent. There are commands that are idempotent like creating a cluster, commands that keep creating new entries like register task definition, and now creating a service is in fact not allowed if already created... such that you have to use update-service. Very bad... I guess this is why terraform and pulumi exist. And ultimately it seems these tools are intended for interactive usage not machine to machine usage (which would use a desired state concept instead).

CMCDragonkai:
The `ecs create-service` does however support the `clientToken` which allows us to make it similar to `create-cluster`. Now it behaves idempotently as we create a sha256sum of the templated file.

I'm beginning to think that these infrastructure setup should be done in stages. That is, the creation of this infrastructure should be staged AOT, and then on each commit, the only thing that needs to be done is actually just pushing up the image, and forcing a new deployment.

That would mean:

1. Cluster creation
2. Cloudwatch resource
3. VPC creation
4. Task definition registration
5. Service creation

Is all done ahead of time, and on the CI/CD we only do `update-service` to trigger a new deployment after image is uploaded.

CMCDragonkai:
Ok for creating the log group, the `awslogs-create-group` was needed to be enabled to be true. I believe the wizard automatically creates the log group in the beginning using whatever console permissions is available, but the `ecsTaskExecutionRole` actually needs this permission as well, and firstly this role isn't created by default, and isn't given the an additional policy of being able to create the log group.

To avoid issues in the future, the IAM user `matrix-ai-polykey` should probably create the log group in the first place.

CMCDragonkai:
Quick tip from MatrixAI/Polykey#396, you can add extra ignores to the `src` in `utils.nix` which will speed up image building because unnecessary updates won't trigger another update of the nix image.

Also make sure you have the `/root/.aws/credentials` setup to access our nix cache.

```nix
  # Filter source to only what's necessary for building
  src = nix-gitignore.gitignoreSource [
    # The `.git` itself should be ignored
    ".git"
    # Hidden files
    "/.*"
    # Nix files
    "/*.nix"
    # Benchmarks
    "/benches"
    # Docs
    "/docs"
    # Tests
    "/tests"
    "/jest.config.js"
  ] ./.;
```

tegefaulkes:
There seems to be a problem with the `gitlab-runner`. it's failing to connect to the `DIND` service now. I don't understand why since there hasn't been any changes to the docker/dind support in the `gitlab-runner` since I've last tested it. I'll have to do some testing in the `gitlab-runner` repo to find out why.

tegefaulkes:
Looks like the `entrypoint.sh` has changed slightly in the mean time. We're failing to connect to the docker daemon on port `2375` It *should* be connecting on `2376` for TLS. I think a bug has been introduced.

CMCDragonkai:
> Looks like the `entrypoint.sh` has changed slightly in the mean time. We're failing to connect to the docker daemon on port `2375` It _should_ be connecting on `2376` for TLS. I think a bug has been introduced.

I did make a minor change to the logic of acquiring the `DOCKER_HOST`. Not sure if it would produce a bug.

tegefaulkes:
Small logic mistake. When you removed `_should_tls` you changed a `|| to a `&&`

```
if _should_tls || [ -n "${DOCKER_TLS_VERIFY:-}" ]; then
// to
if [ -n "$DOCKER_TLS_CERTDIR" ] && \
     [ -n "$DOCKER_TLS_VERIFY" ]; then
```

I'm checking if changing that back fixes it.

Update: Yeah that was the problem. I'mm pushing a fix commit to `gitlab-runner` now.

tegefaulkes:
In order to streamline the testing on CI I need to do two things.
1. Add logging to the `pkXTarget` and `pkX` commands.
2. We need to skip the building steps in the CI since we're not testing them at this stage and they take the lion's share of the time. We need a way to download the pre-made executables. and use them instead so we can skip the build jobs.

The docker image is only 109M, I could just include it in the repo for the duration of testing.

Github doesn't like the file size but gitlab is handling it fine. I only need to push to gitlab to test for now.

emmacasolin:
> 2. We need to skip the building steps in the CI since we're not testing them at this stage and they take the lion's share of the time. We need a way to download the pre-made executables. and use them instead so we can skip the build jobs.

None of the `build` jobs that run tests actually save the built executable anymore - it's just done in the `build:dist` job. So you can disable the `build:platforms-generate` and `build:platforms` jobs to save time

tegefaulkes:
I've just checked manually with docker. js-polykey is properly capturing and handling the signals when I use `docker kill ID --signal=sigterm`. This is working for all of the registered handlers `SIGINT`, `SIGTERM`, `SIGQUIT` and`SIGHUP`. Using `ctrl + c` when running the docker image with `-it` works as well.

I think the only problems we've had with the signals is the special handling on windows platorm. I haven't checked mac but I assume it uses the same or similar system to linux.

I'll make a new issue specifying the problem for windows.

CMCDragonkai:
What about when `-it` isn't used. Does kill still work? The `-it` can change behaviour of programs.

On 14 July 2022 1:58:26 pm AEST, Brian Botha ***@***.***> wrote:
>I've just checked manually with docker. js-polykey is properly capturing and handling the signals when I use `docker kill ID --signal=sigterm`. This is working for all of the registered handlers `SIGINT`, `SIGTERM`, `SIGQUIT` and`SIGHUP`. Using `ctrl + c` when running the docker image with `-it` works as well.
>
>I think the only problems we've had with the signals is the special handling on windows platorm. I haven't checked mac but I assume it uses the same or similar system to linux.
>
>I'll make a new issue specifying the problem for windows.
>
>-- 
>https://github.com/MatrixAI/js-polykey/pull/391#issuecomment-1183958251
>
>Message ID: ***@***.***>
-- 

tegefaulkes:
Just did a quick test without `-it`. it works fine.

tegefaulkes:
```ts
runTestIf(
    process.platform === 'linux' || 
    global.testPlatform === 'DOCKER'
)
```

Is excessively verbose. with some small changes and a gobal variable `global.platform = testPlatform ?? process.platform` I could simplify it down to...
```ts
runTestIf(['linux', 'docker'])
```
To specify what platforms to run the test on. However there will be other conditions besides platform such as in the nat tests. So I could give it the signature `runTestIf(Array<string> | boolean)`?

CMCDragonkai:
No it's better to pass a callback. Don't special case it like that. I think right now it just takes a boolean, but you can make it also take a callback. But I wouldn't bother with such minor optimisation, explicitness here is a feature not a bug.

tegefaulkes:
I created a function that wraps the other one.
```ts
function runTestIfPlatforms(...platforms: Array<string>) {
  return runTestIf(platforms.includes(testPlatform));
}
``` 

If I need more flexibility then I can use `runTestIf`. This way I can be explicit with `runTestIfPlatforms('linux', 'docker')` but avoid all the noise.

CMCDragonkai:
There's 3 places where there are command execution usages:

* `tests/bin`
* `tests/nat`
* `tests/nodes/NodeConnection.test.ts` - this one was missed and wasn't using the `typecript-cached-transpile` and that is a reason why this test would have been very slow. It takes 6+ seconds to run full compilation

We need to move and centralise ALL "command" execution utilities into a directory:

```
/tests/utils/exec.ts
/tests/utils/utils.ts
/tests/utils/index.ts
```

Basically we have to make `tests/utils` a directory, and then move some of those files into relevant files under that directory. All the `pk*` commands should be in `tests/utils/exec.ts`, and we can centralise the configuration there.

tegefaulkes:
I'll create an issue for this to spec this out and track the changes.

tegefaulkes:
General estimates for time left
- MatrixAI/Polykey#404 - 1 day
- MatrixAI/Polykey#410 - done, needs review.
- MatrixAI/Polykey#407 - need to convert rest of tests, 1-2 days
- MatrixAI/Polykey-CLI#13 - Needs research and changes made 1-2 days
- MatrixAI/Polykey-CLI#11 - waiting on MatrixAI/Polykey-CLI#13, and MatrixAI/Polykey#410 test fixes, Otherwise most of the changes needed are made by MatrixAI/Polykey#407, just need 1 day to enable and confirm tests
- MatrixAI/Polykey-CLI#12 - waiting on MatrixAI/Polykey-CLI#14 mac test fixes, otherwise same as above, 1day
- MatrixAI/Polykey-CLI#14 - need to look at this to have a better idea. I'm guessing 2-3 days. depends on how much the failures overlap with each other.

CMCDragonkai:
Try rebasing on staging, as I've incorporated swc into ts-node.

CMCDragonkai:
With the `--root-key-file`, you should also check if encrypted, if so, decrypt with the required `--password`.

tegefaulkes:
After re-basing the test agent is failing to start with 
```
TypeError: Cannot read properties of undefined (reading 'isLocked')
    at KeyManager.getNodeId (/home/faulkes/matrixcode/polykey/js-polykey/node_modules/@matrixai/async-init/src/CreateDestroyStartStop.ts:233:28)
    at Function.createPolykeyAgent (/home/faulkes/matrixcode/polykey/js-polykey/src/PolykeyAgent.ts:214:36)
    at async /home/faulkes/matrixcode/polykey/js-polykey/src/bin/agent/CommandStart.ts:219:19
    at async CommandStart.<anonymous> (/home/faulkes/matrixcode/polykey/js-polykey/src/bin/CommandPolykey.ts:75:7)
    at async CommandPolykey.parseAsync (/home/faulkes/matrixcode/polykey/js-polykey/node_modules/commander/lib/command.js:923:5)
    at async main (/home/faulkes/matrixcode/polykey/js-polykey/src/bin/polykey.ts:54:5)
    at async /home/faulkes/matrixcode/polykey/js-polykey/src/bin/polykey.ts:103:5
```

It's referring to this snippet in `CreateDestroyStartStop`.

```ts
if (this[initLock].isLocked('write') || !this[_running]) {
     resetStackTrace(errorNotRunning, descriptor[kind]);
     throw errorNotRunning;
}
```

This is very odd. us `npm ls @matrixai/async-init` gives us the same version. so it's not a weird dependency issue.
```
[faulkes@faulkeswork:~/matrixcode/polykey/js-polykey]$ npm ls @matrixai/async-init
@matrixai/polykey@1.0.0 /home/faulkes/matrixcode/polykey/js-polykey
├── @matrixai/async-init@1.8.1
├─┬ @matrixai/db@4.0.5
│ └── @matrixai/async-init@1.8.1 deduped
├─┬ @matrixai/workers@1.3.3
│ └── @matrixai/async-init@1.8.1 deduped
└─┬ encryptedfs@3.5.3
  └── @matrixai/async-init@1.8.1 deduped
```

Was `pkSpawn` working? I don't know what's wrong here.

tegefaulkes:
It's a problem in staging too. the start.test.ts start in foreground test is failing for the same reason. But the code builds fine and `createPolykeyAgent` works fine too. So it might be a problem with the `pkSpawn` command?

I'll have to come back to this later.

tegefaulkes:
I've separated the issues for the mac and win side of things out of this PR. The issues have been updated to track what's blocked. What still needs to be done is all part of the MatrixAI/Polykey-CLI#10 Epic.

tegefaulkes:
This still needs to be re-based. However there is an issue with the staging branch were we're getting errors with the `async-init` lib. While I was exploring the problem I found that the `PolykeyAgent` runs fine but running polykey in a separate process for the bin tests would run into the error.

I'll re-base after completing the remaining issues. I can't do proper testing while this is still a problem.

CMCDragonkai:
Right now `swc` has problems running with decorators. See:

* https://github.com/MatrixAI/js-workers/pull/4#issuecomment-1192205602
* https://github.com/TypeStrong/ts-node/issues/1844.

We had to disable the `swc` usage in js-logger atm.

Upstream claims that it should be working, so we will need to have a look at it, and see if we can replicate the problem.

This is why there's bugs occurring with using `async-init`.

CMCDragonkai:
See https://github.com/TypeStrong/ts-node/issues/1844#issuecomment-1192694183, see if you can take is repo there and add in the usage of our `js-async-init` package, then run with `ts-node` using `swc`.

Then we can turn back on `swc` @emmacasolin https://github.com/MatrixAI/js-async-init/pull/18/files#r928370963. Note that `swc` should still be enabled wherever decorators are not used. So `js-async-locks` and `js-async-init` should be able to use `swc`, it's downstream repos that cannot...

tegefaulkes:
Todo for merge:
- [x] 1. finish up docker integration tests MatrixAI/Polykey#407
	- that means just get the ones to work working.
	- disable any tests using `createPolykey` or mocking 
- [x] 2. collate execution commands into test/utils/exec.ts. https://github.com/MatrixAI/Polykey/pull/391#issuecomment-1186712188
- [x] 3. re-base on staging.

src/bin/bootstrap/CommandBootstrap.ts
```diff
@@ -21,19 +22,23 @@ class CommandBootstrap extends CommandPolykey {
         options.recoveryCodeFile,
         this.fs,
       );
+      const privateKeyPem = await binProcessors.processRootKey(
```
CMCDragonkai, line 25: Test comment from code see.

jest.config.js
```diff
@@ -31,7 +31,8 @@ const globals = {
   // Timeouts rely on setTimeout which takes 32 bit numbers
   maxTimeout: Math.pow(2, 31) - 1,
   testCmd: process.env.PK_TEST_COMMAND,
-  testPlatform: process.env.PK_TEST_COMMAND_DOCKER,
+  testPlatform: process.env.PK_TEST_PLATFORM ?? process.platform,
+  tmpDir: process.env.PK_TEST_TMP_DIR ?? os.tmpdir(),
```
CMCDragonkai: The setting `tmpDir` here doesn't quite make sense to me.

Any code should be using `os.tmpdir()` should behave according to whatever `TMPDIR` is set to.

On Linux and containers, this is usually set to `/tmp`.

And all of our containers have `/tmp` built in.

Now if we want to "share" some data on the filesystem between container code and outside the container (whether that's a different container). Then the right way to do this is to:

1. Not use the `/tmp` for this.
2. Use a new path, make this `/srv`
3. Then mount a special path into `/srv`.

So for example when you are doing `docker run`.

You'd want to use:

```
docker run \
  --env PK_NODE_PATH='/srv' \
  --mount type=bind,src="$XXX",dst='/srv'
```

This is how we do this in the `testnet.polykey.io` task definition. Telling explicitly PK to use a particular node path, while mounting into the node path if we want to persist that data.

What is `$XXX`? This location is currently set to `PK_TEST_DATA_PATH`. This variable seems unnecessary. I think the `scripts/docker-run.sh` is intended to act as a "executable file proxy" so it can be used as a file target to execute. This variable should be passed in from the JavaScript code. It should not be hardcoded in `scripts/docker-run.sh`.

This means `tests/utils/exec.ts` (the location which should contain all the `pk*` functions, should decide where this location should be.

Now because you want to share the same location between the host (potentially a host container, and not host machine), and other containers. Then it should be sufficient to mount a temporary directory from `/tmp/...`.

CMCDragonkai: Here's a visualisation (the 3 boxes are each separate environments, they may all be containers, or the host may be a machine/VM):

```
                                    ┌─────────────────────┐
                                    │    PolykeyAgent     │
                                    │                     │
                              ┌─────┼─►/srv               │
                              │     │  PK_NODE_PATH=/srv  │
                              │     └─────────────────────┘
                              │
┌────────────────────────┐    │
│ Host                   │    │
│                        │    │     ┌─────────────────────┐
│                        │    │     │   PolykeyClient     │
│ /tmp                   │    │     │                     │
│ /tmp/polykey-test-XXX──┼────┴─────┼─►/srv               │
│                        │          │  PK_NODE_PATH=/srv  │
└────────────────────────┘          └─────────────────────┘
```

The end result is that we do not change anything about the TMPDIR or `os.tmpdir`, and this variable should not be required.

This means the tests should look very similar as before, and you can use `--node-path` instead of `PK_NODE_PATH` to be more explicit.

.gitlab-ci.yml
```diff
@@ -321,16 +321,22 @@ integration:docker:
...
-    - image="$(docker load --input ./builds/*docker* | cut -d' ' -f3)"
-    - docker run "$image"
+    - docker info
+    - >
+      nix-shell --run $'
+        image="$(docker load --input ./builds/*docker* | cut -d\' \' -f3)"
+        PK_TEST_DOCKER_IMAGE=$image \
+        PK_TEST_COMMAND=scripts/docker-run.sh \
+        PK_TEST_PLATFORM=docker \
+        PK_TEST_TMP_DIR=/builds/$CI_PROJECT_PATH/tmp \
```
CMCDragonkai: So the directory here has to exist under `/builds`.

Any directory that is mounted into the Polykey containers subsequently have to be under this directory.

So the issue is that normally our test code would use `os.tmpdir()` to create the data directories.

But now this has to be changed to use this location instead.

To resolve this, the tests may require an override on the data directory creation, rather than creating from `os.tmpdir()`. But don't override `TMPDIR` env variable, that should be kept as is.

CMCDragonkai: Therefore `PK_TEST_TMP_DIR` can actually be `PK_TEST_DATA_DIR`. Which would then override `os.tmpdir()`.

jest.config.js
CMCDragonkai: If you already have a `pk*Target` let's just special case docker then.

Then you can specify the `--mount` option directly and also keep the paths the same.

I don't actually see a huge issue with using `/srv` as the directory node path inside the containers. But it can be set to the same path outside.

Therefore on CI/CD, we would see something like `/build/$CI_PROJECT_PATH/tmp/polykey-test-XXX` being used both inside the `PolykeyAgent` and `PolykeyClient` containers and as well as the same build/host container.

CMCDragonkai:
```
                                    ┌─────────────────────────┐
                                    │    PolykeyAgent         │
                                    │                         │
                              ┌─────┼─►/tmp/polykey-test-XXX  │
                              │     │  PK_NODE_PATH=""        │
                              │     └─────────────────────────┘
                              │
                              │
┌────────────────────────┐    │
│ Host                   │    │     ┌─────────────────────────┐
│                        │    │     │   PolykeyClient         │
│ /tmp                   │    │     │                         │
│ /tmp/polykey-test-XXX──┼────┴─────┼─►/tmp/polykey-test-XXX  │
│                        │          │  PK_NODE_PATH=""        │
└────────────────────────┘          └─────────────────────────┘
```

The `""` means the same path.

CMCDragonkai: Alright so if `scripts/docker-run.sh` can be deleted and the same stuff done inside the `tests/utils/exec.sh`, we remove a couple environment variables.

Note however there is still these global/env variable:

1. `PK_TEST_PLATFORM` - defaults to `os.platform()` can be overridden to `'docker'`.
2. `PK_TEST_COMMAND` - target file to execute (it's supposed to be a shell command actually), in the case of `docker` execution, this could be set to the path to `docker` itself...
3. `PK_TEST_DATA_PATH` - the data path to be used by the tests, right now all the tests are using `os.tmpdir()`, it should default to using `PK_TEST_DATA_PATH`, which itself defaults to `os.tmpdir()`, on the CI/CD this is set to under the `/build/$CI_PROJECT_PATH` because only this location shareable between the the dind containers

CMCDragonkai: Try and see what happens if you were to allow `shell` execution for `PK_TEST_COMMAND`. I believe we normally use `execFile`, but it can be changed to `exec`. Some things may not work. This will end up allowing arbitrary shell commands.

CMCDragonkai: Ok `PK_TEST_COMMAND` is ignored when `PK_TEST_PLATFORM='docker'`.

.gitlab-ci.yml
```diff
@@ -321,16 +321,22 @@ integration:docker:
...
-    - image="$(docker load --input ./builds/*docker* | cut -d' ' -f3)"
-    - docker run "$image"
+    - docker info
+    - >
+      nix-shell --run $'
+        image="$(docker load --input ./builds/*docker* | cut -d\' \' -f3)"
+        PK_TEST_DOCKER_IMAGE=$image \
+        PK_TEST_COMMAND=docker \
+        PK_TEST_PLATFORM=docker \
+        PK_TEST_DATA_DIR=/builds/$CI_PROJECT_PATH/tmp \
```
CMCDragonkai: I don't think the additional env variable for `PK_TEST_DOCKER_IMAGE` as it adds more configuration.

I propose something like this:

```
PK_TEST_PLATFORM='docker'
PK_TEST_COMMAND="$(which docker) run \${options} someimage:sometag
```

The `someimage:sometag` is specified outside.

Now you need to use a simple utility function to "template" the options and args.

```
const template = (str, vars) => str.replace(
  /\${(.*?)}/g, 
  (_, p) => vars[p]
);

template(pkTestCommand, { options: '--x y z' })
```

CMCDragonkai: That way the `PK_TEST_COMMAND` remains an actual command, it's just that options can be provided in an interpolated fashion.

And this allows future applications of `PK_TEST_COMMAND` to the other platforms in case there are special options required too.

Now where should the options be "templated"? It should be done in your `tests/utils/exec.ts`, you can create a common function to do it. The `template` can be a useful utility function put into `tests/utils/utils.ts`. Might even be useful elsewhere too.

CMCDragonkai: You would need to make sure you're using `exec` and not `execFileSync`.

tegefaulkes: I don't think that's going to work. when using `child_process.spawn` it takes (command, args, options). So the command will need to be `docker` on it's own, and the args need to be `['run', ...options, 'someimage:sometag', ...pkargs]`.

If we really need to get rid of the extra env variable, I can get the image id inside the test by execing `docker image ls -f "reference=polykey*" -q`

CMCDragonkai: Doesn't the spawn take an arbitrary string as the first parameter? You should be able to pass a full command into `spawn`. You just need to enable `shell: true`.

CMCDragonkai: spawn

CMCDragonkai: So it should work like `child_process.spawn('ls -l | cat', [], { shell: true });`

CMCDragonkai: BTW I just noticed `killSignal` is set to `SIGTERM`, that doesn't work on windows right? Seems strange that it would default to signal that has no effect on windows. Or maybe the docs is not up to date.

tegefaulkes:
This has been re-based on `feature-swc_fix`. Which should be merged first at MatrixAI/Polykey#430.

The only thing left here is to;

1. collate execution commands into test/utils/exec.ts. https://github.com/MatrixAI/Polykey/pull/391#issuecomment-1186712188
2. update nat tests to use runTestIf, runDescribeIf changes. 
3. `PK_TEST_DOCKER_IMAGE` usage still needs to be removed. I think the simplest solution is to just get the image ID using `docker image ls -q` we can use it with the `--filter` option but there should only be 1 image anyway.

From there it's general commit cleaning and merge prep.

.gitlab-ci.yml
tegefaulkes: `PK_TEST_TMP_DIR` is now `PK_TEST_DATA_DIR`.

jest.config.js
tegefaulkes: These changes have been made. Mainly the `docker-run.sh` has been removed and inlined within the `pkXTarget` functions.

CMCDragonkai:
We cannot rely on that. There may be multiple images and tags that are different. It just needs to be set explicitly via `PK_TEST_COMMAND`.

CMCDragonkai:
Can you start squashing.

tegefaulkes:
Commits have been cleaned up.

.gitlab-ci.yml
CMCDragonkai: Ok so some new ideas since we discussed:

1. Try using `execFile` and `spawn` with `shell: true` option, and see if it's possible to pass an arbitrary command as the first argument. And what the second argument of `args` is used for, possibly just concatenated arguments after the string?
2. This allows us to use `PK_TEST_COMMAND='ls -l | docker run'` or any arbitrary shell command string. Do note that on windows, this will be a `CMD` shell and not a bash/sh shell.
3. The `PK_TEST_COMMAND` originally was just a path to a file, now its a shell command. If it's a shell command, it should support the ability to interpolate a shell variable like: `PK_TEST_COMMAND='docker run $DOCKER_OPTIONS image:tag'`. Where you can then interpolate `$DOCKER_OPTIONS` during `execFile` or `spawn` by using the env option such as `{ env: { ...process.env, DOCKER_OPTIONS: '--a --b --c' } }`.
4. The `PK_TEST_PLATFORM` and `PK_TEST_COMMAND` should be coupled together as all or nothing. Meaning they can only either be both set, or both unset (`undefined`). This check should be done on the `jest.config.js`. This ensures that the user has to set both, the platform for how it is run, and the command for what is run. This means that `PK_TEST_PLATFORM` should not be defaulted to `os.platform()` instead it defaults to undefined.
5. The end result is that target based tests will end up with an intermediate shell program, where as regular testing without the 2 env variables will use the original way of executing PK. We have to be wary of any kill signals now, since for docker there could be 2 intermediate processes the shell and the docker's shell (if using a shell), but if doing a direct execution, then it should be just the shell itself.
6. As for the `PK_TEST_DATA_DIRECTORY`, that should be renamed to `PK_TEST_TMPDIR` so it's obvious that it's overwriting the `os.tmpdir()` being used bby the tests. But it does not overwrite `TMPDIR` normally, as we don't want to change the meaning of the `TMPDIR`.

tegefaulkes: Doing some testing with `child_process.spawn`. Looks like 1, 2, 3 above work as expected and should be doable. I'll make the changes and do some more testing.

tegefaulkes: Just ran all of the tests locally. It works with 1 minor difference breaking a single notification test. When running with `shell: false` a sing arg in the args array can contain spaces and be considered a single arg. In the problem test we have the following args;

```ts
        [
          'notifications',
          'send',
          nodesUtils.encodeNodeId(receiverId),
          'test message 3',
        ],
```

With `shell: false` the arg `test message 3` is considered a single arg. as if we passed it to the command like `"test message 3"`. However  with `shell: true` each word is treated as a seperate arg. so the command things the message is just `test`.

This can be fixed by setting the message to `'"test message 3"'` but this will break the test when it runs normally. Since it's a pretty niche problem affecting only 1 test I'll just replace the spaces in the message with `_`.

Just keep this difference in behaviour in mind for the future.

tegefaulkes: tracking what's been done here,
- [X] 1.
- [X] 2.
- [X] 3.
- [x] 4.
- [x] 5.
- [X] 6.

tegefaulkes: For 4, we have 1 case where we can't set both. That's for normal testing in the CI. This means we'd either have to;
1. allow PK_TEST_PLATFORM to be set to `linux` with  `PK_TEST_COMMAND` being undefined.
2. Have `runTestifPlatform` run by default if `PK_TEST_PLATFORM` is not set.

I'm leaning towards 2. here but I may need to remove `'linux'` from every `runTestIfPlatform` usage.

CMCDragonkai:
Regarding the usage of `shell: true`. There are some differences now in the command executions.

There are 3 functions for executing things:

* `child_process.exec` - defaults to `shell: true`
* `child_process.execFile` - defaults to `shell: false`
* `child_process.spawn` - defaults to `shell: false`

The `exec` is never used. Because it's the most basic one, but we do not have this used anywhere in our tests.

Now the problem is that `shell: true` affects the interpretation of spaced arguments. By default `shell: false`, means that `['a b']` is one single atomic argument value. So `child_process.spawnSync('touch', ['a b'], { shell: true });` means `touch a b` and not `touch 'a b'`. This means you end up with 2 files being created rather than 1 file `a b` being created.

Furthermore, when using `child_process.execFile`, this actually interprets the arguments incorrectly. It expects the first argument to be a file path. For example: `child_process.execFileSync('touch', ['a b'], { shell: true });`, this fails as it tries to execute `/bin/sh -c touch a b`, and it doesn't properly quote the command as `/bin/sh -c 'touch a b'`.

Therefore, if we are switching to using `shell: true`, we cannot actually use `execFile` or `execFileSync`. They won't be portable.

Instead any usage of `shell: true` must be using `spawn`.

We are using `execFile` in a few places, it's possible to replace these usages with `spawn` instead, since `spawn` can do everything that `execFile` does, and we end up **only** using `spawn`. This means `pkExec` should be swapped to using `childProcess.spawn` in order to maintain consistency so that when the `PK_TEST_COMMAND` is being used, it will continue to using `spawn`.

However as for the `'a b'` interpretation, when the `shell: true`, (and this occurs when `PK_TEST_COMMAND` is being used), then any spaced arguments must be either quoted around the argument value OR it has be escaped.

The simplest solution is to escape every space character when `shell: true`, so you maintain consistent behaviour between `shell: false` and `shell: true`. That way `['a b']` can contineu be used as a normal argument and test writers don't have to change their behaviour. Inside your testing utilities targetting `PK_TEST_COMMAND`, you have to map every argument to an escaped version of the argument using a regex replace ` ` with `\ `.

Something like:

```ts
arg.replace(/ /g, '\\ ')
```

CMCDragonkai:
Note that there could still be issues in the future, it's best to make sure that our test arguments doesn't use too many fancy stuff in the future to avoid problems.

.gitlab-ci.yml
CMCDragonkai: What case is that? I was thinking that `PK_TEST_PLATFORM` is only useful during targeted execution. Otherwise tests should work normally across all platforms.

tegefaulkes: Resolving this. It has been addressed.

CMCDragonkai:
This might be a better escaping function:

```
arg.replace(/(["\s'$`\\])/g,'\\$1');
```

It also escapes a few other special variables that may be interpreted specially when `shell: true`.

I just tested it with:

```
touch \"\ \$\`\\\'
```

It works quite nicely to make sure it's all interpreted as a raw string.

tegefaulkes:
I'm consolidating all of the child_process spawning commands inside of `tests/utils/exec`. Nat test utils contain 2 such commands `pkExecNs` and `pkSpawnNs`. I can move them over but they depend on a function that's used extensively in `nat/utils.ts` called `nsenter()`. 

I don't want to split this function between the two files and I'm not sure if it's a good idea to have the two files depend on each other. I could;
1. Move `nsenter()` to the `exec.ts` and have `nat/utils` import it. this would be the easiest but `nsenter` would be out of place.
2. `nsenter()` just creates some extra args, we could just pass them into `pkXNs`. and make a wrapper `pkXNs` inside `nat/utils.ts` to recreate the original signature.

Not sure what to do here, I'm leaning towards 2.

@CMCDragonkai

CMCDragonkai:
We're moving `nsenter` into `tests/utils/exec.ts`.

Future enhancement requires the `tests/nat` to also use the `PK_TEST_PLATFORM` and `PK_TEST_COMMAND` when both are set. That would be the case where nat tests can run against a linux binary, and this would most likely occur on `integration:linux` job, which runs with an ubuntu container.

tegefaulkes:
Ok, this should be good to merge now. all the tests should be passing. I haven't enabled the integration tests to check them for a little bit but they are passing locally so they should be passing there as well.

PR #406, opened 2022-07-12
feat: adding `workerManager` parameter to `createPolykeyAgent` and `createKeyManager`

### LINKED ISSUE #399: WorkerManager is being reported to be created **after** the agent is already started and reported its status
### Describe the bug

In the deployed testnet, I can see that `WorkerManager: Created WorkerManager` is logged after the STDOUT of the agent. I'm not sure if this is a bug or if its just due to cloud watch reporting the 2 messages at the same time.

```
2022-07-04T07:13:37.048000+00:00 ecs/polykey-testnet/e6c450d3cf2443f785eaa0d98d122671 INFO:WorkerManager:Creating WorkerManager
2022-07-04T07:13:37.137000+00:00 ecs/polykey-testnet/e6c450d3cf2443f785eaa0d98d122671 {"pid":1,"nodeId":"v7osqo4g9uca06qb48ra7pbpergp59qtdh9584a0j0a4ud80qgaig","clientHost":"0.0.0.0","clientPort":1315,"agentHost":"127.0.0.1","agentPort":36631,"proxyHost":"0.0.0.0","proxyPort":1314,"forwardHost":"127.0.0.1","forwardPort":42899,"recoveryCode":"adjust broom select dish major portion arm horn kit thrive onion coil margin brother arctic skin equip soup mango club thought like guard sister"}
2022-07-04T07:13:37.137000+00:00 ecs/polykey-testnet/e6c450d3cf2443f785eaa0d98d122671 INFO:WorkerManager:Created WorkerManager
```

Do note that the 2 message timestamps are exactly the same, even the milliseconds `2022-07-04T07:13:37.137000+00:00`. It could just due to the fact that the task is set to `mode: non-blocking` logging which causes the message to arrive out of order.

Part of the problem is that STDOUT and STDERR are interleaved in cloudwatch, so maybe that's the issue too.

Check the code too as we should definitely wait for worker manager to be fully created before we print out the agent status.

### To Reproduce

1. Run `pk agent start`
2. Observe `WorkerManager: Created WorkerManager`

### Expected behavior

1. Message should arrive and be printed out before the final STDOUT.

### Additional context

* #396 - discovered while deploying testnet

### Notify maintainers

@CMCDragonkai

tegefaulkes:
### Description

#399 was not actually a problem, just a race condition in the logs since the two outputs were on `stdout` and `stderr` at exactly the same time. I checked and there was no issue. However I did discover that the `WorkerManager` was only set after starting up polykey. So anything such as `KeyManager` that used it during startup never had it. I've made some modifications to fix this. `workerManager` can be provided to `createPolykeyAgent` and propagated to the relevant domains. Ultimately the `WorkerManager` is started before `PolykeyAgent` now when doing `polykey agent start`. This should fix the order of the logs too.

### Issues Fixed

* Fixes #399

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -97,6 +97,7 @@ class PolykeyAgent {
     sessionManager,
     grpcServerClient,
     grpcServerAgent,
+    workerManager,
     fs = require('fs'),
     logger = new Logger(this.name),
     fresh = false,
@@ -143,6 +144,7 @@ class PolykeyAgent {
     sessionManager?: SessionManager;
     grpcServerClient?: GRPCServer;
     grpcServerAgent?: GRPCServer;
+    workerManager?: PolykeyWorkerManagerInterface;
     fs?: FileSystem;
     logger?: Logger;
     fresh?: boolean;
@@ -204,6 +206,7 @@ class PolykeyAgent {
           keysPath,
           password,
           fs,
+          workerManager,
           changeCallback: (data) =>
             events.emitAsync(PolykeyAgent.eventSymbols.KeyManager, data),
           logger: logger.getChild(KeyManager.name),
@@ -232,6 +235,7 @@ class PolykeyAgent {
           logger: logger.getChild(DB.name),
           fresh,
         }));
+      if (workerManager != null) db.setWorkerManager(workerManager);
       identitiesManager =
         identitiesManager ??
         (await IdentitiesManager.createIdentitiesManager({
@@ -348,6 +352,7 @@ class PolykeyAgent {
           logger: logger.getChild(VaultManager.name),
           fresh,
         }));
+      if (workerManager != null) vaultManager.setWorkerManager(workerManager);
       sessionManager =
         sessionManager ??
         (await SessionManager.createSessionManager({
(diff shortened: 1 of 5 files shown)


CMCDragonkai:
We will need to consider now with the speed up, whether it is worth it to send it to the workers for encryption tasks. The latency spent can be alot more than the time required to just compute the data. To do this, we will need to put rules of thumb thresholds into the code as constants. That is if we the encrypted data is `> X bytes`, then we send it to the worker.

However that's not really entirely consistent due to platform differences. So it may be better to do a sort of JIT. Where we measure the time to compute and the time to sent to worker and back, and then compare, and adjust this threshold accordingly.

We can test latency pretty easily regularly and we just need to compute things on the main thread or even in the worker thread and subtract the time...

CMCDragonkai:
This is superseded by the work in #446. In there, the worker managers are still set dynamically but only used for certain things.

Furthermore, they can set as part of the creation routines, so that bootstrapping can make use of it.

This means, simply by changing up our order of bootstrap, this will be resolved. So I'm going to close this PR.

PR #411, opened 2022-07-12
ci: merge staging to master

maxwell-aiden:
This is an automatic PR generated by the pipeline CI/CD. This will be automatically fast-forward merged if successful.


maxwell-aiden:
Pipeline Attempt on 585509433 for ceaebe657231e327f60a6395d33e76b8e089720a

https://gitlab.com/MatrixAI/open-source/js-polykey/-/pipelines/585509433

CMCDragonkai:
@emmacasolin if any intermittent test failures occur here, it's best to be solved by solving the underlying problem, not just making fixes here in this branch.

That will be all the new issues that were created and some of the PRs that @tegefaulkes created.

* MatrixAI/Polykey#398
* MatrixAI/Polykey#399 
* MatrixAI/Polykey#391
* MatrixAI/Polykey-CLI#14 
* MatrixAI/Polykey#407
* MatrixAI/Polykey-CLI#11 
* MatrixAI/Polykey-CLI#12

You can also help @tegefaulkes flesh out those issues too.

maxwell-aiden:
Pipeline Attempt on 586532953 for 0894eabe97ad4406649a0e4f5a0ba26978236005

https://gitlab.com/MatrixAI/open-source/js-polykey/-/pipelines/586532953

maxwell-aiden:
Pipeline Attempt on 586804088 for 2b79b48b053f9b5af7fcdada9629588f97e23919

https://gitlab.com/MatrixAI/open-source/js-polykey/-/pipelines/586804088

maxwell-aiden:
Pipeline Attempt on 588459726 for 37f03d9137259a682fb2b4644b6e2de52c0b6f53

https://gitlab.com/MatrixAI/open-source/js-polykey/-/pipelines/588459726

maxwell-aiden:
Pipeline Attempt on 590199895 for b9d7f383658366e50415238dc2a7ff0aaa8de37c

https://gitlab.com/MatrixAI/open-source/js-polykey/-/pipelines/590199895

maxwell-aiden:
Pipeline Attempt on 595123941 for 3dce49302371a8b62f754e3ae05410750b5ede1c

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/595123941

maxwell-aiden:
Pipeline Attempt on 596835327 for ce07c78db52ffb8ef144f4303172816dbb3a925d

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/596835327

maxwell-aiden:
Pipeline Attempt on 598042628 for 4117a049eee956b78ef2ec5ccd1bae9e2b22bb7d

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/598042628

maxwell-aiden:
Pipeline Attempt on 598744741 for 21aca8ab17618ff8d50c07b2f194aab7b50346b9

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/598744741

CMCDragonkai:
**Because we are deprioritising Mac and Windows integration tests for now. Then next commit here should disable the MacOS and Windows tests for both build stage and integration stage.**

At the same time, I want to refactor the conditional test utilities to use boolean conditional parameter again, but instead reify the conditions into things like `isWindows`, `isLinux`... etc, then you can combine them up easily with `isWindows || isLinux`... etc. String parameters are not obvious.

Furthermore, it's apparent now that we have situations where a test should run normally, and also under a particular platform, but no other platforms are allowed. And there are tests that we want to run only for a given platform. This behaviour should be standardised.

A standard set of functions should be used:

```
testIf
describeIf
```

The condition should be `testIf(boolean, ...args)` where `...args` is sent to `test`. This makes it easily extendable.

An alternative is `testIf(() => bool, ...args)` where it's a function. A unary function by itself is sort of useless, but it can be useful if we do `testIf((meta: Metadata) => bool, ...args)`, where metadata is actually passed into the predicate which is derived from the usage of the test. However I cannot think of any metadata right now that would be useful. Maybe the test name, test location, `__dirname`... etc.

A predicate can also be asynchronous, which allows it to do runtime side-effects (mostly queries).

However @tegefaulkes says it's annoying to have to write `os.platform() === 'linux'` all the time, so to avoid this, we can "reify" the booleans. This is more obvious and is better for performance as it eliminates a bunch of checks. Basically we can have `isPlatformLinux` and `hasNsenter`... etc. This is already used in `js-logger` under `hasCaptureStackTrace` and `hasStackTraceLimit`. See: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig and https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables.

Finally this means that through anyone's interpretation:

```
testIf(isPlatformLinux, ...)
```

Should **only** run if the platform is Linux.

Therefore if we want to run without a platform requirement, then:

```
testIf(isPlatformEmpty || isPlatformLinux, ...args)
```

Which means that you can run under any platform and under platform linux.

Finally for and conditions:

```
testIf(hasNsenter && isPlatformLinux, ...args)
```

Which allows us to wire up conditions.

All of this is lot easier to understand, and isn't that much more to write than what is done already.

CMCDragonkai:
A few extra things:

1. Update js-encryptedfs 3.5.5 when it is released, it has the new logger as well
2. This should mean any EFS `// @ts-ignore` shouldn't be needed
3. When MatrixAI/Polykey#419 is merged all `//@ts-ignore` should be removed from PK (only EFS will have `// @ts-ignore`

CMCDragonkai:
@tegefaulkes I changed the `integration:docker` job to do use gitlab variables:

```
  variables:
    DOCKER_TLS_CERTDIR: "/certs"
    FF_NETWORK_PER_BUILD: "true"
    PK_TEST_PLATFORM: "docker"
    PK_TEST_TMPDIR: "${CI_PROJECT_DIR}/tmp/test"
  script:
    - docker info
    - >
      nix-shell --arg ci true --run $'
      image_and_tag="$(docker load --input ./builds/*docker* | cut -d\' \' -f3)";
      PK_TEST_COMMAND="docker run \$DOCKER_OPTIONS $image_and_tag /bin/polykey" npm run test -- tests/bin;
      '
```

Using `npm run test -- tests/bin` as trigger for integration testing docker is interesting...

But I can't think of a good way otherwise right now. Mainly because if `npm run test` by itself is sufficient to do integration testing, then all tests must be using the `testIf` and `describeIf` conditionals and that doesn't seem right.

I think for the integration tests in particular... we may need to in fact group them, in this case it is grouped by directory `/tests/bin`, but there's also the group annotation `jest-runner-groups` if directory separation isn't sufficient but I've shied away from using that due to poor documentation.

This comes down to MatrixAI/Polykey#434 where we now have a situation where conditions mean "do run this only if X, Y Z is true", but then there are tests without any conditions whatsoever that always run. That allows unit tests to be structured with a base layer of common tests, with exceptional platform tests.

Integration testing seems different though, we don't want re-run the common tests, but now we want to run a specific set of tests. A directory is a better of indicating separation then any kind of comment annotation.

We should explore ways of exporting tests or describe conditions similar to how we export the `main` function. This way we could import tests from another file and have this as a way of grouping relevant tests. Integration tests can also have their own test conditionals, but in a similar fashion of having a base set of integration tests that always run, and then some tests with exceptional conditions.

So atm, `tests/bin` seems like a catch all, but it could get messy in the future. So we should consider `tests/integration/docker/`, `tests/integration/windows/`... etc, an these can then bring in relevant tests through `import` calls.

Perhaps jest can work on modules as is, if we just do:

```
// This becomes a declarative re-usable specification for tests, rather than the imperative `describe` and `test` functions
const tests = {
  'test group': [{
    'test function': [() => {}, ...options]
  }, ...options]
};

if (require.main === module) {
  void executeTests(tests);
}

export default tests;
```

Alternatively we could just rewrite the test functions themselves as re-usable functions.

The idea being, you could then compose a test structure from other tests:

```
import binTests from '../bin`;

const tests = {
  ...binTests.X
};
```

This will require some experimentation in the future.

Note that it's always possible to just import/require another test file, and that will just run those tests as part of the new file, but that is not fine-grained.

CMCDragonkai:
Windows and MacOS tests have been disabled in staging now so we can focus on getting `testnet.polykey.io` fully tested @emmacasolin

maxwell-aiden:
Pipeline Attempt on 601108657 for 450d65689d9fc05deafe6d301665a2f449778750

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/601108657

CMCDragonkai:
This is the list of all the `// @ts-ignore` that should be removed when MatrixAI/Polykey#419 is merged:

```sh
ag '// @ts-ignore - version of' | less
```

CMCDragonkai:
@emmacasolin the NAT tests taking too long https://gitlab.com/MatrixAI/open-source/Polykey/-/jobs/2793709838

maxwell-aiden:
Pipeline Attempt on 601122123 for 5105ec95a8dad1244bc655e8a801749fb60426e5

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/601122123

CMCDragonkai:
It is possible to to have an exported module:

```ts
if (process.env.JEST_WORKER_ID) {
  describe(Queue.name, () => {
    const logger = new Logger(`${Queue.name} test`, LogLevel.WARN, [
      new StreamHandler(),
    ]);
    test('do it', async () => {
      console.log('wee');
    });
  });
}

export default 3;
```

One issue, is that `process.env.JEST_WORKER_ID` is true even if you only imported the file, there would need to be a way to check if the test is directly executed or only imported. Then one can then only execute the tests if it is one of the found test files. However there is no variable I can find for this atm.

CMCDragonkai:
It appears `require.main` is `null` in the scripts directly executed, and it is not `null` when not directly executed. So we could check if `require.main` is null or not.

```
if (require.main == null) {
  describe(...);
}

export default tests;
```

CMCDragonkai:
MatrixAI/Polykey#435 is created to address reusing unit tests for integration testing.

maxwell-aiden:
Pipeline Attempt on 605881353 for 3c206c5892cf02398dca15e04d081acdc0ebd44c

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/605881353

tegefaulkes:
Just a note, `tests/agent/service/nodesCrossSignClaim.test.ts:52` still has a polykey agent doing key generation. This should be updated to used the method from MatrixAI/Polykey#420

maxwell-aiden:
Pipeline Attempt on 607021759 for 94ec1d2a6971f66b2140fded3543656ef385f2a4

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/607021759

CMCDragonkai:
@emmacasolin needs to rebase on top to deal with the changes to the `globalThis` and `window`.

maxwell-aiden:
Pipeline Attempt on 609109647 for 976f9135ba1bdb7f4f02c3a6b13170a2f466d134

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/609109647

maxwell-aiden:
Pipeline Attempt on 609126080 for e425469b561a307b5cd28b53c7b07e0cc86b2707

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/609126080

maxwell-aiden:
Pipeline Attempt on 609960309 for e80624e153da5235a9c968b73fd0378a0a06585a

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/609960309

maxwell-aiden:
Pipeline Attempt on 609987169 for 7b1edbd6ec77c1a5705563de7a940f7441d1ac53

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/609987169

maxwell-aiden:
Pipeline Attempt on 609996933 for 27df70f351c45d73193eccee2e1cb448977074f3

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/609996933

emmacasolin:
Almost all of the docker tests are failing on the pipeline. It looks like any usage of `globalThis.tmpDir` throws an `ENOENT` error, but only when run on docker. The few tests that are passing on docker all use `globalThis.testDir` or `os.tmpdir()` instead.

tests/utils/exec.ts
```diff
@@ -0,0 +1,600 @@
...
+  shell?: boolean;
+};
+
+const tsConfigPath = path.resolve(
+  path.join(globalThis.projectDir ?? '', 'tsconfig.json'),
+);
+
+const polykeyPath = path.resolve(
+  path.join(globalThis.projectDir ?? '', 'src/bin/polykey.ts'),
+);
```
tegefaulkes, line 27: `globalThis.projectDir ?? ''` had to be added to prevent an error for `globalThis.projectDir` being undefined. Likely the same needs to be done for `globalThis.testDir` somewhere.

Details here https://github.com/MatrixAI/Polykey/pull/436#issuecomment-1210251745

CMCDragonkai: Why would this be affected by https://github.com/MatrixAI/Polykey/pull/411#issuecomment-1212635793?

This problem was due to `ts-node` running an internal test script that did not have the test environment.

The docker tests I feel like has too much required environment variables, so we want to remove them as much as  possible. At least the `PK_TEST_PLATFORM` should disappear with #435.

tegefaulkes: The problem was that `globalThis` variables are not defined when running as a sub process. That means it won't be defined when running as docker as well. If the docker image is trying to use `globalThis.testDir` somewhere then we will run into an error `globalThis.testDir` is undefined. As far as the type system knows  the `globalThis` variables are always defined so it's not caught when building.

CMCDragonkai: What docker process would be using `globalThis`? It would only be possible if you were running a test script under `tests` dir that imports the `tests/utils`. The docker tests are not running test scripts are they? They should be running `src/bin/polykey.ts` which wouldn't be using `globalThis.testDir`...

tegefaulkes: It shouldn't be using `globalThis`. But if this problem is anything like the previous problem where just importing `tests/utils/exec.ts` was causing `global.this` variables to be used then we would get this error. I'll have to dig into why this is happening but it's too similar to the previous problem to not check for that first.

maxwell-aiden:
Pipeline Attempt on 620069468 for 5d90a03b57d92d1e86a2bbed79e022f487551c12

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/620069468

CMCDragonkai:
Disable windows and mac for now. We are only going to focus on Linux and Docker CI/CD until we have released.

maxwell-aiden:
Pipeline Attempt on 638238451 for a7225100291d6b8c2bc97d86bb75ecbe16678113

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/638238451

maxwell-aiden:
Pipeline Attempt on 643351200 for 33b4660492bf81c4d3835d0b819ab801f513dc5c

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/643351200

maxwell-aiden:
Pipeline Attempt on 644999693 for 54f0c2b671229361e0f7b75c21a1547655cf7ae4

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/644999693

CMCDragonkai:
1.1.5 of js-errors should be brought in (skipped 1.1.4 due to https://github.com/MatrixAI/js-errors/pull/8#issuecomment-1252154621).

CMCDragonkai:
@tegefaulkes 

- [x] 1. Update js-errors to 1.1.5
- [x] 2. Disable windows and macos CI builds
- [x] 3. Fix docker integration tests
- [x] 4. Disable discovery and `tests/nat` if these are not passing atm
- [x] 5. Get the build pipeline working so we can get a usable binary before we start on new features
- [x] 6. Create new issue for `timedCancellable` integration - commandeered MatrixAI/Polykey#450 for this issue
- [ ] 7. Discovery and `timedCancellable` PR MatrixAI/Polykey#451
- [ ] 8. Testnet integration tests MatrixAI/Polykey#441 
- [ ] 9. Try using https://swc.rs/docs/usage/jest to see if improves the test speed and doesn't introduce any issues, we are already using swc for `ts-node` and it seems to be working fine for the code right now, so I'm guessing it should work for jest too. - this is being tested in `feature-swc-jest`, @tegefaulkes please merge into staging if it passes.

maxwell-aiden:
Pipeline Attempt on 646414230 for 586750e98e8104e18430b39cd7a3e36e8ff4291e

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/646414230

maxwell-aiden:
Pipeline Attempt on 647558180 for cc7ee37ea8157870630dde0501a3a53c29db2326

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/647558180

tegefaulkes:
For task 2, I think Win and Mac integration jobs are already disabled? The change was done in commit 626f03e911c594dfda7e82522b704bf4c5d93ebe

CMCDragonkai:
Look into disabling the build tests for mac/windows.

tegefaulkes:
I've doubled the timeouts for the nat tests. I also disabled one of them due to it throwing an error.

That leave the nodes domain. Seems like all of the errors we're seeing is due to the connection errors. For now I'm just going to increase the default timeout for Ping. That seems to be causing the problem but itself is not the source of the bug. 

Right now I'm taking the quickest action to get the CI passing so we can test the pipeline, We can tear into the meat of the problems afterwards.

CMCDragonkai:
Yea please keep track of a list of tests that have been disabled (quick fixes) at the top of this PR.

maxwell-aiden:
Pipeline Attempt on 648462229 for e45261c3a937f9ec3c2c95d11683849c3697ebad

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/648462229

tegefaulkes:
There are bugs in the `nodes` domain that just don't make sense. But I can recreate it by making all of the `HandlerId`'s the same. This suggests that for some reason in the CI using `this.refreshBucketHandler.name` doesn't return the name like we expect. Seems like `this.refreshBucketHandler.name`, `this.gcBucketHandler.name` and `this.pingAndSetNodeHandler.name` are identical when running in CI.

This is quite odd but it's something we've seen before with decorators overwriting the `name` parameter before. but this is not using decorators, the decorator problem was fixed and this is only happening on CI...

As a solution for now, I'm going to hard code a descriptor string to the end as a failsafe.
```ts
// this
  public readonly refreshBucketHandlerId =
    `${this.basePath}.${this.refreshBucketHandler.name}` as TaskHandlerId;

// becomes this
  public readonly refreshBucketHandlerId =
    `${this.basePath}.${this.refreshBucketHandler.name}.refreshBucketHandlerId` as TaskHandlerId;
```

So if we do have this problem we still end up with a unique ID. I'll also add a test to check this behaviour for regression.

maxwell-aiden:
Pipeline Attempt on 648502683 for 2e4fe59101240893e397fc8eac00726dbc23db2d

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/648502683

maxwell-aiden:
Pipeline Attempt on 648508993 for 0e24e530daab1e001a5f8077f2847b73b6918ea7

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/648508993

CMCDragonkai:
> There are bugs in the `nodes` domain that just don't make sense. But I can recreate it by making all of the `HandlerId`'s the same. This suggests that for some reason in the CI using `this.refreshBucketHandler.name` doesn't return the name like we expect. Seems like `this.refreshBucketHandler.name`, `this.gcBucketHandler.name` and `this.pingAndSetNodeHandler.name` are identical when running in CI.

I'd want to know whether this is something we cannot rely upon. That is the names of arrow function properties in classes. There's no decorator applied to them... and the decorators should be preserving the names generally.

To make it easier to simulate random things in the CI, we can go with a `tests/scratchpad.test.ts`, and have a `test('scratchpad', async () => {})` for arbitrary test simulations.

And yes this may result in some useless history while we try things out. That can be solved by having throwaway feature branches or by squashing history. If we are squashing here, then you want to do it by marking out the commits before we merge.

CMCDragonkai:
If it means we cannot rely on the names of the functions, then yes, we should use constants then. But remember we are also using `this.name` and `this.constructor.name` for all the DB levels, so I would expect `this.prop.name` or `this.method.name` should also work the same even in the presence of decorators.

maxwell-aiden:
Pipeline Attempt on 648567023 for 801f79f43e1676021e443033990cc5dd0552a762

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/648567023

tegefaulkes:
All tests have passed now. Just waiting to see how the rest of the CI jobs go.

Two tests have been disabled until some bugs are addressed.
- tests/nat/endpointIndependentNAT.test.ts:358, `'node1 behind EIM NAT connects to node2 behind EIM NAT via seed node'` was disabled
- tests/nodes/NodeConnectionManager.general.test.ts:326, `'cannot find node (contacts remote node)'` was disabled

I've made a commit for the scratch job and testing the name problem but I'm waiting for the current CI pipeline to finish before pushing that up.

CMCDragonkai:
What was the results of the scratchpad?

tegefaulkes:
Docker integration test failed. Everything else looks good. I'm pushing up the scratch commit now.

maxwell-aiden:
Pipeline Attempt on 648746706 for 18cd6bbd7d1fc76f165b179891da2c10474ab124

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/648746706

CMCDragonkai:
So docker integration tests you made a comment about this earlier. Have a look and we may make a quick fix for this until we can restructure our integration tests into directories.

CMCDragonkai:
Is there a way to also disable the CI when we have a `wip` or `WIP` as the prefix to the commit message?

tests/utils/exec.ts
CMCDragonkai: Let's do a quickfix for this. But long term we need to move all the integration tests into a separate directory and discard these hacks. Instead we would do things like:

```
import parameterisedTest from '..somecommontestmodule..';

describe('', () => {
  test('', parameterisedTest(p1, p2));
});
```

The idea is that `parameterisedTest(p1, p2)` should return an `() => Promise<void>` function to execute as the actual test.

Doing this though, we need to be careful about any side-effects coming from `beforeEach` and `beforeAll` and `afterEach` and `afterAll`. In fact, it would make sense to ensure that all of our tests are functional, which means if things need to be established before a test, they musts be passed as a pointer into the test function itself, so they are still pointing to the same objects being mutated.

This way we separate the "test structure" from the actual test itself which can be shared among different test files.

CMCDragonkai:
The `this.basePath` used in side `NodeManager` should just use `this.constructor.name`.

And all public properties should be up top of the class. Try to give some spacing between the arrow function properties though especially since they are going to have docblocks subsequently.

tegefaulkes:
The scratch is passing.

```
INFO:NodeManager test:checking names
INFO:NodeManager test:NodeManager.refreshBucketHandler
INFO:NodeManager test:NodeManager.gcBucketHandler
INFO:NodeManager test:NodeManager.pingAndSetNodeHandler
INFO:NodeManager test:end of names
```

https://gitlab.com/MatrixAI/open-source/Polykey/-/jobs/3081733042

tegefaulkes:
BUT it's failing in the nodes domain.

```
  ● NodeManager test › Should have unique HandlerIds
    expect(received).not.toEqual(expected) // deep equality
    Expected: not "NodeManager."
      1052 |       logger,
      1053 |     });
    > 1054 |     expect(nodeManager.gcBucketHandlerId).not.toEqual(
           |                                               ^
      1055 |       nodeManager.refreshBucketHandlerId,
      1056 |     );
      1057 |     expect(nodeManager.gcBucketHandlerId).not.toEqual(
      at Object.<anonymous> (tests/nodes/NodeManager.test.ts:1054:47)
```

CMCDragonkai:
Ok investigating that later, move to the docker stuff.

It's interesting that basically the constructor name is fine, but the arrow function property name is not.

tests/utils/exec.ts
tegefaulkes: I've narrowed it down. the `--node-path` parameter we're giving in the test is not properly relative to the working directory. Leading us to try and create the agent at a path that doesn't exist. 

```
// docker
// node path
'tmp/test/polykey-test-NCPSFm/polykey'

// working directory
/home/faulkes/matrixCode/polykey/js-polykey/tmp/test/polykey-test-NCPSFm
```

```
// normally
// node path
'/tmp/polykey-test-WWjNME/polykey'

// working directory
/tmp/polykey-test-WWjNME
```

Ah I see, normally we're using a full absolute path, but with the docker testing we're using a relative path. To fix this we should be using consistent in the style of path for both cases. The simplest change would be to make dataDir an absolute path in all cases by using `path.resolve`.

Also the `testCmd` was not getting passed to the `pkSpawn` opts.

maxwell-aiden:
Pipeline Attempt on 650341300 for 724fd1221acce7bbc73487baf48a4a5fc08d38cb

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650341300

tegefaulkes:
The Docker integration tests ***should*** be passing now. it will take a while for the CI to confirm this.

maxwell-aiden:
Pipeline Attempt on 650368198 for 9a58452142850ad1bbf48445901ab3d42fa5c93c

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650368198

maxwell-aiden:
Pipeline Attempt on 650409110 for ec5aa578e928cfc9def8e2e9418558a25a864630

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650409110

tests/utils/exec.ts
tegefaulkes: There's a single test failure now. 

```
FAIL tests/bin/polykey.test.ts
  ● polykey › format option affects STDERR
    SyntaxError: Unexpected token d in JSON at position 0
        at JSON.parse (<anonymous>)
      63 |       rlErr.once('close', reject);
      64 |     });
    > 65 |     const stderrParsed = JSON.parse(stderrStart);
         |                               ^
      66 |     expect(stderrParsed).toMatchObject({
      67 |       level: expect.stringMatching(/INFO|WARN|ERROR|DEBUG/),
      68 |       key: expect.any(String),
      at Object.<anonymous> (tests/bin/polykey.test.ts:65:31)
      
```

This is odd, I'm not sure what's causing it and it's passing locally. I'm running the test again with logging enabled to try and sus it out. But if I can't see a quick solution I'll just disable the test.

maxwell-aiden:
Pipeline Attempt on 650466358 for c918000118b92b959a11345522c759a11ae69e61

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650466358

maxwell-aiden:
Pipeline Succeeded on 650466358 for c918000118b92b959a11345522c759a11ae69e61

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650466358

tests/utils/exec.ts
tegefaulkes: Found and fixed the problem. Resolving this.

CMCDragonkai:
Avoid wip commits on staging, one it is on staging, we generally don't squash history. We just roll forward.

maxwell-aiden:
Pipeline Attempt on 650511030 for de757738bd71a3f1e212bfc44470f00a9fc254f0

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650511030

CMCDragonkai:
To finish up 5. https://github.com/MatrixAI/Polykey/pull/411#issuecomment-1253467448 we need to do a prerelease tag to observe the GH release page.

Remember the steps are:

```
npm version prerelease --preid alpha
git push --follow-tags
```

CMCDragonkai:
To finish 4. https://github.com/MatrixAI/Polykey/pull/411#issuecomment-1253467448, make sure to only disable the discovery tests. Apparently you don't need to disable the `tests/nat`?

CMCDragonkai:
It looks like `swc-jest` is choking on decorator syntax: https://gitlab.com/MatrixAI/open-source/Polykey/-/jobs/3083242775

```
  ● Test suite failed to run
      x Unexpected token `@`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, `
      | for template literal, (, or an identifier
        ,-[/builds/MatrixAI/open-source/Polykey/src/acl/ACL.ts:22:1]
     22 | @CreateDestroyStartStop(
        : ^
        `----
```

CMCDragonkai:
Ok I think I remember what the problem was.

1. ts-node uses swc, and properly fulfills the swc config
2. swc by itself does not respect `tsconfig.json`
3. which means swc needs its own configuration

Need to create something like this https://swc.rs/docs/configuration/swcrc

Map it from our current `tsconfig.json`. We could do this dynamically inside `jest.config.js` by putting it into the `config` parameter:

```
const fs = require('fs')

const config = JSON.parse(fs.readFileSync(`${__dirname}/.swcrc`, 'utf-8'))

module.exports = {
  transform: {
    '^.+\\.(t|j)sx?$': ['@swc/jest', { ...config, /* custom configuration in Jest */ }],
  },
}
```

CMCDragonkai:
@tegefaulkes please create an issue for the arrow function property names that occurred in the `NodeManager`.

Also can you keep track of all the tests that been disabled, this should be done in MatrixAI/Polykey#441.

maxwell-aiden:
Pipeline Succeeded on 650511030 for de757738bd71a3f1e212bfc44470f00a9fc254f0

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/650511030

CMCDragonkai:
Ok this PR closed, all the remaining comments above, we will do on the next staging commits.

PR #419, opened 2022-07-15
Integrate js-db and concurrency testing

### LINKED ISSUE #244: Support read-after-write consistency for `NodeGraph` bucket operations
### Specification

There's currently an issue of "read-after-write" consistency with sequences of batch operations in the `NodeGraph` buckets database, as `setNode` operations rely on the state of the bucket on insertion.

> Recall that a bucket has a fixed capacity of `k` nodes. If we attempt to add a new node to a full bucket, we find the least active node in the bucket and remove it, in favour of the new node.
> 
> **However**, we come across an issue if we want to create a batch operation of multiple `setNodeOps` within the same bucket. 
> 
> **Because the `ops` are created before any additions, the `bucket` value is statically defined, based on the initial state of the bucket.**
> 
> For example, if I were to batch add 3 nodes (`A`, `B`, `C`) into the same bucket (that already has some node `Z`, we'd have the following array of ops:
> 
> ```
>     [
>       {
>         type: 'put',
>         domain: this.nodeGraphBucketsDbDomain,
>         key: bucketIndex,
>         value: {'Z':9.9.9.9, 'A':1.1.1.1},
>       },
>       {
>         type: 'put',
>         domain: this.nodeGraphBucketsDbDomain,
>         key: bucketIndex,
>         value: {'Z':9.9.9.9, 'B':2.2.2.2},
>       },
>       {
>         type: 'put',
>         domain: this.nodeGraphBucketsDbDomain,
>         key: bucketIndex,
>         value: {'Z':9.9.9.9, 'C':3.3.3.3},
>       },
>     ]
> ```
> 
> That is, the final `setNode` op would overwrite the entire bucket with only nodes `Z` and `C` in the bucket.

This problem can be solved by utilising the same system in js-encryptedfs. We'd need to use:
* separate sublevels for each individual bucket index
* `Transaction` snapshot system 

In this process, it would also be worthwhile to change our keys to be buffers.

### Additional context

* original thread from MR that brought up this issue https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/209#note_680113619
* mid-sprint meeting has some good discussions of this https://vimeo.com/609500458:
  * overall summary at ~1:00:30
  * `Transaction` overview at 11:00
* original nodes claims issue #213

### Tasks

1. Investigate the snapshot system in the EFS `Transaction` class
2. Change buckets to use individual sublevels for each bucket index (instead of a single sublevel indexed by bucket indexes - see how RFS)
3. Add "higher" metadata level for storing the size of a particular bucket. See INodeManager sublevel vs INodeManager dir sublevel. The INodeManager sublevel is capable of storing metadata.
3. Incorporate transaction system into `NodeGraph` operations
4. Remove `lexicographic-integer` dependency from PK (see original comments regarding this here https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/205#note_707121227)

### LINKED ISSUE #294: Fine Grained Concurrency Control Standardisation

### LINKED ISSUE #382: Use fast-check to test concurrency consistency invariants
### Specification

We are likely to have alot of concurrency consistency invariants to test. EFS showed us alot of potential pitfalls and bugs in race conditions. SI helps at least ensure that we always have consistency at the DB layer, however it turns out testing concurrency is always really complicated and involves some overhead. Simple `expect` tests are not enough.

Here comes https://github.com/dubzzz/fast-check, it's a property based testing based on quickcheck. This can be used for quick light-weight model based testing.

Primarily quick-check related libraries provide 2 things:

1. A quick way of generating random data according to some schema, which is useful for any kind fuzzing (and we saw how fuzzing was useful in solving our DB encoding issues)
2. And a way to specify models and testing software if they fit our models.

> While other model-based testing tools such as ModelJUnit 1 are perfectly valid alternatives, as a model-based testing tool, we selected QuickCheck [4] for the fact that it provides test case shrinking functionality whereby upon detecting a failure, it searches for smaller test cases which also generate the fault to aid debugging. Although initially developed as a tool for the Haskell programming language, it has been implemented on a number of platforms and for this work, we utilise the Erlang implementation [2]. QuickCheck is primarily a random test data generation tool but it also supports model-based testing. Given a model and a set of pre and postconditions, the tool is able to automatically generate test cases which are used to verify that the conditions hold in the context of different sequences of actions.

So we should see if fast-check can help us do concurrency testing in a number of domains:

1. Our network domain has lots of concurrency issues
2. Vaults domain - lots of reads/writes here
3. Nodes domain - lots of management of nodes
4. Gestalts and ACL domain

### Additional context

* https://github.com/dubzzz/fast-check
* https://github.com/dubzzz/fast-check/issues/564 - issue about delaying scheduled functions so that the calls themselves can be made concurrent

### Tasks

1. Investigate how `fast-check` can help with concurrent tests and fuzz tests
2. Investigate how to integrate `fast-check` into jest testing including timeouts and reporting
3. Apply fast-check to EFS tests and PK tests

CMCDragonkai:
### Description

This integrates the new js-db 5.0.0 which brings in rocksdb. It now has correct implementation of optimistic transactions using Snapshot Isolation.

In bringing this, all of our concurrency issues should start with a correct foundation. This means any race conflict at the DB layer will throw an `ErrorDBTransactionConflict` exception. These exceptions will need to either be handled with an automatic retry where it is appropriate, or when it is not appropriate, it should be bubbled all the way the user interface (CLI and GUI) and the user can decide what to do.

Write skews can be resolved with new methods `getForUpdate` which materialises the read-write conflict into a write-write conflict. Where these may occur, transaction locks should also be used.

The `DBTransaction` supports both optimistic transactions and pessimistic transactions in the same transaction object. Locking within the transaction should only be done in order to guard against write-skew thrashing (like counter races).

Because the `DBTransaction` has its own lock, and most operations should be done with respect to the DB, most usages of in-memory locking can be replaced with reifying it to the database. This means some in-memory constructs will be replaced with in-DB structures instead.

Transaction mutual exclusion should start at the boundary of the program. This means at the GRPC service request handler level. A transaction will be created there, and this context will be shared by the transitive closure of all nested functions and methods. This should be sufficient for most usages.

The EFS will continue to use the old leveldb until it also transitions to using the new 5.0.0 js-db. The EFS runs its own database atm, so it's independent of the main PK database. Conversion of EFS to using the new db will be a separate PR, it requires separate testing because it has to remove its own locking in favour of using transaction-native locking.

While integrating this new DB, we will also start experimenting with more formal model-checking tests against our codebase in order to test concurrent phenomena.

### Issues Fixed

* Fixes #294
* Fixes #244 
* Related #382

### Tasks

- [x] 1. keypath and options are swapped for relevant DB and trans methods.
- [x] 2. GRPC handlers need to begin the transaction contexts.
- [x] 3. Try and simplfy the `if(tran == null) db.withTranF/G` blocks. use `tran => function()` to avoid code bloat. For generators where possible just pass the generator down with a arrow function. Generally try and avoid wrapping it in a withTransactionG if possible.
- [x] 4. Any simple get/set operations can just use DB directly if tran isn't provided. for single operations a transaction is pointless.
- [x] 5. lockbox with re-enter-ency  is provided, this can be used to avoid conflicts on counters and creation. 
- [x] 6. Remove `ACL` and `Sigchain` implementation of `withTransactionF` with the db version of it. The `Sigchain` version needs to use the transaction's locking now.
- [x] Update DB, EFS and logger dependencies.
- [x] Remove any `ts-ignore` comments for logger usage.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
For better concurrency testing, as well as `fast-check`, we can bring in `jest-extended`. It's already starting to be used by EFS which has lot of concurrency behaviour to verify.

CMCDragonkai:
Rebased on master.

This has now jest-extended, fast-check, DB 5.0.1.

CMCDragonkai:
> Transaction mutual exclusion should start at the boundary of the program. This means at the GRPC service request handler level. A transaction will be created there, and this context will be shared by the transitive closure of all nested functions and methods. This should be sufficient for most usages.

This means the `DBTransaction` object is passed into domain methods. So the methods of `Queue` for example might be:

```
public addSomething(...args, tran?: DBTransaction) {
}
```

We're using the last parameter as the optional argument here... but it may be rolled into a general `options` object too in the future.

But this is the design I've found that works as this also allows the domain objects to be used atomically as each method will create their own transaction context level.

CMCDragonkai:
@tegefaulkes @emmacasolin this PR is an exploration of the concepts of queue, db and tracing. Several PRs are likely to be split off from this PR which will target staging. I'll be setting up a framework from this PR, and then there will be pieces of work sliced out.

* The DB integration will require reworking how transactions are used for atomicity.
* Tracing can be its own domain that creates span that abstracts over how we currently use `logger.getChild`
* Queue involves a lot of interesting ideas

CMCDragonkai:
By using `raw: true`, you can insert `Buffer.from(lexi.pack(num))` directly as a key without first encoding it as a hex value. (Hex value encoding can still be useful though in case it has to maintain some defined set of values independent from other values when it is concatenated together).

Doing this currently with the timestamped level for the tasks.

CMCDragonkai:
When presenting task ids to the end user, in a similar vein to claim ids, we have to show multibase encoded IDs, as that's the only way for users to be able to easily copy-paste them around. But they are decoded to regular `IdSortable` when possible. If they are carried around in JSON and record keys they are carried around with string encoding. Something like `TaskIdString`.

```
TaskId - the regular `IdSortable` type
TaskIdString - string encoded to be used in record keys
TaskIdEncoded - multibase encoded
```

CMCDragonkai:
I'm using `extractTs` from `IdSortable` to acquire the timestamp portion of the `TaskId`, this way the ID and the timestamp that is being stored is the same. The timestamp can never change, but we could allow the `delay` to be dynamically updated, but this is not an essential feature for how we envision using the `Queue`.

CMCDragonkai:
This will be the PR where we do both the DB integration and Queue/Scheduler experiments initially, then depending on whichever one is ready, they can be separated out to be merged in.

tegefaulkes:
As discussed, I'll separate the changes here into two PRs, The queue changes will move into a new branch `feature-queue`.

CMCDragonkai:
You can separate the changes already here into several smaller commits too like introducing jest-extended, and moving `PromiseDeconstructed` out.

CMCDragonkai:
Don't forget to rebase on staging too. We are then making a stacked PR from this.

tegefaulkes:
Summary of changes that need to be made for DB 5.x.x integration. 
- keypath and options are swapped for relevant DB and trans methods.
- GRPC handlers need to begin the transaction contexts.
- Any transaction conflicts bubble up to the user for now while we assess how much conflicting may occur.
- Try and simplfy the `if(tran == null) db.withTranF/G` blocks. use `tran => function()` to avoid code bloat. For generators where possible just pass the generator down with a arrow function. Generally try and avoid wrapping it in a withTransactionG if possible.
- Any simple get/set operations can just use DB directly if tran isn't provided. for single operations a transaction is pointless.
- lockbox with re-enter-ency  is provided, this can be used to avoid conflicts on counters and creation.

tegefaulkes:
Come to think of it, would the transaction locking actually prevent conflicts on counters? we'd have to lock before creating the transaction don't we? If both T1 and T2 edit a counter C, then if T1 finishes any time during the lifetime of T2 then a conflict should happen. So we can only prevent the conflict by delaying creation of T2 until T1 finishes. Am I understanding that right?

tegefaulkes:
If I recall, for some client commands we decided on exposing primitive operations and composing them within the bin commands. This was applying the porcelain plumbing approach. However this means for some of the bin commands we can't use a single transaction to the operation since it would have to begin within the GRPC handler. 

I tried to find an example of this in the nodes bin commands but I didn't see it being an issue there. It's possible that this is not currently a problem but it's something we should watch out for.

tegefaulkes:
Reviewing the service handlers. Looks like they all already create the transaction and pass it down. So not much to change with that respect. One of the service handlers was missing it however and has been updated.

CMCDragonkai:
> Come to think of it, would the transaction locking actually prevent conflicts on counters? we'd have to lock before creating the transaction don't we? If both T1 and T2 edit a counter C, then if T1 finishes any time during the lifetime of T2 then a conflict should happen. So we can only prevent the conflict by delaying creation of T2 until T1 finishes. Am I understanding that right?

We delay by using a lock. That's how serialisation is enforced. So transaction by itself is not sufficient.

CMCDragonkai:
> If I recall, for some client commands we decided on exposing primitive operations and composing them within the bin commands. This was applying the porcelain plumbing approach. However this means for some of the bin commands we can't use a single transaction to the operation since it would have to begin within the GRPC handler. 
> 
> I tried to find an example of this in the nodes bin commands but I didn't see it being an issue there. It's possible that this is not currently a problem but it's something we should watch out for.

We won't deal with papering over transaction conflicts or composing client ops over transactions yet. It can be done but when done over a network it requires live session state usually over a connection lifetime. Also called distributed objects.

CMCDragonkai:
> Reviewing the service handlers. Looks like they all already create the transaction and pass it down. So not much to change with that respect. One of the service handlers was missing it however and has been updated.

Make sure to start testing for `ErrorDBTransactionConflict`.

tegefaulkes:
I can see that `acl` and `Sigchain` both define their own `withTransactionF` method that wraps the db transaction methods. The Sigchain version adds some extra locking logic on top but it never uses it. I can just replace these methods with `this.db.withTransactionF`.

 edit, nevermind, the Sigchain version is using the locking logic. but the ACL one I think should be removed.

CMCDragonkai:
Yes, domains no longer require their own transaction contexts.

Any place where a domain is also locking as part of a "transaction" context can be replaced with native `DBTransaction.lock` calls.

tegefaulkes:
Discovery is using a single lock to enforce consistency between `queueIsEmpty`, `pushKeyToDiscoveryQueue` and `removeKeyFromDiscoveryQueue`.  So far as I can tell, this locking is not actually needed due to the use of transactions. I'm going to remove it.

tegefaulkes:
It would be useful if `DBTransaction.lock` could take a `LevelPath` or `keyPath` as the key to lock on.

CMCDragonkai:
I deliberately did not tie the locking method to the key structure of the DB. The reason is that the locks do not support any kind of hierarchical locking atm. So it would provide an incorrect mental model of how it works.

You just need to concatenate your own keypath if you want to do this.

tegefaulkes:
Most of the DB changes has been applied now. We are seeing some transaction conflicts now.

- tests/vaults/VaultManager.test.ts:320 `cannot concurrently create vaults with the same name`
- tests/vaults/VaultManager.test.ts:344 `can concurrently rename the same vault`
- tests/vaults/VaultManager.test.ts:1763 `Concurrently creating vault with same name only creates 1 vault`
- tests/vaults/VaultInternal.test.ts:292 `concurrent write operations prevented`

As expected, they're concurrency tests.

CMCDragonkai:
Great!

Some new ideas regarding concurrency testing should be incorporated.

1. Using jest-extended asymmetric matchers and using regex matchers for describing non-deterministic output. Asymmetric matchers in particular are useful with the `expect().toSatsify` and also the usage of object/array matchers. You can now describe conditions/predicates for subparts of a data structure. https://github.com/jest-community/jest-extended#asymmetric-matchers (use the `toSatisfy` for arbitrary predicates)
2. Using `fast-check` - I haven't tried this just yet, but you want to see how it can do asynchronous scheduling and enumerating different kinds of asynchronous scheduling.
3. Using `js-async-locks` `Barrier` for sequencing concurrent computation deterministically, this is very "primitive" tool, so you're not really doing full concurrent coverage, prefer the above 2 ideas.

CMCDragonkai:
Commit 16c5f9ef7605554680f6a1373eab26cd7b6b0026 is for integrating `jest-extended`, you can say `test: integrating jest-extended`.

No squashing is needed just yet though as we are stacking PRs.

tegefaulkes:
The `tests/vaults/VaultManager.test.ts` conflicts were fixed after replacing the `vaultsNamesLock` with transaction locks on the names themselves.

tegefaulkes:
Do transaction conflicts not reveal what `keyPath` conflicted? That would be useful information to know.

CMCDragonkai:
It may do so inside the cause chain:

```
        if (e.code === 'TRANSACTION_CONFLICT') {
          this.logger.debug(
            `Failed Committing ${this.constructor.name} ${this.id} due to ${errors.ErrorDBTransactionConflict.name}`,
          );
          throw new errors.ErrorDBTransactionConflict(undefined, {
            cause: e,
          });
        } else {
```

CMCDragonkai:
Otherwise the information isn't recorded elsewhere. I'd only be able to extract information about what the transaction is trying to write, but unless the caused exception shows some extra information there's nothing else.

tegefaulkes:
Looks like we're still getting transaction conflicts in the `VaultManager` tests. They only appear if a bunch of test files in parallel. So it's likely a timing issue?

CMCDragonkai:
Each test should be independent.

CMCDragonkai:
Firstly make sure that EFS is not being constructed with our own Polykey DB. It should be using its own DB, which is still on 4.x. This still the case with `VaultManager`.

Secondly because the DB are separate. The transactions in the PK DB do not reach into the EFS DB, they just have completely independent state.

Any locking you do, should be on the PK DB. There is no locking on EFS (and no transactions would be exposed outside the EFS).

Any transaction conflict cannot be occurring due to EFS.

Transaction conflicts are coming from vault metadata.

CMCDragonkai:
In most cases we can just bubble up the transaction conflict.

In edge cases we do need locking:

1. Where we need strict serialisation for prevent counter racing
2. Where we are dealing with multiple independent areas of state, and a transaction context is not shared between all of them, so a lock is needed to coordinate changes, like for example create a vault which includes creating EFS state that isn't transactional, or creating in-memory structures

Even in 2., if it is possible to do "compensation" at the end (see Saga pattern https://microservices.io/patterns/data/saga.html), we can also avoid locking, and just compensate at the end of a conflict. The `DBTransaction` still exposes function hooks that you can add to commit and rollback. For example one could add a function that deletes/removes/garbage-collects in-memory state.

tegefaulkes:
The vault transaction conflicts are fixed now. Just needed to make sure that the locking encompassed all of the DB operations. At this stage all of the tests are passing. 

Here are a list of domains that are using the transaction locking.
- `Discovery` - locking on gestaltKey to avoid adding the same key to the queue.
- `NotificationsManager` - locking on the message counter.
- `Sigchain` - locking on the claimId and sequence number. 
- `VaultInternal` and `VaultManager` - locking metadata along side other mutations.

Domains using `@matrixai/async-locks` 
- `VaultInternal` and `VaultManager` - object map and efs access.
- `Proxy` - not using DB.
- `Schema` - locking reading writing the status file.
- `NodeConnectionManager` - object map.

From this list, I think we can look at removing it for the sigchain `ClaimId` and `Discovery`s `GestaltKey`. Both of these can be handled via the conflict exception. Locking on the counters can be kept just to avoid common conflicts.

CMCDragonkai:
Remember to use `getForUpdate` when doing locking on counters to prevent write skews. Details are in #294.

tegefaulkes:
That covers most of the DB updates. I just need to address task 4 now. Applying fast-check still needs to be done, However since it's kind of separate I may just make a new PR for that. Try to keep the PRs small and all that.

CMCDragonkai:
There will likely be a long review over what has changed about the test semantics. Plus the fact that now that we expect that transaction conflicts occur, new tests should be added to `bin/` tests that acquire these transaction conflicts.

tegefaulkes:
There may be other places where we need `getForUpdate`.  For example updating node information in the `NodeGraph`.

CMCDragonkai:
Make sure to describe the write skew situation you're dealing with whenever you're changing things.

tegefaulkes:
ACL updates entries in the database based on data in the database. Look at a807ee6 for details

tegefaulkes:
This should cover all aspects of the the DB changes. Time for a review of that. As I mentioned previously #382 can be covered in a new PR. It needs to be looked into and specified in more detail.

.gitlab-ci.yml
```diff
@@ -96,7 +96,7 @@ check:test-generate:
     - if: $CI_COMMIT_BRANCH && $CI_COMMIT_BRANCH !~ /^(?:master|staging)$/ && $CI_COMMIT_TITLE !~ /^[0-9]+\.[0-9]+\.[0-9]+(?:-.*[0-9]+)?$/
       when: manual
 
-check:test:
+.check:test:
```
CMCDragonkai: Can you re-enable check tests for this PR. I want to see the CI/CD run tests over this.

CMCDragonkai:
Can you give a dot point list of tests that have been changed, and the expected transaction conflicts that can occur? Perhaps domain by domain. Expand this https://github.com/MatrixAI/Polykey/pull/419#issuecomment-1211549909 to include commentary on relevant RPC methods.

.gitlab-ci.yml
tegefaulkes: Re-enabled

tegefaulkes:
Making #382 related as it won't be addressed in this PR.

CMCDragonkai:
@tegefaulkes when you have merged js-encryptedfs, come back to this PR and update the EFS and logger here too (by removing all the `@ts-ignore` messages for the logger). The `@matrixai/db` should also be `5.0.2`.

Then final tests with this, and we review it here to merge.

CMCDragonkai:
The fast check application can be in #438 so it can be applied to all the new concurrent mechanisms (scheduler, tasks, context decorators) that we are using.

CMCDragonkai:
1. Update to EFS to `3.5.6`
2. Update DB to `5.0.3`
3. Update logger to `3.0.0`
4. Remove all the `@ts-ignore` messages for the logger injection, find the commit to revert, but also just do a search as well
5. Review any of the problems with the DB lifecycle and EFS lifecycle since we changed a bunch there
6. Clean up and squash the WIP commits

src/sigchain/Sigchain.ts
```diff
@@ -93,7 +91,7 @@ class Sigchain {
     // Therefore, with no claims in the sigchain, the previous sequence number
     // is set to 0.
     await withF([this.db.transaction()], async ([tran]) => {
-      const sequenceNumber = await tran.get<number | null>([
+      const sequenceNumber = await tran.getForUpdate<number | null>([
```
CMCDragonkai: Does this need a `getForUpdate`?

This would be in case you're going to do a `put` later involving the same key, and whether this could occur concurrently.

When we are in a singleton context, it shouldn't be necessary. But I'm debating whether it's better to be safe. But my understanding is that `start` is never called concurrently (and if it is, that's handled by the decorator).

What is the race condition here?

tests/vaults/VaultManager.test.ts
```diff
@@ -317,7 +317,7 @@ describe('VaultManager', () => {
     },
     globalThis.defaultTimeout * 2,
   );
-  test('cannot concurrently create vaults with the same name', async () => {
+  test('Concurrently creating vault with same name only creates 1 vault', async () => {
```
CMCDragonkai: Tests should have lowercase strings.

src/vaults/VaultManager.ts
```diff
@@ -985,7 +978,7 @@ class VaultManager {
     tran: DBTransaction,
   ): Promise<Buffer> {
     let key: Buffer | undefined;
-    key = await tran.get([...this.vaultsDbPath, 'key'], true);
+    key = await tran.getForUpdate([...this.vaultsDbPath, 'key'], true);
```
CMCDragonkai: Same with this. Doesn't seem necessary. What is the race condition that we are guarding against here? Remember it's only necessary during "write skews".

tests/acl/ACL.test.ts
```diff
@@ -407,7 +407,7 @@ describe(ACL.name, () => {
   test('transactional operations', async () => {
     const acl = await ACL.createACL({ db, logger });
     const p1 = acl.getNodePerms();
-    const p2 = acl.withTransactionF(async (tran) => {
+    const p2 = db.withTransactionF(async (tran) => {
       await acl.setNodesPerm(
         [nodeIdG1First, nodeIdG1Second] as Array<NodeId>,
         {
```
CMCDragonkai, line 412: We are going to want to rewrite all of our concurrent testing to be standardised under #382 rather than each domain doing their own thing.

CMCDragonkai:
For all the places where you are using `getForUpdate`, I'm thinking that this is only necessary during write skews. And during the instance methods for the managers. During the creation/destruction of singleton instances, it shouldn't be necessary since they are singletons.

It's only when multiple concurrent calls are likely, as in the creation/destruction of non-singleton instances, and during method invocations likely triggered from multiple RPC handlers.

CMCDragonkai:
Also make sure you're that you are sharing the same `DBTransaction` instance in all the lifecycle handlers.

src/sigchain/Sigchain.ts
tegefaulkes: The same key is updated updated here so I used the `getForUpdate`. if it doesn't really make a difference here due to the life cycle locking then I'd lean towards keeping it for consistency.

tests/acl/ACL.test.ts
tegefaulkes: Is this something to resolve or just something to take note of?

CMCDragonkai: Just a note.

src/vaults/VaultManager.ts
tegefaulkes: I was only checking for if a key was updated following a read. I this case we write a key if it didn't already exist. If you feel it's unnecessary here then I can remove it.

src/sigchain/Sigchain.ts
CMCDragonkai: Yes, but at the same time, this is only useful during a write skew. Which means 2 overlapping reading a shared set of data then writing puts that violates constraints.

Does this strictly happen here even without lifecycle locking?

CMCDragonkai: I should clarify:

1. With SI as we have it now, there is no possibility of consistency anomalies anymore. So technically race conditions cannot occur with proper use of transactions.
2. Write skew is the only anomaly left, which is rare.
3. Counter racing is technically not a race condition, we don't break the consistency of our data model with it due to SI. What happens more accurately is that we end up with lots of repeating transaction conflicts. To avoid this, we use counter locking, the `getForUpdate` is never needed here

CMCDragonkai: So an example of a write skew in our system right now is the relationship between vault id and vault name.

If 2 transactions are updating a vault id's vault name. They could both read from the same vault id, and then set different vault names. Since vault id and vault names are meant to be 1 to 1 relationships, this could break consistency without having a transaction conflict.

2 things should be done here:

1. Add a lock to serialise all vault id vault name updates.
2. Change to using `getForUpdate` for the vault id

The 2. step isn't technically necessary because 1. already serialises it. But 2. should be added because it's a second layer of defense and most importantly it hints to the developer that this is a write skew.

src/vaults/VaultManager.ts
CMCDragonkai: Yep remove, it's only for write skews. https://github.com/MatrixAI/Polykey/pull/419#discussion_r951086979 (unless this itself is a write skew!?).

CMCDragonkai:
Wherever it is relevant, we can always add locks to ensure serialisation. In fact one could say that all handlers are serialised in some special way, but it is often too complex, and something that we want to minimise to where it is needed. Instead we are going to bubble up the transaction conflict up the handlers. Then we can then come down later from the PK client and add auto-retry where appropriate.

tegefaulkes:
So If I understand the usage of `getForUpdate`, we'd want to use it where we get from key `A` and update key `B` or `C` based on the contents of `A`? In that case we'd want to use it for cases were we have relationships between data or indexing?

The example above is for renaming vaults were we update vaultId->vaultName, vaultName->vaultId and delete oldVaultName. Another example is the ACL where we have tight relationships between vaultId -> [nodeId, nodeId] --> permId -> permRef.

So we'd want to use `getForUpdate` to help enforce consistency for data we're trying to keep relationships between.

CMCDragonkai:
What's an example of a ACL write skew? But yes that's the general idea. Have a read on write skew on wikipedia again to refresh memory. It also depends on whether there are any consistency constraints that are being violated here.

This is kind of why I wanted SSI to avoid this problem entirely since it is a complicated edge case.

tegefaulkes:
For ACL there are two aspects.
1. we're following a chain of nodeId -> permId -> permRef and end up updating the data in the permRef. 
2. While following this chain we're checking if the nodeIds are missing the `permId` or `permRef` and removing these nodes from the `permRef` data.

Here is a code snippet.

<details>

```ts
 /**
   * Gets the record of node ids to permission for a given vault id
   * The node ids in the record each represent a unique gestalt
   * If there are no permissions, then an empty record is returned
   */
  @ready(new aclErrors.ErrorACLNotRunning())
  public async getVaultPerm(
    vaultId: VaultId,
    tran?: DBTransaction,
  ): Promise<Record<NodeId, Permission>> {
    if (tran == null) {
      return this.db.withTransactionF((tran) =>
        this.getVaultPerm(vaultId, tran),
      );
    }
    const nodeIds = await tran.get<Record<NodeId, null>>(
      [...this.aclVaultsDbPath, vaultId.toBuffer()],
      false,
    );
    if (nodeIds == null) {
      return {};
    }
    const perms: Record<NodeId, Permission> = {};
    const nodeIdsGc: Set<NodeId> = new Set();
    for (const nodeIdString in nodeIds) {
      const nodeId: NodeId = IdInternal.fromString(nodeIdString);
      const permId = await tran.get(
        [...this.aclNodesDbPath, nodeId.toBuffer()],
        true,
      );
      if (permId == null) {
        // Invalid node id
        nodeIdsGc.add(nodeId);
        continue;
      }
      const permRef = (await tran.get(
        [...this.aclPermsDbPath, permId],
        false,
      )) as Ref<Permission>;
      if (!(vaultId in permRef.object.vaults)) {
        // Vault id is missing from the perm
        nodeIdsGc.add(nodeId);
        continue;
      }
      perms[nodeId] = permRef.object;
    }
    if (nodeIdsGc.size > 0) {
      // Remove invalid node ids
      for (const nodeId of nodeIdsGc) {
        delete nodeIds[nodeId];
      }
      await tran.put(
        [...this.aclVaultsDbPath, vaultId.toBuffer()],
        nodeIds,
        false,
      );
    }
    return perms;
  }
```

</details>

CMCDragonkai:
Ok we have identified where write skews may exist:

```
O <-- O
 ^
  \+- O
```

If it is meant to be a 1:1 relationship, then this application constraint is being violated. It's not a transaction conflict.

Solve by adding a lock to force serialisation. Then by adding a `getForUpdate` which would force a conflict if a lock wasn't there.

This applies to assigning a vault name to a vault id, and also creating a new claim on the sigchain, as the sigchain must be linear.

src/sigchain/Sigchain.ts
tegefaulkes: This has been removed, `getForUpdate` is only used for adding new claims to avoid diverging history.

src/vaults/VaultManager.ts
tegefaulkes: This has been removed.

tegefaulkes:
I think this is good now but we'll have to see what the CI shows us. If it's good then I can start squashing and prepping to merge.

tegefaulkes:
Should be good to merge now.

CMCDragonkai:
CI failures, a timeout and a transaction conflict error.

CMCDragonkai:
Does this PR actually resolve #244?

CMCDragonkai:
For #244, the main idea is that `NodeGraph` should be taking a `tran` parameter as per the comment: https://github.com/MatrixAI/Polykey/issues/244#issuecomment-1046587911.... so that way the node graph can be properly transactional and composed with other operations. I don't remember if there was anything else I missed, but @tegefaulkes took over the development of the node graph lately.

CMCDragonkai:
Long term we should remove all the object map locks in favour of the transaction locks. This is because locks don't compose, and all the object maps are just representations of the source of truth on disk. Ideally we should use the `tran.queueSuccess` and such that can do in-memory changes only after a commit occurs. This means local state inside a transactional function is temporary, and only after a commit succeeds, does global (domain state) get mutated.

Generally I reckon we shouldn't have any operations that is mutating the in-memory state without also involving the underlying disk, so there shouldn't be a situation where they can't coordinate with the transaction locks.

The transaction locks are far more useful generally because they are re-entrant, and they don't release until transaction destruction.

Issue created: https://github.com/MatrixAI/Polykey/issues/443

src/vaults/VaultManager.ts
```diff
@@ -711,6 +714,7 @@ class VaultManager {
 
     if ((await this.getVaultName(vaultId, tran)) == null) return;
     await this.vaultLocks.withF([vaultId, RWLockWriter, 'write'], async () => {
+      await tran.lock(vaultId);
```
CMCDragonkai: This needs to use the a unique key path relative to the vault manager.

I recommend using the same key path and just joining it to a `string`.

Otherwise this can result in global namespace conflict.

This goes for any usage of `tran.lock`.

I think you're doing something similar already with:

```
      await tran.lock(
        [...this.vaultsNamesDbPath, newVaultName].toString(),
        vaultId,
      );
```

It would rely on the array `toString()`. I reckon instead using `[].join('')` cause it will ensure deterministic result rather than relying on the magic `toString()`.

CMCDragonkai: Actually maybe we could just take the key path. I just checked:

```
> [Buffer.from('abc'), 'lol', 123].toString()
'abc,lol,123'
```

This seems to work well enough. Let's hope the `toString` of arrays don't change in the mean time.

CMCDragonkai: I think this is more deterministic and better:

```
> [Buffer.from('abc'), 'lol', 123].join('')
'abclol123'
```

So just pass the same key path and and call a `join('')`.

CMCDragonkai:
I reckon we still need to fix that the key paths used for locking and also investigate why:

```
    ✕ fails after receiving undefined singly signed claim (20523 ms)
```

This is taking longer than 20s. Seems like a possible deadlock.

CMCDragonkai:
> I reckon we still need to fix that the key paths used for locking and also investigate why:
> 
> ```
>     ✕ fails after receiving undefined singly signed claim (20523 ms)
> ```
> 
> This is taking longer than 20s. Seems like a possible deadlock.

Actually I saw that other places also take about 20 seconds there. So I guess it's not a big deal. So we can merge without addressing this for now.

We got lots of things like queues, tasks, cancellation that is coming in to help with managing our asynchronous timelines and resources.

src/vaults/VaultManager.ts
tegefaulkes: Resolved. I also updated all usages of locking on a KeyPath to use `join` instead of `toString`

Issue #435, opened 2022-08-05
Reusable (and exportable) Test Cases - grouping check and integration stage tests


CMCDragonkai:
### Specification

As we are moving towards greater amounts of integration testing, there are lot of existing unit tests that we would like to run, but only during integration. Particularly right now, we are using `tests/bin` as both the unit tests for the CLI, but also integration testing in the docker platform, and eventually for windows MatrixAI/Polykey-CLI#11 and macos MatrixAI/Polykey-CLI#12.

Once we standardise conditional testing MatrixAI/Polykey#434, we will have the ability to use `npm test` to automatically run all tests, while conditional utilities would use "feature flags" to determine whether the test should run or not.

However this doesn't allow us to specify a select group of tests to run. The conditional testing only allows us to say what tests should not be running. This means conditional testing utilities is primarily useful for unit testing stage, where we test everything, but disable certain unit tests when certain conditions are not true.

Instead for integration tests, we need a better "grouping specification" to say instead that these group of tests we do want to test. Think of this as the "allow list" in comparison to the "disallow list" that conditional testing utilities represent. The "disallow list" here would be applied after the "allow list" is applied first.

Originally we evaluated the usage of `jest-runner-groups`, but I didn't like it's comment annotation implementation and the lack of defined behaviour on edge cases.

Instead I believe directories are good way of structuring tests as we have historically done, and we should be able to reuse directories for "allow list" grouping that would be useful for the integration testing stage.

This would mean rather than running `npm test -- tests/bin` under docker integration testing, oen could do instead `npm test -- tests/integration/docker`, only allowing tests under the `tests/integration/docker` during the docker platform. At the same time, this means general unit testing would need to ignore the integration directory.

To avoid copying/duplicating all the tests however, we would need to be able to re-use test cases defined in other directories, and this means the ability to export test cases to be lazily executed.

Here's an initial prototype:

```ts
async function tests() {
  describe(...);
}

if (require.main == null) {
  void tests();
}

export default tests;
```

The only issue with this, is that it's very coarse-grained, it just allows one to reimport tests from another file, but one cannot dig into the nested test structure, it's just all side effects. If we wanted to be able to selectively execute individual tests from a different test file, we would need something like:

```
const tests = {
  groupName: {
    beforeAll: [() => { ... }],
    afterAll: [() => { ... }],
    beforeEach: [() => { ... }],
    afterEach: [() => { ... }],
    testname1: [() => { ... }, ...options],
[...]

Issue #329, opened 2022-02-09
Generic Non-Blocking Task Management ("Queue") for discovery and nodes domains


emmacasolin:
### Specification
Unattended discovery was added in #320, however, there is no concept of priority within the queue. There are three ways that a vertex (a node or identity) can be added to the discovery queue, and they should follow this order of priority:
1. Manually, via the discovery methods `queueDiscoveryByNode()` and `queueDiscoveryByIdentity()` (these are called in the commands `identities discover` (explicit discovery) and `identities trust` (explicitly setting a permission, so we want the Gestalt to be updated via discovery).
2. As a step in the discovery process whereby child vertices are added into the discovery queue in order to discover the entire connected gestalt.
3. Automatically by a process of rediscovery when we want to update existing Gestalts in the Gestalt Graph (to be addressed in #328).

Vertexes with a higher priority should be discovered first, either by being placed at the front of the queue or by modifying the traversal method of the queue. The priority queue could also be further optimised by grouping vertices from the same gestalt together when this is known (for example when adding child vertices).

### Additional context

* Original PR for implementing Unattended Discovery: https://github.com/MatrixAI/js-polykey/pull/320
* Issue for re-adding discovered vertices into the queue to update the gestalt graph https://github.com/MatrixAI/js-polykey/issues/328
* #188 - Indexing the LevelDB may be required to be able to implement a persistent priority queue
* https://github.com/matrixai/js-db/issues/1 - general indexing into js-db
* https://nicolas-van.github.io/modern-async/modern-async/1.1.0/Queue.html

### Tasks
1. Modify the existing Discovery Queue to be a Priority Queue
2. Ensure that when a user interactively wants to discover a gestalt vertex that it becomes the highest priority and gets executed first
3. Look into the potential for further optimising the priority queue, for example by having multiple points of comparison with varying levels of importance that can influence the priority of a particular vertex in the queue

CMCDragonkai:
There's a go implementation of persistent priority queue backed by leveldb here: https://github.com/beeker1121/goque. It can be used as a reference for this. Also a JS implementation here: https://github.com/eugeneware/level-q (the goque is probably more comprehensive).

Our priority queue needs to by default maintain order, because we do want to know the sorted list of jobs. But also allow us to add a special priority number on top.

From my imagination:

This reminds me of the indexing problem, where you can ask for a list of rows sorted by several columns. The first column would dictate the base sort, then subsequent columns would sort any ambiguous sub-orders.

[...]

PR #436, opened 2022-08-08
Standardise process execution for tests

### LINKED ISSUE #432: Standardise Process Execution in `tests/utils/exec.ts`
### Specification

With the resolution of MatrixAI/Polykey#410, there is some messiness left in the `tests/utils/exec.ts`.

We want to standardise the implementation of:

* `pkExec`
* `pkSpawn`
* `pkExpect`

Clarify the behaviour of `pk` and `pkStdio` in the comments.

And also incorporate the requirement of `pkExecNs` and `pkSpawnNs` into the main 3 functions above.

Due to the difference in command and argument quoting behaviour when `shell: true` for `child_process.execFile` (https://github.com/MatrixAI/Polykey/pull/391#issuecomment-1196318536), we should no longer using `child_process.execFile` (and nor are we supposed to use `child_process.exec`).

This means all 3 functions will instead be using `child_process.spawn`.

These functions will need to add an additional option for the command string override. Currently the functions all take existing required parameters, this could be rolled into an options object instead for easier extensibility. The below examples assumes this has been done.

Right now we have a `PK_TEST_COMMAND` which overrides the default `ts-node` command string running against `${globalThis.projectDir}/src/bin/polykey.ts`.

However we want to have the standard override idiom: `PARAMETERS > ENVIRONMENT > DEFAULT`.

This allows us to incorporate `pkExecNs` and `pkSpawnNS` into the same system, as they will be explicitly setting network namespaced command string.

To make this nicer to use, the `exec.ts` should also be exporting:

```ts
const polykeyPath = `${globalThis.projectDir}/src/bin/polykey`

export {
  polykeyPath
};
```

So any usage of `pkExecNs` and `pkSpawnNs` should look like:

```ts
import * as testExec from '../utils/exec';

await pkExec(
  ['agent', 'start'], 
  { command: `${nsenter(usernsPid, netnsPid)} ts-node ${testExec.polykeyPath}}`}
)
```

Now the 3 commands are going to be the entrypoints for executing Polykey. We can still export the target variants (which we've renamed to be explicitly about `WithShell` and `WithoutShell`) in case there are edge cases.

If we need to, it should be like this:

```ts
const polykeyPath = `${globalThis.projectDir}/src/bin/polykey`

// By default `globalThis.testCommand` should be `undefined` because `PK_TEST_COMMAND` will not be set
// This is strictly checking for existence, `PK_TEST_COMMAND=''` is legitimate but undefined behaviour
async function pkExec(args, opts: { command: globalThis.testCommand }) {
  if (opts.command == null) {
    await pkExecWithoutShell(args);
  } else {
    await pkExecWithShell(args, { command });
  }
}

// This is the DEFAULT
async function pkExecWithOutShell(args) {
  await childProcess.spawn(`ts-node ${polykeyPath}`, args, { shell: false });
}

// This is the PARAMETER > ENVIRONMENT override
async function pkExecWithShell(args, opts) {
  args = args.map(escapeShellArg);
  await childProcess.spawn(opts.command, args, { shell: true });
}

function escapeShellArg(arg: string): string {
  return arg.replace(/(["\s'$`\\])/g,'\\$1');
}

export {
  polykeyPath,
  pkExec,
  pkExecWithoutShell,
  pkExecWithShell,
  escapeShellArg,
};
```

This requires that the from the user's (tester's) perspective, the execution behaviour should stay the same. This requires us to sanity check a few things:

1. All signals are propagated from the intermediate shell process to the PK agent process
2. All child process events are received as if the `shell: false` was set

Note the argument escaping function `escapeShellArg`, this is necessary to ensure that any shell metacharacters in the argument strings are interpeted literally.

To make the above work for docker integration tests, we also need to change our `release.nix` to use `Entrypoint` instead of `Cmd`:

```nix
      config = {
        Entrypoint = "/bin/polykey";
      };
```

This will require updates to `Polykey-Infrastructure` as well, as there's no more `/bin/polykey` needed. You can just pass arguments directly after, e.g. `docker run image:tag agent start`.

Finally we are left with:

* `pk` - this just calls the `main` function
* `pkStdio` - this just wraps `pk` with mocking
* `exec` - this is intended to call any other program arbitrarily

The `pk` and `pkStdio` should be reverted to back to how it was before. Except with potentially some signature change to abide by changes to `pkExec` and friends. They are intended to be only for function execution, and not real process testing. If any test needs to do integration testing, they can migrate to using `pkExec` instead. Or a new test can be created that does that specifically.

The `exec` should taking input signature similar to `spawn`, this way it is possible for the end user to also specify `shell: true` and `shell: false` and use `escapeShellArg` for their special cases.

### Additional context

* Because `spawn` returns a process reference, any function expecting simple promise results should wrap the process with any relevant promise handlers.
* https://github.com/MatrixAI/Polykey/pull/391#issuecomment-1196318536 - research on shell behaviour differences
* MatrixAI/Polykey-Simulation#1 - NAT tests is currently incompatible with docker integration tests, they are going to be running independently, this is because NAT tests use a different command string, using namespaces, if NAT tests need to be run in integrated fashion, they can only be applied on the Linux platform with a Linux binary in the `PK_TEST_COMMAND`, if NAT tests need to be run during integration for other platforms, separate tests would have to be written for Docker, MacOS, Windows as they cannot simulate network namespaces (except for docker, but I don't think the CI/CD would be capable of this), so instead they would do this as part of MatrixAI/Polykey-CLI#71 which means going for `testnet.polykey.io` instead of local NAT simulation testing

### Tasks

1. Move `test/utils.ts` into `tests/utils/utils.ts` and create `tests/utils/index.ts`:
   ```ts
   export * from './utils.ts';
   export * as exec from './exec.ts';
   ```
2. Change the `pkExec`, `pkSpawn`, `pkExpect` signatures to taken an options object at the very end, required parameters can be specified as part of object, or as initial parameters
3. Refactor `pkExec`, `pkSpawn`, `pkExpect` according to spec.
4. Remove `pkExecNs` and `pkSpawnNS` and replace with `pkExec` and `pkSpawn` with command override options.
5. Command override options should produce a command string to be used in `spawn(cmd)`.
6. Any usage of `global` should be using `globalThis` instead, this is the correct form way of accessing globals.
   * Incorporate https://eslint.org/docs/latest/rules/no-restricted-globals and ensure that `window` and `global` are not allowed to be used.
7. Export functions in order of precedence (always the functions the users are supposed to be using, and then utilities for edgecases).
8. Change `release.nix` to use `Entrypoint`.
9. Deploy container in CI/CD but also update task definition in `Polykey-Infrastructure`
10. Sanity check signalling and process events when using `shell: true` with the intermediate process, post results of these sanity checks along with code doing the check in this issue for reference.
11. No usages of `execFile` nor `exec` should be used at all.
12. Due to linting rules, `child_process` should actually be `childProcess`, make the necessary changes.

emmacasolin:
### Description
There are many ways of running our tests across different platforms, and this should be standardised.

- Our `pkExec`/`pkSpawn`/`pkExpect` methods should have a default and override implementation. Default should run polykey as normal while the override should run polykey from a different entrypoint. The override should be taken from an argument first and then from an environment variable.
- The remaining methods should not have override implementations, as these are meant for special cases (e.g. mocking for `pkStdio` and general process execution for `exec`). All places in the code where such methods are used and these special cases do not apply should switch over to using `pkExec` or similar.
- The override methods should be exported as well as the defaults to be used for edge cases.

We will need to ensure that signals and events are still handled correctly after making these changes.

### Issues Fixed
* Fixes #432 

### Tasks
- [x] 1. Move test/utils.ts into tests/utils/utils.ts and create tests/utils/index.ts
- [x] 2. Change the pkExec, pkSpawn, ~~pkExpect~~ signatures to taken an options object at the very end for the command, env, and cwd
- [x] 3. Refactor pkExec, pkSpawn, ~~pkExpect~~ according to spec in #432 
- [x] 4. Remove pkExecNs and pkSpawnNS and replace with pkExec and pkSpawn with command override options
- [x] 5. Incorporate https://eslint.org/docs/latest/rules/no-restricted-globals

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/tests/agent/service/notificationsSend.test.ts
+++ b/tests/agent/service/notificationsSend.test.ts
@@ -26,7 +26,7 @@ import * as utilsPB from '@/proto/js/polykey/v1/utils/utils_pb';
 import * as notificationsPB from '@/proto/js/polykey/v1/notifications/notifications_pb';
 import * as nodesUtils from '@/nodes/utils';
 import * as notificationsUtils from '@/notifications/utils';
-import { expectRemoteError } from '../../utils';
+import * as testUtils from '../../utils';
 import { globalRootKeyPems } from '../../fixtures/globalRootKeyPems';
 
 describe('notificationsSend', () => {
@@ -225,7 +225,7 @@ describe('notificationsSend', () => {
     };
     const request1 = new notificationsPB.AgentNotification();
     request1.setContent(notification1.toString());
-    await expectRemoteError(
+    await testUtils.expectRemoteError(
       grpcClient.notificationsSend(request1),
       notificationsErrors.ErrorNotificationsParse,
     );
@@ -253,7 +253,7 @@ describe('notificationsSend', () => {
       .sign(privateKey);
     const request2 = new notificationsPB.AgentNotification();
     request2.setContent(signedNotification);
-    await expectRemoteError(
+    await testUtils.expectRemoteError(
       grpcClient.notificationsSend(request2),
       notificationsErrors.ErrorNotificationsValidationFailed,
     );
@@ -279,7 +279,7 @@ describe('notificationsSend', () => {
     );
     const request = new notificationsPB.AgentNotification();
     request.setContent(signedNotification);
-    await expectRemoteError(
+    await testUtils.expectRemoteError(
       grpcClient.notificationsSend(request),
       notificationsErrors.ErrorNotificationsPermissionsNotFound,
     );
--- a/tests/bin/agent/status.test.ts
+++ b/tests/bin/agent/status.test.ts
@@ -4,7 +4,6 @@ import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
 import Status from '@/status/Status';
 import * as nodesUtils from '@/nodes/utils';
 import config from '@/config';
-import * as execUtils from '../../utils/exec';
 import * as testUtils from '../../utils';
 import { globalRootKeyPems } from '../../fixtures/globalRootKeyPems';
 
@@ -41,7 +40,7 @@ describe('status', () => {
         fs,
         logger,
       });
-      const agentProcess = await execUtils.pkSpawn(
+      const agentProcess = await testUtils.pkSpawn(
         [
           'agent',
           'start',
@@ -54,22 +53,26 @@ describe('status', () => {
           '--verbose',
         ],
         {
-          PK_NODE_PATH: path.join(dataDir, 'polykey'),
-          PK_PASSWORD: password,
-          PK_ROOT_KEY: globalRootKeyPems[0],
+          env: {
+            PK_NODE_PATH: path.join(dataDir, 'polykey'),
+            PK_PASSWORD: password,
+            PK_ROOT_KEY: globalRootKeyPems[0],
+          },
+          cwd: dataDir,
         },
-        dataDir,
         logger,
       );
       await status.waitFor('STARTING');
       let exitCode, stdout;
-      ({ exitCode, stdout } = await execUtils.pkStdio(
+      ({ exitCode, stdout } = await testUtils.pkExec(
         ['agent', 'status', '--format', 'json'],
         {
-          PK_NODE_PATH: path.join(dataDir, 'polykey'),
-          PK_PASSWORD: password,
+          env: {
+            PK_NODE_PATH: path.join(dataDir, 'polykey'),
+            PK_PASSWORD: password,
+          },
+          cwd: dataDir,
         },
-        dataDir,
       ));
       expect(exitCode).toBe(0);
       // If the command was slow, it may have become LIVE already
@@ -78,17 +81,19 @@ describe('status', () => {
         pid: expect.any(Number),
       });
       await status.waitFor('LIVE');
-      const agentProcessExit = execUtils.processExit(agentProcess);
+      const agentProcessExit = testUtils.processExit(agentProcess);
       agentProcess.kill('SIGTERM');
       // Cannot wait for STOPPING because waitFor polling may miss the transition
       await status.waitFor('DEAD');
-      ({ exitCode, stdout } = await execUtils.pkStdio(
+      ({ exitCode, stdout } = await testUtils.pkExec(
         ['agent', 'status', '--format', 'json'],
         {
-          PK_NODE_PATH: path.join(dataDir, 'polykey'),
-          PK_PASSWORD: password,
+          env: {
+            PK_NODE_PATH: path.join(dataDir, 'polykey'),
+            PK_PASSWORD: password,
+          },
+          cwd: dataDir,
         },
-        dataDir,
       ));
       expect(exitCode).toBe(0);
       // If the command was slow, it may have become DEAD already
@@ -97,13 +102,15 @@ describe('status', () => {
         status: expect.stringMatching(/STOPPING|DEAD/),
       });
       await agentProcessExit;
-      ({ exitCode, stdout } = await execUtils.pkStdio(
+      ({ exitCode, stdout } = await testUtils.pkExec(
         ['agent', 'status', '--format', 'json'],
         {
-          PK_NODE_PATH: path.join(dataDir, 'polykey'),
-          PK_PASSWORD: password,
+          env: {
+            PK_NODE_PATH: path.join(dataDir, 'polykey'),
+            PK_PASSWORD: password,
+          },
+          cwd: dataDir,
         },
-        dataDir,
       ));
       expect(exitCode).toBe(0);
       expect(JSON.parse(stdout)).toMatchObject({
@@ -115,10 +122,10 @@ describe('status', () => {
   testUtils.testIf(
     testUtils.isTestPlatformEmpty || testUtils.isTestPlatformDocker,
   )('status on missing agent', async () => {
-    const { exitCode, stdout } = await execUtils.pkStdio(
+    const { exitCode, stdout } = await testUtils.pkExec(
       ['agent', 'status', '--format', 'json'],
       {
-        PK_NODE_PATH: path.join(dataDir, 'polykey'),
+        env: { PK_NODE_PATH: path.join(dataDir, 'polykey') },
       },
     );
     expect(exitCode).toBe(0);
@@ -131,7 +138,7 @@ describe('status', () => {
     let agentPassword;
     let agentClose;
     beforeEach(async () => {
-      ({ agentDir, agentPassword, agentClose } = await execUtils.setupTestAgent(
+      ({ agentDir, agentPassword, agentClose } = await testUtils.setupTestAgent(
         globalRootKeyPems[1],
         logger,
       ));
@@ -149,13 +156,15 @@ describe('status', () => {
         logger,
       });
       const statusInfo = (await status.readStatus())!;
-      const { exitCode, stdout } = await execUtils.pkStdio(
+      const { exitCode, stdout } = await testUtils.pkExec(
         ['agent', 'status', '--format', 'json', '--verbose'],
         {
-          PK_NODE_PATH: agentDir,
-          PK_PASSWORD: agentPassword,
+          env: {
+            PK_NODE_PATH: agentDir,
+            PK_PASSWORD: agentPassword,
+          },
+          cwd: agentDir,
         },
-        agentDir,
       );
       expect(exitCode).toBe(0);
       expect(JSON.parse(stdout)).toMatchObject({
@@ -187,7 +196,7 @@ describe('status', () => {
       });
       const statusInfo = (await status.readStatus())!;
       // This still needs a `nodePath` because of session token path
-      const { exitCode, stdout } = await execUtils.pkStdio(
+      const { exitCode, stdout } = await testUtils.pkExec(
         [
           'agent',
           'status',
@@ -205,8 +214,10 @@ describe('status', () => {
           'json',
           '--verbose',
         ],
-        {},
-        dataDir,
+        {
+          env: {},
+          cwd: dataDir,
+        },
       );
       expect(exitCode).toBe(0);
       expect(JSON.parse(stdout)).toMatchObject({
--- a/tests/nodes/NodeConnection.test.ts
+++ b/tests/nodes/NodeConnection.test.ts
@@ -2,7 +2,7 @@ import type { AddressInfo } from 'net';
 import type { ConnectionInfo, Host, Port, TLSConfig } from '@/network/types';
 import type { NodeId, NodeInfo } from '@/nodes/types';
 import type { Server } from '@grpc/grpc-js';
-import type * as child_process from 'child_process';
+import type { ChildProcessWithoutNullStreams } from 'child_process';
 import net from 'net';
 import os from 'os';
 import path from 'path';
@@ -38,7 +38,7 @@ import * as testNodesUtils from './utils';
 import * as grpcTestUtils from '../grpc/utils';
 import * as agentTestUtils from '../agent/utils';
 import { globalRootKeyPems } from '../fixtures/globalRootKeyPems';
-import { spawnFile } from '../utils/exec';
+import * as testUtils from '../utils';
 
 const destroyCallback = async () => {};
 
@@ -733,14 +733,25 @@ describe(`${NodeConnection.name} test`, () => {
         | NodeConnection<grpcTestUtils.GRPCClientTest>
         | undefined;
       let testProxy: Proxy | undefined;
-      let testProcess: child_process.ChildProcessWithoutNullStreams | undefined;
+      let testProcess: ChildProcessWithoutNullStreams | undefined;
       try {
-        const testProcess = spawnFile('tests/grpc/utils/testServer.ts');
+        const testProcess = await testUtils.spawn(
+          'ts-node',
+          [
+            '--project',
+            testUtils.tsConfigPath,
+            `${globalThis.testDir}/grpc/utils/testServer.ts`,
+          ],
+          undefined,
+          logger,
+        );
         const waitP = promise<string>();
-        testProcess.stdout.on('data', (data) => {
+        testProcess.stdout!.on('data', (data) => {
           waitP.resolveP(data);
         });
-        // TestProcess.stderr.on('data', data => console.log(data.toString()));
+        testProcess.stderr!.on('data', (data) =>
+          waitP.rejectP(data.toString()),
+        );
 
         // Lets make a reverse proxy
         testProxy = new Proxy({
@@ -799,14 +810,25 @@ describe(`${NodeConnection.name} test`, () => {
         | NodeConnection<grpcTestUtils.GRPCClientTest>
         | undefined;
       let testProxy: Proxy | undefined;
-      let testProcess: child_process.ChildProcessWithoutNullStreams | undefined;
+      let testProcess: ChildProcessWithoutNullStreams | undefined;
       try {
-        const testProcess = spawnFile('tests/grpc/utils/testServer.ts');
+        const testProcess = await testUtils.spawn(
+          'ts-node',
+          [
+            '--project',
+            testUtils.tsConfigPath,
+            `${globalThis.testDir}/grpc/utils/testServer.ts`,
+          ],
+          undefined,
+          logger,
+        );
         const waitP = promise<string>();
-        testProcess.stdout.on('data', (data) => {
+        testProcess.stdout!.on('data', (data) => {
           waitP.resolveP(data);
         });
-        // TestProcess.stderr.on('data', data => console.log(data.toString()));
+        testProcess.stderr!.on('data', (data) =>
+          waitP.rejectP(data.toString()),
+        );
 
         // Lets make a reverse proxy
         testProxy = new Proxy({
--- a/tests/utils/exec.ts
+++ b/tests/utils/exec.ts
@@ -1,20 +1,30 @@
 import type { ChildProcess } from 'child_process';
 import type ErrorPolykey from '@/ErrorPolykey';
-import type { PrivateKeyPem } from '@/keys/types';
-import type { StatusLive } from '@/status/types';
-import child_process from 'child_process';
+import childProcess from 'child_process';
 import fs from 'fs';
 import path from 'path';
 import process from 'process';
 import readline from 'readline';
-import os from 'os';
 import * as mockProcess from 'jest-mock-process';
 import mockedEnv from 'mocked-env';
 import nexpect from 'nexpect';
 import Logger from '@matrixai/logger';
 import main from '@/bin/polykey';
-import { promise } from '@/utils';
-import * as validationUtils from '@/validation/utils';
+
+type ExecOpts = {
+  env: Record<string, string | undefined>;
+  command?: string | undefined;
+  cwd?: string;
+  shell?: boolean;
+};
+
+const tsConfigPath = path.resolve(
+  path.join(globalThis.projectDir ?? '', 'tsconfig.json'),
+);
+
+const polykeyPath = path.resolve(
+  path.join(globalThis.projectDir ?? '', 'src/bin/polykey.ts'),
+);
 
 const generateDockerArgs = (mountPath: string) => [
   '--interactive',
@@ -48,31 +58,76 @@ const generateDockerArgs = (mountPath: string) => [
 ];
 
 /**
- * Wrapper for execFile to make it asynchronous and non-blocking
+ * Execute generic (non-Polykey) shell commands
  */
 async function exec(
   command: string,
   args: Array<string> = [],
+  opts: ExecOpts = { env: {} },
 ): Promise<{
+  exitCode: number;
   stdout: string;
   stderr: string;
 }> {
+  const env = {
+    ...process.env,
+    ...opts.env,
+  };
   return new Promise((resolve, reject) => {
-    child_process.execFile(
-      command,
-      args,
-      { windowsHide: true },
-      (error, stdout, stderr) => {
-        if (error) {
-          reject(error);
-        } else {
-          return resolve({
-            stdout,
-            stderr,
-          });
-        }
-      },
-    );
+    let stdout = '',
+      stderr = '';
+    const subprocess = childProcess.spawn(command, args, {
+      env,
+      windowsHide: true,
+      shell: opts.shell ? opts.shell : false,
+    });
+    subprocess.stdout.on('data', (data) => {
+      stdout += data.toString();
+    });
+    subprocess.stderr.on('data', (data) => {
+      stderr += data.toString();
+    });
+    subprocess.on('exit', (code) => {
+      resolve({ exitCode: code ?? -255, stdout, stderr });
+    });
+    subprocess.on('error', (e) => {
+      reject(e);
+    });
+  });
+}
+
+/**
+ * Spawn generic (non-Polykey) shell processes
+ */
+async function spawn(
+  command: string,
+  args: Array<string> = [],
+  opts: ExecOpts = { env: {} },
+  logger: Logger = new Logger(spawn.name),
+): Promise<ChildProcess> {
+  const env = {
+    ...process.env,
+    ...opts.env,
+  };
+  const subprocess = childProcess.spawn(command, args, {
+    env,
+    stdio: ['pipe', 'pipe', 'pipe'],
+    windowsHide: true,
+    shell: opts.shell ? opts.shell : false,
+  });
+  // The readline library will trim newlines
+  const rlOut = readline.createInterface(subprocess.stdout!);
+  rlOut.on('line', (l) => logger.info(l));
+  const rlErr = readline.createInterface(subprocess.stderr!);
+  rlErr.on('line', (l) => logger.info(l));
+  return new Promise<ChildProcess>((resolve, reject) => {
+    subprocess.on('error', (e) => {
+      reject(e);
+    });
+    subprocess.on('spawn', () => {
+      subprocess.removeAllListeners('error');
+      resolve(subprocess);
+    });
   });
 }
 
@@ -88,28 +143,23 @@ async function pk(args: Array<string>): Promise<any> {
  * Both stdout and stderr are the entire output including newlines
  * This can only be used serially, because the mocks it relies on are global singletons
  * If it is used concurrently, the mocking side-effects can conflict
- * @param env Augments env for command execution
- * @param cwd Defaults to temporary directory
  */
 async function pkStdio(
   args: Array<string> = [],
-  env: Record<string, string | undefined> = {},
-  cwd?: string,
+  opts: ExecOpts = { env: {} },
 ): Promise<{
   exitCode: number;
   stdout: string;
   stderr: string;
 }> {
-  if (globalThis.testCmd != null) return pkStdioTarget(args, env, cwd);
-
-  cwd =
-    cwd ??
+  const cwd =
+    opts.cwd ??
     (await fs.promises.mkdtemp(path.join(globalThis.tmpDir, 'polykey-test-')));
   // Recall that we attempt to connect to all specified seed nodes on agent start.
   // Therefore, for testing purposes only, we default the seed nodes as empty
   // (if not defined in the env) to ensure no attempted connections. A regular
   // PolykeyAgent is expected to initially connect to the mainnet seed nodes
-  env['PK_SEED_NODES'] = env['PK_SEED_NODES'] ?? '';
+  opts.env['PK_SEED_NODES'] = opts.env['PK_SEED_NODES'] ?? '';
   // Parse the arguments of process.stdout.write and process.stderr.write
   const parseArgs = (args) => {
     const data = args[0];
@@ -151,7 +201,7 @@ async function pkStdio(
     () => process,
   );
   const mockCwd = mockProcess.spyOnImplementing(process, 'cwd', () => cwd!);
-  const envRestore = mockedEnv(env);
+  const envRestore = mockedEnv(opts.env);
   const mockedStdout = mockProcess.mockProcessStdout();
   const mockedStderr = mockProcess.mockProcessStderr();
   const exitCode = await pk(args);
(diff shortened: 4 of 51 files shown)


emmacasolin:
Aside from converting `pkExpect` to use the same style as `pkExec` and `pkSpawn` (https://github.com/MatrixAI/Polykey/issues/432#issuecomment-1207568216), this should be ready for CI/CD redeployment and testing with docker.

CMCDragonkai:
Regarding [501b803](https://github.com/MatrixAI/Polykey/pull/436/commits/501b803ac27eb4c4cc834ebf01c618e672f0e93a), this should be a separate PR, this requires changes to:

1. All docker tests
2. Polykey-Infrastructure - this requires a new PR in the gitlab repo
3. README instructions on container/docker usage

CMCDragonkai:
@tegefaulkes 

Is [fd87744](https://github.com/MatrixAI/Polykey/pull/436/commits/fd87744a46d11b1ef7338e84a5106684a75464ed) going to be rebased with #437?

emmacasolin:
Ah ok. I've dropped that commit and updated the PR description. So this is now just for standardising the process execution utils.

CMCDragonkai:
This requires rebase on staging. The `globalThis` is being used now. Don't use `window` nor `global`.

emmacasolin:
I can't find any mention of there being a `shell` option for `nexpect.spawn`, and trying to add it in results in a TypeError. This is the definition for the `options` object that it expects:

```ts
interface ISpawnOptions {
    cwd?: string | undefined;
    env?: any;
    ignoreCase?: any;
    stripColors?: any;
    stream?: any;
    verbose?: any;
}
```
---

As a side note, the `nexpect` library doesn't actually use `child_process.spawn` in the latest release, and instead uses `cross_spawn.spawn`. We're still using version 0.4.2, but this means we can't update `nexpect` if we specifically care about using `child_process` underneath.

---

Interestingly it seems they made the switch due to compatibility issues on Windows https://github.com/nodejitsu/nexpect/pull/37

emmacasolin:
> I can't find any mention of there being a `shell` option for `nexpect.spawn`, and trying to add it in results in a TypeError. This is the definition for the `options` object that it expects:
> 
> ```ts
> interface ISpawnOptions {
>     cwd?: string | undefined;
>     env?: any;
>     ignoreCase?: any;
>     stripColors?: any;
>     stream?: any;
>     verbose?: any;
> }
> ```
> 
> As a side note, the `nexpect` library doesn't actually use `child_process.spawn` in the latest release, and instead uses `cross_spawn.spawn`. We're still using version 0.4.2, but this means we can't update `nexpect` if we specifically care about using `child_process` underneath.
> 
> Interestingly it seems they made the switch due to compatibility issues on Windows [nodejitsu/nexpect#37](https://github.com/nodejitsu/nexpect/pull/37)

Regardless of whether `nexpect` uses `child_process` or `cross_spawn` (which just wraps `child_process` and passes the options through), the problem is that `nexpect` doesn't simply pass the options as is to the underlying `spawn` call. Rather the options are injected into a different `context` object that then gets passed into all of the underlying methods.

```js
  options = options || {};
  context = {
    command: command,
    cwd: options.cwd || undefined,
    env: options.env || undefined,
    ignoreCase: options.ignoreCase,
    params: params,
    queue: [],
    stream: options.stream || 'stdout',
    stripColors: options.stripColors,
    verbose: options.verbose
  };
```

So I don't think it's possible for `pkExpect` to have a `shell: true` override if we're using `nexpect`. Since we aren't using `pkExpect` at the moment, and since this is blocking https://github.com/MatrixAI/Polykey/pull/439, I think converting `pkExpect` can go into a separate issue and be put on hold for now. Everything else for this PR is done, just need to squash down the commits.

tests/utils/exec.ts
```diff
@@ -48,31 +62,40 @@ const generateDockerArgs = (mountPath: string) => [
...
+      env,
+      windowsHide: true,
+      shell: opts.shell ? opts.shell : false,
+    });
+    subprocess.stdout.on('data', (data) => {
+      stdout += data.toString();
+    });
+    subprocess.stderr.on('data', (data) => {
+      stderr += data.toString();
+    });
```
tegefaulkes, line 89: Don't we want to return the exit code here? I see it wasn't part of this function before but I think having the exit code is needed here.

tests/utils/exec.ts
```diff
@@ -180,219 +198,134 @@ async function pkStdio(
...
-  const rlOut = readline.createInterface(subprocess.stdout!);
-  rlOut.on('line', (l) => logger.info(l));
-  const rlErr = readline.createInterface(subprocess.stderr!);
-  rlErr.on('line', (l) => logger.info(l));
-  return subprocess;
+  if (opts.command == null) {
+    return await pkSpawnWithoutShell(args, opts, logger);
+  } else {
+    return await pkSpawnWithShell(args, opts, logger);
+  }
```
tegefaulkes: This seems odd to me. We have two commands here that do mostly the same thing. Wouldn't it be cleaner if we had 1 more generic spawn command that can take the command, args, opts and logger as parameters. Then here we can pass in the needed command and augment the augments depending if opts.command is set?

The less we duplicate code the better but I'm not sure if it's worth changing at this point. Thoughts?

tegefaulkes (commented):
Just 2 comments for now. I need to head off for now but I'll come back and do a more thorough review.

tests/utils/exec.ts
CMCDragonkai: If this is the worker and wrapper pattern then it's fine to do this.

tests/utils/exec.ts
CMCDragonkai: Where should the exit code go?

CMCDragonkai:
Ok `pkExpect` can be done in a later issue.

CMCDragonkai:
Add a comment to `pkExpect` to indicate this.

tests/utils/exec.ts
emmacasolin: I suppose we could return the code, but since this is just for running shell commands the code itself probably wouldn't be too informative. Wouldn't it just be 0 or 1 for the most part?

tests/agent/service/notificationsSend.test.ts
```diff
@@ -26,7 +26,7 @@ import * as utilsPB from '@/proto/js/polykey/v1/utils/utils_pb';
 import * as notificationsPB from '@/proto/js/polykey/v1/notifications/notifications_pb';
 import * as nodesUtils from '@/nodes/utils';
 import * as notificationsUtils from '@/notifications/utils';
-import { expectRemoteError } from '../../utils';
+import { expectRemoteError } from '../../utils/utils';
```
CMCDragonkai: Just do `import * as testUtils from '../../utils';`.

CMCDragonkai: Same with the other places.

tests/utils/exec.ts
CMCDragonkai: If this is intended to return a `Promise`, you need to capture the error and throw it as a promise if the process failed to spawn.

A successful exit code is always `0`. See: https://nodejs.org/api/child_process.html#event-error.

And also:

> If this method is invoked as its [util.promisify()](https://nodejs.org/api/util.html#utilpromisifyoriginal)ed version, it returns a Promise for an Object with stdout and stderr properties. The returned ChildProcess instance is attached to the Promise as a child property. In case of an error (including any error resulting in an exit code other than 0), a rejected promise is returned, with the same error object given in the callback, but with two additional properties stdout and stderr.

CMCDragonkai: So no need to return the `exitCode`. It should be part of the child process exception.

Can you find out what the callback returns in the case of a failing call. I think it's something like `ChildProcessError`. Prototype this.

tests/utils/exec.ts
CMCDragonkai: If there are no catch handler, it's fine to `return pkSpawn...`, no need to do an intermediate `await`.

tests/utils/exec.ts
CMCDragonkai: I just checked the old `pkStdio`, it does in fact return the `exitCode`, but this is only because it's only executing a function. As long as it is doing that, returning the `exitCode` should be done.

However if you are executing a child process, any command that doesn't exist with `0` is considered an error right?

tests/utils/exec.ts
```diff
@@ -629,171 +604,33 @@ async function setupTestAgent(privateKeyPem: PrivateKeyPem, logger: Logger) {
 }
 
 function spawnFile(path: string) {
-  return child_process.spawn('ts-node', [
+  return childProcess.spawn('ts-node', [
     '--require',
     'tsconfig-paths/register',
     path,
```
CMCDragonkai: This is no longer necessary because it is in the `tsconfig.json`. Can you make sure that the nodes tests are also no longer doing this. I don't think @tegefaulkes fixed that up in the last PR.

tests/nodes/NodeConnection.test.ts
```diff
@@ -733,7 +733,7 @@ describe(`${NodeConnection.name} test`, () => {
         | NodeConnection<grpcTestUtils.GRPCClientTest>
         | undefined;
       let testProxy: Proxy | undefined;
-      let testProcess: child_process.ChildProcessWithoutNullStreams | undefined;
+      let testProcess: childProcess.ChildProcessWithoutNullStreams | undefined;
       try {
         const testProcess = spawnFile('tests/grpc/utils/testServer.ts');
```
CMCDragonkai: I'm wondering what the purpose of these are? Can these not be done through our other exec utilities? Like `pkExec` now that it's for generic executions? Why not just directly call `childProcess.spawn`. This utility seems useless.

tests/notifications/NotificationsManager.test.ts
```diff
@@ -22,7 +22,7 @@ import * as notificationsErrors from '@/notifications/errors';
 import * as vaultsUtils from '@/vaults/utils';
 import * as nodesUtils from '@/nodes/utils';
 import * as keysUtils from '@/keys/utils';
-import * as testUtils from '../utils';
+import * as testUtils from '../utils/utils';
```
CMCDragonkai: This should be `import * as testUtils from '../utils';`. No need to specify the internal module. The `index.ts` takes care of it.

tests/utils/exec.ts
emmacasolin: The new `pkStdio` returns the exit code. This is for `exec` which is only intended to run shell commands, not Polykey commands. So there is no `exitCode` property, only `code`.

tests/utils/exec.ts
```diff
@@ -16,6 +15,21 @@ import main from '@/bin/polykey';
...
+type ExecOpts = {
+  env: Record<string, string | undefined>;
+  command?: string | undefined;
+  cwd?: string;
+  shell?: boolean;
+};
+
+const tsConfigPath = path.resolve(
+  path.join(globalThis.projectDir, 'tsconfig.json'),
+);
```
CMCDragonkai: This is not necessary anymore. All configuration of `ts-node` is automatically resolved in `tsconfig.json`. So no need to use `-p`... etc.

tests/utils/exec.ts
```diff
@@ -180,219 +198,134 @@ async function pkStdio(
...
+    });
+    subprocess.stderr.on('data', (data) => {
+      stderr += data.toString();
+    });
+    subprocess.on('exit', (code) => {
+      resolve({ exitCode: code ?? -255, stdout, stderr });
+    });
+    subprocess.on('error', (e) => {
+      reject(e);
+    });
```
CMCDragonkai, line 318: Is there any overlap betwee non-0 exit codes and the error event?

tests/utils/exec.ts
```diff
@@ -410,42 +343,79 @@ async function pkExecTarget(
...
+    ...opts.env,
+  };
+  // Recall that we attempt to connect to all specified seed nodes on agent start.
+  // Therefore, for testing purposes only, we default the seed nodes as empty
+  // (if not defined in the env) to ensure no attempted connections. A regular
+  // PolykeyAgent is expected to initially connect to the mainnet seed nodes
+  env['PK_SEED_NODES'] = env['PK_SEED_NODES'] ?? '';
+  const subprocess = childProcess.spawn(
+    'ts-node',
+    ['--project', tsConfigPath, polykeyPath, ...args],
```
CMCDragonkai, line 400: Does not require `'--project', tsConfigPath`.

tests/utils/exec.ts
```diff
@@ -410,42 +343,79 @@ async function pkExecTarget(
...
-  env = {
+  const env = {
     ...process.env,
-    ...env,
-    DOCKER_OPTIONS: generateDockerArgs(cwd).join(' '),
+    ...opts.env,
   };
+  if (globalThis.testPlatform === 'docker') {
+    env.DOCKER_OPTIONS = generateDockerArgs(cwd).join(' ');
+  }
```
CMCDragonkai, line 446: I'm wondering should this be part of the utilities, or should this be in the test cases? This seems weird to put here. It seems the test cases themselves should be the ones changing their execution options by checking the `testPlatform`. @tegefaulkes

tests/agent/service/notificationsSend.test.ts
emmacasolin: I thought we were moving away from index imports? Or is it fine for these utils?

CMCDragonkai:
I think the `setupTestAgent` doesn't belong in `tests/utils/exec.ts`. That isn't about executing some arbitrary command, but is more like a fixture itself.

You can keep it in `tests/utils/utils.ts` for now.

Plus everywhere should be using `import testUtils from '../utils';`. And the `index.ts` already exports (un-namespaced) all the utility functions anyway.

If it doesn't exist, please create `tests/utils/index.ts` that rexports all the relevant functions:

```
export * from './utils.ts';
export * from './exec.ts';
```

This is similar to `src/utils/index.ts`.

Any `tests/utils.ts` should be now be `tests/utils/utils.ts`.

tests/agent/service/notificationsSend.test.ts
CMCDragonkai: Fine for utilities here.

tests/utils/exec.ts
CMCDragonkai: Yea, I know but we want the APIs to be similar. So the question is, what does spawn's error event look like and does it cover non-0 exit codes?

CMCDragonkai:
> > I can't find any mention of there being a `shell` option for `nexpect.spawn`, and trying to add it in results in a TypeError. This is the definition for the `options` object that it expects:
> > ```ts
> > interface ISpawnOptions {
> >     cwd?: string | undefined;
> >     env?: any;
> >     ignoreCase?: any;
> >     stripColors?: any;
> >     stream?: any;
> >     verbose?: any;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As a side note, the `nexpect` library doesn't actually use `child_process.spawn` in the latest release, and instead uses `cross_spawn.spawn`. We're still using version 0.4.2, but this means we can't update `nexpect` if we specifically care about using `child_process` underneath.
> > Interestingly it seems they made the switch due to compatibility issues on Windows [nodejitsu/nexpect#37](https://github.com/nodejitsu/nexpect/pull/37)
> 
> Regardless of whether `nexpect` uses `child_process` or `cross_spawn` (which just wraps `child_process` and passes the options through), the problem is that `nexpect` doesn't simply pass the options as is to the underlying `spawn` call. Rather the options are injected into a different `context` object that then gets passed into all of the underlying methods.
> 
> ```js
>   options = options || {};
>   context = {
>     command: command,
>     cwd: options.cwd || undefined,
>     env: options.env || undefined,
>     ignoreCase: options.ignoreCase,
>     params: params,
>     queue: [],
>     stream: options.stream || 'stdout',
>     stripColors: options.stripColors,
>     verbose: options.verbose
>   };
> ```
> 
> So I don't think it's possible for `pkExpect` to have a `shell: true` override if we're using `nexpect`. Since we aren't using `pkExpect` at the moment, and since this is blocking #439, I think converting `pkExpect` can go into a separate issue and be put on hold for now. Everything else for this PR is done, just need to squash down the commits.

Did you create a separate issue for this?

emmacasolin:
> > > I can't find any mention of there being a `shell` option for `nexpect.spawn`, and trying to add it in results in a TypeError. This is the definition for the `options` object that it expects:
> > > ```ts
> > > interface ISpawnOptions {
> > >     cwd?: string | undefined;
> > >     env?: any;
> > >     ignoreCase?: any;
> > >     stripColors?: any;
> > >     stream?: any;
> > >     verbose?: any;
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > As a side note, the `nexpect` library doesn't actually use `child_process.spawn` in the latest release, and instead uses `cross_spawn.spawn`. We're still using version 0.4.2, but this means we can't update `nexpect` if we specifically care about using `child_process` underneath.
> > > Interestingly it seems they made the switch due to compatibility issues on Windows [nodejitsu/nexpect#37](https://github.com/nodejitsu/nexpect/pull/37)
> > 
> > 
> > Regardless of whether `nexpect` uses `child_process` or `cross_spawn` (which just wraps `child_process` and passes the options through), the problem is that `nexpect` doesn't simply pass the options as is to the underlying `spawn` call. Rather the options are injected into a different `context` object that then gets passed into all of the underlying methods.
> > ```js
> >   options = options || {};
> >   context = {
> >     command: command,
> >     cwd: options.cwd || undefined,
> >     env: options.env || undefined,
> >     ignoreCase: options.ignoreCase,
> >     params: params,
> >     queue: [],
> >     stream: options.stream || 'stdout',
> >     stripColors: options.stripColors,
> >     verbose: options.verbose
> >   };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I don't think it's possible for `pkExpect` to have a `shell: true` override if we're using `nexpect`. Since we aren't using `pkExpect` at the moment, and since this is blocking #439, I think converting `pkExpect` can go into a separate issue and be put on hold for now. Everything else for this PR is done, just need to squash down the commits.
> 
> Did you create a separate issue for this?

Not yet. Will reference here once I have.

tests/utils/exec.ts
CMCDragonkai: It seems that only the `types` are used now in nodes, so this is OK. But it should be better to do `import type { ... } from 'child_process';` instead.

```
nodes/NodeConnection.test.ts
5:import type * as child_process from 'child_process';
736:      let testProcess: child_process.ChildProcessWithoutNullStreams | undefined;
802:      let testProcess: child_process.ChildProcessWithoutNullStreams | undefined;
```

CMCDragonkai:
There's also:

```
nat/utils.ts
1:import type { ChildProcess } from 'child_process';
5:import child_process from 'child_process';
134:  const subprocess = child_process.spawn(
159:  const subprocess = child_process.spawn(
```

Should these be using our exec utilities?

emmacasolin:
> There's also:
> 
> ```
> nat/utils.ts
> 1:import type { ChildProcess } from 'child_process';
> 5:import child_process from 'child_process';
> 134:  const subprocess = child_process.spawn(
> 159:  const subprocess = child_process.spawn(
> ```
> 
> Should these be using our exec utilities?

Yeah it should be possible for those to switch over to using the new `exec`. I'll make that change.

CMCDragonkai:
Beware that the new `exec` only executes right, whereas you have the subprocess there. If you need the subprocess, you may need to keep the `childProcess.spawn`.

tests/agent/service/notificationsSend.test.ts
emmacasolin: Fixed now for all places

tests/utils/exec.ts
emmacasolin: It looks like the `error` event is for higher-level things to do with starting/stopping/communicating with the child process https://nodejs.org/api/child_process.html#event-error - it returns a plain old error object. The `exit` event returns an exit code and signal https://nodejs.org/api/child_process.html#event-exit.

CMCDragonkai: If that's the case, then the error event should result in you `reject(e)`, and then for non-0 exit code, you just need to resolve with the `exitCode` as a returned property.

I think there's actually a similar example of this already being done elsewhere.

tests/utils/exec.ts
emmacasolin: Removed the intermediate awaits from `pkExec` and `pkSpawn`

tests/utils/exec.ts
CMCDragonkai: Make sure to cleanly separate things like:

* `testUtils.exec` - just executes things (returns promise)
* `testUtils.pkExec` - executes PK

If we need a generic spawn then:

* `testUtils.spawn` - generic spawning (returns process)
* `testUtils.pkSpawn` - specifically pk spawning

All functions end up using `childProcess.spawn`.

tests/utils/exec.ts
CMCDragonkai: Here it is, this is the example of what I'm talking about. So `testUtils.exec` and `testUtils.pkExec` should be similar in this regard.

tests/utils/exec.ts
CMCDragonkai: @tegefaulkes perhaps this change will need to be assigned into #435. Because at that point we can group all docker tests into `tests/integration/docker`. And there is where we should be doing things like the above. In fact... I believe the `PK_TEST_PLATFORM` would be removed too.

tegefaulkes: I'm resistant to putting it in the test cases. It means adding it to a lot of places, having to update all those places if it changes and the worst aspect, having to remember to include it when writing new tests. Basically it would add more surface area for making mistakes.

I think it makes sense to include it along side other logic for switching behaviour based on when `globalThis.testCmd` is set.

CMCDragonkai: I've added some details into #435.

But now that I see this was added in this PR.

So can these things be in the test cases in `tests/bin`, or is there some other reason this is here? Perhaps to avoid having to write it out in each test case?

@emmacasolin was this piece of code brought in from a different location?

CMCDragonkai: Oh ok I see it now under `generateDockerArgs`. I reckon this is indeed something that would have to go into #435.

tests/utils/exec.ts
CMCDragonkai: Resolving because this is a fine way of doing worker and wrapper pattern. Just remember to export the `pkSpawnWithoutShell` and `pkSpawnWithShell` as well.

tests/utils/exec.ts
emmacasolin: We now have a generic `exec` and generic `spawn`. The NAT utils no longer need to use `child_process` directly and now use either `testUtils.exec` or `testUtils.spawn` when calling generic shell commands.

tests/utils/exec.ts
emmacasolin: Yep, `testUtils.exec` now follows this same format.

tests/utils/exec.ts
CMCDragonkai: Great. @tegefaulkes we should then avoid using `childProcess` directly (only for its types with `import type { ... } from 'child_process';`.

Any executions goes to `testUtils.spawn` or `testUtils.exec`.

tests/nodes/NodeConnection.test.ts
CMCDragonkai: Use `import type { ... } from 'child_process';` here instead. That way no confusion with using `childProcess` directly.

tests/utils/exec.ts
CMCDragonkai: Use `import type { ... } from 'child_process';` here instead. That way no confusion with using childProcess directly.

tests/nodes/NodeConnection.test.ts
emmacasolin: Done

tests/bin/agent/status.test.ts
```diff
@@ -54,22 +53,26 @@ describe('status', () => {
...
+          },
+          cwd: dataDir,
         },
-        dataDir,
         logger,
       );
       await status.waitFor('STARTING');
       let exitCode, stdout;
-      ({ exitCode, stdout } = await execUtils.pkStdio(
+      ({ exitCode, stdout } = await testUtils.pkExec(
```
CMCDragonkai, line 67: So previously @tegefaulkes changed `pkStdio` to be using `execFile` right?

What we are doing here is:

1. Reverting `pkStdio` to be how it was before, calling the functions and having mocks.
2. The tests that were using `pkStdio` and didn't have problems with the lack of mocking is now using `pkExec`.
3. So `pkStdio` is still preserved for tests that require mocking.

Now I'm a bit confused, how did the tests that required mocking continue to work when `pkStdio` was swapped out to calling `execFile`?

emmacasolin: I don't think `execFile` was ever used? `pkStdio` was only changed to have a `pkStdioTarget` option, which called `child_process.spawn`. The change here was to remove `pkStdioTarget` and anywhere that `pkStdio` was being used for a docker integration test I changed it to use `pkExec` instead. None of the tests that use mocking were ever affected and continue to use `pkStdio` with mocking as they always have.

CMCDragonkai:
Please add this in the end to ignore the `src/proto` from coverage reports: https://jestjs.io/docs/configuration#coveragepathignorepatterns-arraystring

tests/utils/exec.ts
```diff
@@ -629,171 +613,29 @@ async function setupTestAgent(privateKeyPem: PrivateKeyPem, logger: Logger) {
...
+  pk,
+  setupTestAgent,
   processExit,
   expectProcessError,
-  setupTestAgent,
+  pkExecWithShell,
+  pkExecWithoutShell,
+  pkSpawnWithShell,
+  pkSpawnWithoutShell,
   spawnFile,
```
CMCDragonkai: Can this be replaced with `testUtils.spawn`?

tests/utils/exec.ts
```diff
@@ -629,171 +613,29 @@ async function setupTestAgent(privateKeyPem: PrivateKeyPem, logger: Logger) {
...
   pkSpawn,
-  pkStdioTarget,
-  pkExecTarget,
-  pkSpawnTarget,
   pkExpect,
+  pkStdio,
+  exec,
+  spawn,
+  pk,
+  setupTestAgent,
```
CMCDragonkai: See if you can move this to `tests/utils/utils.ts`.

tests/utils/exec.ts
```diff
@@ -629,171 +613,29 @@ async function setupTestAgent(privateKeyPem: PrivateKeyPem, logger: Logger) {
...
+  pkExecWithShell,
+  pkExecWithoutShell,
+  pkSpawnWithShell,
+  pkSpawnWithoutShell,
   spawnFile,
-  nsenter,
-  pkExecNs,
-  pkSpawnNs,
+  escapeShellArgs,
 };
```
CMCDragonkai, line 600: I think the order of exports should be:

```ts
export {
  polykeyPath,
  exec,
  spawn,
  pk,
  pkStdio,
  pkExec,
  pkExecWithShell,
  pkExecWithoutShell,
  pkSpawn,
  pkSpawnWithShell,
  pkSpawnWithoutShell,
  pkExpect,
  processExit,
  expectProcessError,
  escapeShellArgs,
};
```

tests/utils/exec.ts
emmacasolin: Not without some changes. The API for `testUtils.spawn` is slightly different since it's not meant to be used for non-polykey processes. Those four aren't used anywhere at the moment, they're just there for edge cases.

tests/utils/exec.ts
emmacasolin: `setupTestAgent` has been moved.

tests/utils/exec.ts
emmacasolin: Fixed

tests/utils/exec.ts
CMCDragonkai: > The API for testUtils.spawn is slightly different since it's not meant to be used for non-polykey processes.

I thought `testUtils.spawn` and `testUtils.exec` are intended for non-polykey processes?

tests/utils/utils.ts
```diff
@@ -62,6 +68,75 @@ async function setupGlobalKeypair() {
   }
 }
 
+/**
+ *
+ * @param privateKeyPem - Optional root key override to skip key generation
+ * @param logger
+ */
```
CMCDragonkai: Remove the docblock, unnecessary.

emmacasolin:
The nodes tests are timing out locally as well. I'm not sure what caused them to suddenly start doing this - the only change between the last time they were passing and the first time they started failing was changing the imports. I tried reverting this for `NodeConnection.test.ts` (where the timeouts are) but this didn't fix the issue.

tests/utils/exec.ts
emmacasolin: Oh sorry I thought you were referring to the `pkExec...`/`pkSpawn...` variants. I'm not really sure what `spawnFile` does. It's only used in `NodeConnection.test.ts`, but I didn't write those tests.

CMCDragonkai:
Have a look at `NodeConnection.test.ts`. It's just acquiring the test process as a dummy process for GRPC. If your `testUtils.spawn` is simple enough, then it should work fine. You can swap it over and remove `spawnFile`.

Obviously you want to look into why nodes tests are timing out now. Run them individually as well. Maybe see if this was brought in by some previous commits @tegefaulkes any ideas?

tests/utils/exec.ts
CMCDragonkai: Have a look at `NodeConnection.test.ts`. It's just acquiring the test process as a dummy process for GRPC. If your `testUtils.spawn` is simple enough, then it should work fine. You can swap it over and remove `spawnFile`.

tegefaulkes:
Node connection tests will be using agents to test against. Check that all of the agents have the key generation override applied to them. It should look like `privateKeyPemOverride: globalRootKeyPems[2],`, check the key managers too. 

If not that then are some tests timeout a connection when they shouldn't be?

`NodeConnection.test.ts` are passing locally in the `feature-concurrent-state` branch. All the tests individually complete within 5 seconds.

CMCDragonkai:
These 2 locations:

```
»» ~/Projects/Polykey/tests
 ♖ ag 'spawnFile'                                                         (staging) pts/4 16:04:44
utils/exec.ts
631:function spawnFile(path: string) {
795:  spawnFile,

nodes/NodeConnection.test.ts
41:import { spawnFile } from '../utils/exec';
738:        const testProcess = spawnFile('tests/grpc/utils/testServer.ts');
804:        const testProcess = spawnFile('tests/grpc/utils/testServer.ts')
```

Should be using `ts-node` directly. The files don't have a shebang and they don't have executable permissions. They are just scripts that must be executed with `ts-node`. Windows doesn't support shebangs.

Furthermore the path to the file is not robust. It should be using `globalThis.testDir`.

```ts
const testProcess = testUtils.spawn('ts-node', [ `${globalThis.testDir}/grpc/utils/testServer.ts` ]);
```

Also @tegefaulkes has mentioned that there is no error event handler for these functions. You need to attach an error event handler for the `testUtils.spawn` and `testUtils.exec`. So that when `await testUtils.spawn` it should be throwing an exception if an error event is thrown... (this is only relevant at startup).

CMCDragonkai:
Ok so I think the solution is this.

Both `testUtils.exec` and `testUtils.spawn` use `childProcess.spawn`.

Now in the case of exec, there are 2 events you are listening on:

1. `subprocess.on('error')` - here you reject the `Promise<{...}>`
2. `subprocess.on('exit')` - here you resolve the `Promise<{...}>`

In the case of `spawn`, there are also 2 events you need to listen on:

1. `subprocess.on('error')` - here you reject the `Promise<ChildProcess>`
2. `subprocess.on('spawn')` - here you resolve `Promise<ChildProcess>` but you actually need to deregister the error event handler, because that event may still be emitted under other conditions

This way you can do `await testUtils.spawn()` and `await testUtils.exec()` and you can get an exception if the execution of the command fails.

See: https://nodejs.org/api/child_process.html#event-spawn

CMCDragonkai:
If you're finishing up for the day @emmacasolin, @tegefaulkes can take over this, and get this merged.

CMCDragonkai:
Note that this means the `NodeConnection.test.ts` should be changing to using `await testUtils.spawn(...)` in order to get a proper exception telling us **why** `grpc/utils/testServer.ts` is failing to be spawned.

Individual tests **should** still be listening on the error event in case the spawned process itself still results in some kind of error. The `NodeConnection.test.ts` is pretty legacy now, and I wish to refactor that entirely after #329 is resolved.

CMCDragonkai:
When `shell` option is specified, the above behaviour applies to the shell process itself. But that's a caveat that has to be understood by the tester when using the `shell` option.

tegefaulkes:
So, the command is not actually failing to spawn but it is throwing an error in `stderr`.  The real problem is this snippet in `tests/utils/exec.ts:21`.

```ts
const polykeyPath = path.resolve(
  path.join(globalThis.projectDir, 'src/bin/polykey.ts'),
);
```
This is sneaky one. This used to be inline in each function and was only run when the function was called. But here it's run during import of `tests/utils/exec.ts`. This combined with the fact that `globalThis.projectDir` is only defined during jests tests was causing a type error when the testserver was run as a subprocess.

```
Error: thrown: "TypeError [ERR_INVALID_ARG_TYPE]: The \"path\" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:372:5)
    at validateString (node:internal/validators:120:11)
    at Object.join (node:path:1172:7)
    at Object.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/tests/utils/exec.ts:22:8)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Module.m._compile (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/ts-node/src/index.ts:1618:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Object.require.extensions.<computed> [as .ts] (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}
"
```

CMCDragonkai:
So I was wrong about the `-p tsconfigPath` not being needed. It is needed because the CWD is being changed when we run the commands. It's just the `-r` isn't needed anymore.

So everywhere `ts-node` is used, the `--project ${tsconfigPath}` should be used.

A similar const can be used like `const tsconfigPath`. But the same fix to `globalThis.projectDir` would need to be used.

Also `NodeConnection.test.ts` is directly using `ts-node`, that also requires the `--project` option too.

PR #437, opened 2022-08-08
Standardise conditional testing fix

### LINKED ISSUE #434: Standardise Conditional Testing Utilties
### Specification

I want to refactor the conditional test utilities to use boolean conditional parameter again, but instead reify the conditions into things like `isWindows`, `isLinux`... etc, then you can combine them up easily with `isWindows || isLinux`... etc. String parameters are not obvious.

Furthermore, it's apparent now that we have situations where a test should run normally, and also under a particular platform, but no other platforms are allowed. And there are tests that we want to run only for a given platform. This behaviour should be standardised.

A standard set of functions should be used:

```
testIf
describeIf
```

The condition should be `testIf(boolean, ...args)` where `...args` is sent to `test`. This makes it easily extendable.

An alternative is `testIf(() => bool, ...args)` where it's a function. A unary function by itself is sort of useless, but it can be useful if we do `testIf((meta: Metadata) => bool, ...args)`, where metadata is actually passed into the predicate which is derived from the usage of the test. However I cannot think of any metadata right now that would be useful. Maybe the test name, test location, `__dirname`... etc.

A predicate can also be asynchronous, which allows it to do runtime side-effects (mostly queries).

However @tegefaulkes says it's annoying to have to write `os.platform() === 'linux'` all the time, so to avoid this, we can "reify" the booleans. This is more obvious and is better for performance as it eliminates a bunch of checks. Basically we can have `isPlatformLinux` and `hasNsenter`... etc. This is already used in `js-logger` under `hasCaptureStackTrace` and `hasStackTraceLimit`. See: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig and https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables.

Finally this means that through anyone's interpretation:

```
testIf(isPlatformLinux, ...)
```

Should **only** run if the platform is Linux.

Therefore if we want to run without a platform requirement, then:

```
testIf(isPlatformEmpty || isPlatformLinux, ...args)
```

Which means that you can run under any platform and under platform linux.

Finally for and conditions:

```
testIf(hasNsenter && isPlatformLinux, ...args)
```

Which allows us to wire up conditions.

All of this is lot easier to understand, and isn't that much more to write than what is done already.

### Additional context

* Originally posted https://github.com/MatrixAI/Polykey/pull/411#issuecomment-1197616576

### Tasks

- [x] 1. Move `tests/utils.ts` to `tests/utils/utils.ts` and add `tests/utils/index.ts`.
- [x] 2. create a set of reifyed booleans for common test conditions in test utils.
	- platforms
	- expected shell programs for nat testing
- [x] 3. update all `runTestIf` to use `testIf` with reifyed booleans.
- [x] 4. remove`runTestIfPlatform` and `runDescribeIfPlatform`, replace with reifyed booleans and `testIf`

tegefaulkes:
### Description
This PR standardises how conditional testing with `testIf` and `describeIf` is used with new reified booleans.

### Issues Fixed

- Fixes #434 
- Related #435 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
Can you make sure `global` is changed to `globalThis`?

CMCDragonkai:
You can use https://eslint.org/docs/latest/rules/no-restricted-globals to add a restriction to `global`. And provide a replacement for `globalThis`. I think it can be autofixed then.

tests/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './utils';
+export * as exec from './exec';
+export * as platform from './platform';
```
CMCDragonkai: I suggest that these are all:

```
export * from './exec';
export * from './platform';
```

This matches how `src/utils/index.ts` does it.

CMCDragonkai: Only need to namespace here if the functions will conflict.

tests/nat/endpointIndependentNAT.test.ts
```diff
@@ -1,19 +1,21 @@
 import os from 'os';
 import path from 'path';
 import fs from 'fs';
-import process from 'process';
-import shell from 'shelljs';
 import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
 import * as testNatUtils from './utils';
-import { runDescribeIf } from '../utils';
+import { testIf } from '../utils';
```
CMCDragonkai: This should be `import * as testsUtils from '../utils';`. Should be consistent with the rest of usage of utils.

tests/nat/endpointIndependentNAT.test.ts
```diff
@@ -1,19 +1,21 @@
...
+import {
+  isPlatformLinux,
+  hasIp,
+  hasIptables,
+  hasNsenter,
+  hasUnshare,
+} from '../utils/platform';
```
CMCDragonkai: Yea replace this with just `testsUtils.isPlatformLinux`... etc. Once you have `import * as testsUtils from '../utils';`.

tests/utils/platform.ts
```diff
@@ -0,0 +1,29 @@
+import shell from 'shelljs';
+
+const isTestPlatformLinux = global.testPlatform === 'linux';
+const isTestPlatformMacOs = global.testPlatform === 'macos';
+const isTestPlatformWindows = global.testPlatform === 'windows';
+const isTestPlatformDocker = global.testPlatform === 'docker';
+const isTestPlatformEmpty = global.testPlatform == null;
```
CMCDragonkai: This should be using `globalThis`. And also, can you add a comment above that this is temporary until issue #435 is done. Afterwards these replaced as directories for grouping allowed tests.

CMCDragonkai: Use the docblock comment. Not `//`. Give it whitespace in between like:

```
/**
 * ...
 */

const isTestPlatformDocker ...;
const isTestPlatformMacOS ...;
const isTestPlatformWindows ...;

const isPlatformLinux ...;
```

CMCDragonkai:
I think the PR still needs a task list to be ticked off. The PR can have tasks that indicate pending things to be done that isn't about the issue.

tegefaulkes:
I didn't include the task list here since this PR leans more to review than drafting out the problem. I'll add it back in but any `todo`s are covered by the review comments right now.

CMCDragonkai:
I just did a quick check, and it is actually:

```
import * as testUtils from '../../utils';
```

That is the most common.

Not:

```
import * as testsUtils from '../../utils';
```

So I guess we keep it as `testUtils` for now?

This is because I can see there are other similar `test.+Utils`.

Hard to remember cause the directory is `tests` not `test`.

tegefaulkes:
Changed it to `testUtils` now.

.eslintrc
```diff
@@ -28,6 +28,13 @@
     "no-constant-condition": 0,
     "no-useless-escape": 0,
     "no-console": "error",
+    "no-restricted-globals": [
+      "error",
+      {
+        "name": "global",
+        "message": "Use globalThis instead"
```
CMCDragonkai:
```
"Use `globalThis` instead"
```

tegefaulkes: Updated the message.

tegefaulkes:
I'm re-triggering agent/service and nat tests in CI. Seems like they failed due to timeouts. It's just 3 tests in total so I think we're good to merge here.

tegefaulkes:
They timed out again.

<details>

```
 ● nodesCrossSignClaim › fails after receiving undefined singly signed claim

    thrown: "Exceeded timeout of 20000 ms for a hook.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      31 |   let localId: NodeId;
      32 |   let remoteId: NodeId;
    > 33 |   beforeEach(async () => {
         |   ^
      34 |     dataDir = await fs.promises.mkdtemp(
      35 |       path.join(os.tmpdir(), 'polykey-test-'),
      36 |     );

      at tests/agent/service/nodesCrossSignClaim.test.ts:33:3
      at Object.<anonymous> (tests/agent/service/nodesCrossSignClaim.test.ts:20:1)
      
  ● endpoint independent NAT traversal › node1 behind EIM NAT connects to node2 behind EIM NAT via seed node
    thrown: "Exceeded timeout of 40000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
      292 |     globalThis.defaultTimeout * 2,
      293 |   );
    > 294 |   testUtils.testIf(supportsNatTesting)(
          |                                       ^
      295 |     'node1 behind EIM NAT connects to node2 behind EIM NAT via seed node',
      296 |     async () => {
      297 |       const {
      at tests/nat/endpointIndependentNAT.test.ts:294:39
      at Object.<anonymous> (tests/nat/endpointIndependentNAT.test.ts:15:1)
      
```

</details>

I don't think there's a problem with the tests themselves. I know the `agent/service` test is passing locally. Barring a deeper dive into the problem I think we can only increase the timeout for this. 

Want me to look into it now or merge now and fix later?

CMCDragonkai:
Ok leaving the time outs to a later PRs.

PR #438, opened 2022-08-08
Feature TaskManager Scheduler and Queue and Context Decorators for Timed and Cancellable

tegefaulkes:
### Description

At the same time, we may apply the DB to new a generic queuing system. Automatic indexing still isn't available yet, and we may only provide that after transparent encryption/decryption is implemented in the DB. So for now any DB usage has to continue using its own indexes. However SI transactions will help realise this, since we can now be user that any reads and writes are all consistent within the transaction body.

Queue does not use `WorkerManager`, it doesn't know about the nature of the task, instead it executes tasks according to a user-specified concurrency limit, task handlers can decide on their own accord to execute tasks by passing it to the `WorkerManager` for parallel processing, this allows tasks to be IO-bound to be sent into the scheduler/queue without worrying about it being inefficiently sent to the worker manager

### Issues Fixed

* Fixes #329 
* Fixes #297
* Related #382 

### Tasks

Asynchronous Promise Cancellation and Timer:

- [x] 1. `Timer` class with integration into `PromiseCancellable`
- [x] 2. tests for `Timer`
  - [x] timer can be cancelled without unhandled promise rejection
  - [x] timer defaults to eager but can also be lazily cancelled
  - [x] timer does not keep the event loop live
- [x] 3. `@timed` decorator using `Timer` and `AbortSignal`
- [x] 4. tests for the `timed` decorator testing `Signal` and `AbortController`
  - [x] runtime validation
  - [x] decorator syntax
  - [x] decorator expiry
  - [x] decorator propagation
  - [x] explicit cancellation or signal abortion
  - [x] `timed` error stack provides information on where the decorated function is called
- [x] 4. `timed` HOF variant to allow function composition
- [x] 5. `@cancellable` decorator using `AbortSignal` and `PromiseCancellable`
- [x] 6. tests for the `cancellable` decorator
- [x] 7. `cancellable` HOF variant
- ~[ ] 8. `@transactional` decorator using `DBTransaction`~ - not doing due to https://github.com/MatrixAI/Polykey/pull/438#issuecomment-1239446753
- ~[ ] 9. tests for `@transactional`~
- ~[ ] 10. `transactional` HOF variant~
- [x] 11. `timedCancellable` as he combination of timed and cancellable

Tasks System:

- [x] 1. Scheduler
  - [x] Processing task loop has to involve a while loop (even if it is using an async generator)
  - [x] Processing task loop should be blocked based on a `Lock` from `async-locks`
[...]

Issue #353, opened 2022-02-27
Reduce the timeout for establishing a Node Connection within the Discovery domain (by adding timer override to `NodeConnectionManager`)


emmacasolin:
### Specification
The default timeout for node connections is currently set inside the `NodeConnectionManager` for 20s. This can only be set when creating a new Polykey agent by setting the `nodeConnectionManagerConfig` when calling `PolykeyAgent.createPolykeyAgent()`:
```ts
// From src/config.defaults
nodeConnectionManagerConfig: {
  connConnectTime: 20000, // this is the relevant setting
  connTimeoutTime: 60000,
  initialClosestNodes: 3,
},
```
This value is subsequently used for all node connections, regardless of what the node connection is used for or where it originates from. This is a problem for the Discovery domain, where a default timeout of 20s is too long. The Discovery domain calls `NodeManager.requestChainData()`, which calls `NodeConnectionManager.withConnF()`, which subsequently creates the node connection that we use. We need to be able to change the timeout for creating this node connection from inside the Discovery domain to something smaller, for example 1-2s. This does not need to be a modifiable option that is available to the user, since it should not affect the behaviour of the Discovery domain, so the timeout can just be a constant.

In order for the Discovery domain to be able to set its own timeout for Node Connections, we need to propagate the timeout through  a few different places in the code. Specifically:

1. The chosen timeout must be set as a constant property of the `Discovery` class
2. `NodeManager.requestChainData()` needs to have an extra parameter for this timeout (alternatively, since this `requestChainData()` method is not used anywhere else outside of `Discovery` we could move it into the `Discovery` class and methods called by it could just directly use the `Discovery`'s timeout property)
3. `NodeConnectionManager.withConnF()` (called by `requestChainData()`) needs an additional `timeout` parameter
4. `NodeConnectionManager.acquireConnection()` (called by `withConnF()`) needs an additional `timeout` parameter
5. `NodeConnectionManager.createConnection()` (called by `acquireConnection()`) needs an additional `timeout` parameter
6. `NodeConnectionManager.establishNodeConnection()` (called by `createConnection()`) needs an additional `timeout` parameter which defaults to `NodeConnectionManager.connConnectTime` if not set explicitly - all instances of using `NodeConnectionManager.connConnectTime` inside this method need to switch over to using this parameter instead

### Additional context
 - Old issue for handling exceptions during Node Connection timeouts: https://github.com/MatrixAI/js-polykey/issues/348

### Tasks
[...]

Issue #328, opened 2022-02-09
Discovery - revisiting Gestalt Vertices and error handling


emmacasolin:
### Specification
Unattended Discovery was implemented in MatrixAI/Polykey#320, however, nodes/identities that have been previously discovered are currently stored in a set of `visitedVertices` and cannot be re-added to the queue for discovery. While we don't want to add these vertices back into the queue immediately, we need to design some sort of policy for revisiting these vertices when the discovery queue is empty (and/or after a certain amount of time has passed), so that new claims are reflected in the Gestalt Graph for existing Gestalts. This could potentially be achieved by adding TTLs or a `lastVisited` tag to vertices.

Note that, since child vertices are added into the queue during discovery, only one vertex (either a node or identity) per Gestalt needs to be added back into the queue for rediscovery.

### Additional context
- Original PR for Unattended Discovery: https://github.com/MatrixAI/js-polykey/pull/320
- Issue for incorporating a system of priority into the queue (vertices queued for rediscovery should have the lowest priority): https://github.com/MatrixAI/js-polykey/issues/329

### Tasks
1. Add TTLs or a `lastVisited` tag to gestalt vertices
2. Design a policy for revisiting vertices - this should consider the current length of the queue, the vertices currently in the queue, and the amount of time since a given vertex was in the queue

CMCDragonkai:
Comparing to the `NodeGraph`, where buckets have TTLs, the bucket TTLs are eagerly refreshed. Whereas most cache TTLs are lazily refreshed.

But the nodes themselves do not have TTLs, they just a last visited tag.

Therefore gestalt vertexes are similar to nodes in this regard. They will have last visited tags, not TTLs. This is because when the TTL expires, we don't actually want to eagerly "discover" the gestalt vertex.

The discovering protocol is not driven by TTL expiry. It should be driven by kinetic priority determined by proximity and trust and interactivity.

CMCDragonkai:
A bucket which is a group of nodes is therefore similar to a gestalt which is group of gestalt vertices.

> Note that, since child vertices are added into the queue during discovery, only one vertex (either a node or identity) per Gestalt needs to be added back into the queue for rediscovery.

This is an interesting comment, because that would mean each gestalt is consider uniformly.

However unlike buckets in the node graph. Gestalts themselves have a priority and this is based on proximity.

So imagine your gestalt as the highest priority.

Then directly trusted gestalts as the second priority.

Then transitively trusted gestalts as subsequent priorities.

[...]

PR #441, opened 2022-08-11
Integration tests for `testnet.polykey.io`

emmacasolin:
### Description
Adding a suite of integration tests for interacting with a deployed agent at `testnet.polykey.io`. These will go under `tests/integration/testnet`.

### Issues Fixed
* Related MatrixAI/Polykey-CLI#71 
* Fixes MatrixAI/Polykey#388 

### Tasks
- [x] 1. Add diagrams to existing NAT tests
- [x] 2. Create new subdirectory `tests/integration/testnet`
- ~[ ] 3. Write tests as specified in MatrixAI/Polykey-CLI#71~ 

Required tests:
- [X] Connecting to `testnet.polykey.io` seed node (by setting as seed node during `pk agent start`)
- ~[ ] Verification of seed node in local keynode's `NodeGraph` (`pk nodes getall`)~
- ~[ ] Sending notification to seed node (expect failure because no trust established, but this shouldn't show up locally)~
- ~[ ] Establishing trust of seed node with local keynode (`pk identities trust`)~
- ~[ ] Verifying node has been added to seed node's gestalt graph after trust has been established (`pk identities get/list`)~
- ~[ ] Sending notification to seed node, and verifying seed node received (will need to connect to testnet and run `pk notifications read`)~
- [X] Pinging the seed node
- [X] Pinging another node through the seed node

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Updated inline-comment documentation
* [x] Lint fixed
* [x] Squash and rebased
* [x] Sanity check the final build

PR #445, opened 2022-09-02
Integrate TaskManager into NodeGraph and Discovery

### LINKED ISSUE #307: Process exit handler for tracking unresolved asynchronous work ("promise deadlocks"/"why is node terminating for no reason?")
### Is your feature request related to a problem? Please describe.

Coming from https://github.com/MatrixAI/js-polykey/pull/296#issuecomment-989926237.

Promises by themselves don't keep the process open.

This means if you create a promise that is intended to be resolved by an event emitter, or some other event. If that event is not emitted, or never emitted. Then the entire Node process will EXIT at that point, and the rest of the code after awaiting the promise **will not be executed**. This is known as "promise deadlock". https://github.com/nodejs/node/issues/22088#issuecomment-748410302

I encountered this while trying to create a `FlowCountInterceptor` that was not coded correctly and resulted in a promise that never resolves.

One way to fix this, is to use `poll` instead of awaiting for an event driven promise. It turns out looping `setTimeout` keeps the event loop alive whereas waiting for an internal event does not. In fact you can `unref` a timer: https://nodejs.org/api/timers.html. Waiting for IO events usually does block the event loop. Unfortunately the `EventEmitter` does not. There's some extra stuff about node timers and event loop internals that one has to learn here: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

Another thing to help us to is to use our `ExitHandlers` to attach handlers for "unresolved" promises, which can help us detect this kind sneaky bug in the future. This would have to use a combination of `beforeExit` and also the new async hooks module.

Quick hack to detect these deadlocks:

```ts
process.on('beforeExit', (code) => {
  if (process.exitCode == null) {
    console.log('EXIT WITH DEADLOCK', code);
  }
});
```

Note that if you were to put `setTimeout` to then emit the event, nodejs won't terminate until the `setTimeout` is resolved. This is because a timer event is put on the event loop. It's not that node knows that the promise will eventually resolve, it's simply that there's still work to do.

**This only happens when you are awaiting on a promise or expecting a callback to be called, but no work is available on the event loop**.

### Describe the solution you'd like

Incorporate the `beforeExit` handler into the `ExitHandlers` and create a relevant exception that indicates when this occurs. Something like `ErrorPolykeyAsynchronousDeadlock`. This should tell us that somewhere we are waiting for a promise that never resolves, and Node decides to just exit there because there is no other work on the event loop.

We won't have a way to write into the exception where this promise being awaited for is coming from though. So using a tracing debugger would be necessary. But it's better than not having any message at all.

Also doing this will require mocking in `pkStdio` as well to prevent this handler from being attached to the jest process.

### Describe alternatives you've considered

Consider using the `async_hooks` API that is available to track it.

### Additional context

1. https://github.com/nodejs/node/issues/22088
2. https://github.com/nodejs/node/issues/29355
3. https://github.com/nodejs/promises-debugging/issues/16

Note that for the opposite problem: "Why is node still running?", there are these solutions:

* https://github.com/myndzi/wtfnode

### LINKED ISSUE #413: `Starting Connection Forward` infinite loop
### Describe the bug
This bug has been observed on a deployed agent on AWS attempting to connect to a local agent that has been stopped. The deployed agent continuously logs the message `INFO:ConnectionForward [host]:[port]:Starting Connection Forward` (where the host and port are those of the local agent that was shut down until the deployed agent eventually stops. There appears to be some sort of infinite loop here.

```
...
  | 2022-07-13T16:29:58.863+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:30:19.864+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:30:40.865+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:31:01.880+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:31:22.858+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:31:43.874+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:32:04.858+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:32:25.855+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
  | 2022-07-13T16:32:46.858+10:00 | INFO:ConnectionForward 120.18.194.227:3806:Starting Connection Forward
...
```

### To Reproduce
1. Connect to a deployed agent (`npm run polykey -- agent start --seednodes="[deployedAgentNodeId]@[deployedAgentHost]:[deployedAgentPort] --verbose --format json`)
2. At some point after a connection is established kill the local agent

### Expected behavior
We do not expect to see any sort of infinite looping or reconnecting. If an agent goes offline we shouldn't continue trying to re-connect to it.

### Tasks
- [X] 1. Add an exponential back off delay to contacting offline nodes. With implementation and testing, could take 0.5 - 1 day.
  - [X] Use an in memory Map for `NodeIdString` -> `{lastAttempt: number, delay: number}`.
  - [X] If a node is within it's 'backoff' period then we don't attempt to connect to it. This will have to be applied to the `findNode` and `getClosestGlobalNodes`.  We may want the ability to ignore these offline nodes or check every time. So we can add a flag to toggle skipping these nodes.
  - [X] This should be updated in `getClosestGlobalNodes` when pinging nodes.
  - [X] Create a test for this feature.

### LINKED ISSUE #414: Connection dropped/timed out when connecting to deployed agent
### Describe the bug
Occasionally, when attempting to connect to a deployed agent, we see connection timeouts. Sometimes, this also results in a connection being dropped (and shutting down both agents if the connection is attempted during `pk agent start` with the deployed agent specified as a seed node). A specific occurrence of both agents shutting down ended with this error on the local agent:

```
{"type":"ErrorPolykeyRemote","data":{"message":"","timestamp":"2022-07-13T06:34:55.448Z","data":{},"cause":{"type":"ErrorPolykeyRemote","data":{"message":"","timestamp":"2022-07-13T06:34:55.409Z","data":{},"cause":{"type":"ErrorConnectionStartTimeout","data":{"message":"","timestamp":"2022-07-13T06:34:55.363Z","data":{},"description":"Connection start timed out","exitCode":68}},"description":"Remote error from RPC call","exitCode":68,"metadata":{"nodeId":"v5neee032adia1v5u8lj483hhb5bp45ejb8knlmtbuei00lvupudg","host":"54.206.24.79","port":1314,"command":"nodesHolePunchMessageSend"}}},"stack":"ErrorPolykeyRemote\n    at toError (/home/emma/Projects/js-polykey/src/grpc/utils/utils.ts:203:29)\n    at Object.callback (/home/emma/Projects/js-polykey/src/grpc/utils/utils.ts:381:17)\n    at Object.onReceiveStatus (/home/emma/Projects/js-polykey/node_modules/@grpc/grpc-js/src/client.ts:351:26)\n    at /home/emma/Projects/js-polykey/node_modules/@grpc/grpc-js/src/call-stream.ts:206:27\n    at Object.onReceiveStatus (/home/emma/Projects/js-polykey/src/grpc/utils/FlowCountInterceptor.ts:57:13)\n    at InterceptingListenerImpl.onReceiveStatus (/home/emma/Projects/js-polykey/node_modules/@grpc/grpc-js/src/call-stream.ts:202:19)\n    at Object.onReceiveStatus (/home/emma/Projects/js-polykey/node_modules/@grpc/grpc-js/src/client-interceptors.ts:462:34)\n    at Object.onReceiveStatus (/home/emma/Projects/js-polykey/node_modules/@grpc/grpc-js/src/client-interceptors.ts:424:48)\n    at /home/emma/Projects/js-polykey/node_modules/@grpc/grpc-js/src/call-stream.ts:330:24\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)","description":"Remote error from RPC call","exitCode":68,"metadata":{"nodeId":"v0ntccuvju21203s05i11jd9o3d35ae08d29csledv3ntpbjjatcg","host":"54.253.9.3","port":1314,"command":"nodesHolePunchMessageSend"}}}
```

And this series of errors/warnings on the deployed agent:

```
INFO:NodeConnectionManager:Getting connection to vjrd8n87opao3ctcnirccdo9e6crpq5r92m17umnv9ku00od9vdpg
INFO:NodeConnectionManager:existing entry found for vjrd8n87opao3ctcnirccdo9e6crpq5r92m17umnv9ku00od9vdpg
INFO:NodeConnectionManager:withConnF calling function with connection to vjrd8n87opao3ctcnirccdo9e6crpq5r92m17umnv9ku00od9vdpg
WARN:ConnectionReverse 129.94.8.25:31843:Reverse Error: ErrorConnectionTimeout
INFO:ConnectionReverse 129.94.8.25:31843:Stopping Connection Reverse
WARN:ConnectionForward 129.94.8.25:31843:Forward Error: ErrorConnectionTimeout
INFO:ConnectionForward 129.94.8.25:31843:Stopping Connection Forward
INFO:clientFactory:Destroying GRPCClientAgent connected to 129.94.8.25:31843
INFO:NodeConnection 129.94.8.25:31843:Destroying NodeConnection
INFO:NodeConnection 129.94.8.25:31843:Destroyed NodeConnection
INFO:clientFactory:Destroyed GRPCClientAgent connected to 129.94.8.25:31843
{     "type": "ErrorGRPCClientCall",     "data": {         "message": "14 UNAVAILABLE: Connection dropped",         "timestamp": "2022-07-13T06:35:15.212Z",         "data": {             "code": 14,             "details": "Connection dropped",             "metadata": {}         },         "stack": "ErrorGRPCClientCall: 14 UNAVAILABLE: Connection dropped\n    at toError (/lib/node_modules/@matrixai/polykey/dist/grpc/utils/utils.js:174:24)\n    at Object.callback (/lib/node_modules/@matrixai/polykey/dist/grpc/utils/utils.js:331:25)\n    at Object.onReceiveStatus (/lib/node_modules/@matrixai/polykey/node_modules/@grpc/grpc-js/build/src/client.js:189:36)\n    at /lib/node_modules/@matrixai/polykey/node_modules/@grpc/grpc-js/build/src/call-stream.js:111:35\n    at Object.onReceiveStatus (/lib/node_modules/@matrixai/polykey/dist/grpc/utils/FlowCountInterceptor.js:80:29)\n    at InterceptingListenerImpl.onReceiveStatus (/lib/node_modules/@matrixai/polykey/node_modules/@grpc/grpc-js/build/src/call-stream.js:106:23)\n    at Object.onReceiveStatus (/lib/node_modules/@matrixai/polykey/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:365:141)\n    at Object.onReceiveStatus (/lib/node_modules/@matrixai/polykey/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:328:181)\n    at /lib/node_modules/@matrixai/polykey/node_modules/@grpc/grpc-js/build/src/call-stream.js:187:78\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)",         "description": "Generic call error",         "exitCode": 69     } }
```

### To Reproduce
The exact cause of the error is still unknown, however, it may be reproducible by connecting to a deployed agent as a seed node `npm run polykey -- agent start --seednodes="[deployedAgentNodeId]@[deployedAgentHost]:[deployedAgentPort] --verbose --format json`

### Expected behavior
Even if a connection times out, this shouldn't kill an agent, especially not the seed node. The seed node should be able to ignore a failed connection, even if it was the cause of the failed connection, since we need seed nodes to be stable.

### LINKED ISSUE #415: Continuous connection attempts between a deployed seed node and local agent
### Describe the bug
After initially connecting to a deployed seed node, the seed node will also attempt to open a forward connection back to the local agent that initialised the original connection. The two nodes will then continue to connect to each other (how long this will continue for is yet to be determined, but it is at least several minutes).

### To Reproduce
1. Connect to a deployed agent as a seed node using `npm run polykey -- agent start --seednodes="[deployedAgentNodeId]@[deployedAgentHost]:[deployedAgentPort] --verbose --format json`

### Expected behavior
Connections should not be re-attempted continuously, regardless of whether a successful connection was established in the beginning. Once a connection is established there should be no more logs bout the state of the connection, unless it changes. Some of the connections may also be to other nodes as part of syncing the node graph in the background, however, if each of these connections is logged it might start to get confusing, so maybe we should rethink this.

### Additional context
- May be related to #414 

### Tasks
- [x] 1. Remove logging messages for getting existing connections, Should only be logging life-cycle of connections. trivial
- [x] 2. Add 'jitter' to the `refreshBucket` delays to avoid large batches happening at once. trivial
- [x] 3. Add a 'heuristic' for `refreshBucket` to avoid asking the same node over and over again.  1 hour
  - [x] if a `refreshBucket` operation returns less than 20 nodes then that's the WHOLE network it can see. reset the delays for all other buckets.
  - ~If a node has been queried recently then skip asking it again. might not need this one.~ This will be addressed along side #413 if at all.

### LINKED ISSUE #418: Destroyed agent shuts down remote agent performing background tasks
### Describe the bug
We have several asynchronous background queues, some of which involve establishing grpc connections (node connections) with remote agents. However, if the remote agent we're contacting is destroyed during connection establishment then this will shut down our own agent.

Example:
```
{"type":"ErrorAgentClientDestroyed","data":{"message":"","timestamp":"2022-07-15T05:59:49.016Z","data":{},"stack":"ErrorAgentClientDestroyed\n    at /home/emma/Projects/js-polykey/src/nodes/NodeConnectionManager.ts:567:39\n    at /home/emma/Projects/js-polykey/src/nodes/NodeConnectionManager.ts:205:22\n    at withF (/home/emma/Projects/js-polykey/node_modules/@matrixai/resources/src/utils.ts:24:18)\n    at async constructor_.withConnF (/home/emma/Projects/js-polykey/src/nodes/NodeConnectionManager.ts:197:12)\n    at async constructor_.getClosestGlobalNodes (/home/emma/Projects/js-polykey/src/nodes/NodeConnectionManager.ts:495:28)\n    at async constructor_.findNode (/home/emma/Projects/js-polykey/src/nodes/NodeConnectionManager.ts:412:8)\n    at async constructor_.refreshBucket (/home/emma/Projects/js-polykey/src/nodes/NodeManager.ts:580:5)\n    at async constructor_.startRefreshBucketQueue (/home/emma/Projects/js-polykey/src/nodes/NodeManager.ts:711:9)","description":"Agent Client is destroyed","exitCode":64}}
```

### To Reproduce
The timing is quite finicky, but you just need two agents running and you need to kill one of them at the time when the other is doing something like
```
INFO:NodeConnectionManager:Getting connection to vutea98s5hv7qcde3elv4vc9qpqsv2oph374ql8i0ogiu106nia2g
INFO:NodeConnectionManager:existing entry found for vutea98s5hv7qcde3elv4vc9qpqsv2oph374ql8i0ogiu106nia2g
INFO:NodeConnectionManager:withConnF calling function with connection to vutea98s5hv7qcde3elv4vc9qpqsv2oph374ql8i0ogiu106nia2g
```

### Expected behavior
For operations that are occurring in the background, potentially even without the user being aware of them, this should not cause the agent to shut down. While this behaviour makes sense for an operation the user chose to initiate, background tasks that the user has no control over should not be able to kill the agent,

### Tasks
- [x] 1. Review `findNode`, `pingNode`, `getClosestGlobalNodes`, `getRemoteNodeClosestNodes` and `syncNodeGraph` can't throw an error due to a connection error. 1 hour
- ~ MatrixAI/Polykey-CLI#8 ~
	- Any GRPC calls made downstream of GRPC handlers don't need this treatment, any errors will bubble up to the connection and not crash the agent.

tegefaulkes:
### Description

This PR focuses on updating `Nodes`, `Discovery` domain to use the new `Tasks` system.

Generally there are 3 places where background queues are being used. Each of these need to be updated to use the new tasks system from #438.

1. Discovery system
2. Nodes pinging and authenticating nodes.
3. Nodes refreshing buckets on a timeout.
4. `NodeConnectionManager`  Does the network entry procedure with `syncNodeGraph`.

#### SetNode details
In `NodeConnectionManager` when adding a node to the `NodeGraph` with `nodeManager.setNode` we can end up with the case where a bucket is full. When this happens we need to ping nodes within the bucket to determine if they're still alive and remove any nodes that don't respond so we can add the new one. We need to convert this to use the scheduler/queue.

By default the `nodeManager.setNode` doesn't ping a node to check if it's online before adding it. It is expected that you ping the node before using `setNode` to add it. We add nodes whenever we interact or discover a node. This can happen in the cases of... We learn about a node from other nodes, the a connection to the learned node hasn't been made so it needs to be pinged. A node has connected to us so it just needs to be added. We connected to a node, details need to be updated. Given that we needed the ability to ping and `setNode` or just `setNode`. 

#### RefreshBuckets details
`nodeManager.refreshBucket` needs to updated to use the new scheduler/queue system. In this case we will be making use of the `Scheduler` features. A refresh bucket operation needs to be run on a bucket if the bucket hasn't seen activity for an hour. Given this we need schedule each bucket for an hour delay. If a bucket is updated we need to reset the timer for the bucket. To do this we need to make use of the `taskPath` and timer updating features of the `Scheduler`.  These `refreshBucket` tasks should be the lowest priority.

Having `refreshBuckets` system use the `Tasks` system is a little complex since it works on a kind of watchdog system. here are the requirements of the `refreshBucket` system
1. A `refreshBucket` operation selects a random `NodeId` within the target bucket's range of nodes and preforms a search for that node.
2. Each bucket needs to be scheduled to do a `refreshBucket` operation every hour.
3. If data is updated within a bucket then this scheduled delay is reset to an hour.

here are some relevant constraints of the tasks system.
1. A task is scheduled with a delay and is removed when cancelled or completed.
2. A task state can be checked. It can be `scheduled`, `queued`, `active`, `success` or `failure`.
3. A scheduled task's delay can be updated if it is in the `scheduled` state.
4. A task can be 'grouped' using the task's `path`, We can track a task for a single bucket using a path such as `['refreshBucket', bucketIndex]`. This can enable us to find existing tasks for buckets.

Given these constraints I think we need to do the following.
1. During start we need to iterate over the existing `refreshBucket` tasks using the `tasks.getTasksByPath`, reset the delay on existing tasks and create ones for buckets missing them.
2. When any bucket get updated using `nodeManager.setNode` then we need to update the delay of that refresh bucket task. This can be done by getting the task for the bucket using `tasks.getTasksbyPath` and updating the scheduled delay if the task is in the `scheduled` state. If it is `queued` or `Active` then ignore, if not task exist create one. The tasks's delay can be updated by either canceling it and re-creating or using a provided `updateDelay` method.
3. The tasks themselves are technically emphemiral so we could remove all `refreshBuckets` tasks during `nodeManager.stop()`.  Minor detail, won't really make a difference to operation.

#### NodeConenctionManager details
In the `NodeConnectionManager` we have a method `syncNodeGraph` that does the following proceedure. If `syncNodeGrap` is run with blocking = false then we need to run it in the background. 
1. for each seed node we request the closest nodes to our `NodeId`. Then go through this list pinging and adding their details to our nodeGraph.
2. Then for every bucket above the closet node's bucket we run a `refreshBucket` operation.

Looking over this I think it should be part of `NodeManager` and making calls to `NodeConnectionManager` for pinging.

As for how to implement this using `Tasks`.
1. Do an initial getRemoteNodeClosestNodes to the seed node. This is blocking but it could also be a task.
2. The first step then schedules `pingAndAdd` tasks for each of the closest nodes.
3. The first step then schedules with 0 delay a `refreshBucket` operation for every bucket above the closest node's bucket.

#### Other details
Other aspects to consider. The Kademlia `findNode` operation in the `NodeConnectionManager` is considered a single operation but is in essence a priority queue search for the target node. We can consider splitting this up into a compound task where each step in the search can be it's own task within the process. This would apply to the `refreshBucket` operation since that is doing a `findNode` as well.

Handlers will need to support cancel-ability. They must take an abort signal and quickly end operation.

### Issues Fixed

* Fixes #418
* Fixes #415
* Fixes #414
* Fixes #413
* Fixes #307
* Related #353 - This will be a separate PR after this one. After we address updating `Discovery` to use `TaskManager`.
* Related #328 
* Related #329 - this issue contains notes on how to refactor the discovery and prioritising gestalt vertices
* Related #297 - this issue contains notes on what should be made timed/cancellable

### Tasks
 - [x] 1. Add`TaskManager` to `PolykeyAgent`
   - At the end of PolykeyAgent.start you need to do scheduler.startProcessing(), it should be a lazy creation and lazy start from the beginning to allow handlers to be registered. This is why PolykeyAgent would be creating a lazy scheduler so it doesn't start the processing loop.
 - [x] 2. Update nodes pinging/authentication queue to use the new tasks system
 - [x] 3. Update refresh buckets to use new tasks system.
 - ~4. Update `Discovery` domain to use the new tasks system.~
 - [x] 5. Update and check nodes tests.
 - ~6. Update and check `Discovery` tests.~
 - ~7. Address issue #353~ 
 - [x] 8. Address issue #418
 - [x] 9. Address issue #415
 - ~10. Address issue #414~ - Needs more testing, can't address this now. I've added #307 to help with part of this.
 - [x] 11. Address issue #413
 - [x] 12. refactor `setNode` garbage collection https://github.com/MatrixAI/Polykey/pull/445#issuecomment-1246175185
 - [x] 13. Address issue #307  

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
Looking at the `syncNodeGraph` in the `NodeConnectionManager` I think we can move that functionality into the `NodeManager` and avoid having the `NodeConnectionManager` depend on the `Tasks`. But it does fit within the context of other `NodeConnectionManager` methods such as `findNode`, `getClosestGlobalNodes` and `getRemoteNodeClosestNodes` so I'm not sure.

CMCDragonkai:
What does this function do again?

tegefaulkes:
That function does the initial network entry procedure for `Kademlia` by getting the closes nodes to yourself, pining and adding them to your grapth. Then setting up initial refreshBuckets operations to fill in the rest of the NodeGraph.

CMCDragonkai:
@tegefaulkes during integration into node graph. I'd like to ensure that we have understood all the problems that the nodegraph has such as:

* #418 
* #415 
* #414
* #413

CMCDragonkai:
Then we redeploy testnet and focus on #441.

CMCDragonkai:
Discovery has a bug according to vscode:

```
    // If we don't have one then we can't request data so just skip
    if (authIdentityIds === [] || authIdentityIds[0] == null) {
      return undefined;
    }
```

It has never had a proper review, so it's architecture should be reviewed and probably refactored to align to the model we have developed in the Tasks system, since it's also something that has background tasks. In fact... now that the tasks system centralises background task processing, it's possible that we can have discovery and nodegraph both delegate repeat-processing to the tasks system, and remove their own internal loops, thus simplifying how discovery and nodegraph works!!

CMCDragonkai:
* [x] At the end of PolykeyAgent.start you need to do scheduler.startProcessing(), it should be a lazy creation and lazy start from the beginning to allow handlers to be registered. This is why PolykeyAgent would be creating a lazy scheduler so it doesn't start the processing loop.

CMCDragonkai:
This requires a full squash and rebase on top. Don't squash the existing commits from before in #438, only the ones you added on top in the mean time. It may require dropping commits before rebase. Not sure.

tegefaulkes:
Re-base is done. I did some cleaning up of the history as well.

CMCDragonkai:
Can you also start speccing out the tasks that needs to be done here tomorrow?

CMCDragonkai:
The tasks here look pretty minimal, I'd like to incorporate all the new issues I've added to this PR. So it should be fleshed out.

Also it looks like node manager integration is completely done? Surely there's more to it right? Plus all the testing required.

CMCDragonkai:
Make sure to address:

* The bug issues coming from testnet here: https://github.com/MatrixAI/Polykey/pull/445#issuecomment-1240605105
* Backport this problem to #438 - it's breaking the CI/CD
* Integrating it into the PK agent startup and shutdown process https://github.com/MatrixAI/Polykey/pull/445#issuecomment-1242997226

I also added #353 as that relates to the whole idea of deadlines and timeouts which should start using `@timed` and `@cancellable` and `@timedCancellable`.

More work in #328 should be addressed in a later PR, that looks like new features. I'd like to focus this PR on specifically refactoring background task execution and bug fixing.

tegefaulkes:
re-based and targeting staging now.

src/nodes/NodeConnectionManager.ts
```diff
@@ -605,46 +631,54 @@ class NodeConnectionManager {
...
-              await this.nodeManager!.setNode(nodeId, nodeData.address, true);
-            }
-          };
-
-          if (!block) {
-            this.queue.push(pingAndAddNode);
-          } else {
+          const pingAndSetTask = await this.taskManager.scheduleTask({
+            delay: 0,
+            handlerId: this.pingAndSetNodeHandlerId,
```
CMCDragonkai: Suggest using arrow function property on the class for this. Then the name of the property could be the this handler ID too?

tegefaulkes:
`NodeManager.setNode` needs to be refactored for handling full buckets with the TaskManager.
- Nodes are added right away, if a bucket is full then it's put into a pending index for that bucket.
- When nodes are added to a pending index we schedule a job for garbage collecting nodes and adding them in for that bucket. This tasks is a singleton, if a task exists for a bucket then we don't create a new one.
- This task will iterate over pending nodes, for each pending we ping the oldest node. If old node responds we update the time on that and drop the new one. if no response we remove the old one and add the new one. 
- Might just make the pending nodes in memory only, not much point doing this for old data after starting up.
- Locking needs to be applied when updating buckets. Since garbage collection happens in the background it's very likely it could conflict.
- These garbage collection tasks are cleaned up when nodeManager stops.

CMCDragonkai:
So this PR will focus only on getting task manager into nodes and discovery, and fixing testnet bugs. Then merge into staging.

Then disable the staging mac and windows cicd and get staging passing so we can deploy to master and finally build an application that connects to testnet successfully. Docker and linux tests is what we'll focus on.

Afterwhich we can then tackle the remaining discovery issues and other features.

src/nodes/NodeConnectionManager.ts
CMCDragonkai:
```
[this.constructor.name, this.pingAndSetNodeHandler.name].join('.')
// NodeConnectionManager.pingAndSetNodeHandler
```

We would just be careful about this if we ever do a schema migration...

And schema migration is relevant upon doing a 1.0 release.

If a class name/handler name changes, then a schema migration will be necessary as well.

We can code in some "constraints" into our `schema` domain later to see if it DB's schema matches up with what is expected... But we do have a place to code up forward migrations.

tegefaulkes: I have addressed this.

tegefaulkes:
I'm just considering how to handle the pending nodes.  My first thought was to store the pending nodes within buckets the same way as normal nodes in the NodeGraph. But looking at the NodeGrapth it seems I'd have to replicate a lot of logic to handle resetting nodes. 

But now that I think about it we don't really need to track pending nodes within buckets or even keep them persistent. I could keep a single Map of node id to node address and reference that when garbage collecting buckets. The garbage collection process can iterate over the pending nodes and attempt to add them to the nodeGraph. If the bucket is full it can do the usual check against the oldest node in that bucket and update accordingly. 

When doing `getNode` from the `NodeManager` we can check this map first for the and then the `NodeGraph`.

Although this map will only contain nodes for buckets that are full. It could grow in size quickly since half of all nodes will fit in bucket 255. This means 255 will be constantly full and end up with half of all nodes going into the pending map. Is this a problem? Do we want some way to limit this?

src/nodes/NodeManager.ts
```diff
@@ -410,6 +459,7 @@ class NodeManager {
     force: boolean = false,
     timeout?: number,
     tran?: DBTransaction,
+    options: { signal?: AbortSignal } = {},
```
CMCDragonkai: The `timeout` and the `signal` should be replaced with a `ContextTimed` object.

Use the `timedCancellable` decorator here.

This is a bit more complicated than original design.

The reason is that the `setNode` can schedule background tasks or not.

In the case of `block: false`, then everything runs within the same promise lifecycle. In which case the `timer` and the `signal` gets propagated down to all the downstream functions. Now this would mean:

```ts
public setNode(
  nodeId: NodeId,
  nodeAddress: NodeAddress,
  block?: boolean,
  force?: boolean,
  ctx?: Partial<ContextTimed>
): PromiseCancellable<void>;
@ready(new nodesErrors.ErrorNodeManagerNotRunning())
@timedCancellable(false, 1000)
public async setNode(
  nodeId: NodeId,
  nodeAddress: NodeAddress,
  block: boolean = true,
  force: boolean = false,
  @context ctx: ContextTimed
): Promise<void> {

  // ...

}
```

CMCDragonkai: Once this change takes place. The meaning of `timeout` is different. The original `timeout` corresponds to the connection establishment timeout that occurs for every proxy connection.

However in this case, that's not what what the timer means. Instead timer is now defaulted to 1000 ms. Which is the total amount of time allowed to complete a `setNode` operation.

CMCDragonkai: The `ctx.timer.getTimeout()` can then be used to pass into `pingNode` which has independent timeouts that is all occurring in parallel. This works for the `block: true`.

What about `block: false`, in this case, we would propagate the `ctx.timer.delay` into the `TaskDeadline`, which itself is then propagated to the `pingNode` running inside the task handler function.

CMCDragonkai: This applies to the intermediate `garbageCollectOldNode` too.

In the new garbage collection mechanism where we are maintaining pending nodes. We would be iterating the existing nodes in a bucket in parallel and pinging them. However we may wish still abide by the k-parallel constant.

This `k parallel constant` should be written to the class as a parameter.

src/nodes/NodeManager.ts
```diff
@@ -474,45 +536,48 @@ class NodeManager {
...
   }
 
+  // FIXME: make cancellable
   private async garbageCollectOldNode(
     bucketIndex: number,
     nodeId: NodeId,
     nodeAddress: NodeAddress,
     timeout?: number,
+    options: { signal?: AbortSignal } = {},
   ) {
```
CMCDragonkai: The garbage collection mechanism is changing to become like a while loop that goes over all the existing and pending nodes. We will need a `Semaphore` exposed in the `async-locks` library so the while loop can iterate over the combined bucket upon acquiring a lock. The semaphore limit is set to the k parallel constant.

This should allow us to collapse the existing and pending nodes into a single bucket with the remaining nodes.

CMCDragonkai: The maximum number of pings that could occur is the 20 bucket limit, but the minimum is however many nodes there are in the pending bucket partition multiplied by the k parallel constant.

CMCDragonkai: Actually max is 20, min will be 3. The true number of pings will be somewhere in between.

CMCDragonkai:
If we are going to do a reset buckets, and the pending node are just in-memory. We can finish the reset, then just iterate over pending nodes and try to insert it into the system using `setNode` again.

Reset buckets only occur when we reset the root key, so it will be rare, and when that happens we can have a sort of "stop the world" thing to reset a bunch of stuff. Although this happens in the background.

At the same time, pending nodes can also be dropped. We're already dropping a bunch of existing nodes during the reset bucket.

src/nodes/NodeManager.ts
```diff
@@ -573,169 +641,153 @@ class NodeManager {
...
-      bucketIndex,
-      Date.now() + this.refreshBucketTimerDefault,
-    );
-    // If the bucket was pending a refresh we remove it
-    this.refreshBucketQueueRemove(bucketIndex);
-    if (bucketIndex === this.refreshBucketNext) {
-      // Bucket is same as next bucket, this affects the timer
-      this.refreshBucketRefreshTimer();
+  @ready(new nodesErrors.ErrorNodeManagerNotRunning())
+  public async updateRefreshBucketDelay(
```
CMCDragonkai, line 891: In the process of refreshing the bucket delay, the `getTask` and `updateTask` cannot share the same transaction context.

```ts
const task = await getTask();
try {
  await updateTask(task.id);
} catch (e) {
  if (e instanceof ErrorTaskMissing) {
     // create it
  } else if (e instanceof ErrorTaskRunning) {
    // If it is already running
    // Do nothing!
  }
}
```

This is because if the `getTask` and `updateTask` shared a context, even though `updateTask` is locking the `TaskId`, the snapshot is created at `getTask`, and 2 transactions doing a get, then put, can have a conflict. We don't want a conflict here because we don't expect a conflict exception due to task mutation when interacting with the nodes domain.

So here, we are relying on the exceptions of `updateTask` to know whether we can do what we want.

src/nodes/NodeManager.ts
```diff
@@ -573,169 +641,153 @@ class NodeManager {
...
+      const exists = existingTasks[bucketIndex];
+      if (!exists) {
+        // Create a new task
+        this.logger.debug(
+          `Creating refreshBucket task for bucket ${bucketIndex}`,
+        );
+        const spread =
+          (Math.random() - 0.5) *
+          this.refreshBucketDelay *
+          this.refreshBucketDelaySpread;
```
CMCDragonkai: Can this be a utility function. Call it `nodesUtils.refreshBucketsDelayJitter`. It should a function that can give you this number that you adding to the `delay`.

src/nodes/NodeManager.ts
```diff
@@ -36,37 +36,67 @@ class NodeManager {
...
+    const spread =
+      (Math.random() - 0.5) *
+      this.refreshBucketDelay *
+      this.refreshBucketDelaySpread;
+    await this.taskManager.scheduleTask({
+      delay: this.refreshBucketDelay + spread,
+      handlerId: this.refreshBucketHandlerId,
+      lazy: true,
+      parameters: [bucketIndex],
+      path: ['refreshBucket', `${bucketIndex}`],
```
CMCDragonkai: This should become a class property. It should also have the class name as the prefix. Similar to the task handler name as well. That means using `this.constructor.name`.

CMCDragonkai:
> 3. The tasks themselves are technically emphemiral so we could remove all refreshBuckets tasks during nodeManager.stop(). Minor detail, won't really make a difference to operation.

This is because refresh bucket tasks are only used to maintain the network liveness during non-busy times.

Therefore once we stop PK and start it again after some time, there's no need to respect the existing delays.

Therefore, we don't care about the existing refresh tasks in the database.

Should we delete this state then? I think we should do so, to explicitly indicate they are ephemeral. Furthermore if the task manager starts again, and our schema changed, we may end up with phantom tasks running. It's better to be safe here and just delete the refresh bucket tasks since they are meant to be ephemeral.

CMCDragonkai:
A similar situation occurs with the `setNode` tasks. Because the pending partition is in-memory, as soon as we stop and start PK, we will have an empty pending partition. This should mean that we should also delete all the `setNode` tasks.

Same with the `pingAndSetNode` task. That should also be deleted at the end.

This means nodes domain doesn't have any persistent tasks at all then.

CMCDragonkai:
Ok if the `syncNodeGraph` goes into `NodeManager`, then `NodeManager` centralises all the background task management.

Ensure that all task handlers are arrow properties. Group them together.

Then ensure that all methods relating to the task handlers and associated tasks are grouped together as well. These methods should probably be all protected methods, and therefore should near the bottom to the class.

CMCDragonkai:
> Other aspects to consider. The Kademlia findNode operation in the NodeConnectionManager is considered a single operation but is in essence a priority queue search for the target node. We can consider splitting this up into a compound task where each step in the search can be it's own task within the process. This would apply to the refreshBucket operation since that is doing a findNode as well.

I think this requires the same semaphore that will be used for garbage collection. So I need to get the `Semaphore` into `js-async-locks`.

Nothing this does requires to be in the background.

tests/client/service/nodesAdd.test.ts
```diff
@@ -128,16 +130,17 @@ describe('nodesAdd', () => {
     });
   });
   afterEach(async () => {
+    await taskManager.startProcessing();
```
CMCDragonkai: This should be `stopProcessing()`. Check all the rest as well.

CMCDragonkai:
The new `js-async-locks` at `3.2.0` when released will have the `Semaphore`. Use it like `new Semaphore(limit)`.

In this case, I guess you do not need a timeout on the semaphore locking operation.

tegefaulkes:
About half of the `setNode` garbage collection refactor is done. the `garbageCollectBucket` and `addPendingNode` functions are done. Just need to update `setNode` to schedule a garbage collection task and add the pending node. I need to check if the blocking flag for `setNode` is needed anymore. We can update `nodeManager.getNode` to check the pending nodes as well.

CMCDragonkai:
Once you have finished Discovery integration with TaskManager, we go ahead and merge this. The #328 will be low priority atm, so we put that till later.

I want to have say by Tuesday/Wed that we go back to #441, fix the problems in the staging pipeline, and in particular making our staging CI/CD working again. That way our staging should build, and we should start getting built binaries for Linux.

src/nodes/NodeManager.ts
tegefaulkes: Garbage collection has been updated. As a result, `setNode` is no longer has optional blocking or abortion with signal.

src/nodes/NodeGraph.ts
```diff
@@ -151,6 +151,15 @@ class NodeGraph {
     return space;
   }
 
+  @ready(new nodesErrors.ErrorNodeGraphNotRunning())
+  public async lockBucket(bucketIndex: number, tran: DBTransaction) {
```
CMCDragonkai, line 155: Should this be `protected`?

src/tasks/TaskManager.ts
```diff
@@ -1148,6 +1148,7 @@ class TaskManager {
     this.logger.debug(`Requeued Task ${taskIdEncoded}`);
   }
 
+  @ready(new tasksErrors.ErrorTaskManagerNotRunning())
```
CMCDragonkai: This should not be needed. It is a protected function so you don't need a `ready` decorator.

src/nodes/utils.ts
```diff
@@ -310,6 +313,14 @@ function generateRandomNodeIdForBucket(
   return xOrNodeId(nodeId, randomDistanceForBucket);
 }
 
+function isConnectionError(e): boolean {
```
CMCDragonkai, line 326: Need a comment here what this is for. What constitutes a connection error and why is it being checked here?

src/nodes/NodeManager.ts
```diff
@@ -390,13 +431,13 @@ class NodeManager {
     );
   }
 
+  // FIXME: make cancelable
```
CMCDragonkai: Use `timedCancellable`. Make sure to connect it to the task deadline.

CMCDragonkai (changes requested):

src/bin/errors.ts
```diff
@@ -49,6 +49,12 @@ class ErrorCLIPolykeyAgentProcess<T> extends ErrorCLI<T> {
   exitCode = sysexits.OSERR;
 }
 
+class ErrorCLIPolykeyAsynchronousDeadlock<T> extends ErrorCLI<T> {
```
CMCDragonkai: This should not be a CLI error. This should be part of all the other errors like UnhandledPromiseRejection. We could have them in ExitHandlers.

Actually all of these can be in bin/errors.ts. But it needs to be separate from ErrorCLI, also we should change ErrorCLI to ErrorBin.

src/client/service/nodesAdd.ts
```diff
@@ -79,7 +79,6 @@ function nodesAdd({
             host,
             port,
           } as NodeAddress,
-          true,
```
CMCDragonkai, line 82: What is the impact of this change?

CMCDragonkai (changes requested):

src/nodes/NodeManager.ts
tegefaulkes: I've converted functions called inside of the task handlers to be a `timedCancellable`. So far it's working fine. More work needs to be done  with converting methods to `timedCancellable` but I think this should be left to a separate PR that focuses on implementing deadlines and cancellablility where needed.

src/nodes/NodeManager.ts
tegefaulkes: This has been fixed.

src/nodes/NodeManager.ts
tegefaulkes: This has been addressed for all handlers.

tests/client/service/nodesAdd.test.ts
tegefaulkes: This has been checked for all tests and fixed.

src/nodes/NodeGraph.ts
tegefaulkes: It's getting used inside of `NodeManager` to lock the bucket during `setNode` and `garbageCollectBucket`.

src/tasks/TaskManager.ts
tegefaulkes: It's protected here but it's callable at any time with `task.cancel()` or `task.promise().cancel()`. I added it while debugging but it's not really needed.

tegefaulkes: I've removed it.

tegefaulkes:
Personal reference of TODOs.
todos:
- [x] review error usage
- [x] review TODOs
- [x] review the removal of blocking from `setNode` and check how it affects it's use.
- [x] clean up existing tasks when stopping a domain. Only if tasks are ephemeral - This is because NodeManager tasks are ephemeral and therefore can be removed without problems
- [x] Move `syncNodeGraph` into `NodeManager`
- ~Consider making  `getGlobalClosestNodes`  concurrent.~

notes:
- wider timedCancellable work needs to be done across GRPC and NodeConnectionManager. This should be it's own PR?

src/client/service/nodesAdd.ts
tegefaulkes: Blocking has been added back in. This change has been reverted as a result.

src/nodes/NodeManager.ts
tegefaulkes: I have extracted this into the utils.

tegefaulkes:
There is a build error coming from a dependency.

```
[faulkes@matrix-vostro-5410-1:~/matrixCode/polykey/js-polykey]$ npm run build

> polykey@1.0.0 build
> shx rm -rf ./dist && tsc -p ./tsconfig.build.json

npm ls ts-custom-error
node_modules/ts-custom-error/dist/custom-error.d.ts:17:151 - error TS2344: Type 'Properties' does not satisfy the constraint 'CustomErrorProperties'.

17 export declare function customErrorFactory<Properties>(fn: CustomErrorFunction<Properties>, parent?: GenericErrorConstructor): CustomErrorConstructor<Properties>;
                                                                                                                                                         ~~~~~~~~~~

  node_modules/ts-custom-error/dist/custom-error.d.ts:17:44
    17 export declare function customErrorFactory<Properties>(fn: CustomErrorFunction<Properties>, parent?: GenericErrorConstructor): CustomErrorConstructor<Properties>;
                                                  ~~~~~~~~~~
    This type parameter might need an `extends CustomErrorProperties` constraint.


Found 1 error in node_modules/ts-custom-error/dist/custom-error.d.ts:17


[faulkes@matrix-vostro-5410-1:~/matrixCode/polykey/js-polykey]$ npm ls ts-custom-error
polykey@1.0.0 /home/faulkes/matrixCode/polykey/js-polykey
└─┬ @matrixai/errors@1.1.3
  └── ts-custom-error@3.2.0

```

CMCDragonkai:
Might be a regression. Check if staging builds too. It could be due to changes in the typescript version too. You might want to bisect.

tegefaulkes:
It's happening on the staging branch as well.
```
[faulkes@matrix-vostro-5410-1:~/matrixCode/polykey/js-polykey]$ npm run build

> polykey@1.0.0 build
> shx rm -rf ./dist && tsc -p ./tsconfig.build.json

node_modules/ts-custom-error/dist/custom-error.d.ts:17:151 - error TS2344: Type 'Properties' does not satisfy the constraint 'CustomErrorProperties'.

17 export declare function customErrorFactory<Properties>(fn: CustomErrorFunction<Properties>, parent?: GenericErrorConstructor): CustomErrorConstructor<Properties>;
                                                                                                                                                         ~~~~~~~~~~

  node_modules/ts-custom-error/dist/custom-error.d.ts:17:44
    17 export declare function customErrorFactory<Properties>(fn: CustomErrorFunction<Properties>, parent?: GenericErrorConstructor): CustomErrorConstructor<Properties>;
                                                  ~~~~~~~~~~
    This type parameter might need an `extends CustomErrorProperties` constraint.


Found 1 error in node_modules/ts-custom-error/dist/custom-error.d.ts:17

```

CMCDragonkai:
Can you explain how the timeout system works in the `setNode`?

Right now, we have decided that each node connection establishment only is given 1 second to establish a connection.

The entire garbage collection is not given a timeout to complete.

It's only cancellable. Meaning that there's no deadline.

However if each node connection is given 1 second to establish, and the maximum possible number of nodes is a bucket is 20. Then at most there will be 20 seconds to run for a `setNode` GC operation.

I was confused by the fact that you separated out GC of nodes from the `setNode`. It sounds more like that `setNode` is a unit operation, that does GC under the hood and nothing else does GC of the node buckets.

Therefore unlike the usual timeout mechanism provided by `timedCancellable`, this pattern is a inside-out timeout, rather than outside-in timeout.

Outside-in is where the entire operation is given a timeout to complete.

Inside-out is that the internal operation has timeouts, and they build up to a larger timeout.

If we wanted to preserve outside-in timeouts, that would mean that a 20s `TaskDeadline` is specified, and that then needs to be divided into units for each individual node connection to attempt.

We wouldn't want 1 node connection to inherit the entire 20s, because the purpose of this operation is to find out which connections are live. So we need to have a smaller timeout anyway for each node.

So inside-out is easier to do here. And thus there's no `TaskDeadline`.

Do note that this would not match the pattern of functionality in other parts of the PK app though.

CMCDragonkai:
Ok after some discussion this is how we can do things.

1. Outside-In constraint - this is where the timeout is specified externally and passed in as a constraint
2. Inside-Out constraint - this is where the timeout is constructed internally (but may be parameterised), but the ultimate timeout is dependent on some calculations and other dimensions of the algorithm/data structure

When the CLI command calls the GRPC client and specifies a deadline that is passed to the GRPC service handler, this is an example of a "Outside-In" constraint. The intention is for this deadline to specify how long the RPC call should take, and if it takes longer than that, the call should be cancelled.

Example: `pk nodes set ...`. This is a CLI command that sets a new node. It translates to doing a GRPC call that has a deadline that defaults to `1000` ms. This means it is expected that setting a node should take 1000ms or less.

On the GRPC service handler, it should take the deadline turn it into a `Timer` object and pass that down to `setNode(ctx: ContextTimed)`. This timer should only control how long it takes before `setNode`'s promise resolves.

Now under the hood of `setNode`, it is basically asking nodes if they are alive. But it can do this in the background. This "are you alive?" question can be parameterised by an additional `pingTimeout` parameter.

If we go back to the CLI call, we would expect something like `pk nodes set --ping-timeout=1000`, where one can control how long it should take to ask the "are you alive?" question.

This parameter would propagate down to `setNode(pingTimeout, ctx: ContextTimed)`.

So now there are 2 timeouts. 1 for the are you alive question, and another for the time to finish the promise.

Inside `setNode`, the `ctx.timer` and `ctx.signal` only has effect if the blocking is true. If the blocking is not true, the scheduling of the garbage collection occurs in the background and therefore the whole thing would finish very quickly.

If the blocking is true, then the `ctx.timer` does have an effect.

The `pingTimeout` however will always be passed down to the proxy connection establishment, and is the independent amount of time that is allowed for the "are you alive?" question.

This is a Inside-Out constraint because the total amount of time taken for garbage collection is dependent on the `pingTimeout` multiplied by the total max nodes in a bucket divided by the parallel connection establishment size.

CMCDragonkai:
The end goal is that all of our GRPC service handlers should then process deadlines into `Timer` objects and pass them into `ContextTimed` contexts.

CMCDragonkai:
In the future, when using the semaphore for locking things and where there's a presence of signal cancellation, you want to race both the locking operation and the signal abortion.

See the `TaskManager` as examples of this implementation. There's an existing issue for making this easier in https://github.com/MatrixAI/js-async-locks/issues/21.

For now, you're creating a the semaphore internally and you are waiting for the lock before checking the signal. But in the future I believe you should really be locking and passing a context directly. This will make the code simpler and more consistent.

In the mean time, please add a comment explaining why you have decided to wait for the lock before checking the signal and then locking, and **why you didn't just lock semaphore while waiting for the abortion**.

src/nodes/NodeManager.ts
```diff
@@ -559,183 +771,251 @@ class NodeManager {
...
+   * from each seed node and add them to this database
+   * Establish a proxy connection to each node before adding it
+   * By default this operation is blocking, set `block` to false to make it
+   * non-blocking
+   */
+  public syncNodeGraph(
+    block?: boolean,
+    ctx?: Partial<ContextTimed>,
+  ): PromiseCancellable<void>;
+  @ready(new nodesErrors.ErrorNodeConnectionManagerNotRunning())
```
CMCDragonkai: This changes to `ErrorNodeManagerNotRunning`.

src/nodes/NodeConnectionManager.ts
```diff
@@ -426,24 +419,26 @@ class NodeConnectionManager {
    * port).
    * @param targetNodeId ID of the node attempting to be found (i.e. attempting
    * to find its IP address and port)
-   * @param timer Connection timeout timer
-   * @param options
+   * @param ignoreRecentOffline skips nodes that are within their backoff period
+   * @param ctx
    * @returns whether the target node was located in the process
    */
+  public getClosestGlobalNodes(
```
CMCDragonkai, line 432: This function is not doing the operation concurrently. This never got done concurrently.

src/nodes/NodeManager.ts
```diff
@@ -559,183 +771,251 @@ class NodeManager {
...
+            throw e;
+          }
+        }
+        this.logger.debug(
+          `Updating refreshBucket task for bucket ${bucketIndex}`,
+        );
+      } else {
+        // These are extra, so we cancel them
+        // TODO: make error
+        task.cancel(Error('TMP, cancel extra tasks'));
```
CMCDragonkai: This should be it's own error class.

I would also change the async generator to speicfically deal with the first item, then all subsequent items.

tests/network/Proxy.test.ts
```diff
@@ -311,16 +312,16 @@ describe(Proxy.name, () => {
...
       proxy.openConnectionForward(
         nodeIdABC,
         localHost,
         utpSocketHangPort as Port,
-        timer,
+        { timer },
       ),
     ).rejects.toThrow(networkErrors.ErrorConnectionStartTimeout);
-    timerStop(timer);
+    timer.cancel('clean up');
```
CMCDragonkai: We are going to change this in a later PR, so this can be rollbacked to how it was. Integrating timers all over the codebase.

CMCDragonkai:
I've fixed the build errors in staging via https://github.com/MatrixAI/Polykey/pull/411/commits/54f0c2b671229361e0f7b75c21a1547655cf7ae4.

I've also updated js-errors to 1.1.4 to force the usage of 3.2.2 and above for ts-custom-error. So upon updating to it, it will force 3.2.2, which should be important for all transitive dependencies.

Upon rebasing, it should fix the build errors.

src/nodes/NodeManager.ts
tegefaulkes: This has been fixed.

src/nodes/NodeConnectionManager.ts
tegefaulkes: Is this something we want to fix in this PR?

src/nodes/NodeManager.ts
tegefaulkes: I would, but it's not the first task we're interested in, it's the first task that is still running, queued or scheduled. The first one could throw `ErrorTaskMissing` and that wouldn't count as a found task.

I've updated it to use a boolean flag to make it more clear. using a counter wasn't needed.

tegefaulkes:
After the timedCancellable changes i'm starting to see connection errors cropping up when running a bunch of tests at once. I think all of the tests pass when running test files one at a time with `--runInBand`. 

These errors may be due to connections failing/closing from the client side. I'm not sure we're fulling testing these conditions. I haven't fully explored the problem but my first guess is that the service handlers are not properly handling connection issues coming from the client side. I haven't seen these kinds of connection errors before so it's also possible that the client side is not handling it properly either. 

We should address this problem in more detail during the wider `timedCancellable` changes. I should make a new issue for it.

tests/network/Proxy.test.ts
tegefaulkes: I've reverted this. This change is left to the wider `timedCancellable` changes PR.

src/nodes/NodeConnectionManager.ts
tegefaulkes: I think I should make a new issue for this. This PR is getting too large and we should merge it sooner than later.

src/nodes/NodeManager.ts
CMCDragonkai: The custom error here can be a special task error that throws into the task handler for cancellation.

Can you create a error here under `src/nodes/errors.ts` for this specific purpose, then document it similar to `ErrorTaskRetry` and `ErrorTaskStop`. See `src/tasks/errors.ts`. These errors are "signals".

CMCDragonkai:
What do you mean by connections failing/closing on client side. Are you talking about proxy connections or GRPC connections or node connections?

src/nodes/NodeConnectionManager.ts
CMCDragonkai: Depends on how easy it is to do it. And if it can use the same pattern we are using for refresh buckets, and the semaphore... etc.

tegefaulkes:
The erros look like
```
Error: Socket is closed
    at UTP.Object.<anonymous>.UTP.send (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/index.js:121:50)
    at /home/faulkes/matrixCode/polykey/js-polykey/src/utils/utils.ts:178:9
    at new Promise (<anonymous>)
    at UTP.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/src/utils/utils.ts:163:12)
    at constructor_.send (/home/faulkes/matrixCode/polykey/js-polykey/src/network/Connection.ts:79:11)
    at Timeout.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/src/network/ConnectionReverse.ts:147:20)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)

```

```
Error: Unhandled error. (Error: UTP_ECONNRESET
    at createUTPError (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/lib/connection.js:238:15)
    at Connection.Object.<anonymous>.Connection._onerror (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/lib/connection.js:175:16) {
  code: 'UTP_ECONNRESET',
  errno: 1
})
    at new NodeError (node:internal/errors:372:5)
    at Connection.emit (node:events:516:17)
    at Connection.Object.<anonymous>.Connection._onclose (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/lib/connection.js:169:25)
```

This one is trying to bind socket `55555` during a `beforeAll`. Not sure why it's an issue now.

```
Error: address already in use
    at UTP.Object.<anonymous>.UTP.bind (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/index.js:198:13)
    at /home/faulkes/matrixCode/polykey/js-polykey/src/utils/utils.ts:178:9
    at new Promise (<anonymous>)
    at UTP.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/src/utils/utils.ts:163:12)
    at Object.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/tests/nodes/NodeManager.test.ts:83:11)
```

CMCDragonkai:
There are tests in `tests/network/Proxy.test.ts` that change the timeout default parameters and test timed out proxy connections.

But there are no tests currently testing overriding the timeout at `openConnectionForward`. You may need to add additional tests there and see how that timeout conflicts or works with other timeouts that already exist in the `Proxy.test.ts`. There may be a minimum threshold required by punching and keep alive cycles.

CMCDragonkai:
There is actually 1 test doing `openConnectionForward` with a overridden timer. Search for it, and see if that can help us debug the above.

src/nodes/NodeManager.ts
tegefaulkes: The `timedCancellable` conversions are done fr this PR. `setNode` was updated to be a `timedCancellable` but the timer is for the overall task completeion. The `pingNode`s as part of the garbage collection are started with their own timers set by the `pingTimeout` parameter.

Further `timedCancellable` conversion will be done in a later PR.

src/nodes/NodeConnectionManager.ts
tegefaulkes: New issue created at #448

tegefaulkes:
These connection bugs will be addressed in staging after merge.

tegefaulkes:
The connection bugs will be addressed in staging. I'm creating new issues to track loose ends.

tegefaulkes:
New issue for dealing with connection errors at #450

tegefaulkes:
re-based on  `origin/staging`

CMCDragonkai:
Since the cancellation of the tasks that are meant to be singleton. You should describe them as a singleton symbol. Because this is an implementation detail. Look at `TaskManager.ts`, I've used symbols as encapsulated cancellation reasons when they aren't meant to be caught externally.

Put a comment about this too.

```
  const abortSchedulingLoopReason = Symbol('abort scheduling loop reason');
  const abortQueuingLoopReason = Symbol('abort queuing loop reason');
```

So we can do:

```
  const abortSingletonTaskReason = Symbol('abort singleton task reason');
```

tegefaulkes:
I'm adding this as well for when we stop tasks during `nodeManager.stop();
```
const stoppingNodeManagerReason = Symbol('stopping NodeManager reason');
````

**Comment by tegefaulkes** [comment]
Shouldn't `TaskManager` be extending an interface like?
```
interface TaskManager extends CreateDestroyStartStop {}
```

Every other `CreateDestroyStartStop` class is doing this. What happens if we don't do this?

**Comment by CMCDragonkai** [comment]
> I'm adding this as well for when we stop tasks during `nodeManager.stop();
> 
> ```
> const stoppingNodeManagerReason = Symbol('stopping NodeManager reason');
> ```

I think you should use `abortEphemeralTaskReason`. Keep it the same pattern.

**Comment by CMCDragonkai** [comment]
> ```
> interface TaskManager extends CreateDestroyStartStop {}
> ```

Yea that was missed, add it in.

**Comment by tegefaulkes** [comment]
Both of these have been fixed.

File: src/tasks/TaskManager.ts [hunk 18]
```diff
@@ -245,7 +244,6 @@ class TaskManager {
    * Stop the active tasks
    * This call is idempotent
    */
-  @ready(new tasksErrors.ErrorTaskManagerNotRunning(), false, ['stopping'])
```
CMCDragonkai, line 247: This is confirmed to work at any status. So no need for this.

src/tasks/TaskManager.ts
```diff
@@ -236,7 +236,6 @@ class TaskManager {
    * Stop the scheduling and queuing loop
    * This call is idempotent
    */
-  @ready(new tasksErrors.ErrorTaskManagerNotRunning(), false, ['stopping'])
```
CMCDragonkai, line 238: This is confirmed to work at any status. So no need for this.

tegefaulkes:
Here's a run down of the time take to do one of the NAT tests

```
[faulkes@matrix-vostro-5410-1:~/matrixCode/polykey/js-polykey]$ sudo npm run test -- tests/nat/endpointDependentNAT.test.ts 

> polykey@1.0.0 test
> jest "tests/nat/endpointDependentNAT.test.ts"

Determining test suites to run...
GLOBAL SETUP
Global Data Dir: /tmp/polykey-test-global-H5zUdU
  console.time
    setting up env: 12038 ms

      at Object.<anonymous> (tests/nat/endpointDependentNAT.test.ts:194:15)

  console.time
    executing first ping: 9579 ms

      at Object.<anonymous> (tests/nat/endpointDependentNAT.test.ts:215:15)

  console.time
    starting second ping: 4911 ms

      at Object.<anonymous> (tests/nat/endpointDependentNAT.test.ts:238:15)

  console.time
    starting teardown: 2445 ms

      at Object.<anonymous> (tests/nat/endpointDependentNAT.test.ts:246:15)

 PASS  tests/nat/endpointDependentNAT.test.ts (40.502 s)
  endpoint dependent NAT traversal
    ✓ node1 behind EDM NAT cannot connect to node2 behind EDM NAT (29022 ms)
    ○ skipped node1 behind EDM NAT connects to node2
    ○ skipped node1 connects to node2 behind EDM NAT
    ○ skipped node1 behind EDM NAT cannot connect to node2 behind EIM NAT

```

Keep in mind that each ping operation is doing and exec that compiles the code, runs the command with NS enter and waits for the full command to finish. I'd expect ping to take 2 seconds of that to fail. I've confirmed that the timer is set up properly with the 2 seconds. 

5 seconds to complete the ping command seems on the edge of reasonable. But 10 seconds seems way too long. 

For now I'm just expanding timeouts. Digging deeper into this can be left to the wider `timedCancellable` changes.

tegefaulkes:
Actually, increasing the timeouts is sweeping the problem under the rug. I'll ignore the NAT domain for this PR, Fixing this will be addressed by the `timedCancellable` changes PR.

tegefaulkes:
The `Discovery` domain job keeps failing. Consistently it's a timeout in `afterEach()` during clean up. I can't tell why, I checked locally and clean up is reasonably quick. I'm going to leave this to the next `Discovery` tasks integration PR.

CMCDragonkai:
When merging this to staging, disable the discovery domain too and other tests so we can get a working build and deployment.

tegefaulkes:
Found a small bug to watch out for. In some cases where we pop something out of an array it's possibly undefined if the array was empty. In some cases we were using '!' to throw an error when we expected it to be defined. But during runtime the error was never thrown and we ended up with undefined variables in weird places.

```ts
const nodeIds: Array<NodeId> = []
const nodeId = nodeIds.pop()!;
// Typescript thinks this is defined but it is `undefined`. 
nodeId.equals()
```

This resulted with errors in weird places. As to why we're seing the error now? I made `getRemoteNodeClosestNodes` return `[]` when it times out instead of an error. This made the failing parts of the code see an empty array for once.

tegefaulkes:
I've fixed up what I can for now. I'm going to start squashing commits now.

tegefaulkes:
Ok, this is good to merge now. There are 3 test domains that need an eyeball still.
- Discovery
- `tests/nat`
- Nodes - I think I got all the bugs in this but waiting for CI to finish will take a while.

src/bin/utils/ExitHandlers.ts
```diff
@@ -83,6 +52,64 @@ class ExitHandlers {
...
+      binUtils.outputFormatter({
+        type: this._errFormat,
+        data: e,
+      }),
+    );
+    process.exitCode = error.exitCode;
+    // Fail fast pattern
+    process.exit();
+  };
+  protected deadlockHandler = async () => {
```
CMCDragonkai: This requires a newline in between.

src/bin/utils/ExitHandlers.ts
```diff
@@ -83,6 +52,64 @@ class ExitHandlers {
...
+      binUtils.outputFormatter({
+        type: this._errFormat,
+        data: e,
+      }),
+    );
+    process.exitCode = error.exitCode;
+    // Fail fast pattern
+    process.exit();
+  };
+  /**
```
CMCDragonkai: Newline needed.

CMCDragonkai: If properties have doc block comments, separate a newline around them.

Issue #450, opened 2022-09-21
Apply `timedCancellable` across the board to control how long side-effects are allowed to complete


tegefaulkes:
### Specification

With the application of `timedCancellable` to functions within the nodes domain. We have encountered new connection errors randomly during tests. These error started after applying a deadline of 1500 ms to `pingNode` usage. We suspect the problem is coming from `openConnectionForward` after setting such a short deadline. These may be bugs that existed previously that was never revealed by the current testing. Currently there is 1 `NodeConnectionManager` test where we override the deadline for `openConnectionForward`.

The occur randomly and are as follows. the can be re-created by running all of the nodes domain tests concurrently.

<details>

```
Error: Socket is closed
    at UTP.Object.<anonymous>.UTP.send (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/index.js:121:50)
    at /home/faulkes/matrixCode/polykey/js-polykey/src/utils/utils.ts:178:9
    at new Promise (<anonymous>)
    at UTP.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/src/utils/utils.ts:163:12)
    at constructor_.send (/home/faulkes/matrixCode/polykey/js-polykey/src/network/Connection.ts:79:11)
    at Timeout.<anonymous> (/home/faulkes/matrixCode/polykey/js-polykey/src/network/ConnectionReverse.ts:147:20)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)

Error: Unhandled error. (Error: UTP_ECONNRESET
    at createUTPError (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/lib/connection.js:238:15)
    at Connection.Object.<anonymous>.Connection._onerror (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/lib/connection.js:175:16) {
  code: 'UTP_ECONNRESET',
  errno: 1
})
    at new NodeError (node:internal/errors:372:5)
    at Connection.emit (node:events:516:17)
    at Connection.Object.<anonymous>.Connection._onclose (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/lib/connection.js:169:25)

Error: address already in use
    at UTP.Object.<anonymous>.UTP.bind (/home/faulkes/matrixCode/polykey/js-polykey/node_modules/utp-native/index.js:198:13)
[...]

Issue #327, opened 2022-02-08
Sigchain Class API should provide paginated ordered claims by returning Array-POJO and indexed access


CMCDragonkai:
### Specification

The `Sigchain` API requires some rework as discovered here: https://github.com/MatrixAI/js-polykey/pull/310#discussion_r801248769

1. The `ChainData` and `ChainDataEncoded` types are record POJOs, which are not the right data type we should be returning to callers. Record POJOs do not preserve order, and when returning chain data, we would want to have the ability to iterate over the data, and it should be in the same order as the chain data (as ordered by the claim id).
2. When fetching a collection of resources in JS, always return an `Array`. Record POJOs are not ordered. Alternatively to preserve both order and index access, a `Map` can be used. HOWEVER, `Array` and POJOs are pure data. The `Map` and `Set` are not. This is why by convention, we only use `Map` as a persistent rich construct, and when exchanging data between systems, we use arrays and POJOs.
3. This means we need `Sigchain.getChain` and `Sigchain.getChainEncoded`. Both methods should be returning the type `Array<Claim>` and `Array<ClaimEncoded>` respectively. The encoded version is the base64ed JWT, but usually in our own application we would be working off the `Claim` structure. At the same time there should be **obvious** conversion functions from `Claim` to `ClaimEncoded` and back.
4. These 2 functions should expose `seek` and `limit` parameters to enable cursor pagination. Note that `seek` needs to be type of the seek key. The seek key of `Sigchain` is the `ClaimId`. Which we know to be based on `IdSortable` which itself uses timestamps internally. Both parameters are optional. One should be able to synthetically construct a `ClaimId` to represent a point in time cursor as well. This is a secondary priority, do this after the main design works.
5. For indexed access, we would not expect the user to acquire a collection, and then query their collection. Instead we would expect the user to directly call a `getClaim` which you can pass the `ClaimId` to acquire the `Claim` itself. If they want an encoded version, we can expose `getClaimEncoded`.
6. The `ChainData` and `ChainDataEncoded` may no longer be necessary as types. Just use `Array<Claim>` and `Array<ClaimEncoded>`.  **However** it is essential that one can iterate over these 2 arrays, and be able to use extractors/transformers/converters to be able to acquire the `ClaimId`. This means the `ClaimId` must always be acquirable given a `Claim` or `ClaimEncoded`, ideally efficiently.
[...]

PR #451, opened 2022-09-26
Integrating `TaskManager` into `Discovery`

tegefaulkes:
### Description

This PR focuses on updating the `Discovery` domain to make use of the `TaskManager`.

### Issues Fixed

* Related #445 

### Tasks

- [x] 1. Update the Discovery queue to make use of the `TaskManager` queuing.
- [x] 2. Remove any tests that are no longer required, I.E. anything testing the async queue aspect of the `Discovery` since it's handled via `TaskManager` now.
- ~ 5. re-enable Discovery tests in the generated CI tests ~ - The problem is ultimately a connection timeout, not addressed here.
- [x] 3. Update any tests that need it.
- [x] 4. Add any new tests as needed.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
We can limit the scope of discovery operations because we only care about:

1. Our own gestalt
2. Any gestalt we trust
3. Any gestalt that the user triggers manually

We don't need to crawl transitive gestalts (friends of friends...).

CMCDragonkai:
Discovery is likely to be push and pull at the same time. But right now everything in PK is pull-based cause it's easier to implement.

So we can do something simple.

At most we might have 1000 gestalts trusted, and each gestalt is likely a handful of nodes, say 5. Then at the most have 5000 nodes to check in with.

We can say 1 day TTL to revisit each node.

When the push-based gossip protocol is available, we can actually increase the TTL, and that TTL acts like the last-resort in the sense that if no update has occurred from there we can check in.

It could be even simpler of we can expect that any given node of a gestalt should be up to date with their own gestalt's information.

At this point in time, let's make it configurable and see how it goes.

The goal here for this PR is not to make our discovery perfect UI/UX-wise, but to simply integrate the task manager system, simplify the code in discovery so it becomes stateless (with respect to task management), and also incorporate `timedCancellable` where possible.

CMCDragonkai:
So `timedCancellable` for #450 will be in another PR, but this starts it out for the discovery domain at least.

tegefaulkes:
General changes so far:
- The parts of discovery that handles queuing has been removed. This includes the 'plug' promises, database levels and the discovery ID generator.  
- The logic of the queue where it handled the vertex was separated out into a new method. This logic included connecting to a vertex, getting the claims' data, iterating over the claims, verifying the claims and adding them to the gestaltGraph, and adding the new vertices to the list.  
- Adding a vertex to the list now consists of creating a singleton task. each task focuses on processing a single vertex.   
- I tried to cancel all tasks on `discovery.destroy()` but it would require the `taskManager` to be started at the time since we need to iterate over active tasks at the time. For now cancel the tasks when starting with fresh.  
- To avoid re-trying tasks, a `visitedVertices` set is used, this hasn't been updated. A more complex mechanism is need for this that has persistence. I think this will be left with the more complex discovery changes later.  
- Discovery tests have been updated. none removed or added so far. The change was just using `taskManager`.

tegefaulkes:
Re-based on staging and a quick fix. Should pass CI now.

tegefaulkes:
The Discovery failures is due to `proxy.stop()` taking a very long time to complete. I could dig into it, but ultimately the problem will be addressed by the cancel-ability changes PR after this one. For now i'll keep the discovery tests disabled.

```
console.time
    proxy.stop(): 19142 ms
      at Object.<anonymous> (tests/discovery/Discovery.test.ts:240:13)
```

src/discovery/Discovery.ts
```diff
@@ -81,14 +89,33 @@ class Discovery {
...
+    try {
+      await this.processVertex(vertex, ctx);
+    } catch (e) {
+      if (e instanceof tasksErrors.ErrorTaskStop) {
+        // We need to recreate the task for the vertex
+        await this.startDiscoveryForVertex(vertex);
+        return;
+      }
+      // Aborting a duplicate task is not an error
+      if (e === abortSingletonTaskReason) return;
```
tegefaulkes, line 124: We should be ignoring `discoveryDestroyedTaskReason` as well.

src/discovery/Discovery.ts
```diff
@@ -123,36 +153,28 @@ class Discovery {
...
-    await this.discoveryProcess;
+    this.taskManager.deregisterHandler(this.discoverVertexHandlerId);
     this.logger.info(`Stopped ${this.constructor.name}`);
   }
 
   public async destroy(): Promise<void> {
     this.logger.info(`Destroying ${this.constructor.name}`);
-    await this.db.clear(this.discoveryDbPath);
     this.logger.info(`Destroyed ${this.constructor.name}`);
   }
```
tegefaulkes: `discoveryDestroyedTaskReason` is used to cancel a task when destroying. Currently it's only used when starting with `fresh: true`.  Doing it during `discovery.destroy` requires that the `TaskManager` is running when destroying. I haven't made this change yet and I'm not sure it's worth making it.

I'm leaning towards needing it.

src/discovery/Discovery.ts
```diff
@@ -175,299 +197,291 @@ class Discovery {
...
-      nextDiscoveryQueueId = IdInternal.fromBuffer<DiscoveryQueueId>(key);
+    for await (const _ of this.taskManager.getTasks('asc', true, [
+      this.constructor.name,
+      this.discoverVertexHandlerId,
+    ])) {
+      return false;
     }
-    return nextDiscoveryQueueId == null;
+    return true;
   }
```
tegefaulkes: Not used anywhere, might remove it?

src/discovery/Discovery.ts
```diff
@@ -175,299 +197,291 @@ class Discovery {
...
-    gestaltKey: GestaltKey,
-  ): Promise<void> {
-    await this.db.withTransactionF(async (tran) => {
-      const valueIterator = tran.iterator<GestaltKey>(
-        this.discoveryQueueDbPath,
-        { valueAsBuffer: false },
+  protected async startDiscoveryForVertex(
+    vertex: GestaltKey,
+    tran?: DBTransaction,
+  ) {
```
tegefaulkes: Calling it `startDiscoveryForVertex` seems bad, maybe `queueDiscoveryForVertex`?

src/nodes/NodeManager.ts
```diff
@@ -942,9 +942,9 @@ class NodeManager {
             throw e;
           }
         }
-        this.logger.debug(
-          `Updating refreshBucket task for bucket ${bucketIndex}`,
-        );
+        // This.logger.debug(
+        //   `Updating refreshBucket task for bucket ${bucketIndex}`,
+        // );
```
tegefaulkes: This made debugging annoying, This should be removed or reverted.

src/nodes/NodeManager.ts
```diff
@@ -954,9 +954,9 @@ class NodeManager {
       }
     }
     if (!existingTask) {
-      this.logger.debug(
-        `No refreshBucket task for bucket ${bucketIndex}, new one was created`,
-      );
+      // This.logger.debug(
+      //   `No refreshBucket task for bucket ${bucketIndex}, new one was created`,
+      // );
```
tegefaulkes: Same as above, should be removed or reverted.

tests/client/service/gestaltsDiscoveryByNode.test.ts
```diff
@@ -139,16 +139,17 @@ describe('gestaltsDiscoveryByNode', () => {
...
       db,
       keyManager,
       gestaltGraph,
       identitiesManager,
       nodeManager,
       sigchain,
+      taskManager,
       logger,
     });
+    await taskManager.start();
```
tegefaulkes: This should be `taskManager.startProcessing`.

tests/client/service/gestaltsGestaltTrustByNode.test.ts
```diff
@@ -218,6 +218,7 @@ describe('gestaltsGestaltTrustByNode', () => {
       identitiesManager,
       nodeManager,
       sigchain,
+      taskManager,
       logger,
     });
```
tegefaulkes, line 222: This is missing `taskManager.startProcessing()`. Maybe processing isn't needed here, I should check it.

tegefaulkes (commented):
Some small changes and discussion needed.

src/nodes/NodeManager.ts
tegefaulkes: I'm going to revert these, it seems out of scope for this PR.

src/discovery/Discovery.ts
tegefaulkes: If we want to clear all the discovery tasks when destroying then we need `TaskManager` and `DB` to be running at the time. In `polykeyAgent.destroy` the `DB` is already started temporarily. So it's not a major change.

tegefaulkes: `discovery.destroy()` is now cancelling any existing discovery tasks.

src/discovery/Discovery.ts
tegefaulkes: I'll leave it, It may be useful.

tegefaulkes:
This should be ready to merge pending CI. @CMCDragonkai did you still want to review?

CMCDragonkai:
Yes give me a moment.

src/PolykeyAgent.ts
```diff
@@ -754,16 +755,19 @@ class PolykeyAgent {
     this.logger.info(`Destroying ${this.constructor.name}`);
     // DB needs to be running for dependent domains to properly clear state.
     await this.db.start();
+    // TaskManager needs to be running for Discovery to destroy
+    await this.taskManager.start({ lazy: true });
```
CMCDragonkai: This would only happen if we stop then destroy the domains... Right?

Usually when stopping we stop the db. Then if we want destroy, it still needs access to the db so it is started again for the destruction.

So the issue here is that upon destroying discovery, we have to destroy state including state the tasks system.

This doesn't affect nodes cause all tasks are already removed when the node manager is stopped?

src/discovery/types.ts
```diff
@@ -1,11 +0,0 @@
-import type { Opaque } from '../types';
-import type { Id } from '@matrixai/id';
-
-/**
- * Used to preserve order in the Discovery Queue.
- */
-type DiscoveryQueueId = Opaque<'DiscoveryQueueId', Id>;
-
-type DiscoveryQueueIdGenerator = () => DiscoveryQueueId;
```
CMCDragonkai, line 9: What happened to these types? Are they no longer needed or moved to `ids`?

tests/discovery/Discovery.test.ts
```diff
@@ -59,7 +60,16 @@ describe('Discovery', () => {
   let nodeA: PolykeyAgent;
   let nodeB: PolykeyAgent;
   let identityId: IdentityId;
+
+  const mockedRefreshBucket = jest.spyOn(
+    NodeManager.prototype,
+    'refreshBucket',
+  );
```
CMCDragonkai, line 67: What's the purpose of this in discovery testing?

src/discovery/Discovery.ts
```diff
@@ -17,22 +16,28 @@ import type Sigchain from '../sigchain/Sigchain';
 import type KeyManager from '../keys/KeyManager';
 import type { ClaimIdEncoded, Claim, ClaimLinkIdentity } from '../claims/types';
 import type { ChainData } from '../sigchain/types';
+import type TaskManager from '../tasks/TaskManager';
+import type { ContextTimed } from '../contexts/types';
+import type { PromiseCancellable } from '@matrixai/async-cancellable';
```
CMCDragonkai: Move this to the top. Imports from outside packages are above local imports.

src/discovery/Discovery.ts
```diff
@@ -81,14 +89,35 @@ class Discovery {
...
   protected identitiesManager: IdentitiesManager;
   protected nodeManager: NodeManager;
+  protected taskManager: TaskManager;
 
-  protected discoveryDbPath: LevelPath = [this.constructor.name];
-  protected discoveryQueueDbPath: LevelPath = [this.constructor.name, 'queue'];
-  protected discoveryQueueIdGenerator: DiscoveryQueueIdGenerator;
+  // Protected discoveryDbPath: LevelPath = [this.constructor.name];
+  // protected discoveryQueueDbPath: LevelPath = [this.constructor.name, 'queue'];
+  // protected discoveryQueueIdGenerator: DiscoveryQueueIdGenerator;
```
CMCDragonkai: Why do we still have this?

src/discovery/Discovery.ts
```diff
@@ -17,22 +16,28 @@ import type Sigchain from '../sigchain/Sigchain';
...
 import * as networkErrors from '../network/errors';
 import * as gestaltsUtils from '../gestalts/utils';
 import * as claimsUtils from '../claims/utils';
 import * as nodesUtils from '../nodes/utils';
-import { never, promise } from '../utils';
+import { never } from '../utils';
+import { context } from '../contexts/index';
+import TimedCancellable from '../contexts/decorators/timedCancellable';
+
+const abortSingletonTaskReason = Symbol('abort singleton task reason');
```
CMCDragonkai, line 42: Provide doc block comment about what this reason means.

src/discovery/Discovery.ts
```diff
@@ -175,299 +205,291 @@ class Discovery {
...
+          //  the same node/identity. Thus, we simply iterate over this chain
+          //  of cryptolinks.
+          // Now have the NodeInfo of this vertex
+          const vertexNodeInfo: NodeInfo = {
+            id: nodesUtils.encodeNodeId(nodeId),
+            chain: vertexChainData,
+          };
+          // Iterate over each of the claims in the chain (already verified)
+          // TODO: because we're iterating over keys in a record, I don't believe
+          //  that this will iterate in lexicographical order of keys. For now,
```
CMCDragonkai: Can you check this TODO? Is this still relevant?

CMCDragonkai (changes requested):
Initial review. Haven't gone into the discovery code yet.

src/discovery/Discovery.ts
CMCDragonkai: Comment seems incorrect. There is no more discovery queue. And what we are checking here is the presence of discovery tasks. Perhaps this method should be renamed? And comment rewritten.

src/discovery/Discovery.ts
CMCDragonkai: Why not re-use terminology from nodes? Should this be schedule...

CMCDragonkai:
> The Discovery failures is due to `proxy.stop()` taking a very long time to complete. I could dig into it, but ultimately the problem will be addressed by the cancel-ability changes PR after this one. For now i'll keep the discovery tests disabled.
> 
> ```
> console.time
>     proxy.stop(): 19142 ms
>       at Object.<anonymous> (tests/discovery/Discovery.test.ts:240:13)
> ```

Please link the next PR to here. Applying timedCancellable across the entire codebase.

src/discovery/Discovery.ts
```diff
@@ -81,14 +89,35 @@ class Discovery {
...
+        return;
+      }
+      // Aborting a duplicate task is not an error
+      if (e === abortSingletonTaskReason) return;
+      // Destroying tasks is not an error
+      if (e === discoveryDestroyedTaskReason) return;
+      throw e;
+    }
+  };
+  public readonly discoverVertexHandlerId =
```
CMCDragonkai, line 130: Would this not have the same problem as nodes where the task handler ids collide?

src/PolykeyAgent.ts
tegefaulkes: Yeah. All the tasks in `NodeManager` are ephemeral, they are cancelled during stopping. Here the discovery tasks persist. So we only remove them when cleaning up during destruction or starting fresh.

src/discovery/types.ts
tegefaulkes: They are no longer needed. They only existed to enforce discovery order in the database, `Discovery` no longer uses the database directly and order is handled via the `TaskManager` now. So this is no longer needed.

tests/discovery/Discovery.test.ts
tegefaulkes: We want to stub it out. It was mostly to avoid extra logging. I'm not sure it's needed anymore but we do want to avoid a bunch of random connections in the background while testing discovery.

src/discovery/Discovery.ts
tegefaulkes: Ah, that should be removed.

src/discovery/Discovery.ts
tegefaulkes: Yes, but there is only 1 handler in this domain, so the handler never gets overwritten.

src/discovery/Discovery.ts
tegefaulkes: Moved.

src/discovery/Discovery.ts
tegefaulkes: Removed.

src/discovery/Discovery.ts
tegefaulkes: Added comments.

src/discovery/Discovery.ts
tegefaulkes: The comments still seem relevant but I'm not sure why they're TODOs because they don't describe something that needs to be done. They're just descriptive. I can make them normal comments?

src/discovery/Discovery.ts
tegefaulkes: `hasPendingDiscovery`, `discoveryIsActive`? not sure what to call it.

tegefaulkes: Since we use `waitForDrained` maybe call it `isDrained` to complement it?

src/discovery/Discovery.ts
tegefaulkes: I've called it `scheduleDiscoveryForVertex` for now. I don't really see an analogue name in nodes.

src/PolykeyAgent.ts
CMCDragonkai: Make the comment similar in consistency to DB comment too.

src/discovery/Discovery.ts
CMCDragonkai: It says `I don't believe that this will iterate in lexicographical order of keys. For now,`. What does this mean? Can we make confirm what the proper behaviour is and rewrite the comment.

src/discovery/Discovery.ts
CMCDragonkai: Yea, but if we have a hack for #453, then the same hack should be applied consistently. Although before you do so, please check on the CI/CD if the name is applied properly by logging out this property.

And if isn't affecting this domain, update the #453 with this discrepancy. It is concerning.

src/discovery/Discovery.ts
CMCDragonkai: Just do `hasDiscoveryTasks`. And we do have multiple tasks.

tests/discovery/Discovery.test.ts
CMCDragonkai: So the discovery tests is making use of nodes domain which ultimately connects those nodes over the network. To do this, we have to simulate agents running. This is done in this test with `PolykeyAgent` class, and these agents are nodes that are running in-process. But they are bound to the OS network interfaces.

CMCDragonkai:
As per: https://github.com/MatrixAI/Polykey/pull/451/files/2d004eeb5bc7270320ecef7fb0cb3ebfe43877c8#r983028305

It is possible to remove these mocks from our tests:

```ts
const mockedRefreshBucket = jest.spyOn(
  NodeManager.prototype,
  'refreshBucket',
);
```

If we realise that what we need is to have 2-stage `start` for some of our objects.

The `NodeManager.start` is similar to `TaskManager.start` in the sense that there are 2 stages here.

1. Making the object ready to be interacted with
2. Actually starting the background tasks (live side-effects)

We can reuse the pattern of `lazy` parameter, by putting a `lazy` parameter into `NodeManager.start`.

Then we also need to move `NodeConnectionManager.start`'s ops in setting the seed node to node manager into `NodeManager.syncNodeGraph`:

```ts
    // Adding seed nodes
    for (const nodeIdEncoded in this.seedNodes) {
      const nodeId = nodesUtils.decodeNodeId(nodeIdEncoded);
      if (nodeId == null) never();
      await this.nodeManager.setNode(
        nodeId,
        this.seedNodes[nodeIdEncoded],
        true,
      );
    }
```

The above code seems that it can just be done as part of `NodeManager.syncSeedNodes`.

Then in `NodeManager.start`, we enclose the background operations as part of:

```ts
    if (!lazy) {
      await this.startSync();
    }
```

So this `startSync` would be similar to `TaskManager.startProcessing`, in that it would perform these 2 operations:

```ts
    async startSync() {
      await this.syncSeedNodes(); // Replaced name for `syncNodeGraph`
      await this.setupRefreshBucketTasks();
    }
```

Can we then have an equivalent `stopSync`? Similar to `stopProcessing`? Or does this not matter.

I guess it would whatever `NodeManager.stop` is currently doing, but just factored out to `stopSync`.

Atm, it looks like `syncSeedNodeGraph` doesn't actually get cancelled if we call `NodeManager.stop`, therefore there is a place to create a `stopSync` method that `NodeManager.stop` ends up calling as well.

This would basically make `NodeManager` very similar to `TaskManager`.

Then in the `PolykeyAgent`, under `createPolykeyAgent`, it would need the `lazy: true`, and then under `start`, it would do something like:

```ts
      await this.taskManager.start({ fresh, lazy: true });
      await this.nodeManager.start({ lazy: true });
      await this.nodeConnectionManager.start({ 
        nodeManager: this.nodeManager 
      });
      await this.nodeGraph.start({ fresh });
      await this.discovery.start({ fresh });
      await this.vaultManager.start({ fresh });
      await this.notificationsManager.start({ fresh });
      await this.sessionManager.start({ fresh });
      if (lazy) {
        await this.nodeManager.startSync();
        await this.taskManager.startProcessing();
      }
```

Notice the `lazy` boolean here, this propagates the pattern to `PolykeyAgent`. Which would mean upon testing `PolykeyAgent` such as in the discovery tests, rather than mocking out specific side-effects, we can just start the `PolykeyAgent` lazily, meaning none of the background tasks are started. If we want to do that, we would need another method `PolykeyAgent.startX` (not sure what `X` should be called). But that would then start the background tasks for PolykeyAgent.

---

What does it mean to do tests on PK agent without having it run background tasks? It seems that this means that it would avoid interference if we are trying to do a deterministic test.

So imagine you want to test the PK agent to do a specific direct connection or a specific call, and you would prefer that these background tasks aren't running. Then it does make sense to have this ability of staging the second stage of start.

So now we have 3/4 stages:

1. Async creation
2. Sync construction
3. Async start
4. Async start background stuff

tests/discovery/Discovery.test.ts
CMCDragonkai: Ok some notes about how we can eliminate these mocks in the PR comment, but it will require a new PR on changing some parts of the bootstrapping phase for nodes too.

CMCDragonkai:
The `fwxProxy` logger name should just be `Proxy`.

tests/discovery/Discovery.test.ts
```diff
@@ -152,7 +162,6 @@ describe('Discovery', () => {
     });
     await nodeManager.start();
     await nodeConnectionManager.start({ nodeManager });
-    await taskManager.startProcessing();
     // Set up other gestalt
     nodeA = await PolykeyAgent.createPolykeyAgent({
```
CMCDragonkai, line 166: We should abstract this creation of PolykeyAgents for testing purposes. There's a lot of tests that are repeating this bootstrapping code. I thought we had something like this already called `pkAgent`, but I think that was for spawning a separate process. We should have something similar for starting in-process agents.

CMCDragonkai: It was called `setupGlobalAgent`, but it was removed since we no longer required a global agent process. But it does seem having a utility for in-process agents would be helpful here.

CMCDragonkai: Actually I was confused about this, it's the discovery dependencies themselves.

src/discovery/Discovery.ts
```diff
@@ -175,299 +208,291 @@ class Discovery {
...
+        never();
     }
+    this.visitedVertices.add(vertex);
   }
 
   /**
-   * Will resolve once the queue has drained
+   * Will resolve once all discovery tasks have finished
    */
   public async waitForDrained(): Promise<void> {
```
CMCDragonkai: So this function should be renamed and made a bit more precise. It should be about waiting for all currently scheduled, queued and active discovery tasks to finish. It does not signal cancellation to any of the tasks. Therefore it cannot be integrated into the `stop` method.

This is used for testing, so that way we can wait for discovery side-effects to complete.

Let's call this:

```ts
await waitForDiscoveryTasks();
```

This can then be similar to the other function `hasDiscoveryTasks` which just returns a boolean if there are any scheduled, queued or active tasks in the discovery system.

The function should be kept simple in that it only checks once for all the tasks and waits for them. The looping implementation atm cannot guarantee that no more tasks are scheduled and it could also potentially run forever.

src/discovery/Discovery.ts
```diff
@@ -123,36 +158,34 @@ class Discovery {
...
     this.logger.info(`Started ${this.constructor.name}`);
   }
 
   public async stop(): Promise<void> {
     this.logger.info(`Stopping ${this.constructor.name}`);
-    this.queuePlug.resolveP();
-    await this.discoveryProcess;
+    this.taskManager.deregisterHandler(this.discoverVertexHandlerId);
     this.logger.info(`Stopped ${this.constructor.name}`);
   }
```
CMCDragonkai, line 201: It is a good idea to also initiate cancellation of any discovery tasks that still exist. We should not necessarily expect that the tasks have been stopped when calling `Discovery.stop`. This would then make this similar to `NodeManager.stop` which does in fact remove all tasks.

In this case, we don't need to remove all tasks, but just send cancellation signals and then wait for them to resolve/fulfill.

We also should check if `TaskManager` is still processing and if it is, we need to throw an exception. The `NodeManager` should probably be checking for this too.

CMCDragonkai:
@tegefaulkes can you create a new issue for discovery refactoring: https://github.com/MatrixAI/Polykey/pull/451#issuecomment-1257770135 (needs TTL) - this can be done in separate PR

And also a new issue for this: https://github.com/MatrixAI/Polykey/pull/451#issuecomment-1261697179 - this can be done separate PR

tests/discovery/Discovery.test.ts
tegefaulkes: I can but the change will have to be applied to a broad range of tests. This would be out of scope for the PR. I can make an issue for it.

src/discovery/Discovery.ts
tegefaulkes: I'm having these methods return the number of existing tasks just avoid boolean blindness.

tegefaulkes: This has been updated.

src/discovery/Discovery.ts
tegefaulkes: This has been updated.

src/PolykeyAgent.ts
tegefaulkes: Updated the comment.

src/discovery/Discovery.ts
CMCDragonkai: Then is this `getTaskCount()`?

src/discovery/Discovery.ts
tegefaulkes: `Discovery` is now cancelling any active tasks when stopping. Both `NodeManager` and `Discovery` check for if `TaskManager` is still processing when stopping.

src/discovery/Discovery.ts
tegefaulkes: I applied the same fix as for `NodeManager`. I'm skipping checking if the `name` is empty on CI. Makes more sense to check that when addressing the issue directly.

src/discovery/Discovery.ts
tegefaulkes: I'm going to leave these as is. The gist of what they're saying is that if we had revocation then the order here would matter. But since we don't support that right now it doesn't matter what order we iterate over the claims.  If we do want to update it then I'll leave it for the next stage of discovery refactoring.

tegefaulkes:
All review addressed now. I'll clean up the commit history again, merge, and create the new issues/PRs.

tegefaulkes:
Squashed and re-based on staging.

src/discovery/Discovery.ts
CMCDragonkai: Do you mean sigchain revocation of claims?

CMCDragonkai: What order are we iterating claims atm? Wouldn't we always iterate from most recent to most oldest?

tests/discovery/Discovery.test.ts
CMCDragonkai: Not important atm.

CMCDragonkai:
Please link the new issues you've created here too.

src/discovery/Discovery.ts
tegefaulkes: This comment is referenced in #327. I'll update the comment to make it less weird.

tegefaulkes:
Issue #461 created to address https://github.com/MatrixAI/Polykey/pull/451#issuecomment-1261697179.

tegefaulkes:
Issue #462 created to address further discovery refactoring.

PR #446, opened 2022-09-13
Updating Crypto to use WebCrypto API and to replace RSA with ECC

CMCDragonkai:
### Description

This PR focuses on updating the crypto utilities used by PK. We've been hitting problems using RSA and node-forge utilities, and we should start using the standardised WebCrypto API. This won't fully solve cross platform cryptography because that will need to wait until we hit mobile platforms and deal with it by using WASM or other utilities.

There are some new features coming into this PR:

1. Public key encryption now allows arbitrary message size. There is no longer a limit on large the data to be encrypted is. The existing limit is only 446 bytes.
2. Public key encryption supports static-static, ephemeral-static, and ephemeral-ephemeral.
3. Because Ed25519 public keys are 32 bytes, `NodeId` is now finally the public key. This means you no longer have to acquire the public key separately from the `NodeId`. Once you have the `NodeId` you can use it for public key verification, and for encryption.
4. The `@peculiar/webcrypto` is being monkey patched to `globalThis.crypto`. This ensures that every library is using the same webcrypto backend and this includes CSPRNG and encryption/decryption facilities.

Massive performance improvements in all areas:

<details>

Old performance (node-forge):

```
# TYPE keys.asymmetric_crypto_ops gauge
keys.asymmetric_crypto_ops{name="encrypt 446 B of data"} 418.07
keys.asymmetric_crypto_ops{name="decrypt 446 B of data"} 6.49
keys.asymmetric_crypto_ops{name="sign 446 B of data"} 6.53
keys.asymmetric_crypto_ops{name="sign 1 KiB of data"} 6.57
keys.asymmetric_crypto_ops{name="sign 10 KiB of data"} 6.61
keys.asymmetric_crypto_ops{name="verify 446 B of data"} 449.55
keys.asymmetric_crypto_ops{name="verify 1 KiB of data"} 445.57
keys.asymmetric_crypto_ops{name="verify 10 KiB of data"} 428

# TYPE keys.asymmetric_crypto_margin gauge
keys.asymmetric_crypto_margin{name="encrypt 446 B of data"} 0.39
keys.asymmetric_crypto_margin{name="decrypt 446 B of data"} 0.37
keys.asymmetric_crypto_margin{name="sign 446 B of data"} 0.48
keys.asymmetric_crypto_margin{name="sign 1 KiB of data"} 0.45
keys.asymmetric_crypto_margin{name="sign 10 KiB of data"} 0.25
keys.asymmetric_crypto_margin{name="verify 446 B of data"} 0.74
[...]

PR #455, opened 2022-09-27
Integrate `@swc/jest` instead of `ts-jest` and `babel-jest`

CMCDragonkai:
### Description

This brings in the `@swc/jest` as the transpiler to be used when running tests.

We are already using `swc` for `ts-node` and things appear to work fine.

Note that `swc` has its own configuration, and `ts-node` auto maps tsconfig to swc config. But for this jest integration it has to be done manually. I've decided to do this inside the `jest.config.js`.

The `typescript` syntax means only a few settings needs to be set.

```
  transform: {
    "^.+\\.(t|j)sx?$": [
      "@swc/jest",
      {
        "jsc": {
          "parser": {
            "syntax": "typescript",
            "dynamicImport": true,
            "tsx": true,
            "decorators": compilerOptions.experimentalDecorators,
          },
          "target": compilerOptions.target.toLowerCase(),
          "keepClassNames": true,
        },
      }
    ],
  },
```

I'm not entirely sure what `keepClassNames` is needed for, but apparently needed for more sophisticated decorator usage.

On top of this, it requires a lower case target. Note that this takes over both tsx and jsx and ts and js. All JS code is valid TS code syntactically.

Note that:

```
const { pathsToModuleNameMapper } = require('ts-jest');
```

Is still required to "map" the module names - https://github.com/swc-project/jest/issues/107

Finally I do note that apparently 4.x of jose doesn't require special hack for the module name mapper, but I'm not sure about it yet. - https://github.com/panva/jose/discussions/105#discussioncomment-1486390. I will try this out in #446 since I'm working on the JOSE library there.

This improves our test execution from 8 seconds down to 1 second on my dell-vostro-5402 for single test file. And for matrix-ml-1, 5.6 seconds down to 1.7 seconds real time. I think the perf would increase the more files that needs to be compiled together.

### Tasks

- [x] 1. Installed `@swc/jest`
- [x] 2. Removed `babel-jest` and `@babel/preset-env`
- [x] 3. Kept `ts-jest` due to `pathsToModuleNameMapper`
- [x] 4. Mapped `tsconfig.json` to swc configuration inside `jest.config.js` to enable decorator usage
- [x] 5. Tested several things

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,3 +1,0 @@
-module.exports = {
-  presets: ['@babel/preset-env']
-};
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,10 +9,6 @@ const moduleNameMapper = pathsToModuleNameMapper(compilerOptions.paths, {
   prefix: '<rootDir>/src/',
 });
 
-// using panva/jose with jest requires subpath exports
-// https://github.com/panva/jose/discussions/105
-moduleNameMapper['^jose/(.*)$'] = "<rootDir>/node_modules/jose/dist/node/cjs/$1";
-
 // Global variables that are shared across the jest worker pool
 // These variables must be static and serializable
 if ((process.env.PK_TEST_PLATFORM != null) !== (process.env.PK_TEST_COMMAND != null)) throw Error('Both PK_TEST_PLATFORM and PK_TEST_COMMAND must be set together.')
@@ -50,8 +46,21 @@ module.exports = {
   roots: ['<rootDir>/tests'],
   testMatch: ['**/?(*.)+(spec|test|unit.test).+(ts|tsx|js|jsx)'],
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
-    '^.+\\.jsx?$': 'babel-jest',
+    "^.+\\.(t|j)sx?$": [
+      "@swc/jest",
+      {
+        "jsc": {
+          "parser": {
+            "syntax": "typescript",
+            "dynamicImport": true,
+            "tsx": true,
+            "decorators": compilerOptions.experimentalDecorators,
+          },
+          "target": compilerOptions.target.toLowerCase(),
+          "keepClassNames": true,
+        },
+      }
+    ],
   },
   reporters: [
     'default',
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,10 +60,10 @@
         "polykey": "dist/bin/polykey.js"
       },
       "devDependencies": {
-        "@babel/preset-env": "^7.13.10",
         "@fast-check/jest": "^1.1.0",
         "@streamparser/json": "^0.0.13",
-        "@swc/core": "^1.2.215",
+        "@swc/core": "^1.3.62",
+        "@swc/jest": "^0.2.26",
         "@types/cross-spawn": "^6.0.2",
         "@types/google-protobuf": "^3.7.4",
         "@types/jest": "^28.1.3",
@@ -74,7 +74,6 @@
         "@types/readable-stream": "^2.3.11",
         "@typescript-eslint/eslint-plugin": "^5.45.1",
         "@typescript-eslint/parser": "^5.45.1",
-        "babel-jest": "^28.1.3",
         "benny": "^3.7.1",
         "common-tags": "^1.8.2",
         "eslint": "^8.15.0",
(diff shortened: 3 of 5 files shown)


CMCDragonkai:
Decorator problem is fixed. However I'm noticing a problem in `timed.test.ts`:

```
 FAIL  tests/contexts/decorators/timed.test.ts
  ● Test suite failed to run

    TypeError: Invalid value used as weak map key
        at WeakMap.set (<anonymous>)

      13 |     );
      14 |   }
    > 15 |   contextsUtils.contexts.set(method, index);
         |                          ^
      16 | }
      17 |
      18 | export default context;

      at set (src/contexts/decorators/context.ts:15:26)
      at tests/contexts/decorators/timed.test.ts:62:9
      at Object.<anonymous>.__decorate (tests/contexts/decorators/timed.test.ts:57:108)
      at tests/contexts/decorators/timed.test.ts:205:9
      at describe (tests/contexts/decorators/timed.test.ts:51:3)
```

CMCDragonkai:
It's not a problem with `ts-jest`, only `@swc/jest`.

CMCDragonkai:
I think `@swc/jest` has a bug in their decorator implementation. It appears to not be properly doing decorators for symbol methods.

CMCDragonkai:
Note that due to what `tsconfig.json` is configured to, if we try to use prototype files at project root with decorators, there will be TS errors inside vscode. So if you have to do this, it has to be in `./src` directory, not at the project root.

I tried fixing it before but it introduced new problems, so right now that's a limitation of the fact that decorators are still "experimental".

@tegefaulkes

CMCDragonkai:
Ok both `ts-node` and jest that uses swc has a problem with decorators when applied to symbol methods.

It looks like this:

```ts
import type { ContextCancellable } from './contexts/types';
import { PromiseCancellable } from '@matrixai/async-cancellable';
import cancellable from './contexts/decorators/cancellable';
import context from './contexts/decorators/context';

const symbolFunction = Symbol('sym');

class X {
  [symbolFunction](
    ctx?: Partial<ContextCancellable>,
  ): PromiseCancellable<void>;
  @cancellable(false)
  [symbolFunction](@context ctx: ContextCancellable): Promise<void> {
    expect(ctx.signal).toBeInstanceOf(AbortSignal);
    return new Promise((resolve) => void resolve());
  }
}
```

Running gives us `undefined` inside the `context` parameter decorator.

This is not the case with regular functions and async functions.

CMCDragonkai:
Updating swc to see if this works.

CMCDragonkai:
Reporting the bug upstream: https://github.com/swc-project/swc/issues/5962

CMCDragonkai:
This is blocked until the upstream bug can be fixed. No workarounds atm. We don't actually use symbol methods in our code base atm, only in our tests. But we aren't going to change our tests just yet. Let's see what the CI/CD says though.

CMCDragonkai:
I've actually applied this to `js-async-init` now. It uses `@swc/jest` and it went from 3 seconds of testing to 2 seconds of testing. Which is 30% improvement!

https://github.com/MatrixAI/js-async-init/commit/51c83e3c89c927a6f41335e0628ab057783dde74

CMCDragonkai:
So both `ts-jest` and `@swc/jest` is required. The `ts-jest` is simply needed for the module path mapping.

CMCDragonkai:
The symbol methods are only used in one tests in the contexts tests. These have been moved out to js-contexts. This means I think we should be able to actually accomplish this. This would make testing PK much faster!

CMCDragonkai:
@tegefaulkes can take over this since I think this can start being used. Beware of task 2 and the note about testing if we need the hack regarding panva jose in our `jest.config.js`. Solving this can reduce some entropy in our configuration and significantly reduce testing time.

tegefaulkes:
I'm going to keep it to this PR. I've done a re-base and minor updates including removing the `jose` hack from `tsconfig.js`.  I'll push that up in a sec.

I don't know if this fully works. The I don't recall if all the tests were passing since we're between the client and agent migration in the history here. 

So far as I can tell it's mostly working minus some minor fixes where the create functions need to use `new this()`. There might be a problem with some of the jest mocking. I'll need to review.

I'll cherry-pick the change into the agent migration to test it out. But for now I'll leave this un-merged until I know it's fully working.

CMCDragonkai:
Closing this in favour of #525 as it will be merged into there. @tegefaulkes

tegefaulkes:
re-opening....

PR #457, opened 2022-09-27
feat: centralising all ID types and utilities under `ids` domain

CMCDragonkai:
### Description

This makes it easier to prototype breaking changes as other domains that depend on the ID types and utilities can instead depend on the common factor that is `ids` rather than the actual domains. The types and utilities are still re-exported by their respective domain for ease of use, but now generally if you need only the ID type or utility, use the `ids` domain instead.

### Issues Fixed

* Related to #446 (first dealt with there)

### Tasks

- [x] 1. Created `src/ids`
- [x] 2. Centralised types under `src/ids/types.ts`
- [x] 3. Centralised utilities under `src/ids/index.ts`
- [x] 4. Re-exported ids related utilities and types for all domains
- [x] 5. Changed vault id generator to actually be a generator creator to match all other ID behaviour

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyClient.ts
+++ b/src/PolykeyClient.ts
@@ -1,5 +1,5 @@
 import type { FileSystem, Timer } from './types';
-import type { NodeId } from './nodes/types';
+import type { NodeId } from './ids/types';
 import type { Host, Port } from './network/types';
 import path from 'path';
 import Logger from '@matrixai/logger';
--- a/src/acl/ACL.ts
+++ b/src/acl/ACL.ts
@@ -5,7 +5,7 @@ import type {
   Permission,
   VaultActions,
 } from './types';
-import type { NodeId } from '../nodes/types';
+import type { NodeId } from '../ids/types';
 import type { GestaltAction } from '../gestalts/types';
 import type { VaultAction, VaultId } from '../vaults/types';
 import type { Ref } from '../types';
--- a/src/acl/types.ts
+++ b/src/acl/types.ts
@@ -1,10 +1,6 @@
-import type { Opaque } from '../types';
+import type { PermissionId, PermissionIdString } from '../ids/types';
 import type { GestaltAction } from '../gestalts/types';
 import type { VaultActions, VaultIdString } from '../vaults/types';
-import type { Id } from '@matrixai/id';
-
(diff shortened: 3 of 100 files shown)


CMCDragonkai:
Note that in order to properly mock protected properties you have to do `jest.spyOn(x as any, 'privateOrProtectedProperty');`

Problem is, the type itself is incorrect, I'm not sure if I'm supposed to assert it as a spy instance... etc. Anyway, did a quick fix on the vault tests that are doing this. I believe these tests should probably be refactored quite a bit, they are very slow.

CMCDragonkai:
Looking at the map, it shows that alot of the imports on `nodes/types` have been broken which is nice. Less centralisation on `nodes/types`. But replaced by centralisation on `ids/types`. Which is better, since it's a more "simpler" domain and less likely to break.

image

CMCDragonkai:
@tegefaulkes can you start to make sure that `@ready` isn't being used on protected methods. I found a bunch in the vault manager.

CMCDragonkai:
The full tests will be left to the CI/CD to do, if it passes, merge into staging.

CMCDragonkai:
@tegefaulkes I thought discovery tests are being disabled due to timeouts. It's still failing at this PR, or was it only disabled for staging?

Also there's a failure in nodes... can you have a look at this too.

CMCDragonkai:
So the nodes tests failure for bucket reset is possible due to the fact that resetting to new node id may still result in the same bucket structure. Therefore we need to move to more fine grained test. It's good idea to reproduce it first though with fast check, then change the test to be more deterministic, we can't expect the bucket structure to always be the same.

The discovery tests weren't disabled for feature branches, so it should be done for both staging and feature and re-enabled once @tegefaulkes fixes up those tests.

tegefaulkes:
Re-based on staging and applied a fix to the failing nodes test. The original form of the test still exists but it was converted to using fast check so we can re-create the failure if it happens again. I also split up the tests into smaller forms of it that test specific aspects. This way I can run more test runs for the specific tests.

This should be good to merge pending CI passing.

tegefaulkes:
CI has passed, I'm merging this now.

Issue #465, opened 2022-09-29
Update `Nodes` domain to used `timedCancellable`


tegefaulkes:
### Specification

The nodes domain manages the agent-agent communications between nodes. This is mostly done inside of `NodeConnection` and `NodeConnectionManager` but the `NodeManager` wraps some of this as well. All of these needs to be reviewed and updated to make use of the `TimedCancellable` decorator. 

### Additional context

- Related #445 
- Related #450 

### Tasks

1. Where needed the nodes methods need to be updated to support cancellability.
2. where needed the nodes methods need to use the `timedCancellable` decorator.
3. parts of the code making use of these timedCancellable methods needs to be updated to exploit this new functionality.
4. Tests need to be expanded to test a wide range of timeout and cancellation conditions.

tegefaulkes:
This was mostly addressed in #445 . It just needs further review and possibly test updates.

CMCDragonkai:
Regarding 4.

There are several layers of the nodes tests.

Unit-level tests on the nodes domain can make use of mocking for simple testing.

But more complex testing can be moved into the integration tests... and that relates to moving the integration tests into its own directory.

When you're reviewing the nodes tests, can you take note of any tests can be moved around, so that `tests/nodes` is more lighter weight, but also more flexible so we can apply fast check to them.

I suspect mocking will be required there.

tegefaulkes:
`NodeConnectionManager.withConnG` is a generator but still needs to take the `ctx: ContextTimed` because it could establish a new connection. I don't think the decorators support generators yet. Right now i'm just optionally taking the the context and just passing that along.

I'm not sure if there's a better way for this.

tegefaulkes:
The `NodeConnectionManager` has a dynamic parameter for setting the connection timeouts `connConnectTime`. but not all of the `timedCancellable` methods has a corresponding timeout parameter to use. `pingNode` should have it's own timeout parameter in the `NodeConnectionManager`.

The `NodeManager` wraps some methods from the `NodeConnectionManager` in it's own `timedCancellable` methods. They should dynamically get the timeout value as well but it doesn't seem right to set them in the `NodeManager`. I'll make it take the values from the `NodeConnectionManager` as well.

tegefaulkes:
NAT tests are still failing, that will have to be looked at next.

PR #468, opened 2022-09-30
Feature: network cancellability and deadlines

### LINKED ISSUE #464: Update `Proxy` domain to use `timedCancellable`
### Specification

The `Proxy` domain needs to be updated to use `timedCancellable` and better support cancellability. This means that any methods that create connections needs to support cancellation. Robust testing needs to be added to test if cancelled connections are gracefully handled from both the forward and reverse side of the connection. 

### Additional context

- Related #450 

### Tasks

1. Where applicable the `Proxy` methods needs to support cancellation via the abort signal
2. Where applicable the `Proxy` methods need to be updated to use the `timedCancellabe` decorator.
3. `Proxy` tests need to be expanded to test a wide range of cancellation and connection failure conditions.
4. Update punch interval to 50 ms

tegefaulkes:
### Description

This PR focuses on updating the network domain with `timedCancellable`, general cancellability and deadlines.

### Issues Fixed

* Related #450 
* Fixes #464 

### Tasks

- [x] 1. Update methods for establishing connections to use the `timedCancellable` decorator.
- [x] 2. Replace the connection lock maps with `LockBox`.
- [x] 3. Fix tests based on changes
- ~Add tests to check for the problem in https://github.com/MatrixAI/Polykey/pull/445#issuecomment-1253162920 . the problem should also be fixed.~ New issue at #473 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/network/Connection.ts
+++ b/src/network/Connection.ts
@@ -40,7 +40,7 @@ abstract class Connection {
     tlsConfig,
     keepAliveTimeoutTime = 20000,
     endTime = 1000,
-    punchIntervalTime = 1000,
+    punchIntervalTime = 50,
     keepAliveIntervalTime = 1000,
     logger,
   }: {
--- a/src/network/ConnectionForward.ts
+++ b/src/network/ConnectionForward.ts
@@ -1,17 +1,21 @@
 import type { Socket, AddressInfo } from 'net';
 import type { TLSSocket } from 'tls';
 import type UTPConnection from 'utp-native/lib/connection';
+import type { PromiseCancellable } from '@matrixai/async-cancellable';
 import type { Certificate } from '../keys/types';
 import type { Address, Host, NetworkMessage, Port } from './types';
 import type { NodeId } from '../ids/types';
-import type { AbstractConstructorParameters, Timer } from '../types';
+import type { AbstractConstructorParameters } from '../types';
+import type { ContextTimed } from '../contexts/types';
 import tls from 'tls';
 import { StartStop, ready } from '@matrixai/async-init/dist/StartStop';
 import Connection from './Connection';
 import * as networkUtils from './utils';
 import * as networkErrors from './errors';
 import * as keysUtils from '../keys/utils';
-import { promise, timerStart, timerStop } from '../utils';
+import { promise } from '../utils';
+import * as contextsErrors from '../contexts/errors';
+import { timedCancellable, context } from '../contexts/index';
 
 type ConnectionsForward = {
   proxy: Map<Address, ConnectionForward>;
@@ -106,11 +110,7 @@ class ConnectionForward extends Connection {
     this.connections = connections;
   }
 
-  public async start({
-    timer,
-  }: {
-    timer?: Timer;
-  } = {}): Promise<void> {
+  public async start({ ctx }: { ctx: ContextTimed }): Promise<void> {
     this.logger.info('Starting Connection Forward');
     // Promise for ready
     const { p: readyP, resolveP: resolveReadyP } = promise<void>();
@@ -119,6 +119,9 @@ class ConnectionForward extends Connection {
     // Promise for secure connection
     const { p: secureConnectP, resolveP: resolveSecureConnectP } =
       promise<void>();
+    // Promise for abortion and timeout
+    const { p: abortedP, resolveP: resolveAbortedP } = promise<void>();
+    ctx.signal.addEventListener('abort', () => resolveAbortedP());
     this.resolveReadyP = resolveReadyP;
     this.utpSocket.on('message', this.handleMessage);
     const handleStartError = (e) => {
@@ -154,7 +157,7 @@ class ConnectionForward extends Connection {
       await Promise.race([
         Promise.all([readyP, secureConnectP]).then(() => {}),
         errorP,
-        ...(timer != null ? [timer.timerP] : []),
+        abortedP,
       ]);
     } catch (e) {
       // Clean up partial start
@@ -177,7 +180,7 @@ class ConnectionForward extends Connection {
     }
     this.tlsSocket.on('error', this.handleError);
     this.tlsSocket.off('error', handleStartError);
-    if (timer?.timedOut) {
+    if (ctx.signal.aborted) {
       // Clean up partial start
       // TLSSocket isn't established yet, so it is destroyed
       if (!this.tlsSocket.destroyed) {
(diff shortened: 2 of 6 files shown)


tegefaulkes:
This is at the stage for review now.

CMCDragonkai:
Did you update the the intervals to 100ms?

Also this is only for `network` right? And are we applying this to everywhere where it is applicable to the networking:

1. Establishing connections
2. The `endGracefully` which should have its `timeout` replaced too
3. `ConnectionReverse.compose`
4. `ConnectionReverse.start`
5. `Proxy.connectForward`
6. `Proxy.connectReverse`
7. `Proxy.openConnectionForward`
8. `Proxy.openConnectionReverse`

I want to ensure that `timerStart` and `timerStop` utilities are no longer being used, and instead replaced with the `Timer` class.

CMCDragonkai:
Can you start add new tests to test the new timer usage. As well as explicit cancellation signals that should be stopping all side effects.

CMCDragonkai:
> Robust testing needs to be added to test if cancelled connections are gracefully handled from both the forward and reverse side of the connection.

You need to start on this.

tegefaulkes:
Yes this was applied to the whole `network` domain. I've updated everything that had a timeout/timer.

tegefaulkes:
The networking is like an onion, there are many layers that makes it annoying to thing about. For a robust system each of these layers needs to properly handle timeouts and potential errors. For now I'm reviewing the network architecture to get a clear understanding of these layers. 

At almost every layer we will need to handle the following failure conditions. 
1. SRC connection fails to connect to DST due to DST not responding at all
2. SRC connection fails to connect to DST due to DST rejecting connection
3. SRC starts connecting to DST but DST fails to fully establish.
4. SRC starts connecting to DST but SRC rejects part way.
5. SRC starts connecting to DST but SRC times out part way.
6. SRC connects to DST but SRC stops responding
7. Connection fully succeeds but SRC stops responding
8. Connection fully succeeds but SRC forcibly closes the connection
9. Connection fully succeeds but DST stops responding
10. Connection fully succeeds but DST forcibly closes the connection

Changes in #445 applied a shorter timeout to `nodeManager.ping` which resulted in some new errors from the connections. https://github.com/MatrixAI/Polykey/pull/445#issuecomment-1253162920. Ideally we will reveal the source of these problems by expanding the tests here.

tegefaulkes:
From the Proxy's perspective the first layer is a HTTP connection. When this HTTP connection is received, a forward connection is created if it doesn't already exist. It is then composed by piping data between the HTTP socket and the forward connection.

Two parts of the code here strike me as potential problems.
1. Handling for the HTTP client ending connection isn't set up until the connection is composed. This could lead to a race condition where the connection is ended before we compose the connection. This would result in the connection not gracefully ending here. This also means we don't set this up if the forward connection fails to be created.
2. To properly end a connection we should signal end, keep reading any data until the other side signals end, then we can destroy the socket. We have a method in `connectionForward` and `connectionReverse` for this already called `endGracefully`. We should be doing this for the HTTP connection as well.
3. the `connectionForward.compose` method sets up an `end` event, but it doesn't await for `clientSocket.end` to finish before destroying the connection. This wil likely lead to a race conditon,

Another thing to note, the HTTP clientSocket can provide it's own timeout value, should be be using this to some degree, especially if it's shorter than our default?

tegefaulkes:
tests for HTTP side failures.
1. ~HTTP client connects and immediately ends forcibly.~
2. ~HTTP client connects and immediately ends gracefully.~
3. ~HTTP client connects with a set timeout and times out immediately.~
4. ~HTTP client connects and just stops responding~
5. ~HTTP client connects and ends forcibly after composing.~
6. ~HTTP client connects and ends gracefully after composing.~
7. ~HTTP client connects with a set timeout and times out after composing.~
8. HTTP client connects and just stops responding.

tegefaulkes:
Looks like the `HTTP` client doesn't fully establish connection until the `ConnectionForward` is created and composed. The `handleConnectForward` doesn't send the 200 code until then. This means the above points  1 and 2 are moot. Point 3 still stands. 

As for the tests. It essentially means we can't do anything to the HTTP `clientSocket` directly until the forward connection if created and composed. That precludes us from ending or destroying the `clientSocket` early. that just leaves us 'critical' failures where the HTTP client just stops responding at certain stages. 

I think this means the only error a failure on the HTTP client side could cause is a error when writing to it. This would be when we send the 200 code after setting up. This means we only need to add 1 test for this, but it will have to make use of a `exec` to run it in a separate process and kill it.

CMCDragonkai:
> Yes this was applied to the whole `network` domain. I've updated everything that had a timeout/timer.

Yea it looks like the `Timer` from `src/types.ts` is still being used in:

* `src/PolykeyClient.ts`
* `GRPCClientAgent.ts`
* `GRPCClientClient.ts`
* `GRPCClient.ts`
* `ConnectionForward.ts`
* `NodeConnection.ts`
* `NodeConnectionManager.ts`
* `utils.ts`
* `GRPCClientTest.ts`

Most of that is GRPC related changes, and that's coming in the next PR right?

However the `NodeConnection.ts` and `NodeConnectionManager.ts` should be changed in this PR since this those usages are dependent on the network domain. I suppose this goes into #465?

tegefaulkes:
I'm having trouble re-creating the connection bug from #445. Since it's a race condition style bug, it may just be  harder to tigger on my VM. Since this mostly affects the nodes domain, I'm going to defer fixing it to issue #465. Writing a test for it is going to very tricky since it depends heavily on timing. We may not be able to create a consistent test for it.

CMCDragonkai:
Some of the existing network tests simulate a UTP client/server and others simulate a GRPC client/server. Is it possible to use this to reproduce #445?

tegefaulkes:
It's possible, but it's going to take a bit more work to narrow down the exact conditions causing the problem before I can create a test for it. Right now it's delaying this PR but is a separate enough problem that it can be addressed separately. Given the difficulty we've had with network problems like this in the past, it could be it's own issue.

I'd like to get this PR cleaned, reviewed and merged.

CMCDragonkai:
Ok go ahead. You should then create a separate issue to track that bug then. But it really should have instructions as to how to reproduce that bug. Otherwise there's not much context to go off from.

tegefaulkes:
New issue create at #473.

tegefaulkes:
Changes:
- The hole punch packet interval has been changed from 1000ms to 50ms.
- `ConnectionReverse` and `ConnectionReverse` has been update with
  - `start` takes `ContextTimed` parameter for cancellation.
  - `endGracefully` uses `timedCancellable` with `endTime` as default timeout.
  - `compose` takes `ContextTimed` parameter for cancellation.
- `Proxy` changes
  - the forward and reverse connection 'address -> Lock' map was replaced with `LockBox`.
  - `openConnectionForward` was updated with `timedCancellable` to support cancellation.
  - `connectForward` was updated with `timedCancellable` to support cancellation.
  - `establishConnectionForward` takes `ContextTimed` parameter for cancellation..
  - `openConnectionReverse` was updated with `timedCancellable` to support cancellation.
  - `connectReverse` was updated with `timedCancellable` to support cancellation.
  - `establishConnectionReverse` takes `ContextTimed` parameter for cancellation.
- `NodeConnectionManager` had some temp changes to interface with the above changes.

CMCDragonkai:
As clarified in the meeting today, the proxy shouldn't be handling failures in the client. The client and proxy trust each other. If the GRPC client fails to respond, that's the client's responsibility to deal with.

PR #470, opened 2022-10-04
feat: added `npm run start` and `npm run dev`

### LINKED ISSUE #469: End-to-end dev loop: setup a script to like dev-server that re-runs Polykey agent upon changes to source code
When developing web applications we often had a hot-reload system that would recompile code and reload the browser upon any fix or feature to any layer of the stack.

This kept us honest and allowed us to see what was going on in the entire application and reminded us how it works.

However for CLI applications like PK, we haven't been doing this, so we don't see how our features/fixes impact the CLI, especially if something has been breaking. All we have are jest tests that target the specific layer we are working on.

So we need to have an end-to-end loop that re-runs Polykey (not just automated tests in jest) that shows us the final output on the terminal, so we can remember how this actually works from the user's perspective.

Then on the end of each PR feature, we can copy paste the output of all these terminal commands to show if anything usability wise has changed.

### Tasks

- [x] 1. Use `nodemon` and integrate typescript compilation or the usage of `ts-node` to run `pk agent start`. This should be fast now that we are using `swc` and `ts-node` with `swc` doesn't do any caching anyway.
- [x] 2. Allow us to clear the app state to do a full reset of the PK.
- [x] 3. Allow us to run CLI commands repeatedly against the re-run agent repeatedly, but this could also be done by the end-developer.

CMCDragonkai:
### Description

This adds `nodemon` to provide us that end to end loop.

Use it like this:

1. First edit the `.env` and add `PK_PASSWORD=abc123` and `PK_NODE_PATH=./tmp/polykey`
2. Enter `nix-shell`
3. `npm run dev`
4. Now you have a monitored agent server that auto restarts upon source code changes. See the `nodemon.json` for all the other config.

Note that addition env variables can be added into `.env` and you must reload your shell to ensure some repeatable environment.

At the very least this gives us a running server as we are changing code, and we can run commands on the CLI against it. We don't have anything that runs automated commands against it, but that's easy enough with just a bash script somewhere that you just run over and over. But that script is likely to be a case by case basis for each feature you're working on.

### Issues Fixed

* Fixes #469 

### Tasks

- [x] 1. Added `nodemon`
- [x] 2. Ignored it in `.npmignore`
- [x] 3. Setup `npm run dev`
- [x] 4. Setup `npm run start`

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,6 +5,7 @@
 /tsconfig.build.json
 /babel.config.js
 /jest.config.js
+/nodemon.json
 /scripts
 /src
 /tests
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,0 +1,12 @@
+{
+  "restartable": "rs",
+  "verbose": true,
+  "watch": [
+    "./src/**",
+    "./package.json",
+    "./package-lock.json",
+    "./tsconfig.json"
+  ],
+  "ext": "ts,js,json",
+  "signal": "SIGTERM"
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,6 +79,7 @@
         "mocked-env": "^1.3.5",
         "nexpect": "^0.6.0",
         "node-gyp-build": "^4.4.0",
+        "nodemon": "^2.0.20",
         "pkg": "5.7.0",
         "prettier": "^2.6.2",
         "shelljs": "^0.8.5",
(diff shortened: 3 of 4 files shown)


CMCDragonkai:
From now on we add comment shows how things affect the end-user for new features. I'll do the first step @tegefaulkes.

[asciicast](https://asciinema.org/a/yLVDWcV7lepmsp472GLAOIkOZ)

CMCDragonkai:
Take note not to upload anything sensitive to asciinema @tegefaulkes that's a quick mistake that leads to key rotation. So you should authenticate to asciinema so it creates little untitled posts.

PR #474, opened 2022-10-05
ci: merge staging to master

maxwell-aiden:
This is an automatic PR generated by the pipeline CI/CD. This will be automatically fast-forward merged if successful.

### Tasks

1. [x] - Enable automated testnet connection tests
2. ~[ ] - Investigate random crashes~ - unnecessary since CLI tests will go to Polykey-CLI
3. [ ] - Testnet nodes should never crash regardless of what connecting nodes are doing


maxwell-aiden:
Pipeline Attempt on 658670966 for 06df14a2dae5198f38d645dbf5458a9c3cd06ab0

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/658670966

CMCDragonkai:
@tegefaulkes the Nat tests failed here. There's a failure that is not just a timeout.

tegefaulkes:
Looking at it now. The initial problem here is that the pings are failing after the default timeout of 20 seconds. 

```
  console.time
    ping1: 22561 ms

      at Object.<anonymous> (tests/nat/DMZ.test.ts:213:15)

  console.log
    {"success":false,"message":"No response received"}

      at Object.<anonymous> (tests/nat/DMZ.test.ts:214:15)

  console.log
    {"type":"ErrorCLINodePingFailed","data":{"message":"No response received","timestamp":"2022-10-06T02:14:59.406Z","data":{},"stack":"ErrorCLINodePingFailed: No response received\n    at /home/faulkes/matrixcode/polykey/js-polykey/src/bin/nodes/CommandPing.ts:72:19\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async CommandPing.<anonymous> (/home/faulkes/matrixcode/polykey/js-polykey/src/bin/CommandPolykey.ts:81:7)\n    at async CommandPolykey.parseAsync (/home/faulkes/matrixcode/polykey/js-polykey/node_modules/commander/lib/command.js:923:5)\n    at async main (/home/faulkes/matrixcode/polykey/js-polykey/src/bin/polykey.ts:54:5)","description":"Node was not online or not found.","exitCode":1}}

      at Object.<anonymous> (tests/nat/DMZ.test.ts:215:15)

```

So it's not related to the bug from yesterday. I need to look at the bin command closer to know why the ping is failing. But right now it's consistent with failing to connect.

tegefaulkes:
It's odd, the core problem is that the connection is timing out with `UTP_ETIMEDOUT`. So it's UTP timing out here after 20 seconds. I'm not sure what change caused this, I need to review the the last merge for that. 

Another odd thing is that setting a shorter timeout of '10000' doesn't change the outcome, I'd expect it to abort with a different timeout error but it doesn't. So there is a bug in the code here somewhere. But that relates to the nodes domain cancellability i'm currently working on so it should be addressed there.

tegefaulkes:
The only changes was updating the network domain with cancellability. The network and nodes tests are passing so the connection logic is still working, including `pingNode`.  So why is it only failing in the NAT testing? Is it the network namespacing? because we're using alternate IP addresses? Because it's in a separate process?

CMCDragonkai:
Does the test work locally? Maybe you do need to fix the bug in the nodes domain first.

tegefaulkes:
The tests are failing locally. the problem doesn't seem to be nodes domain specifically. So far as I can tell the `ForwardConnection` is just failing to connect and timing out, but only in the NAT tests. The problem does overlap enough that I'll need to fix it for the nodes PR. I may encounter a fix as part of that work. 

I'll come back to this, I need to think on it for a little bit.

maxwell-aiden:
Pipeline Attempt on 660898712 for e428206e08f084ed7cf93923f5b28f2ea8854ce6

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/660898712

maxwell-aiden:
Pipeline Attempt on 669366315 for d0ffca03dd5851bcdb74d9de0f0a304cecb8e018

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/669366315

maxwell-aiden:
Pipeline Attempt on 675968580 for 0421f7320788f1906c467e35af5cec43865b08d4

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/675968580

maxwell-aiden:
Pipeline Attempt on 677224551 for b381e29f7bdf1396e091472fd45a46d18c9ee34e

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/677224551

maxwell-aiden:
Pipeline Attempt on 678321421 for bb9f39b8451329be1cd1492afbe05535b87762df

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/678321421

maxwell-aiden:
Pipeline Attempt on 678384368 for 04737bc2b19f42522ba627bc1400f00980f1f0ed

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/678384368

CMCDragonkai:
Make sure any logs left over should be following the rest of the code in their logging structure.

maxwell-aiden:
Pipeline Attempt on 688345460 for 1890801816e188b514e4b48f9882a4c144e875f8

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/688345460

tegefaulkes:
There seems to be a bug causing the seed nodes to crash when receiving certain connections. 

```
{
    "type": "TypeError",
    "data": {
        "message": "Cannot read properties of null (reading 'finishShutdown')",
        "stack": "TypeError: Cannot read properties of null (reading 'finishShutdown')\n    at JSStreamSocket.finishShutdown (node:internal/js_stream_socket:160:12)\n    at node:internal/js_stream_socket:147:14\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)\n    at runNextTicks (node:internal/process/task_queues:65:3)\n    at processImmediate (node:internal/timers:437:9)"
    }
}
```

It is internal to some node implementation. It may be related to this issue. https://github.com/nodejs/node/issues/35695.

For now I'm going to disable the testnet connection tests due to the seed nodes instability.

maxwell-aiden:
Pipeline Attempt on 689345933 for 2e972e77c235e8b122e11e2855ebd098dcf50498

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/689345933

maxwell-aiden:
Pipeline Attempt on 689348586 for 6a18b6bca6b76aaf475e740f5110944cc96de950

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/689348586

CMCDragonkai:
> There seems to be a bug causing the seed nodes to crash when receiving certain connections.
> 
> ```
> {
>     "type": "TypeError",
>     "data": {
>         "message": "Cannot read properties of null (reading 'finishShutdown')",
>         "stack": "TypeError: Cannot read properties of null (reading 'finishShutdown')\n    at JSStreamSocket.finishShutdown (node:internal/js_stream_socket:160:12)\n    at node:internal/js_stream_socket:147:14\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)\n    at runNextTicks (node:internal/process/task_queues:65:3)\n    at processImmediate (node:internal/timers:437:9)"
>     }
> }
> ```
> 
> It is internal to some node implementation. It may be related to this issue. [nodejs/node#35695](https://github.com/nodejs/node/issues/35695).
> 
> For now I'm going to disable the testnet connection tests due to the seed nodes instability.

I think we saw that exception before? Is this is what is causing the random failures? Or is the random failures coming elsewhere?

Are you sure, it's HTTP2? And not the UDP, uTP or elsewhere?

CMCDragonkai:
Also we should try to replicate that, and use `rr` to record the execution.

tegefaulkes:
Yeah, I think we've seen this one before. I'm not sure exactly where it's happening but I found that issue after a quick search.

Triggering this bug might be a little tricky. Right now it seems to be caused when I connect to a seed node with the `testnetConnection.test.ts` test.

maxwell-aiden:
Pipeline Attempt on 689376099 for ab8d7594a907a0c4271fd5c8ad2677bb31480607

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/689376099

CMCDragonkai:
In order to make it easier for us to see if our manual test is succeeding, we need a `pk nodes connections` command that lists out all the active node connections. This means we don't have to keep looking over the logs to see if we managed to connect to the testnet.

If this succeeds and we don't crash the testnet, then testnet connection tests should succeed automatically too. We just need to ensure we are doing the same thing.

However our networking system is getting too complex and it is resulting in hard to figure out bugs. We may need to start on a networking refactoring soon after we merge in the `feature-crypto`.

CMCDragonkai:
We can also add a `pk network connections` which shows active proxy connections. This can be compared to `pk nodes connections`.

CMCDragonkai:
We'll look into network refactoring after feature-crypto (this should eliminate the situations where we have random failures). That will also need to fix MatrixAI/js-mdns#1.

Also try to rule out this https://aws.amazon.com/premiumsupport/knowledge-center/ecs-resolve-outofmemory-errors/ happening. It seems it never gets past 50% memory usage in the service.

Remove the old `polykey-testnet` service and update the cloudwatch dashboard accordingly. We are now using 1 service per `NodeId`.

CMCDragonkai:
@tegefaulkes re-enable the testnet connection test after manual testing succeeds with evidence from `pk nodes connections` and `pk network connections`.

tegefaulkes:
Do we want these new commands to be simple and just print out a list of active connections? or slightly more complex where it streams the data a provides a live list?

CMCDragonkai:
As simple as possible, it's just a quick hack to see if our manual tests work.

**Seems like it takes half a day to create a single command on PK. It involves creating 5 things: the method on the domain, the service handler, the protobuf message, the client call, the CLI command.**

This is far too complex, and is another reason why we need to refactor our RPC system so that creating a new simple command should be a matter of copy pasting a few decorators and take 1 hr max.

CMCDragonkai:
We currently still don't have any automated connection to the testnet tests.

This is due to a connection failure causing a crash on the testnet.

This crash is replicable locally now. However this is not the same crashes that occurring randomly up on the testnet.

@tegefaulkes the feature-crypto is higher priority now, you should work on that rebase on staging here.

You may need to leave the `pk network connections` command off for now. I'll look into the networking/RPC problems while doing the tutorial.

CMCDragonkai:
By getting MatrixAI/Polykey#446 into this staging branch we may have less likelihood of test timeouts. But network connection flakiness will require a proper refactoring before we start to eliminate any kind of timeouts from occurring.

tegefaulkes:
Created a new issue for the seed nodes error at https://github.com/MatrixAI/Polykey/issues/494.

maxwell-aiden:
Pipeline Attempt on 690487239 for ff2d436ff4ca382a6804f8f30a8aa9053fdb38fd

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/690487239

maxwell-aiden:
Pipeline Attempt on 715233785 for a40378b966f4b7bd348f761e62fc58b3c60a4651

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/715233785

CMCDragonkai:
Rerun docs generation too.

CMCDragonkai:
@tegefaulkes what is needed for this to merge to master?

* Release builds (macos/windows is disabled)
* Test passing
* Testnet deployment
* Any integration tests at all? Or is still testnet broken and pending fixes in network and RPC?

maxwell-aiden:
Pipeline Attempt on 716562368 for fcc077966fc566c8b6d202daee6ead82f8e4724a

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/716562368

tegefaulkes:
Looks like it's failing to package the executable with versel/pkg. I might just need to add `tsyringe` to the pkg assets to make it work.

```
> polykey@1.0.1-alpha.0 pkg
> node ./scripts/pkg.js --no-dict=leveldown.js "--output=out" "--bin=polykey" "--node-version=16" "--platform=linux" "--arch=x64"
Configured pkg with:
{
  assets: [
    'dist/**/*.json',
    'node_modules/tslib/**/*.js',
    'node_modules/@matrixai/db/prebuilds/linux-x64/node.napi.node',
    'node_modules/fd-lock/prebuilds/linux-x64/node.napi.node',
    'node_modules/sodium-native/prebuilds/linux-x64/node.napi.node',
    'node_modules/utp-native/prebuilds/linux-x64/node.napi.node'
  ],
  scripts: [ 'dist/workers/polykeyWorker.js', 'dist/bin/polykey-agent.js' ]
}
Running pkg:
pkg dist/bin/polykey.js --config=/tmp/nix-build-polykey-1.0.1-alpha.0-linux-x64.drv-0/polykey/pkg.json --targets=node16-linux-x64 --no-bytecode --no-native-build --public --public-packages='*' --output=out --no-dict=leveldown.js
> pkg@5.7.0
> Error! --no-bytecode and no source breaks final executable
  /tmp/nix-build-polykey-1.0.1-alpha.0-linux-x64.drv-0/polykey/node_modules/tsyringe/node_modules/tslib/tslib.js
  Please run with "-d" and without "--no-bytecode" first, and make
  sure that debug log does not contain "was included as bytecode".
node:internal/errors:841
  const err = new Error(message);
              ^
Error: Command failed: pkg dist/bin/polykey.js --config=/tmp/nix-build-polykey-1.0.1-alpha.0-linux-x64.drv-0/polykey/pkg.json --targets=node16-linux-x64 --no-bytecode --no-native-build --public --public-packages='*' --output=out --no-dict=leveldown.js
    at checkExecSyncError (node:child_process:828:11)
    at Object.execFileSync (node:child_process:863:15)
    at main (/tmp/nix-build-polykey-1.0.1-alpha.0-linux-x64.drv-0/polykey/scripts/pkg.js:157:16) {
  status: 2,
  signal: null,
  output: [ null, null, null ],
  pid: 30701,
  stdout: null,
  stderr: null
}
error: builder for '/nix/store/d8ps7djwfdagga85hn9ngdxd4pn5lknf-polykey-1.0.1-alpha.0-linux-x64.drv' failed with exit code 1
error: build of '/nix/store/3iq6khkrcv54y6skss2j7x796859h9kv-polykey-1.0.1-alpha.0-macos-arm64.drv', '/nix/store/d8ps7djwfdagga85hn9ngdxd4pn5lknf-polykey-1.0.1-alpha.0-linux-x64.drv', '/nix/store/dx3a2ldhlw13dqz03laix0y08n0b9dz2-polykey-1.0.1-alpha.0-win-x64.exe.drv', '/nix/store/jhik1wivq859qcf2fz1290svm98q6sm2-docker-image-polykey-1.0.1-alpha.0.tar.gz.drv', '/nix/store/zz6hxx336jafvhc0y0y1wsv808k5zhdq-polykey-1.0.1-alpha.0-macos-x64.drv' failed
```

CMCDragonkai:
It's the build script that failed, nothing to do with tsyringe. Can you try bundling on your local machine.

tegefaulkes:
It's failing to include the tsyringe source,

```
 Error! --no-bytecode and no source breaks final executable
  /tmp/nix-build-polykey-1.0.1-alpha.0-linux-x64.drv-0/polykey/node_modules/tsyringe/node_modules/tslib/tslib.js
  Please run with "-d" and without "--no-bytecode" first, and make
  sure that debug log does not contain "was included as bytecode".
```

I'll do some testing on the laptop but I need to charge it a little first. I can't build it in my VM.

CMCDragonkai:
Is this because we are using some options that changed, that made it want to bring in tslib?

Note that the `pkg` confguration is saying to `node_modules/tslib/**/*.js`. However this is not even a package dependency.

CMCDragonkai:
Try installing `tslib` as a runtime dependency, then enabling it:

```
{
    "compilerOptions": {
        "importHelpers": true
    }
}
```

In the `tsconfig.json`.

maxwell-aiden:
Pipeline Attempt on 716628520 for fae762b915386b7c337fdc67e2ee27f0075ca026

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/716628520

maxwell-aiden:
Pipeline Attempt on 716645616 for 9f7484d6b31a4cd0fcabd42e56227104b5118012

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/716645616

tegefaulkes:
Ok, after adding `tslib` and `tsyringe` to the dependencies, adding `tsyringe` to the `versel/pkg` assets, and setting `        "importHelpers": true` in the tsconfig. I was able to run nix-build without any errors.

However running the packaged executable resulted in a bus error after fully starting up.

CMCDragonkai:
It's possible that `pkg` doesn't understand esm packages. And that's why we had to do this. Adding tslib and tsyringe is needed primarily due to x509 which uses both. But other packages also use tslib. So we might as well use tslib and take advantage of its runtime utilities.

I've made a comment in https://github.com/MatrixAI/Polykey/issues/268#issuecomment-1342039442 for further investigation into the problem of `pkg` and alternative constructs. Not a priority right now.

The bus error seems like a native code problem. This is a regression we should investigate. You will need to use the chrome debugger: https://nodejs.org/en/docs/guides/debugging-getting-started/ and apply those special flags to the pkg bundled executable which at the end of the day is just a nodejs interpreter.

tegefaulkes:
CI is failing on deployment. The script is out of date since we updated the service definitions. There used to be a single `polykey-testnet` service, however, there are now multiple services that need to be updated. One for each node with the name `polykey-testnet-<NodeId>`. 

There is a complication with this. To avoid hard coding the nodeIds and making this brittle, the script will need to query the existing services and force a new deployment for each one.

CMCDragonkai:
The node IDs are supposed to be derived from the recovery codes stored up in AWS.

No hardcoding is required.

tegefaulkes:
To avoid polluting the history here I'm going to make a new branch for testing the CI.

tegefaulkes:
We're getting new errors on the testnet. How much memory does password hashing need?

```
{
    "type":"Error",
    "data":{
        "message":"password hashing failed, check memory requirements.",
        "stack":"Error: password hashing failed, check memory requirements.\n
            at hashPassword (/lib/node_modules/polykey/dist/keys/utils/password.js:60:29)\n 
            at Object.wrapWithPassword (/lib/node_modules/polykey/dist/keys/utils/symmetric.js:154:53)\n 
            at constructor_.writeKeyPair (/lib/node_modules/polykey/dist/keys/KeyRing.js:541:38)\n 
            at constructor_.setupKeyPair (/lib/node_modules/polykey/dist/keys/KeyRing.js:372:28)\n 
            at async constructor_.start (/lib/node_modules/polykey/dist/keys/KeyRing.js:102:41)\n 
            at async /lib/node_modules/polykey/node_modules/@matrixai/async-init/dist/CreateDestroyStartStop.js:69:38\n 
            at async withF (/lib/node_modules/polykey/node_modules/@matrixai/resources/dist/utils.js:17:16)\n 
            at async Function.createKeyRing (/lib/node_modules/polykey/dist/keys/KeyRing.js:55:9)\n 
            at async Function.createPolykeyAgent (/lib/node_modules/polykey/dist/PolykeyAgent.js:137:22)\n 
            at async /lib/node_modules/polykey/dist/bin/agent/CommandStart.js:199:27"
    }
}
```

maxwell-aiden:
Pipeline Attempt on 716744206 for 10f5de4020764e4ceab24a13bf68d3e00804d046

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/716744206

tegefaulkes:
This could be reduced by using the lower complexity constraints. Ideally we can set this using CLI parameters. I'll look into adding that feature.

tegefaulkes:
I reverted the `"importHelpers": true` change in tsconfig.json. It broke a bunch of tests. Look at https://gitlab.com/MatrixAI/open-source/Polykey/-/commit/9f7484d6b31a4cd0fcabd42e56227104b5118012/pipelines?ref=staging for details.

maxwell-aiden:
Pipeline Attempt on 716770231 for 64414c15b5b0b5cdab52191fda014425d9267d9e

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/716770231

CMCDragonkai:
> I reverted the `"importHelpers": true` change in tsconfig.json. It broke a bunch of tests. Look at https://gitlab.com/MatrixAI/open-source/Polykey/-/commit/9f7484d6b31a4cd0fcabd42e56227104b5118012/pipelines?ref=staging for details.

Isn't that supposed to be a drop in replacement of TSC generated routines? Did those tests pass locally too?

CMCDragonkai:
> This could be reduced by using the lower complexity constraints. Ideally we can set this using CLI parameters. I'll look into adding that feature.

Libsodium provides 2 complexity parameters to password hashing. I used recommended ones. I think we should be able to vary this for our testnet nodes. 250 GiB should have enough memory. Maybe we default to a lower set of constraints.

CMCDragonkai:
250MiB is suitable for microservices deployment.

CMCDragonkai:
> We're getting new errors on the testnet. How much memory does password hashing need?
> 
> ```
> {
>     "type":"Error",
>     "data":{
>         "message":"password hashing failed, check memory requirements.",
>         "stack":"Error: password hashing failed, check memory requirements.\n
>             at hashPassword (/lib/node_modules/polykey/dist/keys/utils/password.js:60:29)\n 
>             at Object.wrapWithPassword (/lib/node_modules/polykey/dist/keys/utils/symmetric.js:154:53)\n 
>             at constructor_.writeKeyPair (/lib/node_modules/polykey/dist/keys/KeyRing.js:541:38)\n 
>             at constructor_.setupKeyPair (/lib/node_modules/polykey/dist/keys/KeyRing.js:372:28)\n 
>             at async constructor_.start (/lib/node_modules/polykey/dist/keys/KeyRing.js:102:41)\n 
>             at async /lib/node_modules/polykey/node_modules/@matrixai/async-init/dist/CreateDestroyStartStop.js:69:38\n 
>             at async withF (/lib/node_modules/polykey/node_modules/@matrixai/resources/dist/utils.js:17:16)\n 
>             at async Function.createKeyRing (/lib/node_modules/polykey/dist/keys/KeyRing.js:55:9)\n 
>             at async Function.createPolykeyAgent (/lib/node_modules/polykey/dist/PolykeyAgent.js:137:22)\n 
>             at async /lib/node_modules/polykey/dist/bin/agent/CommandStart.js:199:27"
>     }
> }
> ```

Can we also capture that error and give it a specific exception class? Find the right place to capture it.

CMCDragonkai:
1. See if `importHelpers` also causes test failures locally?
2. Propagate the password ops and mem limit to the CLI:
   
   ```
   /**
    * Use the `min` limit during testing to improve performance.
    */
   const passwordOpsLimits: {
     min: PasswordOpsLimit;
     max: PasswordOpsLimit;
     interactive: PasswordOpsLimit;
     moderate: PasswordOpsLimit;
     sensitive: PasswordOpsLimit;
   } = {
     min: sodium.crypto_pwhash_OPSLIMIT_MIN,
     max: sodium.crypto_pwhash_OPSLIMIT_MAX,
     interactive: sodium.crypto_pwhash_OPSLIMIT_INTERACTIVE,
     moderate: sodium.crypto_pwhash_OPSLIMIT_MODERATE,
     sensitive: sodium.crypto_pwhash_OPSLIMIT_SENSITIVE,
   };
   
   /**
    * Use the `min` limit during testing to improve performance.
    */
   const passwordMemLimits: {
     min: PasswordMemLimit;
     max: PasswordMemLimit;
     interactive: PasswordMemLimit;
     moderate: PasswordMemLimit;
     sensitive: PasswordMemLimit;
   } = {
     min: sodium.crypto_pwhash_MEMLIMIT_MIN,
     max: sodium.crypto_pwhash_MEMLIMIT_MAX,
     interactive: sodium.crypto_pwhash_MEMLIMIT_INTERACTIVE,
     moderate: sodium.crypto_pwhash_MEMLIMIT_MODERATE,
     sensitive: sodium.crypto_pwhash_MEMLIMIT_SENSITIVE,
   };
   
   /**
    * These are the default computational parameters for password hashing.
    * They can be changed to increase or decrease the computational cost.
    * This ensures that deriving a key from a password uses
    * 256 MiB of RAM and 0.7 seconds on a 2.8 GHz Intel Core i7.
    * These need to be consistent to ensure the same hash is produced.
    */
   const passwordOpsLimitDefault = passwordOpsLimits.moderate;
   const passwordMemLimitDefault = passwordMemLimits.moderate;
   ```
   
   Ok there'a challenge with this, the ops limit and mem limit will affect the resulting password hash.
   
   Now the password hash isn't saved anywhere except in-memory. So as long as PK agent is alive, these parameters are "saved" and therefore that works.
   
   However if I stop the agent, and restart the agent, and pass in a password, then how does PK know what the original ops limit and mem limit is?
   
   Well in the `keys/utils/symmetric.ts` we have `wrapWithPassword` and `unwrapWithPassword`. Notice the `wrapWithPassword` saves the `ops`, `mem` and `salt` in the protected header. So in the `unwrapWithPassword` it actually uses these saved parameters to try and decrypt.
   
   So anyway I think if you expose these options like `--password-ops-limit` and `--password-mem-limit` at the very beginning during agent start and bootstrap, then it could work. However the question is whether this overrides these 2 parameters when the password is just being checked.
   
   Let's say there are a couple places where this should be limited to work:
   
   * When the agent starts from scratch - automatic bootstrapping
   * When bootstrapping the agent
   * When the agent starts from an existing state, then these can be passed to the `unwrapWithPassword` which is only used as constraints but not as the actual limits
   
   ````
     // If the ops and mem setting is greater than the limit
     // then it may be maliciously trying to DOS this agent
     if (
       header.ops < passwordOpsLimits.min ||
       header.ops > opsLimit ||
       header.mem < passwordMemLimits.min ||
       header.mem > memLimit
     ) {
       return;
     }
   ```
   In every other case it should have no effect.
3. Node IDs should be deterministically derived by the recovery codes in AWS. Can you confirm this?

tegefaulkes:
1. Yes I tested this locally, it was failing.
3. I can only check this on aws after fixing 2.

tegefaulkes:
The `unwrapWithPassword` will check the header to get the mem and ops limits. The password hash stored inside the `KeyRing` just uses the mem and ops limit set in the constructor. Given this these two hashes can be different if we create the agent with one setting and start it later with another.

The question is, is this a problem? They are used seperately. AFAIK the hash in the `KeyRing` is only used for checking sessions? Ultimately starting later with a different limits would only affect the security of the session tokens? Do we want to bother with enforcing the same limits as the keyPair every time we start?

Double checking the code. It looks like the `KeyRing` will fail to start if we use a smaller limit than the keyPair was generated with.

CMCDragonkai:
Let's say you initially hash the password with (A, B) as the ops and mem limit.

For you to get the equivalent hash, you must keep using (A, B). So we have these situations:

1. When first bootstrap (agent start and bootstrap)
2. When we launch an agent from existing state (this uses the ops/mem saved in the JWK)
3. When we reauthenticate to the agent to perform some call (this needs to use the ops/mem saved in-memory of KeyRing).

So if we add `--password-mem-limit` and `--password-ops-limit`, then this must apply in situations 1. and 2., but not in situation 3. In situation 2. there's a special edge case where it only applies as a **constraint** and not as the actual mem and ops limit.

So in 1. it works as you'd expect. The (A, B) is set in both the on-disk JWK and in-memory KeyRing. In 2. it is only a constraint, it actually just uses it from the on-disk JWK, and the in-memory setting **must be copied** from the on-disk JWK. In 3. it has no effect whatsoever, as we always only use the in-memory settings.

CMCDragonkai:
A simpified version of the above 3 situations is where we only apply it to situation 1. And just ignore the parameters for 2 and 3.

CMCDragonkai:
> 1. Yes I tested this locally, it was failing.
> 2. I can only check this on aws after fixing 2.

Can you try enabling both `downlevelIteration` and `importHelpers` and see if local tests pass: https://www.typescriptlang.org/tsconfig#downlevelIteration

CMCDragonkai:
The exception for the password hash should be called `ErrorPasswordHash`, then just give `cause: e`. We can't match on the code or message (message is too brittle). And static description can indicate potential memory limit or other problems. This cannot be tested using jest. We'll just test in production.

CMCDragonkai:
So for the `deploy-service.sh` script. You'll going to need to use `--output json` then use `--query` to basically JQ that output to get exactly what you want. You may end up with an JSON array (maybe there's a way to exact an set of lines), but you can use `jq` to convert that to line-items. https://stackoverflow.com/questions/71518883/convert-json-to-a-newline-delimited-list-of-all-the-things-tagged-name-with-jq

Then you can use xargs to "map" each line to dereferencing to the service name, then xargs again to map to update-service. https://stackoverflow.com/questions/199266/make-xargs-execute-the-command-once-for-each-line-of-input

So then the script only requires access to the cluster name as a hard-coded entry.

CMCDragonkai:
The DNS resolution may not be suitable portable right now. We should 2 variants of this:

```
resolveHostname - pass in a boolean to switch between the 2
resolveHostnameWithOS
resolveHostnameWithDNS
```

And default to the OS version:

https://nodejs.org/api/dns.html#dnspromiseslookuphostname-options

This may be more suitable for as we are deploying to CI/CD and testnet situations, some places may not be correctly setting up their DNS servers.

Furthermore... right now resolving `localhost` involves a real DNS query packet. This is seems inefficient, since usually the OS would shortcut this and respond as 127.0.0.1.

The flag could be `--dns-resolve`? We can decide on this later.

CMCDragonkai:
> Let's say you initially hash the password with (A, B) as the ops and mem limit.
> 
> For you to get the equivalent hash, you must keep using (A, B). So we have these situations:
> 
> 1. When first bootstrap (agent start and bootstrap)
> 2. When we launch an agent from existing state (this uses the ops/mem saved in the JWK)
> 3. When we reauthenticate to the agent to perform some call (this needs to use the ops/mem saved in-memory of KeyRing).
> 
> So if we add `--password-mem-limit` and `--password-ops-limit`, then this must apply in situations 1. and 2., but not in situation 3. In situation 2. there's a special edge case where it only applies as a **constraint** and not as the actual mem and ops limit.
> 
> So in 1. it works as you'd expect. The (A, B) is set in both the on-disk JWK and in-memory KeyRing. In 2. it is only a constraint, it actually just uses it from the on-disk JWK, and the in-memory setting **must be copied** from the on-disk JWK. In 3. it has no effect whatsoever, as we always only use the in-memory settings.

On reflection:

1. We don't need to override the in-memory ops and mem from the unwrapped jwk for situation 2. Because the hash is only in-memory.
2. I think the check for the `opsLimit` and `memLimit` in the `unwrapWithPassword` is not necessary, and can cause a problem if the user sets a limit higher than the default limit at the beginning.

maxwell-aiden:
Pipeline Attempt on 717837701 for cbb07b56d2385bab2dc9ef080289fb6cd86df454

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/717837701

tegefaulkes:
Ok, that's the keys parameters added and hopefully the `deploy-service.sh` script fixed. To fix the password hashing problem on the testnet. The task definitions need to be updated to set the following env variables. `PK_PASSWORD_OPS_LIMIT=min` `PK_PASSWORD_MEM_LIMIT=min`. The valid options are `min`, `max`, `interactive`, `moderate` and `sensitive`.

The `deploy-service` script is now using `jq` to parse the JSON output of the `aws ecs list-services` command. Turns out that `jq` has built in regex so it was really simple to capture the part of the ARN that was needed.

maxwell-aiden:
Pipeline Attempt on 717874909 for 2b8f96e3230ecefb9e2952f0891bc6ef88109316

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/717874909

tegefaulkes:
Remaining TODOs
1. Updating task definitions for the test net to override the password hashing limits.
2. Check and fix the tslib stuff.
3. check the deterministic keys generation for the testnet. Might be different nodeIds now.
4. Check if the testing is faster now with the new key generation.
5. Check the CI is passing now. I updated the deploy-service script to use JQ, It should work but now but needs to be verified. Any tests after deployment haven't run yet so I don't know if they work.
6. DNS resolution fix, Includes fixing any tests that use 'localhost' as Host as the address. should be '127.0.0.1' as Host

maxwell-aiden:
Pipeline Attempt on 766247512 for 634222119c96382936fe30382899ec949fe4cc82

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/766247512

tegefaulkes:
I'm looking over the bug where the agent doesn't fully stop and end the process. I found that the problem is that the `WorkerManager` is still running and keeping the process open. So we simply need to destroy the worker manager somehow when running `pk agent stop`. 

The problem is, the worker manager is managed externally to the `PolykeyAgent`. We start the agent and then pass the worker manager in with `pkAgent.setWorkerManager(workerManager)`. The agent doesn't hold a reference to the worker manager as it's just passed along the the `VaultManager` and `DB` `setWorkerManager` methods. 

The `pk agent start` command creates the `PolykeyAgent`, creates the `WorkerManager` and then sets it with `setWorkerManager`. The command doesn't do anything at this point. In fact it essentially ends here. The only reason the process doesn't end here is that the `GRPC` and `WorkerManager` is holding the process open.  So within the `pk agent start` command we don't know when the `PolykeyAgent` ends. The `PolykeyAgent.stop()` doesn't have a reference to the `WorkerManager` to destroy it and the agent isn't managing the `workerManager'`s lifecycle anyway. The GRPC handler for the `PK agent stop` command doesn't have a reference to the `workerManager` to destroy it either.

So the problem here is, I don't have a clean way to destroy the `WorkerManager`. I can think of 3 ways to do it.
1. The `pk agent start` command needs a way to know the `PolykeyAgent` has stopped. Likely via a resolving promise. It then can clean up the `WorkerManager` in a finally block. 
2. The  `PolykeyAgent` can hold a reference to the `WorkerManager` when it is set. It can then `unsetWorkerManagger()` when stopping and then destroy the worker manager.
3. The `pk agent stop` grpc handler calls `pkAgent.stop()`. It could also destroy the `WorkerManager`. Getting the reference here would be the messiest option though.

I'm leaning towards option 1. in the current design. If the `pk agent start` creates the worker manager then it's up to that command to properly manage the lifecycle of it. So I'd just need to implement a Promise that resolves when the PolykeyAgent ends.

tegefaulkes:
We're going with option 2 where the `PolykeyAgent` manages the lifecycle of the `WorkerManager` now.

CMCDragonkai:
The original reason for `WorkerManager` lifecycle outside of Pk agent was due to bootstrapping where our original key generation used the workers.

Now it no longer does this, libsodium takes care of this.

Therefore the `WorkerManager` lifecycle can be encapsulated into the `PolykeyAgent`.

Furthermore then the options to start a `WorkerManager` or not would be determined by creation options:

* `undefined` means no workers
* `0` means all cores
* `X` means X cores

The `PolykeyAgent` should default to `undefined`.

But the CLI command `pk agent start` should default to `0` all cores.

CMCDragonkai:
By doing this, we end up ensuring worker manager stays within PK agent lifecycle, so then `pk agent stop` will actually cause the PK agent process to shutdown.

It was a good catch of this, since not doing this would result in a process leak.

tegefaulkes:
We want to default to all cores and specifying `--workers 0` means all cores, how do we specify no cores via the CLI? `--workers none`?

tegefaulkes:
Weird behaviour when specifying negative workers. The agent hangs when stopping. So does the `pk agent stop` command. It pretty much locks up. as a fix I've made it so that negative numbers just doesn't start the `WorkerManager`.  Or do we want this as a explicit error?

CMCDragonkai:
So `--workers 0` should be all cores. You could use `--no-workers` to turn it off. Because the default is `--workers 0`.

CMCDragonkai:
Or make it generic: `--workers false|null|none|no`.

tegefaulkes:
It's fixed now.

maxwell-aiden:
Pipeline Attempt on 768172967 for f3092dd0b1d618c5910be95c3e30c6ca49d1bc25

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/768172967

maxwell-aiden:
Pipeline Attempt on 776659712 for dab5cd8a937cabb0eec2ac15a44ab3b7c3fdb25a

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/776659712

maxwell-aiden:
Pipeline Attempt on 776719016 for aaaae016047d19a02d1e358b8a2cfa866df9d1b1

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/776719016

CMCDragonkai:
I think once swc is integrated into the jest tests directly, then everything should work fine, but one should override the CLI parameters to ensure workers aren't running during bin tests otherwise it would use up too many resources.

maxwell-aiden:
Pipeline Attempt on 777801567 for c68d6f9cda8b96535ae4922bc5b6d16811bda7ae

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/777801567

maxwell-aiden:
Pipeline Attempt on 777814045 for 36d518f4398f48beb4f66841aa89330430d2b04a

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/777814045

maxwell-aiden:
Pipeline Attempt on 792052987 for 126957095085b2b32b819ecfe265e2c9eac2b84c

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/792052987

CMCDragonkai:
1. What's needed for staging to pass?
2. MatrixAI/Polykey#515 needs to be working on this branch.
3. Issues relating to RPC migration like TLS rotation must work here.

CMCDragonkai:
@tegefaulkes please update the state of staging here in regards to the above comment. What's still missing here for staging to merge.

I think agent migration is necessary and replacement of grpc and network by agent RPC and js-quic here too.

CMCDragonkai:
Oh and the import of JSON stream parser. That's still pending here MatrixAI/Polykey#516.

maxwell-aiden:
Pipeline Attempt on 816503534 for 4e2ab406e035b9d307122502ef331018b570ea11

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/816503534

maxwell-aiden:
Pipeline Attempt on 829644425 for 67289cd4aaa982f9ad95b090b6609d080bb5f7a4

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/829644425

maxwell-aiden:
Pipeline Attempt on 857985687 for 1c65304b62fb7e89a28eb1767eadfb333ee0e1fc

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/857985687

maxwell-aiden:
Pipeline Attempt on 858056066 for 25b1bac96dd4b8eb82372be0b303d7be03c5f397

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/858056066

maxwell-aiden:
Pipeline Attempt on 865244286 for 78a9faa6119a190174c3a880679b9812104ecefa

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/865244286

maxwell-aiden:
Pipeline Attempt on 881297680 for f25ef30560a96183629fa9d1e50f93a0cda58ccf

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/881297680

maxwell-aiden:
Pipeline Attempt on 881546106 for b901d25de167d2ea8f686ad844f6cb5ff5501550

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/881546106

CMCDragonkai:
@tegefaulkes breaking changes:

1. db and async-locks need `{ timer: X }` where `X` can be a literal number or `Timer` for timeouts now
2. EFS has changed how it deals with non-encapsulated dependencies, the lifecycle of DB... etc that is injected into EFS on creation is now the users's responsibility, EFS won't stop the DB. This is relevant for vaults domain.
3. Sanity check all other things... CI/CD failure due to pkg. That can be resolved by migrating all CLI related things to Polykey-CLI - which is kicking the can down the road... but we will deal with that separately.

maxwell-aiden:
Pipeline Attempt on 923929304 for e34fcd68e500df849a26997078c0ec9d3d097628

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/923929304

maxwell-aiden:
Pipeline Attempt on 925471308 for 48adbca190ad943c23fa1bb293c4fec53fde8f74

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/925471308

maxwell-aiden:
Pipeline Attempt on 927152983 for a60bcce05e7cd8808b651dac5e39dd80bb9cf8ba

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/927152983

maxwell-aiden:
Pipeline Attempt on 951440760 for fc82a3d43578156c4041234f40ec213459977764

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/951440760

CMCDragonkai:
Once #534 is merged, this is expected to pass and merge into master. However CI deployment to testnet and mainnet will be moved to Polykey-CLI repository.

maxwell-aiden:
Pipeline Attempt on 954294586 for f7708f6f748ee816591c3160f00ed9c541700100

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/954294586

maxwell-aiden:
Pipeline Succeeded on 954294586 for f7708f6f748ee816591c3160f00ed9c541700100

https://gitlab.com/MatrixAI/open-source/Polykey/-/pipelines/954294586

Issue #444, opened 2022-09-02
Integrate Observables to coordinate push-based Dataflow (with reactive properties)


CMCDragonkai:
### Specification

The `EventEmitter` is a nodejs native construct. To be more compatible, we should be using Web APIs like `EventTarget`.

We've started to do this in https://github.com/MatrixAI/Polykey/pull/438#issuecomment-1235041552.

There are some differences:

* https://nodejs.org/api/events.html#nodejs-eventtarget-vs-dom-eventtarget
* https://cameronnokes.com/blog/build-your-own-event-emitter-using-only-native-dom-apis/
* https://www.nearform.com/blog/node-js-and-the-struggles-of-being-an-eventtarget/
* https://nodejs.org/api/events.html#eventtarget-error-handling

Our own usage of `EventBus` is pretty simple. It does however add the ability to emit events asynchronously and return a promise to indicate when it has been received and executed. Making it closer to an observer pattern. I believe we will want to use `EventBus` in a number of ways:

1. Default usecase of a synchronous observer pattern (by default, emitting an event equals executing all synchronous listeners before proceeding)
2. Asynchronous pub/sub (emitting events inside asynchronous code, and expecting handlers to be asynchronous)
3. Asynchronous observer pattern, emitting events and then waiting on all listeners to complete

Basically the event emitter already has the ability to both synchronous listeners and asynchronous listeners. The event bus just extended event emitter with the ability to emit asynchronously compared to emitting synchronously.

Once we change to using event target, the same exact usecases must still be available.

Still our primary usecase will be to wire up reactivity between the different domains in Polykey without hardcoding them. This represents a push-based API.

### Additional context

* https://github.com/MatrixAI/Polykey/pull/438#issuecomment-1235041552 - using event target as part of abort signal and also in our tasks system
* https://stackoverflow.com/questions/24340892/the-difference-between-a-message-queue-eventbus-and-a-pub-sub?noredirect=1&lq=1
* MatrixAI/Polykey#386 - dealing with root key pair changes
* #237 - Streaming has an impact on pagination APIs, both styles are useful

### Tasks

1. Create `js-events` to introduce a `Evented` decorator that can be used to augment classes with `EventTarget` functionality, and `AbstractEvent` to allow a common foundation for our typed events.
[...]

PR #491, opened 2022-10-31
Multi-Host DNS and Multi NodeID resolution - for network entry and general usage

### LINKED ISSUE #483: feat: generic many to many connection ability for `NodeConnectionManager`
### Specification

The `NodeConnectionManager` needs a new method to establish multiple connections at once given a set of `NodeId` and addresses. This will be similar to `withConnF` and `withConnG` but it can take an array of desired `Nodeids` and a second array of addresses. It will then attempt to establish connections with the set of addresses and find any node in the provided set of nodes.

This would enable us to do generic network entry were we can specify a single DNS  hostname that resolves to all of our seed nodes. Then we can connect to one or more of them at the same time using this method.

We will also need some options for this. In some situations we can want all possible connects, any of them, the first 3 connections or just the first connection made.

This would depend on the ability to resolve a hostname to multiple 'A' or 'AAAA' records.

```
withMultiConnF([node1, node2, node3], [testnet.polykey.io], (conns: Map<NodeId, NodeConnection>) => {});
```

### Additional context

* Related MatrixAI/Polykey-CLI#71 
* * Related MatrixAI/Polykey#484 

### Tasks

1. Add a feature to the `NodeConnectionManager` to generically establish one or more connections when looking for one or more nodes.

### LINKED ISSUE #484: feat: resolve `hostname` to multiple addresses when connecting to a node
### Specification

We need to expand the functionality of establishing connections to handle multiple possible addresses when resolving a host name. The most notable use case here is when connecting to the test net. The hostname `testnet.polykey.io` can resolve to all of our seed nodes. We need to resolve this to multiple `A` or `AAAA` records and attempt connections to each one when connecting to a node.

### Additional context

* Related MatrixAI/Polykey-CLI#71 
* Related MatrixAI/Polykey#483 

### Tasks

1. Add ability to resolve a hostname to multiple addresses.
2. add functionality to connect to multiple addresses when connecting to a node.

CMCDragonkai:
### Description

This brings in `resolveHostname` that will resolve a hostname to multiple IPs including IPv4 and IPv6, and do a BFS over CNAMEs.

It uses the `dns.resolve*` functions instead of the lookup. This means it doesn't abide by the local OS's get name lookup. Which also means it cannot be affected by `/etc/hosts`. I didn't think `/etc/hosts` is useful here. This is also better our performance is better here. See: https://nodejs.org/api/dns.html#implementation-considerations

However it still uses the OS provided initial DNS server configuration. Which means local DNS servers are still used, and local resolvers can still cache the records.

In the future, we can add a flag to switch to using lookup for edge cases

### Issues Fixed

* Fixes MatrixAI/Polykey#484
* Fixes MatrixAI/Polykey#483

### Tasks

- [x] 1. Replace `resolveHost` with `resolveHostname`
- [x] 2. Add tests to `resolveHostname`, use common domains like `google.com` and `localhost`
- [x] 3. Fix up all usages of this in relation to MatrixAI/Polykey#483
- [x] 4. Test the timed and cancellable variants on this, you may need to pass a custom server set 

todo: 
1. [x] 1. improve logging for life cycle NCM and the creation/destruction flow of the node connections 
2. [x] 2. defence in depth for using our own nodeId in the node graph. We need to filter out our own nodeId from information when we receive it, resolve it or attempt to use it.
3. [x] 3. relaying nodes should reject requests from nodes to themselves. Should check the nodeId and the address.
4. [x] 4. Make the node connection creation for the multi-node resolving serial instead of concurrent. Need to double check the destroy callback logic here.
5. [x] 5. ICE needs to be extracted from the NC and done along side its creation instead.
6. [x] 6. Special case seed nodes to avoid TTL clean up on connections.
7. ~Any of the odd methods used within`syncNodeGraph` should be in-lined to avoid noise. Also add logging outlining the process of syncing the node graph.~ Anything I could in-line is used in multiple places.
8. [x] 8. Filter out any Ipv6 addresses when resolving hostnames.
9. [x] 9. Add key path to logger messages.
10. [x] 10. Add detailed logging to `syncNodeGraph`
11. [X] 11. Separate timeouts for each sub connection.
12. [x] 12. All nodes need to re-establish to connection seed nodes when they disconnect.
13. [x] 13. Serial connection establishment needs to be updated to work concurrently.
14. [x] 14. Proxy needs to cancel forward connection establishment if the underlying GRPC client connection is destroyed.
15. [x] 15. Special case GC of seed nodes in the NodeGraph.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -573,6 +573,8 @@ class PolykeyAgent {
         PolykeyAgent.eventSymbols.Proxy,
         async (data: ConnectionData) => {
           if (data.type === 'reverse') {
+            console.log(`${nodesUtils.encodeNodeId(this.keyManager.getNodeId())} VS ${nodesUtils.encodeNodeId(data.remoteNodeId)}`)
+            if (this.keyManager.getNodeId().equals(data.remoteNodeId)) return;
             const address = networkUtils.buildAddress(
               data.remoteHost,
               data.remotePort,
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -573,7 +573,6 @@ class PolykeyAgent {
         PolykeyAgent.eventSymbols.Proxy,
         async (data: ConnectionData) => {
           if (data.type === 'reverse') {
-            console.log(`${nodesUtils.encodeNodeId(this.keyManager.getNodeId())} VS ${nodesUtils.encodeNodeId(data.remoteNodeId)}`)
             if (this.keyManager.getNodeId().equals(data.remoteNodeId)) return;
             const address = networkUtils.buildAddress(
               data.remoteHost,
--- a/src/agent/service/nodesHolePunchMessageSend.ts
+++ b/src/agent/service/nodesHolePunchMessageSend.ts
@@ -71,7 +71,7 @@ function nodesHolePunchMessageSend({
               call.request.getProxyAddress(),
             );
             logger.debug(
-              `Received signalling message to target ${call.request.getSrcId()}@${host}:${port}`,
+              `Received signaling message to target ${call.request.getSrcId()}@${host}:${port}`,
             );
             // Ignore failure
             try {
@@ -81,7 +81,7 @@ function nodesHolePunchMessageSend({
             }
           } else {
             logger.error(
-              'Received signalling message, target information was missing, skipping reverse hole punch',
+              'Received signaling message, target information was missing, skipping reverse hole punch',
             );
           }
         } else if (await nodeManager.knowsNode(sourceId, tran)) {
@@ -95,19 +95,19 @@ function nodesHolePunchMessageSend({
           // Checking if the source and destination are the same
           if (sourceId?.equals(targetId)) {
             // Logging and silently dropping operation
-            logger.warn('Signalling relay message requested signal to itself');
+            logger.warn('Signaling relay message requested signal to itself');
             callback(null, response);
             return;
           }
           call.request.setProxyAddress(proxyAddress);
           logger.debug(
-            `Relaying signalling message from ${srcNodeId}@${
+            `Relaying signaling message from ${srcNodeId}@${
               connectionInfo!.remoteHost
             }:${
               connectionInfo!.remotePort
             } to ${targetNodeId} with information ${proxyAddress}`,
           );
-          await nodeConnectionManager.relaySignallingMessage(call.request, {
+          await nodeConnectionManager.relaySignalingMessage(call.request, {
             host: connectionInfo!.remoteHost,
             port: connectionInfo!.remotePort,
           });
--- a/src/agent/service/nodesHolePunchMessageSend.ts
+++ b/src/agent/service/nodesHolePunchMessageSend.ts
@@ -92,6 +92,13 @@ function nodesHolePunchMessageSend({
             connectionInfo!.remoteHost,
             connectionInfo!.remotePort,
           );
+          // Checking if the source and destination are the same
+          if (sourceId?.equals(targetId)) {
+            // Logging and silently dropping operation
+            logger.warn('Signalling relay message requested signal to itself');
+            callback(null, response);
+            return;
+          }
           call.request.setProxyAddress(proxyAddress);
           logger.debug(
             `Relaying signalling message from ${srcNodeId}@${
--- a/src/nodes/NodeConnection.ts
+++ b/src/nodes/NodeConnection.ts
@@ -75,6 +75,7 @@ class NodeConnection<T extends GRPCClient> {
       clientFactory,
       nodeConnectionManager,
       destroyCallback = async () => {},
+      destroyTimeout = 2000,
       logger = new Logger(this.name),
     }: {
       targetNodeId: NodeId;
@@ -86,6 +87,7 @@ class NodeConnection<T extends GRPCClient> {
       clientFactory: (...args) => Promise<T>;
       nodeConnectionManager: NodeConnectionManager;
       destroyCallback?: () => Promise<void>;
+      destroyTimeout?: number;
       logger?: Logger;
     },
     @context ctx: ContextTimed,
@@ -136,19 +138,21 @@ class NodeConnection<T extends GRPCClient> {
       }
       // TODO: this needs to be updated to take a context,
       //  still uses old timer style.
+      const clientLogger = logger.getChild(clientFactory.name);
       client = await clientFactory({
         nodeId: targetNodeId,
         host: targetHost,
         port: targetPort,
         proxyConfig: proxyConfig,
         // Think about this
-        logger: logger.getChild(clientFactory.name),
+        logger: clientLogger,
         destroyCallback: async () => {
+          clientLogger.info(`GRPC client triggered destroyedCallback`);
           if (
             nodeConnection[asyncInit.status] !== 'destroying' &&
             !nodeConnection[asyncInit.destroyed]
           ) {
-            await nodeConnection.destroy();
+            await nodeConnection.destroy({ timeout: destroyTimeout });
           }
         },
         // FIXME: this needs to be replaced with
@@ -158,7 +162,7 @@ class NodeConnection<T extends GRPCClient> {
       // 5. When finished, you have a connection to other node
       // The GRPCClient is ready to be used for requests
     } catch (e) {
-      await nodeConnection.destroy();
+      await nodeConnection.destroy({ timeout: destroyTimeout });
       // If the connection times out, re-throw this with a higher level nodes exception
       if (e instanceof grpcErrors.ErrorGRPCClientTimeout) {
         throw new nodesErrors.ErrorNodeConnectionTimeout(e.message, {
@@ -198,15 +202,20 @@ class NodeConnection<T extends GRPCClient> {
     this.destroyCallback = destroyCallback;
   }
 
-  public async destroy() {
+  public async destroy({
+    timeout,
+  }: {
+    timeout?: number;
+  } = {}) {
     this.logger.info(`Destroying ${this.constructor.name}`);
     if (
       this.client != null &&
       this.client[asyncInit.status] !== 'destroying' &&
       !this.client[asyncInit.destroyed]
     ) {
-      await this.client.destroy();
+      await this.client.destroy({ timeout });
     }
+    this.logger.info(`${this.constructor.name} triggered destroyedCallback`);
     await this.destroyCallback();
     this.logger.info(`Destroyed ${this.constructor.name}`);
   }
--- a/src/nodes/NodeConnection.ts
+++ b/src/nodes/NodeConnection.ts
@@ -44,6 +44,7 @@ class NodeConnection<T extends GRPCClient> {
       proxy,
       clientFactory,
       destroyCallback = async () => {},
+      destroyTimeout,
       logger = new Logger(this.name),
     }: {
       targetNodeId: NodeId;
@@ -53,6 +54,7 @@ class NodeConnection<T extends GRPCClient> {
       proxy: Proxy;
       clientFactory: (...args) => Promise<T>;
       destroyCallback?: () => Promise<void>;
+      destroyTimeout?: number;
       logger?: Logger;
     },
     ctx?: Partial<ContextTimed>,
--- a/src/nodes/NodeConnection.ts
+++ b/src/nodes/NodeConnection.ts
@@ -1,10 +1,8 @@
 import type { NodeId } from './types';
 import type { Host, Hostname, Port } from '../network/types';
-import type KeyManager from '../keys/KeyManager';
 import type { Certificate, PublicKey, PublicKeyPem } from '../keys/types';
 import type Proxy from '../network/Proxy';
 import type GRPCClient from '../grpc/GRPCClient';
-import type NodeConnectionManager from './NodeConnectionManager';
 import type { ContextTimed } from '../contexts/types';
 import type { PromiseCancellable } from '@matrixai/async-cancellable';
 import Logger from '@matrixai/logger';
@@ -44,9 +42,7 @@ class NodeConnection<T extends GRPCClient> {
       targetPort,
       targetHostname,
       proxy,
-      keyManager,
       clientFactory,
-      nodeConnectionManager,
       destroyCallback = async () => {},
       logger = new Logger(this.name),
     }: {
@@ -55,9 +51,7 @@ class NodeConnection<T extends GRPCClient> {
       targetPort: Port;
       targetHostname?: Hostname;
       proxy: Proxy;
-      keyManager: KeyManager;
       clientFactory: (...args) => Promise<T>;
-      nodeConnectionManager: NodeConnectionManager;
       destroyCallback?: () => Promise<void>;
       logger?: Logger;
     },
@@ -71,9 +65,7 @@ class NodeConnection<T extends GRPCClient> {
       targetPort,
       targetHostname,
       proxy,
-      keyManager,
       clientFactory,
-      nodeConnectionManager,
       destroyCallback = async () => {},
       destroyTimeout = 2000,
       logger = new Logger(this.name),
@@ -83,9 +75,7 @@ class NodeConnection<T extends GRPCClient> {
       targetPort: Port;
       targetHostname?: Hostname;
       proxy: Proxy;
-      keyManager: KeyManager;
       clientFactory: (...args) => Promise<T>;
-      nodeConnectionManager: NodeConnectionManager;
       destroyCallback?: () => Promise<void>;
       destroyTimeout?: number;
       logger?: Logger;
@@ -118,24 +108,6 @@ class NodeConnection<T extends GRPCClient> {
     });
     let client: T;
     try {
-      // Start the hole punching only if we are not connecting to seed nodes
-      const seedNodes = nodeConnectionManager.getSeedNodes();
-      const isSeedNode = !!seedNodes.find((nodeId) => {
-        return nodeId.equals(targetNodeId);
-      });
-      if (!isSeedNode) {
-        // FIXME: this needs to be cancellable.
-        //  It needs to timeout as well as abort for cleanup
-        void Array.from(seedNodes, (seedNodeId) => {
-          return nodeConnectionManager.sendSignallingMessage(
-            seedNodeId,
-            keyManager.getNodeId(),
-            targetNodeId,
-            undefined,
-            ctx,
-          );
-        });
-      }
       // TODO: this needs to be updated to take a context,
       //  still uses old timer style.
       const clientLogger = logger.getChild(clientFactory.name);
--- a/src/nodes/NodeConnection.ts
+++ b/src/nodes/NodeConnection.ts
@@ -121,7 +121,7 @@ class NodeConnection<T extends GRPCClient> {
         // Think about this
         logger: clientLogger,
         destroyCallback: async () => {
-          clientLogger.info(`GRPC client triggered destroyedCallback`);
+          clientLogger.debug(`GRPC client triggered destroyedCallback`);
           if (
             nodeConnection[asyncInit.status] !== 'destroying' &&
             !nodeConnection[asyncInit.destroyed]
@@ -189,7 +189,7 @@ class NodeConnection<T extends GRPCClient> {
     ) {
       await this.client.destroy({ timeout });
     }
-    this.logger.info(`${this.constructor.name} triggered destroyedCallback`);
+    this.logger.debug(`${this.constructor.name} triggered destroyedCallback`);
     await this.destroyCallback();
     this.logger.info(`Destroyed ${this.constructor.name}`);
   }
--- a/src/nodes/NodeManager.ts
+++ b/src/nodes/NodeManager.ts
@@ -1132,7 +1132,7 @@ class NodeManager {
       )}`,
     );
     logger.info(
-      `and addresses addresses ${filteredAddresses.map(
+      `and addresses ${filteredAddresses.map(
         (address) => `${address.host}:${address.port}`,
       )}`,
     );
--- a/src/nodes/NodeManager.ts
+++ b/src/nodes/NodeManager.ts
@@ -21,7 +21,6 @@ import * as nodesErrors from './errors';
 import * as nodesUtils from './utils';
 import * as tasksErrors from '../tasks/errors';
 import { timedCancellable, context } from '../contexts';
-import * as networkUtils from '../network/utils';
 import * as validationUtils from '../validation/utils';
 import * as utilsPB from '../proto/js/polykey/v1/utils/utils_pb';
 import * as claimsErrors from '../claims/errors';
@@ -107,18 +106,8 @@ class NodeManager {
       );
       never();
     }
-    const host_ = await networkUtils.resolveHost(host);
-    if (
-      await this.pingNode(nodeId, { host: host_, port }, { signal: ctx.signal })
-    ) {
-      await this.setNode(
-        nodeId,
-        { host: host_, port },
-        false,
-        false,
-        2000,
-        ctx,
-      );
+    if (await this.pingNode(nodeId, { host, port }, { signal: ctx.signal })) {
+      await this.setNode(nodeId, { host, port }, false, false, 2000, ctx);
     }
   };
   public readonly pingAndSetNodeHandlerId: TaskHandlerId =
@@ -237,10 +226,9 @@ class NodeManager {
     if (targetAddress == null) {
       return false;
     }
-    const targetHost = await networkUtils.resolveHost(targetAddress.host);
     return await this.nodeConnectionManager.pingNode(
       nodeId,
-      targetHost,
+      targetAddress.host,
       targetAddress.port,
       ctx,
     );
--- a/src/nodes/NodeManager.ts
+++ b/src/nodes/NodeManager.ts
@@ -311,6 +311,7 @@ class NodeManager {
           ) as PublicKeyPem;
           return [unverifiedChainData, publicKey];
         },
+        connectionTimeout,
         { signal: ctx.signal, timer },
       );
 
@@ -344,6 +345,7 @@ class NodeManager {
             async (connection) => {
               return connection.getExpectedPublicKey(endNodeId) as PublicKeyPem;
             },
+            connectionTimeout,
             { signal: ctx.signal, timer },
           );
           if (!endPublicKey) {
@@ -1038,44 +1040,65 @@ class NodeManager {
     pingTimeout: number | undefined,
     @context ctx: ContextTimed,
   ): Promise<void> {
-    this.logger.info('Syncing nodeGraph');
+    const logger = this.logger.getChild('syncNodeGraph');
+    logger.info('Syncing nodeGraph');
     // Getting the seed node connection information
     const seedNodes = this.nodeConnectionManager.getSeedNodes();
-    const addresses = await Promise.all(
-      await this.db.withTransactionF(async (tran) =>
-        seedNodes.map(
-          async (seedNode) =>
-            (
-              await this.nodeGraph.getNode(seedNode, tran)
-            )?.address,
-        ),
-      ),
-    );
-    const filteredAddresses = addresses.filter(
-      (address) => address != null,
-    ) as Array<NodeAddress>;
-    // Establishing connections to the seed nodes
-    const connections =
-      await this.nodeConnectionManager.establishMultiConnection(
-        seedNodes,
-        filteredAddresses,
-        pingTimeout,
-        undefined,
-        { signal: ctx.signal },
-      );
+    // FIXME: remove or uncomment, need to decide
+    // const addresses = await Promise.all(
+    //   await this.db.withTransactionF(async (tran) =>
+    //     seedNodes.map(
+    //       async (seedNode) =>
+    //         (
+    //           await this.nodeGraph.getNode(seedNode, tran)
+    //         )?.address,
+    //     ),
+    //   ),
+    // );
+    // const filteredAddresses = addresses.filter(
+    //   (address) => address != null,
+    // ) as Array<NodeAddress>;
+    // logger.info(
+    //   `establishing multi-connection to the following seed nodes ${seedNodes.map(
+    //     (nodeId) => nodesUtils.encodeNodeId(nodeId),
+    //   )}`,
+    // );
+    // logger.info(
+    //   `and addresses addresses ${filteredAddresses.map(
+    //     (address) => `${address.host}:${address.port}`,
+    //   )}`,
+    // );
+    // // Establishing connections to the seed nodes
+    // const connections =
+    //   await this.nodeConnectionManager.establishMultiConnection(
+    //     seedNodes,
+    //     filteredAddresses,
+    //     pingTimeout,
+    //     undefined,
+    //     { signal: ctx.signal },
+    //   );
+    // logger.info(`Multi-connection established for`);
+    // connections.forEach((address, key) => {
+    //   logger.info(
+    //     `${nodesUtils.encodeNodeId(key)}@${address.host}:${address.port}`,
+    //   );
+    // });
     // Using a map to avoid duplicates
     const closestNodesAll: Map<NodeId, NodeData> = new Map();
     const localNodeId = this.keyManager.getNodeId();
     let closestNode: NodeId | null = null;
-    for (const [nodeId] of connections) {
+    logger.info('Getting closest nodes');
+    for (const nodeId of seedNodes) {
       const closestNodes =
         await this.nodeConnectionManager.getRemoteNodeClosestNodes(
           nodeId,
           localNodeId,
           { signal: ctx.signal },
         );
-      // Setting node information into the map
-      closestNodes.forEach((item) => closestNodesAll.set(...item));
+      // Setting node information into the map, filtering out local node
+      closestNodes.forEach(([nodeId, address]) => {
+        if (!localNodeId.equals(nodeId)) closestNodesAll.set(nodeId, address);
+      });
 
       // Getting the closest node
       let closeNodeInfo = closestNodes.pop();
@@ -1089,9 +1112,15 @@ class NodeManager {
       const distB = nodesUtils.nodeDistance(localNodeId, closestNode);
       if (distA < distB) closestNode = closeNode;
     }
+    logger.info('Starting pingsAndSet tasks');
     const pingTasks: Array<Task> = [];
     for (const [nodeId, nodeData] of closestNodesAll) {
       if (!localNodeId.equals(nodeId)) {
+        logger.info(
+          `pingAndSetTask for ${nodesUtils.encodeNodeId(nodeId)}@${
+            nodeData.address.host
+          }:${nodeData.address.port}`,
+        );
         const pingAndSetTask = await this.taskManager.scheduleTask({
           delay: 0,
           handlerId: this.pingAndSetNodeHandlerId,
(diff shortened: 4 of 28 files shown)


CMCDragonkai:
The `resolveHostname` is ready to be used.

The references are:

```
nodes/NodeConnectionManager.ts
297:        const targetHost = await networkUtils.resolveHost(targetAddress.host);
803:    host = await networkUtils.resolveHost(host);

nodes/NodeManager.ts
110:    const host_ = await networkUtils.resolveHost(host);
240:    const targetHost = await networkUtils.resolveHost(targetAddress.host);
```

Along with the tests:

```
network/utils.test.ts
28:      networkUtils.resolveHost('www.google.com' as Host),
30:    const host = await networkUtils.resolveHost('www.google.com' as Host);
33:      networkUtils.resolveHost('invalidHostname' as Host),
```

The above usages need to be changed, but it requires investigating MatrixAI/Polykey#483.

So I suggest @tegefaulkes that you take over this PR and commit a fix for both MatrixAI/Polykey#483 and MatrixAI/Polykey#484 in this PR and merge to staging for integration tests.

CMCDragonkai:
Here's an important realisation.

```ts
import type { Hostname } from './src/network/types';
import { Timer } from '@matrixai/timer';
import * as networkUtils from './src/network/utils';

async function main(){
  const timer = new Timer({ delay: 10000 });
  const hosts = await networkUtils.resolveHostname(
    'localhost' as Hostname,
    {
      timer,
    }
  );
  timer.cancel();
  console.log('HOSTS', hosts);

}

main();
```

Since the timer is passed from the outside, the HOF will not cancel the timer. So it's important to explicitly close the timer.

This is why I wanted to provide some convenience utility that is if there is a default delay it would work automatically, but also if I just pass a number, then the expectation is that the timer can just be cancelled automatically.

Like:

```ts
await networkUtils.resolveHostname(..., { timer: 1000 });
// And in this case, rather than inheriting `Timer`, it inherits just a parameter
// and thus it is equivalent to "overriding" the default and passing `timer: undefined`
```

But that can be done later.

CMCDragonkai:
Note that using `dns.resolve` may have issues with mDNS later in MatrixAI/js-mdns#1. Also this may result in some changes to the docker configuration because I don't know what the docker container or how it will receive DNS servers.

Can always change back to `dns.lookup`, if we do so, we won't be really doing anything with `CNAME` at all, as that will all up to the OS to manage that, instead it will just be an array of IPv4 and IPv6 addresses.

CMCDragonkai:
Do the simplest thing to maintain node connections to the seed nodes at all times. If there's a connection loss to the seed nodes, try to reconnect.

Seed node connections must not be subject to node connection TTL.

Seed nodes must not be removed from the node graph.

Last manual test is the priority here.

tegefaulkes:
Most of the changes have been applied now and it is mostly working. There are a lot more regressions than I was expecting outside of the nodes domain.

One of the main issues is, In the case where we have a hostname that resolves to multiple host addresses. How do we handle the connection errors? If we connect to 2 hosts and 1 fails to connect and the other fails the verification, what did it really fail as? Since we're tying to connect to a single node in this case then it could just be the verification error. When trying to find multiple nodes at once however the failure becomes more ambiguous. 

There are about 20 errors still left over. I'm not sure if they will affect manual testing or not. I'm still looking into them

CMCDragonkai:
If you are connecting to `NodeIdA`, and you end up trying 10 different connections, as long as 1 connection succeeds, then the connection to `NodeIdA` is successful, the other failures should be ignored.

CMCDragonkai:
If none of the connections succeed, then you don't need an aggregate error, you can just say that you couldn't connect to `NodeIdA`.

tegefaulkes:
Some tests are expecting a specific reason for failing to connect. I guess I could make them more generic.

CMCDragonkai:
You can still provide specific reasons. You can provide a reason for why none of the connections succeeded, just like `AggregateError`, although we wouldn't use this, we just a specialised exception inside the `network` or `nodes` to represent this. You can also use the `cause`.

Those tests might be too **specific** in that case, those tests should be made more generic, or completely deleted.

tegefaulkes:
Ok, so after that refactoring it's mostly working again. I resolved most of the test failures. What ever failures are left are known problems and stuff that won't interfere with manual testing.

tests/nodes/NodeConnectionManager.lifecycle.test.ts
```diff
@@ -274,22 +310,19 @@ describe(`${NodeConnectionManager.name} lifecycle test`, () => {
       await taskManager.startProcessing();
       // @ts-ignore: kidnap connections
       const connections = nodeConnectionManager.connections;
-      // @ts-ignore: kidnap connectionLocks
-      const connectionLocks = nodeConnectionManager.connectionLocks;
       const initialConnection = connections.get(remoteNodeIdString1);
       expect(initialConnection).toBeUndefined();
       await nodeConnectionManager.withConnF(remoteNodeId1, async () => {
-        expect(connectionLocks.isLocked(remoteNodeIdString1)).toBe(true);
+        // Do nothing
```
CMCDragonkai, line 280: Why run this if it does nothing?

tests/nodes/NodeConnectionManager.lifecycle.test.ts
```diff
@@ -323,21 +354,17 @@ describe(`${NodeConnectionManager.name} lifecycle test`, () => {
...
-      // Connection is not created yet, no locking applied
+      // Connection is not created yet
       expect(connections.get(remoteNodeIdString1)).not.toBeDefined();
 
       // Iterating over generator
       for await (const _ of gen) {
-        // Should be locked for duration of stream
-        expect(connectionLocks.isLocked(remoteNodeIdString1)).toBe(true);
+        // Do nothing
       }
```
CMCDragonkai, line 327: Same here, why run this if it does nothing?

tests/nodes/NodeConnectionManager.seednodes.test.ts
```diff
@@ -453,8 +453,8 @@ describe(`${NodeConnectionManager.name} seed nodes test`, () => {
       await nodeConnectionManager.start({ nodeManager });
       await taskManager.startProcessing();
       // This should complete without error
-      await nodeManager.syncNodeGraph(true, 5000, {
-        timer: new Timer({ delay: 20000 }),
+      await nodeManager.syncNodeGraph(true, 2000, {
+        timer: new Timer({ delay: 15000 }),
       });
```
CMCDragonkai, line 458: What do these constants here mean?

CMCDragonkai:
We should push up from this branch in case manual testing reveals bugs to be fixed here.

tests/nodes/NodeConnectionManager.lifecycle.test.ts
tegefaulkes: It creates the connection for us. the expectation of locking wasn't relevant anymore.

tests/nodes/NodeConnectionManager.seednodes.test.ts
tegefaulkes: 2000 is the time for each individual connection, 15000 is for the overall operation.

tests/nodes/NodeConnectionManager.lifecycle.test.ts
CMCDragonkai: I see, so there's no more direct way of connecting to the other side?

CMCDragonkai: Does the test require this connection?

CMCDragonkai:
Do any of the tasks above need to be ticked off?

tegefaulkes:
I've just run a test where a node connects to the seed node to see how the changes work. I can see that we're attempting 2 connections, one for each seed node on the `testnet.poklykey.io` domain.  1 connection is getting established however its disconnecting far too soon afterwards. I need to look into this.

I'll add logs of the attempt in a moment.

CMCDragonkai:
Don't we need to first test if both seed nodes are connecting to each other?

tegefaulkes:
For that i'll need to apply some small changes to the infrastructure and config.

tests/nodes/NodeConnectionManager.seednodes.test.ts
CMCDragonkai: Should this be taken from overall app configuration? Or is there something special about this test that requires special cased times?

CMCDragonkai:
Yes, go with that, you need to deploy to the testnet.

CMCDragonkai:
This PR is missing changes to `src/config.ts`. It should have the NodeIds included in the source code with `testnet.polykey.io`.

CMCDragonkai:
You'll need to add those first before deploying.

CMCDragonkai:
The test shows 4 points of confusion.

1. Seed node A cannot connect to Seed node B, it never finishes the `Starting Connection Forward`
2. Seed node A terminates the connection from Seed node B
3. Seed node A tries to create 4 NodeConnection for some reason...?
4. It also tries to connect twice to 2 different IPv6 addresses. Where did these IPv6 addresses come from? They aren't on the testnet. It attempts the connection forward to the 2 ipv6 addresses. How is this possible?

CMCDragonkai:
Make sure check your VMs and shutdown any PK agents, don't leave anything running to ensure that our testnet manual tests have clean logs.

tegefaulkes:
todo: 
1. [x] 1. improve logging for life cycle NCM and the creation/destruction flow of the node connections 
2. [x] 2. defence in depth for using our own nodeId in the node graph. We need to filter out our own nodeId from information when we receive it, resolve it or attempt to use it.
3. [x] 3. relaying nodes should reject requests from nodes to themselves. Should check the nodeId and the address.
4. [x] 4. Make the node connection creation for the multi-node resolving serial instead of concurrent. Need to double check the destroy callback logic here.
5. [x] 5. ICE needs to be extracted from the NC and done along side its creation instead.
6. [x] 6. Special case seed nodes to avoid TTL clean up on connections.
7. ~Any of the odd methods used within`syncNodeGraph` should be in-lined to avoid noise. Also add logging outlining the process of syncing the node graph.~ Anything I could in-line is used in multiple places.
8. [x] 8. Filter out any Ipv6 addresses when resolving hostnames.
9. [x] 9. Add key path to logger messages.
10. [x] 10. Add detailed logging to `syncNodeGraph`
11. [X] 11. Separate timeouts for each sub connection.
12. [x] 12. All nodes need to re-establish to connection seed nodes when they disconnect.
13. [x] 13. Serial connection establishment needs to be updated to work concurrently.
14. [x] 14. Proxy needs to cancel forward connection establishment if the underlying GRPC client connection is destroyed.
15. [x] 15. Special case GC of seed nodes in the NodeGraph.


Moving list to PR description

CMCDragonkai:
Avoid the optimisation of proxy connection **then** node connection. It's not working atm for the seed nodes.

Make the sync node graph directly start a node connection.

There's too many cross cutting concerns and interference with other side effects atm. So it's important to ensure that we don't add our own NodeId to the NodeGraph on handling a reverse connection.

tegefaulkes:
I've deployed and done some testing. Here are some notes.

testing locally shows that we have to problem doing network entry with `syncNodeGraph`. I can see it doing multi-node establishment to both seed nodes and succeeding. it then starts the node connections to the seed nodes just fine.

The seed nodes however are acting a little weird. It's doing the multi-node establishment. It's own nodeId is filtered out so it's only trying to connect to the other seed. I can see that we are establishing a proxy connection to the other node just fine.

```
local seed: vg9a9e957878s2qgtbdmu2atvli8ms7muukb1dk4dpbm4llkki3h0 52.62.176.59
remote seed: vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng 52.65.240.37

{"level":"INFO","key":"syncNodeGraph","msg":"Multi-connection established for"}
{"level":"INFO","key":"syncNodeGraph","msg":"vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng@52.65.240.37:1314"}
```

First oddity is that it's trying to add itself to the nodeGraph when it connects to itself. The nodeGraph should prevent it from adding it's own NodeId so i don't think it's a problem.

```
{"level":"INFO","key":"ConnectionForward 52.62.176.59:1314","msg":"Starting Connection Forward"}
{"level":"INFO","key":"Proxy","msg":"Handling connection from 52.62.176.59:1314"}
{"level":"INFO","key":"ConnectionReverse 52.62.176.59:1314","msg":"Starting Connection Reverse"}
{"level":"INFO","key":"ConnectionReverse 52.62.176.59:1314","msg":"Started Connection Reverse"}
{"level":"INFO","key":"ConnectionReverse 52.62.176.59:1314","msg":"Composing Connection Reverse"}
{"level":"INFO","key":"ConnectionReverse 52.62.176.59:1314","msg":"Composed Connection Reverse"}
{"level":"INFO","key":"PolykeyAgent","msg":"Reverse connection adding vg9a9e957878s2qgtbdmu2atvli8ms7muukb1dk4dpbm4llkki3h0:52.62.176.59:1314 to NodeGraph"}
```

The second oddity is that the NodeConnection is failing to establish to the other seed node even though we have an established proxy connection.

(commented this for a paper trail, we've already discussed it)

tegefaulkes:
I think I know what the problem is, it could be timing out before it completes the 2nd connection. Both nodeConnections are being created with the same ctx provided by `getConnection` it's called in. If the first one times out the 2nd will time out immediately.

CMCDragonkai:
So generally whenever we have a collection of things that may time out, we have to both:

1. Have a global timeout
2. Have a per-thing timeout

The global timeout indicates how long to wait for the entire operation. The per-thing timeout would be a parameter timeout for each individual operation.

The "per-thing" timeout cannot always be automatically derived from the global timeout. Because the number of things may be dynamic. There are too many variations of this that could work. A fixed division is possible sometimes, othertimes each per-thing can use the entire timeout and each operation just consumes the global timeout.

So any operation like this would work like this:

```
doManyThings({ thingTimeout: 1000 }, ctx);
```

Where `ctx` is the global timeout, and individual operations have their own timeout.

The only issue is mixing static values here and dynamic values acquired by the `ctx`. It would be useful if the `thingTimeout` were derived somehow by the `ctx.timer.getTimeout()` operation as well.

CMCDragonkai:
We need a idiom for dealing for structured timeouts. Consider 4 different situations. Each function can take an timeout for downstream operations. For hetero, they may take multiple downstream timeouts. The below deals with "default" behaviour if they aren't set.

1. Fixed Homo - t = T/n
2. Dynamic Homo - t = T - the default is to inherit the global T if it is not set
3. Fixed Hetero - t = T/n AND sum(ts) <= T
4. Dynamic Hetero - t = T - the default is to inherit the global T if it is not set

Situation 4 is likely the rarest. Because that implies dynamic number of heterogenous operations, in which case you would probably need a fixed number of classifications.

Here are some examples (pseudocode, I haven't tried):

```
@timedCancellable()
fixedHomo(timeout: number | undefined, ctx: ContextTimed) {
  timeout ??= ctx.timer.getTimeout() / n;
}

@timedCancellable()
dynamicHomo(timeout: number | undefined, ctx: ContextTimed) {
  timeout ??= ctx.timer.getTimeout();
}

@timedCancellable()
fixedHetero(timeouts: [number, number] | undefined, ctx: ContextTimed) {

}

@timedCancellable()
dynamicHetero(timeouts: [number, number] | undefined, ctx: ContextTimed) {

}
```

Connecting to multiple hosts for a given seed node is equivalent to `fixedHomo`. Meaning that the default is the division of the global timeout by the number of hosts being contacted.

This only applies if we assume that the timeouts are serialised, meaning the ops are not concurrent.

If the underlying ops are 100% concurrent, then the timeout for each operation is equal the global timeout. That is the whole thing collapses to as if the global function called a single local function. This means it inherits the global timeout.

However there are also cases in between fully sequential and fully concurrent.

In the case where there is a concurrency pool, this ends up being a mix of sequential and concurrent. If this is a fixed homo situation, then each "pool" is given the divided timeout. That is let's say there's 10 ops, and we are doing 2 at a time. If the global timeout is 100, then the division is `100 / (10/2)` meaning 20 per pool. A 20 per pool, means that every operation entering the pool is given a 20.

If the pool size gets larger, then the timeout per pool increases.

CMCDragonkai:
The NCM must be understood as a "supervisor" (in the sense of process supervision or erlang supervisors).

The child processes in this case are the NC. This is different from other manager objects which manage static structures like `VaultManager`. In the case of static structures, their lifecycle is carefully controlled by the manager object.

However in a supervisor, it is expected that the child processes can fail at any time, and it is up to the parent supervisor to catch these events and handle them appropriately.

Thus with node connections, they can fail at any time, and these events must propagate up to the NCM for them to handle. Different handling policies are possible. A number of different restart strategies are documented here: https://www.erlang.org/doc/man/supervisor.html.

In our case, it only makes sense to restart connections to the seed nodes if they are ever dropped.

Connections to other kinds of nodes may disconnect for other reasons, and in that case we shouldn't restart those connections automatically. The main reason being that once failed/failing, any calls using those connections will also fail, and subsequently the entire call will fail. If the call is re-attempted, connections may be re-established automatically.

---

All the additional bells and whistles added on top of node connections are something that should be enabled and tested only after the core foundation of connections are working:

1. Reference counting
2. TTL for node connections - delayed expiry (only activated after reference count goes to 0)
3. Coalescing concurrent creation with locks
4. Convenience methods for resource allocation and bracketing of node connection

See: https://en.wikipedia.org/wiki/Connection_pool and https://en.wikipedia.org/wiki/Process_supervision

Therefore ICE should be done in the NCM, and not in the NC.

Furthermore in the future we will eliminate the difference between NC, GRPC client and proxy connections. All of these should be the same thing.

CMCDragonkai:
Manual test shows that the seed nodes are connecting to each other now, it was a due to the timeouts problem.

Now what we need to do:

1. Ensure that proxy connections don't continue to start, if the GRPC client has already terminated.
2. Have a policy built into the NCM to always retry on the connections to the seed nodes no matter what. This is can be true for both the seed nodes and also the non-seed nodes.
3. Now we can go back to doing things in parallel for the multi-host connector.

tegefaulkes:
The seed nodes are succesfully connecting to each other. Here is a snippet of the logs.

```
This is for the following nodes
Local node: vg9a9e957878s2qgtbdmu2atvli8ms7muukb1dk4dpbm4llkki3h0 52.62.176.59
Remote node: vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng 52.65.240.37

2022-11-07T15:22:32.316+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeManager.syncNodeGraph","msg":"Syncing nodeGraph"}
2022-11-07T15:22:32.317+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeManager.syncNodeGraph","msg":"Getting closest nodes"}
2022-11-07T15:22:32.319+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager","msg":"Getting NodeConnection for vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng"}
2022-11-07T15:22:32.320+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager","msg":"Finding address for vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng"}
2022-11-07T15:22:32.321+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager","msg":"Found address for vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng at testnet.polykey.io:1314"}
2022-11-07T15:22:32.397+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager","msg":"Creating NodeConnection for vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng"}
// Creating NodeConnecton to the correct address
2022-11-07T15:22:32.398+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager.NodeConnection [vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng@52.65.240.37:1314]","msg":"Creating NodeConnection"}
2022-11-07T15:22:32.399+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager.NodeConnection [vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng@52.65.240.37:1314].clientFactory","msg":"Creating GRPCClientAgent connecting to 52.65.240.37:1314"}
2022-11-07T15:22:32.419+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Proxy","msg":"Handling CONNECT to 52.65.240.37:1314"}
2022-11-07T15:22:32.423+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Proxy.ConnectionForward 52.65.240.37:1314","msg":"Starting Connection Forward"}
2022-11-07T15:22:32.500+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Proxy.ConnectionForward 52.65.240.37:1314","msg":"Started Connection Forward"}
2022-11-07T15:22:32.500+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Proxy.ConnectionForward 52.65.240.37:1314","msg":"Composing Connection Forward"}
2022-11-07T15:22:32.501+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Proxy.ConnectionForward 52.65.240.37:1314","msg":"Composed Connection Forward"}
2022-11-07T15:22:32.502+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Proxy","msg":"Handled CONNECT to 52.65.240.37:1314"}
// NodeConnection established
2022-11-07T15:22:32.510+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager.NodeConnection [vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng@52.65.240.37:1314].clientFactory","msg":"Created GRPCClientAgent connecting to 52.65.240.37:1314"}
2022-11-07T15:22:32.511+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager.NodeConnection [vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng@52.65.240.37:1314]","msg":"Created NodeConnection"}
2022-11-07T15:22:32.541+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager","msg":"Created NodeConnection for vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng"}
2022-11-07T15:22:32.673+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeManager.syncNodeGraph","msg":"Starting pingsAndSet tasks"}
2022-11-07T15:22:32.674+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.scheduler","msg":"Starting Scheduling Loop"}
2022-11-07T15:22:32.674+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.scheduler","msg":"Started Scheduling Loop"}
2022-11-07T15:22:32.675+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.queue","msg":"Starting Queueing Loop"}
2022-11-07T15:22:32.675+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.queue","msg":"Started Queueing Loop"}
2022-11-07T15:22:32.675+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Status","msg":"Finish Status STARTING"}
2022-11-07T15:22:32.675+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Status","msg":"Writing Status to /srv/status.json"}
2022-11-07T15:22:32.679+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.Status","msg":"Status is LIVE"}
2022-11-07T15:22:32.679+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent","msg":"Started PolykeyAgent"}
2022-11-07T15:22:32.679+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent","msg":"Created PolykeyAgent"}
2022-11-07T15:22:32.679+11:00	{"level":"INFO","keys":"polykey.WorkerManager","msg":"Creating WorkerManager"}
2022-11-07T15:22:32.691+11:00	{"level":"INFO","keys":"polykey.WorkerManager","msg":"Created WorkerManager"}
2022-11-07T15:22:32.695+11:00	{"pid":1,"nodeId":"vg9a9e957878s2qgtbdmu2atvli8ms7muukb1dk4dpbm4llkki3h0","clientHost":"0.0.0.0","clientPort":1315,"agentHost":"127.0.0.1","agentPort":34595,"proxyHost":"0.0.0.0","proxyPort":1314,"forwardHost":"127.0.0.1","forwardPort":36627,"recoveryCode":"loop bright forum neck author frog toy violin pilot round short clip else hundred rebuild sponsor indicate come awesome oblige horn skin monkey rent"}
2022-11-07T15:23:32.719+11:00	{"level":"INFO","keys":"polykey.PolykeyAgent.NodeConnectionManager","msg":"Destroying NodeConnection for vh9oqtvct10eaiv3cl4ebm0ko33sl0qqpvb59vud8cngfvqs4p4ng"}
```

tegefaulkes:
The original tasks on the todo list have been complete. I'm taking a moment to clean up the code before addressing the rest,

CMCDragonkai:
Task 14. is lower priority, it could be done later as a new issue. Once the others are done, should move to getting the staging fixed and merged to master.

CMCDragonkai:
Suggest mapping into the node connection promise and returning the array index that way. You can easily get the index plus array elements with `arr.entries`.

CMCDragonkai:
Task 14. may need to be done if there's a chance of a proxy connection leak. This means if the GRPC client already has terminated, and the proxy connection isn't being disconnected, it may keep that proxy connection all the time. My understanding is that proxy connections don't have a TTL, they only have a keep alive.

tegefaulkes:
We want to handle the case where we are creating a proxy connection and the client connection drops before composing. Right now i think we only clean up proxy connections when the proxy connection fails or the composed client connection is closed. I need to double check this.

Right now I'm thinking we can prevent resource leak here by applying a TTL to un-composed connections. This should be pretty simple to implement. Composed connections can be handled by the `NodeConnectionManager`s TTL logic since shutting down the client connection will clean up the proxy connection.

There is a concurrency concern with the TTL timing out while composing. Composing sets up some events and the data pipes. Stopping Cleans up a bunch of stuff. I'm sure we can end up with some undefined behaviour if we do both of them concurrently. To fix this the `compose` and `stop` method needs to share a lock. If we can share the lifecycle lock here that would work.

Do we need to apply this to the `ReverseConnection` as well? I'm leaning towards no. It's life-cycle is mostly driven by the complementary `ForwardConnection`.


Since `connectForward` is a `PromiseCancellable` we can just cancel it if the GRPC connection ends. This can be a first order fix for the problem. We will still need the TTL as a final stop for handling the connection leaks.

tegefaulkes:
I've implemented the 2nd option here. when the client socket closes we cancel the `connectForward` promise. That should be enough to fix 14. for now.

CMCDragonkai:
I think we need to keep this as simple as possible because we are planning replace this system with a more robust networking solution later.

So the problem right is that starting a forward connection is being triggered before a `compose` is called. This is because in the `Proxy.establishConnectionForward`, it's calling:

```ts
    await conn.start({ timer });
```

While this is occurring, it's possible that `clientSocket` has **already** terminated before composition.

```ts
    // HERE, it is already started the connection
    const conn = await this.establishConnectionForward(
      nodeId,
      proxyHost,
      proxyPort,
      timer,
    );
    // What if the `clientSocket` is **already** terminated?
    conn.compose(clientSocket);
```

Therefore the simplest solution here would be to attach an event handler to the client socket, which would then trigger a cancellation on the `conn.start` promise.

CMCDragonkai:
> I've implemented the 2nd option here. when the client socket closes we cancel the `connectForward` promise. That should be enough to fix 14. for now.

This sounds similar to what I proposed above, except I was saying that `conn.start` itself could be `PromiseCancellable`?

tegefaulkes:
It's pretty much what you suggested. `conectForward` propagates the `ctx` down to the `conn.start` ultimately.

CMCDragonkai:
Ok, well once composition is enabled, that event handler should be removed. There's no need to keep around that event handler which was only used between `start` and `compose`.

tegefaulkes:
It gets un-registered in the finally block.

tegefaulkes:
12 can be addressed by a re-occurring ephemeral task that checks if all of the seed nodes has a connection and re-attempts them. If no seed nodes have an active connection we can re-attempt a network entry. 

Now that I think about it, if we fail to connect to any seed nodes then we want to end the `syncNodeGraph` early. There's no point in starting refresh buckets when no connections can be made.

There are 4 steps to this.
1. use a task to periodically check the state of the seed node connections.
2. If any seed node connections are down we attempt to re-establish it.
3. if all seed nodes are down we re-attempt the `syncNodeGraph`
4. If no seed node connections are established during `syncNodeGraph` then we end it early.

CMCDragonkai:
I think task 12 can just be done by monitoring the lifecycle of the seed connections. You already have a callback whenever a connection fails right? Why not expand that to include when the seed connection fails, you just immediately retry?

This would require the `syncNodeGraph` to be using the same "data flow" as the rest of the NCM. But it should be done this now anyway.

tegefaulkes:
Right now the callback isn't called if the connection failed to establish in the first place and we don't want to retry immediately since it's very unlikely to connect right after it failed.  Doing it that way would also spread out the logic within the NCM.

Keeping it all in a task keeps it all in one place and clear what is going on. I'm leaning towards that unless we need the functionality of reacting the second the connection goes down or individual restart delays.

CMCDragonkai:
Ok do it the simplest way for now, as long as it works. But I'll mark this to be redesigned when NCM is refactored so that even during the `syncNodeGraph`, all connection management (restarts... etc) has a single control flow.

src/PolykeyAgent.ts
```diff
@@ -573,6 +573,7 @@ class PolykeyAgent {
         PolykeyAgent.eventSymbols.Proxy,
         async (data: ConnectionData) => {
           if (data.type === 'reverse') {
+            if (this.keyManager.getNodeId().equals(data.remoteNodeId)) return;
```
CMCDragonkai, line 576: Do we still need this if the filtering of our own NodeID is all done in the NCM now? Is this doing double filtering? If so, we shouldn't need this.

src/agent/service/nodesHolePunchMessageSend.ts
```diff
@@ -92,6 +92,13 @@ function nodesHolePunchMessageSend({
             connectionInfo!.remoteHost,
             connectionInfo!.remotePort,
           );
+          // Checking if the source and destination are the same
+          if (sourceId?.equals(targetId)) {
+            // Logging and silently dropping operation
+            logger.warn('Signalling relay message requested signal to itself');
```
CMCDragonkai: Use US english `Signaling`.

src/nodes/NodeConnection.ts
```diff
@@ -116,39 +110,23 @@ class NodeConnection<T extends GRPCClient> {
...
-            seedNodeId,
-            keyManager.getNodeId(),
-            targetNodeId,
-            undefined,
-            ctx,
-          );
-        });
-      }
       // TODO: this needs to be updated to take a context,
       //  still uses old timer style.
```
CMCDragonkai, line 114: Is this still relevant?

src/nodes/NodeManager.ts
```diff
@@ -1044,71 +1040,138 @@ class NodeManager {
...
-          signal: ctx.signal,
-        })) === false
-      ) {
-        continue;
-      }
+    const logger = this.logger.getChild('syncNodeGraph');
+    logger.info('Syncing nodeGraph');
+    // Getting the seed node connection information
+    const seedNodes = this.nodeConnectionManager.getSeedNodes();
+    // FIXME: remove or uncomment, need to decide
```
CMCDragonkai: What are all these comments for? Can they be turned into a new issue?

tests/nodes/NodeConnectionManager.seednodes.test.ts
CMCDragonkai: What about our composed timer spec? Should this be updated so that the individual timeouts can be auto-derived?

CMCDragonkai:
Going to copy this to the docs issue; https://github.com/MatrixAI/Polykey/pull/491#issuecomment-1303150991

tegefaulkes:
I just ran a test on the seed nodes, Everything seems to be working including retrying network entry and seed connections. 

Most jests tests are passing, there are a small amount of tests that need fixing due to changes from this PR. 

I'm going to resolve the review comments now.

tests/nodes/NodeConnectionManager.lifecycle.test.ts
tegefaulkes: Yes, we're testing if a connection is created here. Currently `withConnF` is the only way of establishing a connection.

tests/nodes/NodeConnectionManager.lifecycle.test.ts
tegefaulkes: Same reason as above but for `withConnG`.

tests/nodes/NodeConnectionManager.seednodes.test.ts
tegefaulkes: Just custom parameters for the test here, I'm using a 15 second overall timeout so we get a better error than the jest timeout.

src/PolykeyAgent.ts
tegefaulkes: Yes and no, here we're doing it to avoid trying to add it to the node graph when we receive a connection. The `NodeGraph` already checks this so functionally we're just not printing the log message for it.

src/agent/service/nodesHolePunchMessageSend.ts
tegefaulkes: Fixed.

src/PolykeyAgent.ts
CMCDragonkai: Please add a comment about how this is already being filtered out, but it's done here earlier to avoid posting another log message.

It seems not exactly tidy atm... but this will involve thinking about #444

src/nodes/NodeConnection.ts
tegefaulkes: Yes, we still haven't applied `PromiseCancellable` changes to GRPC level stuff. That was put on hold for now.

CMCDragonkai: Can these comments be taken into a new issue instead of leaving it here.

tests/nodes/NodeConnectionManager.lifecycle.test.ts
CMCDragonkai: Ok sure. But it is also possible to directly use the resource allocation API too right? The `with*` variants are usually convenience wrappers.

src/nodes/NodeManager.ts
tegefaulkes: That was part of the `syncNodeGraph` doing strategy 2. It was disabled while testing but I've re-enabled it now.

tegefaulkes:
That's all the review comments addressed.

Last thing to do is to fix any remaining tests that are failing.

tegefaulkes:
Ok, the only tests failing now are
- nat, expected
- testnetConnection, expected
- ping.test.ts, not expected but I can leave it till after the merge.

Let me do some final merge prep and then this is done.

CMCDragonkai:
Yea on staging, we should expect that connecting to the testnet should work, but it will require redeployment.

PR #498, opened 2023-01-03
Transport Agnostic RPC implementation

### LINKED ISSUE #249: Transport Agnostic RPC

### LINKED ISSUE #500: `RPCServer` specification
### Specification

The `RPCServer` takes incoming streams and handles the using a provided handler. There are 6 points of focus regarding this.

1.  Stream Handlers
2.  JSON Transformer
3.  Metadata
4.  Authentication
5.  Error Handling
6.  Types

#### Stream handlers

<details>

The stream handlers are core to how the RPC server works. When an stream is created by the quic system it is provided to the RPCServer to be handled. There is one point of interaction here `stream creation event -> handle stream`. When a stream creation event happens it should call `rpcServer.handleStream(streamPair: ReadableWritabePair<Uint8Array, Uint8Array>, connectionInfo: ConnectionInfo)`. This method should check the first message of the stream for the method and metadata. Hand off the metadata to the metadata interceptor. Get the required handler from the registered handlers and call the handler with the input stream and data. Then it should consume the handler output and provide the output data to the output stream.

The most generic form of these handlers will be a `DuplexStreamHandler<I extends JSONValue, O extends JSONValue>`. Since we will be handling `ReadableWritablePair<Uint8Array, Uint8Array>`,  this can be consumed 1:1 by a duplex stream handler.  This handler will take the form of...

```ts
type Handler<I, O> = (  
  input: I,  
  container: POJO,  
  connectionInfo: ConnectionInfo,  
  ctx: ContextCancellable,  
) => O;  
type DuplexStreamHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  AsyncGenerator<I>,  
  AsyncGenerator<O>  
>;
```

This can be implemented as ...
``` ts
const duplexHandler: DuplexStreamHandler<JSONValue, JSONValue> =  
  async function* (input, _container, _connectionInfo, _ctx) {  
    yield* input;
  };
```

Ultimately there are 4 kinds of stream handlers, DuplexStream, ClientStream, ServerStream, Unary. The latter 3 can be implemented as a `DuplexStreamHandler`. Their types are as follows.

```ts
// Receives 1 value and returns many
type ServerStreamHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  I,  
  AsyncGenerator<O>  
>;  
// Receives many values and returns 1
type ClientStreamHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  AsyncGenerator<I>,  
  Promise<O>  
>;  
// Receives 1 value and returns 1.
type UnaryHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  I,  
  Promise<O>  
>;
```

Implementing these 3 as a `DuplexStreamHandler` is done by wrapping each respective function within a `DuplexStreamHandler`. For example a `ServerStreamHandler` can be wrapped like below. The others can be implemented as variations of this.

```ts
const wrapperDuplex: DuplexStreamHandler<I, O> = async function* (  
  input,  
  container,  
  connectionInfo,  
  ctx,  
) {  
  for await (const inputVal of input) {  
    yield* handler(inputVal, container, connectionInfo, ctx);  
    break;  
  }  
};
```

The key features of this are
1. All handlers are ultimately handled as a `DuplexStreamHandler` , the client, server and unary handlers are internally converted to a duplex handler when registered.
2. The RPC server interfaces with incoming streams through a single call to `handleStream()`. This should be called during a stream created event.

</details>

#### JSON Transformer

<details>

The stream pair we're working with will be pretty low level. It is a stream of `Uint8Array`. This means we need to convert from something akin to a byte stream to parsed JSON RPC messages. This can happen in stages.
1. Message separation: `Uint8Array` -> separated JSON RPC messages in `Uint8Array` form.
2. Message parsing: `Uint8Array` -> parsed `JsonRPCMessage`'s Where each JSON RPC message is parsed into a JSON object and validated to match the specified JSON RPC message structure.
4. The reverse needs to be done where `JsonRPCMessage` -> `Uint8Array`.

Each stage of transformation makes use of the webstreams `TransformationStream`. The first stage transformation takes the raw `Uint8Array` and emits seperated JSON RPC messages in JSON object form. This is done by making use of the JSON stream parser from https://www.npmjs.com/package/@streamparser/json. We can feed in the input data to the stream parser and it should emit JSON that it detects. The parser should output the top level JSON objects if we provided the following options `{ separator: '', paths: ['$'] }`.  The separator of `''` Allows the parser to process back to back JSON objects of `{JsonRpcMessage}{JsonRpcMessage}`. The paths parameter of `['$']` Will have the parser only output the top level JSON object. ultimately stream conversion should be `{jso`|`nMessa`|`ge}{JsonMess`|`age}` uintArray form -> `{JsonRpcMessage}`|`{JsonRpcMessage}` JSON object form
The 2nd stage focuses on validating the messages. This is also a transformation stream that takes each 'chunk' of the JSON object and uses the generic parsing function to validate the message structure. If any message fails validation it will result in an error the next time the stream is read.

</details>

#### Metadata

<details>

Metadata needs to be sent along side the normal data. This is for communicating information relating the to communication that isn't directly parameters for the RPC call. The metadata must be included in the JSON RPC message `params` parameter. 

The JSON message `params` parameter can contain any `JSONValue`. We should reserve the `metadata` key within this structure for the metadata use. The metadata should be structured as a `POJO`. 

There is a general order to the metadata messages. 

1. Server send leading metadata
2. client responds with leading metadata
3. server and client can send arbitrary metadata at any time
4. client sends trailing metadata
5. server responds with trailing metadata
6. communication ends.

When a call is made the handler side responds with server side leading metadata. The structure of this is not set in stone but it can include information about handler side limitations and data for constructing a authentication response. The Client can then respond with it's leading metadata. This will include the authentication information requred by the server side if needed. 

Metadata can be sent at any time by both the client and server. This will likely be done by just including the metadata within the message parameters. 

While it is simple enough to provide the metadata for each call or call handling. there are common metadata communications that need to be done across all calls and ideally should only be implemented in one place. For this we need to support metadata interceptors.  These will be functions that we can register to the server that will be called on every stream. 

There are 3 kinds of interceptors.
1. And initial interceptor for generating the initial metadata to be sent before any data communication. 
2. initial responding metadata. Called with the first message. will be provided the initial metadata if it was received
3. General interceptor called on any received metadata after the initial 2 interceptors. Bit unsure about this one, It happens in reaction to any received metadata but should the response be it's own metadata message or attached to the next data message? If it's attached I can't guarantee there will be another message. Maybe if the messages end we just send a metadata only message before fully ending?

There can be multiple registered interceptors with each one adding to the metadata. The interceptor should take metadata and return it's own metadata `(metadata: POJO) => Promise<POJO>`. Each stage of metadata will add to the add to the outgoing metadata with a spread operation `const newMetadata = {...oldMetadata, ...interceptorMetadata}` such that the interceptor can overwrite the metadata if needed. This should overwrite any metadata that was part of the message to begin with.

</details>

#### Authenticating

<details>

Authentication is generally left to the user of the RPCServer to set up. It can be handled by the interceptor system and the metadata. For example when doing basic bearer token authentication we can use the type 2 interceptor on the client to add the bearer token to the metadata of the first message. The server can use the type 2 interceptor to authenticate authenticate this token when the first message is received.

Authenticating via the TLS connection certificate should be handled via the quic system. We may want to check the nodeId of a handled connection. I don't see this being useful for every connection but it's needed for checking ACL for certain RPC calls so authenticating based on nodeId should be done within the handler.

 A slightly more advanced method of token authentication and be done with a combination of the type 1 and 2 interceptors. The server side can use the type 1 interceptor to send metadata containing a secret for generating a authentication token. The client can use the type 2 to use the metadata to generate the expected token. The server can then use the type 2 to authenticate the token. This should allow for a more secure version of the token authentication.

</details>

#### Error handling

<details>

There are several sources of errors.

1. Errors from the input stream. Usually indicates a failure of communication. These should not be thrown through the output stream.
2. Parsing error. These come from the input stream parsing stage. They should be clearly parsing errors and are intended to be converted to error message thrown through the output stream.
3. handler internal errors. These come out of the handler itself. These should be converted to error messages and thrown through the output stream.
4. abort signal. The handler is provided an abortion signal. If it chooses to throw the reason then it should be thrown through the output. These errors are either timeouts or the RPC stopping.

There is a clear flow to any error that can happen. An error can come through the input stream, get thrown within the handler, caught by the stream handler and converted to an JSON rpc reponse error message and sent through the output stream. Most errors will be caught, converted and sent though the output stream. If an error can't be handled or isn't intended for the client caller then it should be emitted via the error event system.

When an error is intended for the client caller. it must be caught and converted to a JSON RPC error message. the error itself should be stringified and added to the message along with the relevant fields filled in.

The error event system should be an `EventTarget`. Anyone using the `RPCServer` can register an error event handler for processing these errors. Generally this will just be for logging out the error that couldn't be handled. No errors should bubble up outside the `RPCServer` unless it's a critical failure of the `RPCServer` itself.

The `EventTarget` could be a property of the `RPCServer` or the `RPCServer` can extend the `EventTarget` to streamline the usage. 

</details>

#### Types

<details>

Strict typing is very important for this system and as such the types must be enforced as much as possible.

The JSON RPC messages have a strict structure. There are 4 kinds of messages, Request, Notification, response result, response error.  These are defined within the JSON RPC spec

There are 4 kinds of handlers. When you register a handler you provide them in the form of.

```ts
type Handler<I, O> = (  
  input: I,  
  container: POJO,  
  connectionInfo: ConnectionInfo,  
  ctx: ContextCancellable,  
) => O;  
type DuplexStreamHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  AsyncGenerator<I>,  
  AsyncGenerator<O>  
>;  
type ServerStreamHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  I,  
  AsyncGenerator<O>  
>;  
type ClientStreamHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  AsyncGenerator<I>,  
  Promise<O>  
>;  
type UnaryHandler<I extends JSONValue, O extends JSONValue> = Handler<  
  I,  
  Promise<O>  
>;
```

The types above will enforce the types within the handlers but the stream parsing can't check if the message data matches these types. As a result, the data within the handler will be typed as expected but could potentially be anything. It is up to the implementation of the handler to validate the data or trust it on blind faith. 

</details>

### Additional context

* Related  #249 
* Related #498 

### Tasks

- [x] 1. Stream handling functionality
- [x] 2. stream transformation
- [x] 3. metadata and interceptors
- [x] 4. Implementing authentication.
- [x] 5. Error handling
- [x] 5. Types
- [x] 7. Raw handlers

### LINKED ISSUE #501: `RPCClient` specification
### Specification

There are 3 parts to the client spec

1.  Creation
2.  Calling Interface
3.  Manifest

#### Creation

The `RPCClient` should be a `CreateDestory` class. It's dependencies should be the logger a generic callback that returns a new stream pair. When used with quic this would be a quic stream. The stream creation callback should take the form of `() => Promise<ReadableWriteablePair<Uint8Array, Uint8Array>`

#### Calling interface

There are the 4 types of calls that can be made. Duplex stream, Client stream, Server stream and unary call. Each of these call types should be supported by a generic caller method. For example the duplex caller can take the form of 

```ts
public duplexCaller<I extends JSONValue, O extends JSONValue>(  
  method: string,  
  metadata: POJO,  
): ReadableWritablePair<O, I>{}
```

Much like the RPCServer, the duplex caller is the generic form that the others are based on. The other 3 callers will wrap the duplex caller. The other 3 types are..

```ts
public serverStreamCaller<I extends JSONValue, O extends JSONValue>(
 method: string,
 params: I
 metadata: string,
): Promise<ReadableStream<O>> {}

public clientStreamCaller<I extends JSONValue, O extends JSONValue>(
 method: string,
 metadata: string,
): Promise<{output: Promise<O>, writable: WriteableStream<I>> {}

public unaryCaller<I extends JSONValue, O extends JSONValue>(
 method: string,
 params: I,
 metadata: string,
): O {}

```

Making a rpc call this way can be done like the following. Ideally the typing would be enforced by a method unique to that call but that needs to be enforced by some manifest some how.

```ts
const result = await rpcClient.unaryCaller<I, O>('someMethod', [1, 2, 3], {});

// Ideally the call will be made like so with all the types enforced.
const result = await rpcClient.someMethod([1,2,3], {});
```

When a generic caller method is called it needs to create a Quic stream and hook itself up to that. This should be done by calling the stream creation callback provided during the RPCClient construction.

There will be two forms of the of the caller interfaces. A more generic RWP form that provides a `ReadableWritablePair` as the interface that is returned when called. The other form will be a co-routine style that mirrors how the handlers are set up. The Co-routine style of `CO` for short can be implemented as the following. 

```ts
type WithDuplexCaller<O, I> = (output: AsyncGenerator<O>) => AsyncGenerator<I>  
function withDuplexCaller<O, I>(method: string, caller: WithDuplexCaller<O, I>, metadata: POJO): Promise<void>  {
 // ...
}
  
await withDuplexCaller<number, number>('method', async function * (output) {  
  for await (const value of output) yield value;  
}, {})
```

This style can be applied to the `DuplexStream`,`ClientStream` and `ServerStream` methods. The unary only makes sense as a simple async function call.

The client stream version would have to provide the output as a promise since in this style we expect to receive the output before inputting data. Alternatively the output can be returned by the function and we only input within it. 

#### Manifest

To avoid boilerplate and enforce typing we need some kind of manifest that will dictate the available methods and their types. 

My first thought was to extend the `RPCClient` and add in the specific caller implementations that way. But that will have too much boilerplate where we create the specific caller methods by wrapping the generic ones. 

If we could provide the manifest in the form of a object type that would be ideal.

```ts
Type Manifest = {
	myUnaryCall: UnaryCaller<number, Array<string>>;
	myDuplexCall: DuplexCaller<{a: string, b: number}, number>;
	...
}
```

This way we could construct the RPCClient as `RPCCllient.createRPCClient<Manifest>()` and have all our caller methods available. This would be a very advanced way of enforcing the types and plenty of prototyping needs to be done on this.

If the manifest can be shared between the client and server then that would be the best outcome.

### Additional context

* Related #249 
* Related #498 

### Tasks

* [x] 1. Class life cycle and construction 
* [x] 2. Generic calling WRP interface
* [X] 3. Generic calling CO style interface
* [x] 4. Specific caller interface support with manifest.
* [x] 6. Raw callers, counterpart to the server's raw handlers.

### LINKED ISSUE #502: Middeware specification
### Specification

Still fleshing out the concept.

The middleware is shared between the Client and Server side implementation. So I'm making a single issue for just the middlewear.

### LINKED ISSUE #503: Websocket Client and Server for Client Service API
### Specification

The client based communication is to be implemented using websockets. Currently we have a proof of concept using the node `ws` library. but a lot of the code needs to be formalised into a client and server class for managing the state. I will expand this spec when I come back to it, This is placeholder for now to track the task.

We need to create a client and server class for client communication. 

#### Websocket Server

We need to create a Websocket server class for abstracting the logic and life-cycle of the Websocket server code. The class doesn't really hold state outside of itself so it doesn't require a `destroy` method. I think It can be made a `StartStop`? otherwise `CreateDestroy` should do. 

The server needs to take a stream creation callback of `(streamPair, connectionInfo) => void`. This will be used to call the stream handler when a connection is established. We may have limited connection information on the connecting side. So far as I can tell the remote port can't be obtained from the connection. The client certificate isn't applicable here but I think we can obtain the remote IP address. For local information, the host and port are easily obtained. 

The server needs to support secure connection over HTTPS/SSL and reject any insecure connections. We also need the ability to update the certificates at any time though that may not be possible. The uWebsocket library requires that we write the certs to disk to access them. For this we need to encrypt the files with a random password and pass that password to the server when loading them. I am unsure what method is used to encrypt them at this time. 

Back-pressure propagation needs to be applied to both the readable and writable stream. The websocket has some signalling for when the buffer is filling up. for the writable stream the `ws.send()` returns a number for success, back-pressure and message dropped. We can also check how full the buffer is at any time. There is also a `drain` event for when the back-pressure has drained some. For the readable stream we need to use a BYOB readable stream and signal to the websocket back-pressure by corking it. I'm still a bit fuzzy on this.

Half open connection needs to be supported. if the stream handler closes the writable before the readable or vice versa. the websocket shouldn't be closed until both streams are closed. If the websocket closes abruptly then both streams are closed. If the writable is still used this will result in an error. it is up to the stream handler to deal with this.

The server needs to refuse any normal HTTP requests.

#### Websocket client

We need to create a Websocket client class for abstracting the logic of the client code. The class should extend either `StartStop` or `CreateDestory` depending how the server is implemented or other constraints.

Creation of the the client should take host, port address and `NodeId` of the target server. This should be used to establish the connection.

The client needs to connect over a secure connection using a `wss://` url. We need to add custom logic for authenticating the server's certificate against the expected `NodeId` of the server we're connecting to. This can be done via the upgrade event.

Back-pressure needs to be signalled to the streams. I need to prototype how to do this with the `ws` library. 

Half open connections need to be supported. The websocket should only be ended when both the readable and writable are closed. If both are closed early then this should result in the websocket ending. If the websocket ends early then both streams should be closed. If the writable is still used this will result in an error. it is up to the stream caller to deal with this.

### Additional context

- Related #498 
- Related discussion starting at https://github.com/MatrixAI/Polykey/pull/498#issuecomment-1408172493

### Tasks

- [x] 1. Create a `ClientServer` for websocket based communication.
  - [x] Create a `ClientServer` class for handling life-cycle of the server.
  - ~Add Secure connection with TLS, use encryption when writing files and reading them with the server. 0.5 days~ - Differed to #508 
  - [x] Apply back-pressure propagation to the streams. 0.5 days
  - [x] Support half open connections by only closing the websocket when both streams have closed.
  - [x] Handle abruptly dropped connections. 0.5 days
  - [x] Refuse any non-websocket communication. trivial
  - [x] Support IPv6
- [x] 2. Create a `ClientClient` for websocket based communication.
  - [x] Create a `ClientClient` class for handling life-cycle of the client. 0.5 days
  - [x] Make the connection secure using TLS. 0.5 days
  - [x] Add custom authentication logic for checking the server's certificate against a NodeId. 0.5 days
  - [x] Apply back-pressure propagation to the streams. 1 days
  - [x] Support half open connections by only closing the websocket when both streams have closed. 0.5 days
  - [x] Handle abruptly dropped connections. 0.5 days
  - [x] Connection timeout logic
  - [x] Support IPv6

tegefaulkes:
### Description

This PR addresses adding our own custom transport agnostic RPC implementation. 

The transport agnostic RPC needs to cover everything from the byte stream to the service handlers. This includes
1. Converting a byte stream to RPC messages.
2. Parsing and validating RPC messages.
3. Creating the service handler interfaces.

#### Stage 1 - Converting the byte stream

Each RPC message is a single JSON RPC message object. There are 4 forms to this, a request, notification, response result and a response error. These messages are specified in the JSON-RPC 2.0 spec at https://www.jsonrpc.org/specification.

We need convert a continuous stream of these messages into distinct messages. The stream is made up of JSON-RPC objects that are `JSON.stringify()`ed and converted to buffers. The messages are structured like `{JsonRpcMessage}{JsonRpcMessage}` but in buffer form.

These object strings are delimited by their top level `{}` bracket pair. No delimiter character exists between each message. So to extract a single message out we need to find the top level matching `{}` take the sub-array of that part of the incoming buffer.

Ultimately the resulting transformation stream takes in arbitrary sized buffers of the incoming messages and outputs buffers of single complete message buffers. E.G. `{jso`|`nMessa`|`ge}{JsonMess`|`age}` => `{JsonRpcMessage}`|`{JsonRpcMessage}` This can then be piped through a message validation stream.

#### Stage 2 - parsing and message validation
In this stage we are taking single JSON-RPC messages and converting them into objects and validating them. If the message is still in the buffer form then the conversion is done with `Json.parse(Buffer.toString('utf-8'))`.

These objects need to be valid JSON-RPC message objects. So we need to create validation functions for each type of message. There are 4 kinds of JSON-RPC messages specified in https://www.jsonrpc.org/specification. Each one will need it's own validation function.

Solid types for each type of message need to be created. These message types need to be generic on the data so we can enforce a strict message schema for each service handler. The message data can't be validated at this stage. This should be left for the handlers to validate.

On top of the JSON-RPC spec, we need to add some more parameters to the messages.
A `type` parameter is needed to distinguish message types from each other for simple pattern matching. We also need some kind of sub type to distinguish control messages from data messages.

#### Stage 3 - service and client handlers

We need a structured way of implementing client call and service handlers the the RPC. The data type of the requests and responses need to be identical for both and enforced by the typescript type system. 

The request and response data structure isn't set in stone yet. Right now it's a `POJO`, there was discussion about it being an `Array` but the JSON-RPC spec allows for any `JSONValue`. When implementing the handler we need to specify the request and response types. This is to enforce the type on input and output data in the handler.

request and response data validation needs to be handled within the handlers. We will need to create a parser function for our message data.

The handler themselves need to be generators that implement our handler interface. The interface will look something like this..

```ts
type TestGen<I extends JSONValue, O extends JSONValue> = (  
  input: AsyncGenerator<I>, // Input generator
  context, // Context data such as connection information  
  ctx, // Timeout and cancellation  
) => AsyncGenerator<O>;
  
const testGen: TestGen<DataA, DataA> = async function* (input, a, b){  
  yield* input; // Echo  
}
```

This will be the most generic form of the duplex stream. Some prototyping needs to be done but it should be possible to have higher order functions that converts the unary, server stream and client stream handlers to this duplex stream form. 

```ts
function unary<I, O>(f: (input: I, context, ctx) => O): TestGen<I, O>;
function serverStream<I, O>(f: (input: AsyncGenerator<I>, context, ctx) => O): TestGen<I, O>;
function clientStream<I, O>(f: (input: I, context, ctx) => AsyncGenerator<O>): TestGen<I, O>;
```

#### Remaining spec
Todo:
- Plumbing new Quic streams into handlers.
- How client calls work, how streaming will work with them.
- How metadata will be handled.
- Details on message streaming and control messages. Do we use request messages at all? or abstract over notifications? if just notifications do we use the response message types?
- Registering handlers, dynamic or static configuration? If dynamic how do we enforce message data structure parity between handlers and client callers?

### Issues Fixed

* Fixes MatrixAI/Polykey#249 
* Fixes MatrixAI/Polykey#500
* Fixes MatrixAI/Polykey#501 
* Fixes MatrixAI/Polykey#502
* Fixes MatrixAI/Polykey#503
* Related MatrixAI/Polykey#495 

### Tasks

1. [x] Try and fix the message splitting stream. Try and use the JSON parser again. Try and catch the error. If not then try a different parser.
2. [x] Post an upstream issue for the wrong offset value in the tokenizer
3. [x] Client and server streaming is backwards. need to swap their names.
4. [x] The RPC class needs an error event emitter. Non-handleable errors should be emitted here such as connection failures and non handled errors in raw handlers.
5. [x] Double check that the input streams are properly closed when done. Streams coming out of the quic system may never 'end' so for cases were our handler is done We can close the input.
6. [x] Optimize the stream parser
7. [x] Change up how the JsonRpcMessage params is structured to include metadata
8. [x] Using async iterators
9. [x] Changing `Buffer` streams to `Uint8Array` streams
10. [x] CREATING THE CLIENT SIDE
    1. Need to create split generator for IO
    2. Create generic call handlers
    3. Template and test
    4. Test out BYOB streams for the input and output streams to allow buffering.
    5. Outline how the stream handler is used and set up in the PR.
    6. remote node id isn't really avaliable in the connection info, should be extracted from the certificate info
    7. Still need to make raw stream handler
11. [x] Change the `RPCServer` to `CreateDestroy`
12. [x] The `RPCClient` needs to take a callback to create a stream
13. [x] Create a websocket based client for client communication

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
The plan right  now is to create some test utils using fast-check and the transform stream to convert the raw stream data into distinct JSONRPC messages. 

Looking over the docs we can use generators to transform the stream pretty easily. But the example provided doesn't return a new stream. https://exploringjs.com/nodejs-shell-scripting/ch_web-streams.html#transforming-web-streams-via-generators. So we won't be able to compose it with other transform streams. If we don't care about that then it is fine.  Otherwise we would have to do it the usual way. https://exploringjs.com/nodejs-shell-scripting/ch_web-streams.html#implementing-custom-transformstreams.

After creating the transformer stream I can look into parsing the JSONRPC messages. We need a good way of enforcing the structure of these using the type system. Looking over the JSON RPC spec there isn't a good way to tell the type of a message. For example, a request message contains the fields `jsonrpc`, `method`, `params?`, `id`. However a notification is exactly the same as a request but it doesn't include the `id`. So the only way to tell them apart is to check if the id key exists. This is not a great way to do it.  I think we discussed previously about using a `type` field so I'll likely just go with that. 

The parameters can be any valid `JsonValue`. This means it can be a json object, string, number or array. If I recall from previous discussion we wanted to just use arrays for the parameters? Is that something we still want to enforce?

tegefaulkes:
These web streams don't play very nice with strict typing. While interfaces with proper types are provided. I can't get proper type inference when implementing a transformer class.

For example. when implementing the class `class ChunksToLinesTransformer implements Transformer<Buffer, string>` 
I can implement transform as 

```ts
transform(chunk, controller) {
  chunk //any
  controller //any
};
```

To get proper types on the method I need to implemented as
``` ts
transform: TransformerTransformCallback<Buffer, string> = async (chunk, controller) => {
  chunk // Buffer
  controller // TransformStreamDefaultController<string>
}
```

While this will work it is not ideal. We do still get a warning if the type annotation is wrong.

@CMCDragonkai You've worked with this. Do you have a good way of enforcing the types here?

tegefaulkes:
Slicing out the JSON messages from the stream is going to be a little tricky. I first thought that I could just keep a tally of  `{}` brackets. The problem with this is that `{}` can be within strings in the JSON object. So I'll need to check if the brackets are within strings when I tally them AND account for any escape characters while doing so. 

We may be able to delimit the messages by using the provided ASCII seperators. That being the codes `0x1C`, `0x1D`, `0x1E` and `0x1F`. I don't see these characters listed in the JSON grammar. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#full_json_grammar. So we may be able to use these to delimit messages easily.

CMCDragonkai:
I think there are off the shelf JSON stream-parsers, can we make use of those?

It should be possible to repurpose them to specifically do a "partial parse" that is parse me a JSON message, expect that there may be left over data in the buffer, which can be left to the next iteration of the parsing loop.

CMCDragonkai:
I remember referencing this one: https://github.com/juanjoDiaz/streamparser-json

CMCDragonkai:
No thoughts yet on the type inference.

tegefaulkes:
> I remember referencing this one: https://github.com/juanjoDiaz/streamparser-json

Just took a look at this. It's useful but not ideal here. The parser expects a properly formatted JSON, so it can parse a stream of `[{},{},{},...]` where each object is a JSONRPC message.  The parser will also return every key value pair withing the JSON structure. so we need to filter and return the message objects only. I think I can make a workable solution with this. 

I'm going to try 3 methods of slicing the messages from the stream. Using transform streams should make them interchangeable so we can pick the one we think works the best. Right the 3 approaches are.
1. well defined message header specifying the number of bytes the message contains. The header can contain an enum for the message type so we can mix and match message types. We can share Protobuf, raw binary data and JSON messages within the same stream with this.
2. ASCII separators. ASCII has 4 well defined separators for delimiting data. I checked and they're not valid in the JSON format so they shouldn't be contained within messages. This will not work with binary data but should work fine for the JSON messages.
3. Using the JSON stream parser. We would need to format the stream as a JSON array. so the stream needs to start with `[` and each message separated with `,`. This will only work for JSON message and it's a bit more complex than option 2 so 2 may be a better solution here.

I'll prototype transform streams for each of these methods.

tegefaulkes:
I've been thinking about how to handle bad stream data. Do we even bother with this? A single stream would come from a quic stream. Theses should handle any problems with the data over the socket so I don't expect to see bad or mangled data within the stream.

We should be able to handle bad or mangled data within the stream but I don't think we should bother with that. Any failure to pass data should just throw an error back down the stream and close it out. This would be part of the normal error handling but at a slightly lower level.

CMCDragonkai:
Regarding streaming, based on what we talked about on the whiteboard I want to keep it as vanilla as possible. Try to get the json stream parser to "get you the next token", and then keep the remaining data.

So if the string input is `{ ... }`. As soon as it reaches `}`, it should give you the complete message. If the next character is `{` such as `{ ... }{`, then it even if it throws an error we must ignore it and leave it in the remaining buffer (for the next iteration of the "lexer"). This makes it closer to a compiler lexer which always gets you the next token but does not do more than that.

See if you can configure the streamparser-json to do that. If it can't do that out of the box, consider forking it and making upstream changes to be capable of this.

Finally regarding bad stream data, that should be considered a lexing failure, upon a lexing failure that should result in bad stream data, and while the initial JSON message has been passed onto the downstream consumer, an error should be thrown on the stream, and that means the error gets thrown to the downstream system. The downstream consumer can then decide what to do, which means it could rethrow that exception or close the stream... etc. There should be no automatic decisions made during the mapping operation with respect to the upstream producer.

Visually speaking that means something like:

```
BYTESTREAM -> Transformer -> JSONSTREAM
```

And errors in the transformer is propagated as errors to the `JSONSTREAM`, it is not an error for the input `BYTESTREAM`. Errors flow to the end consumer.

tegefaulkes:
It is possible to separate the messages using the original bracket counting method using the `@streamparser/json` tokenizer to detect valid brackets (Brackets not contained within a string). The tokenizer has a weird quirk however. When returning the offset of a token it doesn't take account of any escaped characters. So while I can work out the offset of the last `}` of a message. It offset value will be off by the number of escaped characters within the message. 

This is an easy fix, I just need to work out the number of escaped characters and add that to the offset.

CMCDragonkai:
Make sure to fuzz test your technique so that it works in all situations of sending JSON frames, or sending invalid data or sending mixed JSON frames and invalid data.

CMCDragonkai:
Are you using `@streamparser/json-whatwg`? Apparently it wraps `@streamparser/json` to use web streams, which may be better suited to what we are doing.

CMCDragonkai:
@tegefaulkes can you start speccing out the tasks for this PR?

tegefaulkes:
I've fleshed out the PR some more. There is still some stuff to add but I need to settle some structure in my head first.

tegefaulkes:
The class can be given a stream and then handle all the control and service handling. The thing is, while this RPC call is open we have state that should be running in the background. Given that we need to shut down these connections gracefully at any time. we need the ability to track these active calls.

We can keep a set of active calls. When they are done they should remove themselves from the set. We can enumerate the set to destroy any active connections when we need to stop the RPC.

These active calls might need some amount of state such as what service handler is needed when processing messages. This could be `CreateDestroy` class for handling this.

CMCDragonkai:
Can you provide a file structure of all the new modules be developed here.

CMCDragonkai:
The MatrixAI/Polykey#249 issue should be updated with the desired specification of this JSON RPC system. Cross out the old GRPC tasks, the old spec can just be separated with `---`.

The new spec should have subsections with `###` for each thing we need to work out.

Refer to the issues in MatrixAI/Polykey#495 for extra detail, I've also updated a comment on each of those issues relating to the migration to JSON RPC.

Make sure to consider:

* Middleware (from the GRPC interceptor)
  - Authentication
  - Method Dispatch
  - Client Side
  - Server Side
* Schema for our `params.meta` and `result.met` for both request and response objects
* Deadline/timeout parameter and the ability to set it to be `null`
* How errors is being serialisation
* How to bubble up errors from QUICStream to RPCClient to whatever upwards, whether should be an `EventTarget` or `Observable`.

CMCDragonkai:
Based on some discussion:

1. There are situations where using Promise and EventTarget as a replacement for callbacks and event emitter means that the errors go in a different direction (to an event handler) rather than to the caller. This is happening right now in js-quic, more details to follow there: https://github.com/MatrixAI/js-quic/issues/1
2. The `RPCServer` should have a `error` event, this can be emitted in the case where there is no appropriate caller. Like if a raw handler of a stream were to throw an error. The more abstract handlers when throwing an error end up going to the stream.
3. If you were to write to the stream, and that fails, that can be an error emitted as an error event.
4. The `PolykeyAgent` is probably going to register an error handler for the `RPCServer`, and in that case it can either just log out the error with nothing else to do. This is because one does not really restart the `RPCServer`, since it's just always going to be handling a stream .
5. Client streaming is where the client sends lots of messages.
6. Server streaming is where the server sends lots of messages.

CMCDragonkai:
We should confirm how to do explicit stream closing for both readable and writable.

In the case of readable, there is way to "cancel" the stream using https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBReader/cancel as well as the `readable.cancel`.

This allows you to provide a reason for the cancellation. But in the case of readable stream the cancel reason I believe I will translate to a stream close.

So on the readable side, this means: https://docs.rs/quiche/latest/quiche/struct.Connection.html#method.stream_shutdown. Which I can pass a `err` number code corresponding a "reason".

But on the writable side, there's actually 2 ways to do this, the `stream_send` with a `fin` and `stream_shutdown`. I'm not entirely clear on the difference yet.

This is more on the QUIC side though, you need to ensure that you're explicitly signalling a close on the readable stream when you are finished reading (for unary, server streaming), and when you are finished writing (for unary, client streaming, server streaming, duplex streaming).

However at the end of the handling, no matter what, both readable and writable sides should be explicitly closed.

CMCDragonkai:
Ok so basically now we just need to fix up the above, and also prepare some middleware for authentication.

CMCDragonkai:
The incremental JSON parser has an inefficient way of figuring the exact source offset. We need to raise an upstream issue to indicate that when we have escaped strings, the offset is not accurate. The offset is only useful if it is correctly indexed against the source input. It will be quite slow if we have to use `JSON.stringify` each time to figure out how many escaped characters that affects the offset.

This seems like a bug/oversight that upstream should be correcting.

CMCDragonkai:
Furthermore another inefficiency is that we have to build up the chunks as an accumulating buffer before we do a `JSON.parse` because we are using the `Tokenizer` and not `TokenParser` from the incremental JSON parser.

We should be able to use the TokenParser to build up the tree internally and not have to do our own buffering. However the problem is that the parser currently throws an error when we write a string into it that has a concatenated message without a delimiter.

What we need to do is to raise an issue upstream as well, to make the parser actually truly incremental, in which case it should not throw an error immediately on writing the string, but upon acquiring the next token.

We can then ignore this error, and restart the parsing for the new message.

CMCDragonkai:
@tegefaulkes it's possible that if you just ignore the error on the `write`, the `onValue` handler might end up being called anyway.

tegefaulkes:
Some small notes.
1. Slight change to JSON RPC message structure. message metadata needs to be included within the params field. We should have a reserved `metadata` keyword but we can just keep the data separate and structure it as `{metadata: JSONValue, data: JSONValue}`.
2. Authentication. The first message of the stream should include authentication details in the metadata field. This could be a bearer token or a password. This needs to be further speced out. Do we want functionality where the server can send some leading metadata and the client replies with a secret generated from that metadata? 
    We need a standard way of handling the authentication. Likely through an authentication function that we should provide the RPC. Or maybe we don't need it as part of the RPC system, we only need the metadata available within the handler. This will tie into the middle-wear stuff.  This will need to tie into the rpc client and some prototyping needs to be done there still.
3. Look into using async iterator type as apposed to generator.
4. client and server streams are backwards
5. It's a stream of uint8 array
6. Create an issue on the json-stream parser about the tokenizer offset problem.
7. Raw handlers should emit on the RPC event emitter

CMCDragonkai:
Just remember that it should be `RPCServer` and `RPCClient` class.

tegefaulkes:
I've created an upstream issue for the tokenizer problem at https://github.com/juanjoDiaz/streamparser-json/issues/24.

tegefaulkes:
I did some more testing with the `JSONParser`. If we catch the error when attempting the `parser.write(data)` then it does output the event for the top level object so that will be an easy way to output the message.

However even with catching the error, writing more input to the parser at this stage will just output more errors. I don't see a way to get the remaining buffer out of the Tokenizer or TokenParser when the error happens. We may be able to take a subarray of the incoming chunk and start from there. I'll see what I can come up with.

CMCDragonkai:
As long as you know the "true offset" where the message was correct, you should be able to destroy the old parser, recreate a new one and then start the parsing from a subarray of the buffer you have.

It may also be possible to extend the `JSONParser` and you can override an existing method internally/property that enables you to ignore the error and "restart" parser state. Notice that the docs also mention you being able to extend the classes.

tegefaulkes:
So it turns out it IS possible to process back to back objects without delimiters. Hidden in the documentation in the parser options

```
{
  paths: <string[]>,
  keepStack: <boolean>, // whether to keep all the properties in the stack
  separator: <string>, // separator between object. For example `\n` for nd-js. If left empty or set to undefined, the token parser will end after parsing the first object. To parse multiple object without any delimiter just set it to the empty string `''`.
}
```

At the very end there `To parse multiple object without any delimiter just set it to the empty string ''`

This solves a lot of our problems.

tegefaulkes:
There is no clean way to check that a stream is closed here. During the unary handler where we only consume 1 chunk and  break out of the loop. I don't see the `flush`  method in the input transport stream being called. If we fully consume the input then we do see this. 

Actually. It seems that if you fully consume the stream then I can see the transformation stream flushing the rest of the data. But if we stop consuming before the end of the data I can see cancel being called in the ReadableStream that's streaming the data.

So it does look like that the streams are being cleaned up in both cases. It's just the case where the consuming side is waiting for more data and the generating side hasn't closed the stream yet that will lead to a stream not being cleaned up.

I can add cancel calls for the input stream when the output steam ends just to be sure but I don't think it is necessary.

CMCDragonkai:
What is this `flush` method?

tegefaulkes:
It's part of a transform stream. It is called when the stream is ending so you can flush any data you are still holding on to.

tegefaulkes:
How will the client interact with the connections and streams. I think the client will be attached to a connection with a 1:1 relationship. AFAIK from previous discussion the connections can be shared across clients of different kinds. So it would be up to the QUIC system to manage the connections? 

In any case, our RPC client will correspond to a single connection. Each call will be a new stream. In my mind that looks like when creating the client it requests a connection to a destination from the underlying quic. The Quic then returns an existing one or creates a new one. The client can then request a new stream from the connection every time it makes a new RPC call. 

Since there are 4 types of calls, Duplex, client, server and unary. We need a way of enforcing the call style and message types. This may be too tricky to do if we dynamically register them. To have nice types I think we need to provide the call methods during construction but I'll need to prototype this.

tegefaulkes:
I'm trying to work out a good way to enforce an interface for the different types of calls. On paper we just need a generic function we can call like `rpcClient.unaryCall<I, O>('method, params: I, metadata): O` but ideally we can call any method with `client.someRPCCall(params: I, metadata): O`.

There are some criteria I want to meet here. 
1. Strict enforced typing. So the input output types of any call and concrete so you know exactly what you're getting.
2. Low boilerplate. Ideally you won't have to create a new method that wraps the more generic caller methods. but we should be able to make a client call with `client.someMethod()`. As far as I know we can only do that by extending the client. It's not the end of the world and we can include the output validation within these methods as well.
3. Defining each RPC method within a single file. Ideally we should be able to define the caller method and handler within the same file with shared types to enforce parity between them.

CMCDragonkai:
> How will the client interact with the connections and streams. I think the client will be attached to a connection with a 1:1 relationship. AFAIK from previous discussion the connections can be shared across clients of different kinds. So it would be up to the QUIC system to manage the connections?

So 1 RPC client should translate to 1 QUIC connection. However every call in the client translates to separate QUIC sream.

CMCDragonkai:
In terms of the lifecycle of connections, upon creating a connection, the connection will basically stay online forever barring any sort of problems that may occur.

So the decision to terminate a connection has to be made by the `RPCClient`, or whatever is managing the `RPCClient`.

Thus this then becomes a `NodeConnectionManager` problem, where I'm guessing a connection to another node corresponds to encapsulating an RPC client. But this is something that requires further thought.

CMCDragonkai:
> Since there are 4 types of calls, Duplex, client, server and unary. We need a way of enforcing the call style and message types. This may be too tricky to do if we dynamically register them. To have nice types I think we need to provide the call methods during construction but I'll need to prototype this.

Because we don't have a "protobuf" IDL anymore in this situation, there are other rpc systems I've seen that provide a manifest.

I think we keep it simple and just make use of Typescript types. Which is to say, that we can have an interface type for any given RPC service, and then that interface type can be utilised by the client. Note that for raw handlers, the method type would have to be very generic.

Alternatively if this is too complex right now, we can skip it and just do dynamic calls from the `RPCClient` and just throw an exception when it doesn't work.

CMCDragonkai:
So on the QUIC side we will have a `QUICClient` that you construct and start, upon doing so this creates a `QUICConnection`. I'm actually unsure whether to have the `QUICClient` at all and it's possible you can just go straight to constructing a `QUICConnection` I'll see.

juanjoDiaz:
Hi guys!

Author of streamparser here in case I can be of any help.

I took a quick look at the PR and I have a couple of comments:

**I think that you could implement this in a cleaner way by using `@streamparser/json-whatwg`.**

You could build the `JsonToJsonMessageStream` as the concatenation of `JsonParser` and a `JsonRpcMessageParse` so you have two streams. One parsing the objects and one converting them to your own format.

You can see [here](https://stackoverflow.com/questions/72612083/using-web-streams-create-a-transformstream-from-several-transformstreams) how to combine two streams:.

The only thing that woudl be missing is that `JsonParser` errrors won't be converted to `rpcErrors.ErrorRpcParse`.

Unfortunatelly, web streams do not offer a good way to handle errors on transform streams. You have to wait until you read the final writtable stream.
I'm thinking that it would be good to add a new `onError` option or something similar that allows you to handle errors and transform them or even skip them.

What do you think?

**Error handling needs some careful consideration**

Which brings me to the next issue: how do you want to handle parsing errors?
JsonParser is left in an `ERROR` state after any error and it won't parse anything else.
This means that you SHOULD close the transform stream if there is any error or subsequent writes will always fail.

In all the use cases that I've seen so far, any error means failing the entire stream.
In your case, it might make sense that errors don't fail the stream but continue trying to see if you can parse anything else from the stream

For example, given `{ "a": 1, b: 2 }{ "c": 3 }` and using a `paths` selector to only emit top level objects, the string will fail on `b` since it's misisng quotes.

What should happen next?
Ideally, it could move to emitting `{ "c": 3 }` but that is not trivial.

If you decide to ignore the error, there are two options:
* Ignore what haven been parsed to far. So, it will error on `:`, emitt `2`, error on `}` and emit `{ "c": 3}`. This is not great because `2` is not really a top level object.
* Ignore the error but keep state. DISASTER! It will error on `:`, on `2`, on `}` because you can't close a JSON after a comma and on every other character.

Of course you could try to count braces to figure out where to close the stream. But I can think of a million cases were that won't work either.
* Any error due to braces `{ "a": 1, b{: 2 }{ "c": 3 }`, `{ "a": 1, b}: 2 }{ "c": 3 }`, etc.
* Braces in strings `{ "a": 1, b: "{ is a brace" }
* Complext nested objects
* etc...

Wdyt? Any ideas here?
Or you think that it's OK to simply fail the stream?

CMCDragonkai:
Hi @juanjoDiaz thanks for the information. In our RPC case, if there's a parsing error, we would expect it to fail the entire stream, and then send a response message indicating a protocol/client failure.

Similar to HTTP servers, if a client sends garbage, we just immediately fail and respond with 4xx error.

CMCDragonkai:
I was also thinking of another failure situation, where the client sends too much data and just builds up memory with the parser. @tegefaulkes I think we need to limit the size of each frame, so that any data greater than that size will also result in a client error. This should prevent adversaries intending to DOS the PK agent.

CMCDragonkai:
Need to spec out this work now:

1. RPCServer
   1. Stream Handlers
   2. JSON Transformer
   3. Metadata
   4. Authentication
   5. Error Handling
   6. Types
2. RPCClient
   1. Creation
   2. Calling Interface
   3. Manifest

This gives us an idea of how the entire system works together (to ensure we are covering all the edge cases), while lower level details can stay as tasks.

juanjoDiaz:
> I was also thinking of another failure situation, where the client sends too much data and just builds up memory with the parser. @tegefaulkes I think we need to limit the size of each frame, so that any data greater than that size will also result in a client error. This should prevent adversaries intending to DOS the PK agent.

@streamparser/json will only accumulate data while it's building the object not after emitting it. I.e., it doesn't matter how huge the stream is. If the objects fit in memory, it will simply emit them one by one and won't keep the memory.
So, you should simply not load all data in-memory but pass it from the transport (network, ipc, etc) to the parse as it comes.
Of course, and attacker could pass a never ending object, array or something that never ends parsing. For that, you could simply count bytes as you pass them to the parser and fail the whole thing once you get to your limit.

tegefaulkes:
I fixed up the message parsing. In stead of checking the message types and filtering them. We're assuming that all messages to a handler are `JsonRpcRequest` messages and are only parsing them. others will throw an error. Likewise the callers will only take responses.

I removed the `type` property from the message structures. Only the response messages need to be told apart and they're distinct on the `error` or `result` property since they're mutually exclusive.

I've added some tests for combined server and client communication to make sure the callers work with the handlers as intended. The tests here are basic and only test the easy path. Edge cases are not checked yet.

Error conversion for the client is done and the error types have been cleaned up. Still need to do an error work over on the server.

CMCDragonkai:
For the middleware, I think we need an equivalent to the `ReadableWritablePair` in the sense of `TransformPair`.

This is a pair of 2 `TransformStream`:

```
{
  readable: TransFormStream; // forward
  writable: TransformStream; // reverse
}
```

This object actually ends up creating 4 streams in total because each `TransformStream` itself is a `Readable` and `Writable` pair`.

```
interface TransformStream<I = any, O = any> {
    readonly readable: ReadableStream<O>;
    readonly writable: WritableStream<I>;
}
```

Compare this to:

```
interface ReadableWritablePair<R = any, W = any> {
    readable: ReadableStream<R>;
    /**
     * Provides a convenient, chainable way of piping this readable stream through a transform stream (or any other { writable, readable } pair). It simply pipes the stream into the writable side of the supplied pair, and returns the readable side for further use.
     *
     * Piping a stream will lock it for the duration of the pipe, preventing any other consumer from acquiring a reader.
     */
    writable: WritableStream<W>;
}
```

So we will need a new type like `TransformPair`.

We will also need the ability to do a `pipeThrough` on the `ReadableWritablePair` to pipe to the `TransformPair`.

The interface doesn't have any additional methods.

I think we can do something like this:

```ts
interface TransformPair {
  readable: TransFormStream; // forward
  writable: TransformStream; // reverse
}

class DuplexStream implements ReadableWritablePair {
  public readable: ReadableStream;
  public writable: WritableStream;

  public pipeThrough(transform: TransformPair);
  // Add in other methods that we need
}

// ReadableStream.pipeThrough(transform: ReadableWritablePair)
// readable => readable/writable

// DuplexStream.pipeThrough(transform: TransformPair)
// readable/writable => readable/writable & readable/writable
```

CMCDragonkai:
Ok so the `metadata` doesn't need to be in the param type anymore. But it is necessary in implementation.

CMCDragonkai:
We need a demonstration of the authentication middleware.

We need something like:

```
rpcServer.registerMiddleware(() => {
  // Check the token
  // If it fails
  // Short circuit with error message
  // Ensure that downstream either doesn't run if it does run (it is just done, because the upstream middleware has handled it)
  // We don't want exceptions downstream if upstream has handled it
  // Close everything
});
```

If you are short-circuiting on the `TransformPair`, it has to `cancel` the upstream `readable` and `abort` the upstream `writable`.

When you're doing the testing, put a `console.log` message under the readable `cancel` and `abort` for writable to see if this is the case.

CMCDragonkai:
Left to do:

1. Client service migration (prototype the `agent status` and `agent unlock`)
2. Fix up the middleware like above
3. Websocket for client service (find a good library for this/preferably on top of HTTP2) - you will need TLS here, that means you have to use web sockets, not TCP
4. Raw handlers for both client side and server side
5. Client server manifest (to auto generate the methods/types)

Just update the todos above.

tegefaulkes:
Shopping for WebSocket implementations. https://www.npmjs.com/package/ws This one is very popular and seems active. It doesn't work on the browser but I think browsers have their native websocket?

CMCDragonkai:
Can you check if it's `WebSocketServer` can take an "arbitrary" socket object? So if we were on a different platform like react native whether it would work. Of course I note that react native supports websocket clients natively, but it doesn't have support for websocket server natively. So the question becomes, would such a library be compatible with https://github.com/Rapsssito/react-native-tcp-socket.

I do note:

```
    if (
      (options.port == null && !options.server && !options.noServer) ||
      (options.port != null && (options.server || options.noServer)) ||
      (options.server && options.noServer)
    ) {
      throw new TypeError(
        'One and only one of the "port", "server", or "noServer" options ' +
          'must be specified'
      );
    }
```

So it's possible to also pass in an http server object, and also this example: https://github.com/websockets/ws#multiple-servers-sharing-a-single-https-server.

It seems that if we had the ability to just run an HTTP server we would have the ability to have websockets, since the rest of the code is just JS?

CMCDragonkai:
I do think that library we should try for now, and as we figure out how WSS works, we can eventually adapt to react native.

CMCDragonkai:
Based on the discussion, the `RPCServer` can hardcode the stack of middleware instead of allowing arbitrary composition.

This way we can specify a pipeline from compression to parsing to to logging to routing to auth... Etc.

In the future a composition operator function can be used and then the whole thing can be set once and the type safety can be enforced.

CMCDragonkai:
Note that compression is not a requirement, just an example of bytestreams to bytestream.

tegefaulkes:
I have a test that incorporates the webstockets with our RPC and it's working.

 I also prototype'd it using a secure connection. to get it working I had to provide a https server to the WebSocketServer.  the server itself can but doesn't need to verify the client's identity. The client however tries to verify the server by default and it doesn't like that the server's certificate is self signed. 

There are two ways around this. First the `rejectUnauthorized: false` option will just skip verifying the server while keeping the connection secure. This will dis-allow us from using the `checkServerIdentity: (servername: string, cert: WebSocket.CertMeta): boolean` callback to check the server's identity.

Alternatively we can use the server's certificate chain as the certificate authority. this requires us to know the server's cert chain. Not idea but still workable.

For now i'm just using the `rejectUnauthorized: false` option. I can demonstrate this tomorrow.

CMCDragonkai:
Yea, we have the ability to override how we verify the server's cert chain. We would do this on the client side and and on the server side. The server side obviously doesn't verify the client. The client should use a custom verification logic.

We did this for GRPC, we do the same for WSS.

CMCDragonkai:
Hey use the `verifyServerCertificateChain` in `network/utils`.

CMCDragonkai:
See how it's done in `grpc/GRPCClient.ts` on line 154.

CMCDragonkai:
So the `WebSocketServer` does re-emit errors from the underlying server.

It's important to use:

```
      this._server = http.createServer((req, res) => {
        const body = http.STATUS_CODES[426];

        res.writeHead(426, {
          'Content-Length': body.length,
          'Content-Type': 'text/plain'
        });
        res.end(body);
      });
```

As well.

CMCDragonkai:
The QUIC stream is changing to the BYOB stream, so when you are pausing the `ws` you should do that according to the desired size similar to `QUICStream`.

CMCDragonkai:
When converting the `WebSocket` into the `ReadableWritablePair` the closing of the WS needs to be coordinated. Because WS doesn't have analogues of stop sending in QUIC, we just need 2 booleans indicating whether readable or writable has been closed. We only close the WS when both readable and writable are closed.

When readable is closed, close the readable stream, but then also drop any incoming message now, this can be done by deregistering the readable handler.

When writable is closed, close the writable stream, nothing needs to happen further.

CMCDragonkai:
During the conversion process, you should check this and just compare if we have covered all the possible edgecases: https://github.com/websockets/ws/blob/2862c2fef1064940c1deb9c4aa961cb76c279feb/lib/stream.js#L49 - you might want to check nodejs lib to see how they convert that to web streams later.

I noticed that the conversion does not make use of `ping/pong` at all here.

When we are converting to the RWP, we also need to keep track of the ping/pong interval. This is part of the control frames in the websocket protocol. This ping/pong is needed to know if the websocket "connection" is still alive. And that means you have something similar to the `QUICConnection` because we have also a `timeout` event there.

CMCDragonkai:
The `timeout` situation for each `WebSocket` will need to translate to a `controller.error` for readable/writable.

CMCDragonkai:
Replace your `createClientServer` with a custom class that follows the `StartStop` decorator. Something like `WebSocketServer`. Then we do `await wsServer.start();` just like the other kind of servers we have.

The starting would then occur inside the `PolykeyAgent`.

Let's make our code consistent with `js-quic`.

So in QUIC we have `QUICClient` and `QUICServer`.

You should have a `WebSocketClient` and `WebSocketServer` as custom classes.

The conversion between RWP to the `WebSocket` provided by `ws` will be flipped.

The `QUICClient` is a `CreateDestroy`, and `QUICServer` is a `StartStop`, you should do the same.

CMCDragonkai:
During `await WebSocketClient.createWebSocketClient()` you need to either override how WS uses the target string, so that we can use `pk://...nodeid...` or change it so that we are passing `host`, `port` and the `nodeIds` directly as parameters, and the Node ID gets checked during `verifyServerCertificateChain`.

Actually because it is `nodeIds` and not one single `nodeId`, you wouldn't want to use `pk://....`, because that would prevent the ability to use multiple Node IDs.

CMCDragonkai:
Make sure to checkout the js-quic codebase to get an idea of the structure of the desired `WebSocketClient` and `WebSocketServer`.

CMCDragonkai:
The https://github.com/uNetworking/uWebSockets has a nodejs bridge and I can see people using it for android and ios, it might be worthwhile exploring that instead if it is suitable for us.

This is the JS bridge: https://github.com/uNetworking/uWebSockets.js/

The ws library is very node specific, won't work outside of nodejs.

CMCDragonkai:
The WS library can take http2 object if you `allowHTTP1` to be true, however it still only uses HTTP 1.1, not HTTP2. The uWebSockets library doesn't have this problem, the HTTP server/websocket stuff is all in C++17 so no composition with nodejs specific servers. This can be a better place to start to avoid migration down the line.

tegefaulkes:
We have a good idea now what needs to be done for client communication with websockets. There is a fair amount of work needed for that but it is also separate enough to be done last. What we have now is more than enough for a proof of concept.

Given the websocket stuff is seperate enough, I'm going to make a new issue for tracking that part of the work. We can keep the spec and discussion about it there.

Quick question, why is the client communication being done with websockets and not with the quic based system we're going to use for agent-agent communication? Isn't the quic sufficient for our needs?

tegefaulkes:
Created issue MatrixAI/Polykey#503 for tracking the websocket client implementation and spec.

tegefaulkes:
I may also make an issue for converting the old GRPC handlers to the new agnostic RPC and ripping out all of the old code.

CMCDragonkai:
Just check the old existing issues to see if there's stuff that should be migrated to it. I think there was an old GRPC review, all of those concerns are still valid here during migration, but some things might be solved by this migration, therefore we should annotate on each item how they are resolved by our migration.

tegefaulkes:
Here is a short list of open GRPC related issues.
- MatrixAI/Polykey#495 This one is already related to this PR but not explicitly fixed by it.
-  MatrixAI/Polykey#466 We provide a `ctx` object with an abort signal But we don't support any deadlines right now. It would be simple to add but needs some planning and discussion within the context of the new agnostic RPC system and how it ties in with the quic and websockets.
- MatrixAI/Polykey#400 Is related but comes down to the quic and websocket implementation. I think IPv6 should be supported by both.
- MatrixAI/Polykey#422 is related
- MatrixAI/Polykey#333 is potentially solved by the last stage of this where we rip out the old GRPC and by extension re-do the  `NodeConnection` and `NodeConnectionManager` stuff.
- MatrixAI/Polykey#332 Becomes irrelevant since connection state depends on the quic and websocket implementation.
- MatrixAI/Polykey-CLI#18 is only slightly related and not addressed here.
- MatrixAI/Polykey#276 should be irrelevant or addressed by this.
- MatrixAI/Polykey#235 Not sure but definitely affected.
- MatrixAI/Polykey#155 
- MatrixAI/Polykey#166 
- MatrixAI/Polykey#354 
- MatrixAI/Polykey-CLI#6 
- MatrixAI/Polykey#148

This list is just a first blush and isn't exhaustive

tegefaulkes:
things todo:
- finish off raw handlers, almost done.
- manifest
- static registering of handlers
- static registering of middleware
- middleware generically composeable, so long as it forms a transformation of `UInt8Array` -> `JsonValue`.

tegefaulkes:
Raw handlers are done now. In the end split the handling and transformation into two stages. we need to read a message to determine the the method to use. So we need need to send a single 'header' message for the raw streams. I used a transformation stream to read off this first message and return it via a promise. It then acts as a pass-through transform that just passes on the `UInt8Array` data. 

Once we have the header data we can determine the method and get the expected header as usual. The header is also passed along to the raw stream handler.

The usual handlers are layered on top of the raw handler now. So all handlers are now raw handlers but are transformed to the usual handlers. The `registerDuplexHandler` method converts a duplex handler into a raw handler now. This also is where all of the middleware setup is done now. The header message is also re-inserted into the middleware strean here via the parsing transformer. 

Note that this means only the normal handlers use the middleware. The raw handlers do not support this. If you HAVE to do authentication on a raw handler you need to process the header's params within the raw handler method. So authentication is possible but we can't respond with any metadata messages. I don't have a nice way of returning any metadata for raw handlers.

tegefaulkes:
next step is to work on the client manifest. I should update some specs at this point as well.

CMCDragonkai:
You should rebase on staging.

tegefaulkes:
Remaining TODOS:
- [x] asynciterable type
- [x] derive manifest
- [x] class structure for handlers
- [x] dependency injection
- [x] error propagation tests
- [x] caller should focus on convenience style and raw RWP
- [x] verify that tests are comprehensive (including authentication middleware)
- [x] indexed types with reserved keys for metadata

tegefaulkes:
```ts
await rpcClients.withMethods.client(async function* (output: Promise<...>) {
  yield 'abc';
}
```

```ts
rpcClient.methods (this should be the convenience style):

unary: (input) => promise<output>
serverS: (input) => asynciterable<output>
clientS: async function* (output: promise<output>) { ... }
duplexS: async function* (output: asynciterable<output>) { ... }
```
Optionally we can drop down to more lower-level expressions
```ts
rpcClient.raw.methods:

rawJSON: {readable, writable} (RWP of JSON)
rawBytes: {readable, writable}: RWP of ByteStream
```
REMOVE these 2
```ts
serverS: (input) => readable 
clientS: {output, writable}
```

tegefaulkes:
For the manifest, in stead of using a tagged structure, we can use classes to allow pattern matching with the `instanceOf` keyword.

We can have a base abstract class and a have that extended by the 5 specific handler classes. Each handler implementation can be an extension of these 5 classes. Then we can tell them apart using the `instanceof` keyword.

The manifest becomes `Record<string, Handler>` and all of the normal type mapping can still be used.

For portability the client manifest needs to be defined separably from the server handlers. The same concept can be used, but use simpler client specific classes that just encode the types.

tegefaulkes:
Class based manifests complete.

CMCDragonkai:
Please add a task regarding the indexed types with reserved keys.

On 8 February 2023 10:59:03 pm GMT-08:00, Brian Botha ***@***.***> wrote:
>Class based manifests complete.
>
>-- 
>https://github.com/MatrixAI/Polykey/pull/498#issuecomment-1423723957
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 

tegefaulkes:
The metadata type looks like this.

```ts
type WithMetadata<T extends Record<string, JSONValue> = Record<string, JSONValue>> = {
  metadata: {
    token: string,
    [Key: string]: JSONValue,
  }
} & Omit<T, 'metadata'> // Omit here prevents us from overriding the metadata type

const asd: WithMetadata<
  {
    hello: string,
    goodbye: string,
  }
  > = {
  metadata: {
    token: 'lol',
    asd: 1,
  },
  hello: 'hi',
  goodbye: 'bye',
}
```

CMCDragonkai:
That's good but I think the type name would be prefixed with `RPC`. Like `type RPCRequestParams` or `RPCResponseResult` for the JSON request & response.

tegefaulkes:
I've changed the type slightly for convenience. This allows us to no explicitly specify the metadata or the token for every call or response. It was annoying since most use cases the metadata was only handled in the middleware.

```ts
// eslint-disable-next-line
type NoData = {};

type WithMetadata<T extends Record<string, JSONValue> = NoData> = {
  metadata?: {
    [Key: string]: JSONValue;
  } & Partial<{ Authorization: string }>;
} & Omit<T, 'metadata'>;
```

tegefaulkes:
There is a slight problem with changing the client caller to the following.

```ts
// From
type ClientCallerImplementation<
  I extends JSONValue = JSONValue,
  O extends JSONValue = JSONValue,
  > = (f: () => AsyncGenerator<I>) => Promise<O>;

// To
type ClientCallerImplementation<
  I extends JSONValue = JSONValue,
  O extends JSONValue = JSONValue,
  > = (f: (output: Promise<O>) => AsyncGenerator<I>) => Promise<void>;
```

Previously signalled ending when the input generator returned.  With the new style that's not really possible since we need to await the output inside the generator.  SO we need a way to signal the end of the input without ending the generator. It's possible to allow yielding `undefined` since that's not an allowed `JSONValue`. 

Another annoyance here is that if we want the output outside of the caller's callback then we'd have to kidnap the value and that's not ideal flow. 

I'll see what I can do.

Ok, to signal the end of writing you need to yield `undefined`. This will close the input stream. Usage is largely unchanged besides this.

CMCDragonkai:
Can this undefined signal be made consistent across other usages? Also are we not able to use `return`?

CMCDragonkai:
To clarify on this problem, this is because when you are in the Async generator, if you call `return`, you end the generator. If `return` is what is used to end the generator and that signals the end of the stream, then you have a problem if you want to signal the end of the stream and then read the `output`.

So you're proposing to have `yield undefined` act as the signal to end the stream instead of `return` (or perhaps in addition to `return`) so you can end the stream to the server and also read the output and do whatever side-effects are necessary.

Here is some issues with this.

1. What happens on subsequent `yield` after `yield undefined;`?
2. On the caller side, one makes a call and puts a callback in it. Suppose we want to use the data provided by the server in subsequent operations, most likely we would not do this inside the async generator context itself because that would result in an callback mountain, we want to simulate synchronous programming with async await. Therefore the only way to do this is to assign the result of the output promise to a let-scoped variable and then continue after awaiting for the call to finish, then use the let-scoped variable.

The second problem is more general to the design of callback-based async generators.

An alternative is to have the async generator on the outside and call `await csg.next(value);` as a way of writing data into the generator and have a separate promise for the output. So basically client streaming call would return `[output, csg]`.

This does make it closer to the RWP construct where you just have the readable and writable stream.

Maybe we even just copy that object and have:

```
{
  readable: Promise<...>;
  writable: AsyncGenerator<...>;
}
```

Which is just a wrapper around the RWP.

Here's the thing about client streaming. It's actually very close the duplex streaming. It has shared constraint which is that you have to write at least 1 thing before you can read. If you don't write at least 1 thing, the first message would never be written and technically the stream would never be created. I think is because stream creation in QUIC is the same as writing the first message.

CMCDragonkai:
See: https://docs.rs/quiche/latest/quiche/struct.Connection.html#method.stream_send

About the last constraint. This means you always need to write at least one message other wise attempting to read the output first would result in a promise deadlock. This applies to both client streaming and server streaming.

We could even throw an exception on read if the writable hasn't had one write yet. This could be done on `QUICStream`.

CMCDragonkai:
I propose a change from:

```
clientS: async function* (output: promise<output>) { ... }
duplexS: async function* (output: asynciterable<output>) { ... }
```

To

```
clientS: [readable: promise<output>, writable: AsyncGenerator]
duplexS: [readable: AsyncGenerator, writable: AsyncGenerator]
```

You can use tuple types to label the properties.

tegefaulkes:
That would essentially be reverting back to the original non-convenient form of these two methods. I don't think we need to change the duplex caller since we already support both the convenient and original style of it. Also there're isn't an issue with the duplex caller?

The client caller I can revert back to the original style of returning `{output: Promise<O>, writable: WriteableStream<I>}` or have the writable be a generator but the usage is functionally the same.

We may need to discuss this change.

CMCDragonkai:
It's mostly the realisation that async generators aren't that convenient on the calling side.

tegefaulkes:
List of current tests

```
[faulkes@matrix-eurocom-p7-pro-se:~/matixWorkspace/gitRepos/Polykey]$ npm run test --maxWorkers=50% /tests/RPC /tests/clientRPC

> polykey@1.0.1-alpha.0 test
> jest "/tests/RPC" "/tests/clientRPC"

Determining test suites to run...
GLOBAL SETUP
Global Data Dir: /tmp/polykey-test-global-HQMrY3
 PASS  tests/RPC/RPC.test.ts (22.75 s)
  RPC
    ✓ RPC communication with raw stream (with seed=1676263408292) (698 ms)
    ✓ RPC communication with duplex stream (with seed=1676263408292) (1480 ms)
    ✓ RPC communication with server stream (with seed=1676263408292) (1744 ms)
    ✓ RPC communication with client stream (with seed=1676263408292) (880 ms)
    ✓ RPC communication with unary call (with seed=1676263408292) (570 ms)
    ✓ RPC handles and sends errors (with seed=1676263408293) (87 ms)

 PASS  tests/clientRPC/handlers/agentStatus.test.ts (29.286 s)
  agentStatus
    ✓ get status %s (576 ms)

 PASS  tests/clientRPC/handlers/agentUnlock.test.ts (29.454 s)
  agentUnlock
    ✓ get status (593 ms)

 PASS  tests/RPC/RPCServer.test.ts (28.576 s)
  RPCServer
    ✓ can stream data with raw duplex stream handler (with seed=1676263408150) (51 ms)
    ✓ can stream data with duplex stream handler (with seed=1676263408150) (900 ms)
    ✓ can stream data with client stream handler (with seed=1676263408150) (1055 ms)
    ✓ can stream data with server stream handler (with seed=1676263408150) (513 ms)
    ✓ can stream data with server stream handler (with seed=1676263408150) (637 ms)
    ✓ Handler is provided with container (with seed=1676263408150) (1317 ms)
    ✓ Handler is provided with connectionInfo (with seed=1676263408150) (931 ms)
    ✓ Handler can be aborted (with seed=1676263408150) (1290 ms)
    ✓ Handler yields nothing (with seed=1676263408151) (565 ms)
    ✓ should send error message (with seed=1676263408151) (848 ms)
    ✓ should send error message with sensitive (with seed=1676263408151) (712 ms)
    ✓ should emit stream error (with seed=1676263408151) (632 ms)
    ✓ forward middlewares (with seed=1676263408151) (805 ms)
    ✓ reverse middlewares (with seed=1676263408151) (774 ms)
    ✓ forward middleware authentication (with seed=1676263408151) (360 ms)

 PASS  tests/clientRPC/websocket.test.ts (30.017 s)
  websocket
    ✓ websocket should work with RPC (1230 ms)

 PASS  tests/RPC/RPCClient.test.ts (29.702 s)
  RPCClient
    ✓ raw caller (with seed=1676263408209) (1044 ms)
    ✓ generic duplex caller (with seed=1676263408210) (1044 ms)
    ✓ generic server stream caller (with seed=1676263408210) (1213 ms)
    ✓ generic client stream caller (with seed=1676263408210) (460 ms)
    ✓ generic unary caller (with seed=1676263408210) (934 ms)
    ✓ generic duplex caller can throw received error message (with seed=1676263408211) (843 ms)
    ✓ generic duplex caller can throw received error message with sensitive (with seed=1676263408211) (937 ms)
    ✓ generic duplex caller can throw received error message with causes (with seed=1676263408211) (577 ms)
    ✓ rawDuplexStreamCaller (with seed=1676263408211) (475 ms)
    ✓ generic duplex caller with forward Middleware (with seed=1676263408211) (738 ms)
    ✓ generic duplex caller with reverse Middleware (with seed=1676263408212) (1001 ms)
    ✓ manifest raw duplex call (with seed=1676263408212) (804 ms)
    ✓ manifest server call (with seed=1676263408212) (754 ms)
    ✓ manifest client call (with seed=1676263408212) (403 ms)
    ✓ manifest unary call (with seed=1676263408212) (332 ms)
    ✓ manifest raw duplex caller (with seed=1676263408212) (643 ms)
    ✓ manifest duplex caller (with seed=1676263408212) (194 ms)
    ✓ manifest without handler errors (27 ms)

 PASS  tests/RPC/utils.test.ts (35.434 s)
  utils tests
    ✓ can parse json stream (with seed=1676263408201) (5435 ms)
    ✓ can parse json stream with random chunk sizes (with seed=1676263408201) (4838 ms)
    ✓ Will error on bad data (with seed=1676263408202) (2916 ms)
    ✓ can parse messages (with seed=1676263408203) (96 ms)
    ✓ Message size limit is enforced (with seed=1676263408203) (1460 ms)
    ✓ merging transformation stream (with seed=1676263408203) (446 ms)
    ✓ can get the head message (with seed=1676263408203) (2977 ms)

Test Suites: 7 passed, 7 total
Tests:       49 passed, 49 total
Snapshots:   0 total
Time:        38.029 s
Ran all test suites matching /\/tests\/RPC|\/tests\/clientRPC/i.
GLOBAL TEARDOWN
Destroying Global Data Dir: /tmp/polykey-test-global-HQMrY3


```

CMCDragonkai:
These are our choices for the caller methods:

```
unary(X): P
serverS(X): Readable
clientS(): [P, Writable] - X
clientS(): {Output, Writable}
duplexS(): [Readable, Writable] - X
duplexS(): {Readable, Writable}
```

Ignore the X marked.

CMCDragonkai:
In the future, client middleware and server middleware should be separated, and can be composed and setted on the client and server after creation. Or if this is a simple change, this can be done now. Then middleware tests can be their own test directory/file because they can be test independently.

**Both handlers and middleware should be parameterisable on creation. They do not actually have to be dynamically settable afterwards.**

CMCDragonkai:
Middleware tests should have descriptive names for specific middleware.

CMCDragonkai:
After ticking off everything, time to start squashing.

tegefaulkes:
The caller methods have been updated again.

tegefaulkes:
Squashed and re-based on staging

tegefaulkes:
Enabling CI

tegefaulkes:
There seems to be some issues with all of the tests and building. But that's something coming from the staging branch so I'll have to fix that there. I should be good to merge now.

PR #504, opened 2023-02-03
Preparing Polykey for Local Demo

tegefaulkes:
### Description

Working software trumps diagrams/presentations/talks. So we want a working demo with associated diagrams running right off staging branch atm.

The relevant commands that we need to see working are below.

For each of the commands we need:

1. Asciinema recording of the CLI execution
2. Diagram on excalidraw showing the control-flow and data-flow from the user's perspective having the components of: User, CLI, Agent, FileSystem, Identity Provider (any thing else that is relevant)
3. Fix up any bugs or minor usability issues here

### Agent Start

```sh
pk agent start
```

This command needs:

1. Bootstrap the `~/.local/share/polykey` if it doesn't exist.
2. Start the agent if the agent state already exists.
3. Ask you to setup a new password and provide you a recovery code if it doesn't exist.
4. Ask you enter the existing password or recovery code if it does exist.
5. Give you the current agent status once booted up.

### Agent Stop

```sh
pk agent stop
```

This command needs:

1. Shutdown the agent gracefully if it is running.
2. Do nothing if the agent is not running.

### Agent Status

```sh
pk agent status
```

This command needs:

1. Show the agent status when the agent is running.

### Vaults Create

```sh
pk vaults create myvault
```

This command needs:

1. Create the vault called `myvault`.

### Secrets Create

```sh
pk secrets create ./the-secret-file.txt myvault:the-secret-file.txt
```

This command needs:

1. Upload the secret file from `./the-secret-file.txt` into `myvault`.

### Secrets Get

```sh
pk secrets get myvault:the-secret-file.txt
```

This command needs:

1. Show the contents of `the-secret-file.txt` on the STDOUT of the CLI.

### Secrets Delete

```sh
pk secrets delete myvault:the-secret-file.txt
```

The command needs:

1. Delete the the file from `myvault`.

### Secrets Env

```sh
pk secrets env myvault:the-secret-file.txt bash
```

This command needs:

1. This needs to inject the value of the secret into the environment of the bash shell.
2. This is optional, it's a stretch goal.
3. Addresses MatrixAI/Polykey-CLI#31.

### Identities Authenticate

```sh
pk identities authenticate github.com cmcdragonkai
```

This command needs:

1. Authenticates with github, and store the provider token

### Identities Claim

```sh
pk identities claim github.com cmcdragonkai
```

This command needs:

1. Claim the identity on the cmcdragonkai gists: https://gist.github.com/CMCDragonkai, and post a cryptolink there

### Tasks

- [x] 1. `pk agent start`
- [x] 2. `pk agent stop`
- [x] 3. `pk agent status`
- [x] 4. `pk vaults create`
- [x] 5. `pk secrets create`
- [x] 6. `pk secrets get`
- [x] 7. `pk secrets delete`
- [ ] 8. `pk secrets env`
- [x] 9. `pk identities authenticate`
- [x] 10. `pk identities claim`

### Final checklist

* [x] Domain specific tests
* [ ] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -394,6 +394,9 @@ class PolykeyAgent {
           logger: logger.getChild(SessionManager.name),
           fresh,
         }));
+      // If a recovery code is provided then we reset any sessions in case the
+      //  password changed.
+      if (keyRingConfig.recoveryCode != null) await sessionManager.resetKey();
       grpcServerClient =
         grpcServerClient ??
         new GRPCServer({
--- a/src/bin/agent/CommandStart.ts
+++ b/src/bin/agent/CommandStart.ts
@@ -10,6 +10,7 @@ import type { PolykeyWorkerManagerInterface } from '../../workers/types';
 import path from 'path';
 import childProcess from 'child_process';
 import process from 'process';
+import * as keysErrors from '../../keys/errors';
 import CommandPolykey from '../CommandPolykey';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
@@ -218,11 +219,18 @@ class CommandStart extends CommandPolykey {
           await workerManager?.destroy();
           await pkAgent?.stop();
         });
-        pkAgent = await PolykeyAgent.createPolykeyAgent({
-          fs: this.fs,
-          logger: this.logger.getChild(PolykeyAgent.name),
-          ...agentConfig,
-        });
+        try {
+          pkAgent = await PolykeyAgent.createPolykeyAgent({
+            fs: this.fs,
+            logger: this.logger.getChild(PolykeyAgent.name),
+            ...agentConfig,
+          });
+        } catch (e) {
+          if (e instanceof keysErrors.ErrorKeyPairParse) {
+            throw new binErrors.ErrorCLIPasswordWrong();
+          }
+          throw e;
+        }
         if (options.workers !== 0) {
           workerManager = await workersUtils.createWorkerManager({
             cores: options.workers,
--- a/src/bin/errors.ts
+++ b/src/bin/errors.ts
@@ -31,6 +31,11 @@ class ErrorCLIClientOptions<T> extends ErrorCLI<T> {
   exitCode = sysexits.USAGE;
 }
 
+class ErrorCLIPasswordWrong<T> extends ErrorCLI<T> {
+  static description = 'Wrong password, please try again';
+  exitCode = sysexits.USAGE;
+}
+
 class ErrorCLIPasswordMissing<T> extends ErrorCLI<T> {
   static description =
     'Password is necessary, provide it via --password-file, PK_PASSWORD or when prompted';
@@ -90,6 +95,7 @@ export {
   ErrorCLI,
   ErrorCLINodePath,
   ErrorCLIClientOptions,
+  ErrorCLIPasswordWrong,
   ErrorCLIPasswordMissing,
   ErrorCLIPasswordFileRead,
   ErrorCLIRecoveryCodeFileRead,
--- a/src/bin/identities/CommandAuthenticate.ts
+++ b/src/bin/identities/CommandAuthenticate.ts
@@ -16,15 +16,10 @@ class CommandAuthenticate extends CommandPolykey {
       'Name of the digital identity provider',
       parsers.parseProviderId,
     );
-    this.argument(
-      '<identityId>',
-      'Digital identity to authenticate',
-      parsers.parseIdentityId,
-    );
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
-    this.action(async (providerId, identityId, options) => {
+    this.action(async (providerId, options) => {
       const { default: PolykeyClient } = await import('../../PolykeyClient');
       const identitiesPB = await import(
         '../../proto/js/polykey/v1/identities/identities_pb'
@@ -59,7 +54,6 @@ class CommandAuthenticate extends CommandPolykey {
         });
         const providerMessage = new identitiesPB.Provider();
         providerMessage.setProviderId(providerId);
-        providerMessage.setIdentityId(identityId);
         await binUtils.retryAuthentication(async (auth) => {
           genReadable = pkClient.grpcClient.identitiesAuthenticate(
             providerMessage,
@@ -90,7 +84,7 @@ class CommandAuthenticate extends CommandPolykey {
               case identitiesPB.AuthenticationProcess.StepCase.RESPONSE: {
                 const authResponse = message.getResponse()!;
                 this.logger.info(
-                  `Authenticated digital identity provider ${providerId} with identity ${identityId}`,
+                  `Authenticated digital identity provider ${providerId}`,
                 );
                 process.stdout.write(
                   binUtils.outputFormatter({
--- a/src/bin/secrets/CommandGet.ts
+++ b/src/bin/secrets/CommandGet.ts
@@ -15,10 +15,6 @@ class CommandGet extends CommandPolykey {
       'Path to where the secret to be retrieved, specified as <vaultName>:<directoryPath>',
       parsers.parseSecretPath,
     );
-    this.option(
-      '-e, --env',
-      'Wrap the secret in an environment variable declaration',
-    );
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
@@ -54,7 +50,6 @@ class CommandGet extends CommandPolykey {
           port: clientOptions.clientPort,
           logger: this.logger.getChild(PolykeyClient.name),
         });
-        const isEnv: boolean = options.env ?? false;
         const secretMessage = new secretsPB.Secret();
         const vaultMessage = new vaultsPB.Vault();
         vaultMessage.setNameOrId(secretPath[0]);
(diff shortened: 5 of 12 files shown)


tegefaulkes:
- `PK agent start` is not prompting for the password twice when starting from clean state. It is however prompting twice when starting with existing state but specifying the `--fresh` flag.
- when starting fresh and specifying the recovery code we do get two prompts as expected.
- when starting with state and specifying the recovery code we get two prompts as expected.

So there is a bug with starting from no state where the password is only prompted once, it should be twice.

- State is maintained when starting with a recovery code from existing state.
- The password is changed when recovering
- The existing session token is still valid after recovering

I think current sessions should be locked if we change the root password of the agent. That doesn't seem to be happening right now.

amydevs:
Asciinemas
1. pk agent start
    - [First Run](https://asciinema.org/a/oYJrDrFQx7bYjEBNG1C4RnqyF)
    - [Subsequent Runs](https://asciinema.org/a/rOhQNHQSttbqDCzOjHkiBwxHv)
2. [pk agent stop](https://asciinema.org/a/Y9zNP8VwiosjMZXwYH2FXS2mG)
3. [pk agent status](https://asciinema.org/a/3GvPhjlyMEHpPWZSJGqFQOYml)
4. [pk vaults create](https://asciinema.org/a/IsoPlkmV2YjG3RToSLEaIOogD)
5. [pk secrets create](https://asciinema.org/a/059CKyhxAme8rYsTGb0oRIa7f)
6. [pk secrets get](https://asciinema.org/a/7LyKkJOdh9BW5Pcv0nlcgF94T)
7. [pk secrets delete](https://asciinema.org/a/tEQ2mGNPEzpvDvuSxhNJ8yNVJ)
8. pk secrets env - not yet ready
9. [pk identities authenticate](https://asciinema.org/a/JUS5HzqXVZBPwEVrPkzcVQSz5)
10. [pk identities claim](https://asciinema.org/a/ORXYE8KViVebZyeo7ddiuS4LO)

tegefaulkes:
Starting the Polykey agent with an invalid password gives a valid but ugly response.

```
[nix-shell:~/matixWorkspace/gitRepos/Polykey]$ npm run polykey -- agent start --node-path /tmp/pkt1

> polykey@1.0.1-alpha.0 polykey
> ts-node src/bin/polykey.ts "agent" "start" "--node-path" "/tmp/pkt1"

✔ Please enter the password … ********
WARN:polykey.PolykeyAgent:Failed Creating PolykeyAgent
ErrorKeyPairParse: Unable to parse root keypair - Private key path /tmp/pkt1/state/keys/private.jwk is not a valid encrypted JWK
  exitCode      74
  timestamp     Fri Feb 03 2023 14:23:23 GMT+1100 (Australian Eastern Daylight Time)
```

This should be a cleaner output with just the message along the lines of `invalid password` and prompt to try again. This would _feel_ broken to an end-user.

We should catch and re-throw the error here as a wrong password error since that makes more sense in this context. Also we should look at less verbose formatting of the errors.

amydevs:
> Starting the Polykey agent with an invalid password gives a valid but ugly response.
> 
> ```
> [nix-shell:~/matixWorkspace/gitRepos/Polykey]$ npm run polykey -- agent start --node-path /tmp/pkt1
> 
> > polykey@1.0.1-alpha.0 polykey
> > ts-node src/bin/polykey.ts "agent" "start" "--node-path" "/tmp/pkt1"
> 
> ✔ Please enter the password … ********
> WARN:polykey.PolykeyAgent:Failed Creating PolykeyAgent
> ErrorKeyPairParse: Unable to parse root keypair - Private key path /tmp/pkt1/state/keys/private.jwk is not a valid encrypted JWK
>   exitCode      74
>   timestamp     Fri Feb 03 2023 14:23:23 GMT+1100 (Australian Eastern Daylight Time)
> ```
> 
> This should be a cleaner output with just the message along the lines of `invalid password` and prompt to try again. This would _feel_ broken to an end-user.

I think also it might be more intuitive to say "Please enter a password:" when the user is initiating PK cos "the" kinda makes it sound like it's something the user should already know.

Also for No. 5 and 6 of https://github.com/MatrixAI/Polykey/pull/504#issuecomment-1414767302, am I doing this correctly? Cos `pk secrets get` returns the secret name without the contents 4 me (Asciinemas are hyperlinked for reference) @tegefaulkes

tegefaulkes:
I think you're doing it correct. It's likely a bug, I'll have a look at the command.

tegefaulkes:
I've fixed `polykey secrets get` command. Give it another go.

It was reading the secreName field of the protobuf message for the secrets contents. We don't return anything in the name field since we provide the name when getting the secret. I just had to use the contents of the `secretContents` field of the protobuf message.

@CMCDragonkai suggested we return the secret name along side the contents but I'm going to defer that for now since it's not a super quick change and it's low priority.

CMCDragonkai:
I forgot that PR MatrixAI/Polykey#265 already exists for `pk secrets env`, but it's quite outdated. I started prototyping the command, and I think the only thing missing is to have a kexec, but we can do something simpler for now, which is just use run a subprocess.

tegefaulkes:
I should check that the output for formatter is used for the output for all of these commands, even for single line outputs.

tegefaulkes:
I think pretty much everything is covered except for the ENV command. 

Just one other thing though. when changing the password on the agent via starting the agent with recovery. any active client session is still valid. I think if the password changes then we should have to unlock it again with the new password.

What behaviour do we want here?

CMCDragonkai:
Yea if the password changes, the session tokens should be reset.

CMCDragonkai:
> Asciinemas
> 
> 1. pk agent start
>    
>    * [First Run](https://asciinema.org/a/oYJrDrFQx7bYjEBNG1C4RnqyF)
>    * [Subsequent Runs](https://asciinema.org/a/rOhQNHQSttbqDCzOjHkiBwxHv)
> 2. [pk agent stop](https://asciinema.org/a/Y9zNP8VwiosjMZXwYH2FXS2mG)
> 3. [pk agent status](https://asciinema.org/a/3GvPhjlyMEHpPWZSJGqFQOYml)
> 4. [pk vaults create](https://asciinema.org/a/IsoPlkmV2YjG3RToSLEaIOogD)
> 5. [pk secrets create](https://asciinema.org/a/059CKyhxAme8rYsTGb0oRIa7f)
> 6. [pk secrets get](https://asciinema.org/a/CHdsyVZEK8h4RrULo8hyL6WyS)
> 7. [pk secrets delete](https://asciinema.org/a/tEQ2mGNPEzpvDvuSxhNJ8yNVJ)
> 8. pk secrets env - not yet ready
> 9. [pk identities authenticate](https://asciinema.org/a/JUS5HzqXVZBPwEVrPkzcVQSz5)
> 10. [pk identities claim](https://asciinema.org/a/ORXYE8KViVebZyeo7ddiuS4LO)

Awesome!

In the future we should have an org account under `enquiry@matrix.ai` and publish our asciicasts there (which can be embedded in our blog). Although, it may be preferable to use https://github.com/asciinema/agg to convert to gifs as that's more easily sharable.

CMCDragonkai:
The `pk secrets get` command should have roundtrip tests that test verbatim file content being saved and printed later that preserves all the newlines of the original content.

Also binary data too. If the secret upload has binary data, we output the binary as is, as if the user were to run `cat binarydatafile`. No special consideration for binary files.

CMCDragonkai:
I'm noticing another error while the agent is running:

```
WARN:polykey.PolykeyAgent.task v0ousk901hlo01ecevdv93usj80:Failed - Reason: TypeError: Reduce of empty array with no initial value
```

It occurs after some time while running the foreground agent.

@tegefaulkes please investigate.

It should:

```
> arr = []
[]
> arr.reduce(() => {})
Uncaught TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
```

Any time we are reducing, we should always have an initial value.

CMCDragonkai:
Another thing we should have:

1. Nested dictionary output formatting, this just means recalling the same output formatter in a recursive way (while incrementing an indentation level...), however this may be complicated due to types that we haven't supported yet
2. When using nested dictionary, if the value is a string, we should avoid printing out the double quote.

```
url	"https://github.com/login/device"
userCode	"04D6-9B53"
```

Better to do:

```
url	https://github.com/login/device
userCode	04D6-9B53
```

Similar to how numbers are printed out now.

I think the output formatter may require some refactoring to make it recursive capable.

CMCDragonkai:
The claim command isn't working for me, it fails after I do authenticate:

```
[nix-shell:~/Projects/Polykey]$ node ./dist/bin/polykey.js identities claim github.com cmcdragonkai
ErrorPolykeyRemote: Remote error from RPC call
  command	identitiesClaim
  nodeId	vve3f83v1qog578bgav365bmmgtt71prpdogs2f2dcqcf59i60br0
  host	127.0.0.1
  port	42123
  timestamp	Fri Feb 03 2023 17:15:59 GMT+1100 (Australian Eastern Daylight Time)
  cause: ErrorProviderUnauthenticated: Provider has not been authenticated or access token is expired or invalid
```

amydevs:
User Flow Diagram
Drawing 2023-02-03 13 42 08 excalidraw

CMCDragonkai:
> The claim command isn't working for me, it fails after I do authenticate:
> 
> ```
> [nix-shell:~/Projects/Polykey]$ node ./dist/bin/polykey.js identities claim github.com cmcdragonkai
> ErrorPolykeyRemote: Remote error from RPC call
>   command	identitiesClaim
>   nodeId	vve3f83v1qog578bgav365bmmgtt71prpdogs2f2dcqcf59i60br0
>   host	127.0.0.1
>   port	42123
>   timestamp	Fri Feb 03 2023 17:15:59 GMT+1100 (Australian Eastern Daylight Time)
>   cause: ErrorProviderUnauthenticated: Provider has not been authenticated or access token is expired or invalid
> ```

The reason for this is because of case insensitivity of the login (possible). But later when in the `identitiesManager` this happens:

```
    if (!identities.includes(identityId)) {
      throw new identitiesErrors.ErrorProviderUnauthenticated();
    }
```

Which is a case sensitive check.

GitHub returns my username as `CMCDragonkai`.

This confusion stems from the fact that `identities authenticate github.com cmcdragonkai` actually does not require my username "identity ID" at all. We should remove it from the `authenticate` subcommand and also remove it from the protobuf message, it isn't used. Maybe in the future other providers will require it, but most providers will just expect the user the login to their portal.

The command ends up telling us the username which is `CMCDragonkai` which I can later use with `pk identities claim github.com CMCDragonkai`.

@tegefaulkes please fix this up by removing it from the command and GRPC message, we can still do a case sensitive check, but our error message can be better instead of saying the provider is not authenticated, it might be better to say something like the identity does not match the list of existing identities... which we could show too.

CMCDragonkai:
Cool so @amydevs can refactor the diagrams to reduce the complexity, label the arrows, remove filesystem where it's not important.

Then re-run the asciinema with the new changes and fixes we've applied.

CMCDragonkai:
Last thing for me is the env command.

CMCDragonkai:
I'm doing `pk secrets env` in a separate branch, going to merge this now.

CMCDragonkai:
> User Flow Diagram Drawing 2023-02-03 13 42 08 excalidraw

This will be great for Polykey Docs. Can you start incorporating this content into the docs too (keep it somewhere in the files for now so later we can properly form the content).

CMCDragonkai:
@amydevs can you get this into Polykey docs too?

PR #506, opened 2023-02-15
Websocket for Client Service API

### LINKED ISSUE #503: Websocket Client and Server for Client Service API

tegefaulkes:
### Description

This PR is for implementing a Websocket client and server pair for polykey client communication. 

### Issues Fixed

* Fixes #503
* Related #508

### Tasks

Looking at about 5 days of work.

- [x] 1. Create a `ClientServer` for websocket based communication.
  - [x] Create a `ClientServer` class for handling life-cycle of the server.
  - ~Add Secure connection with TLS, use encryption when writing files and reading them with the server. 0.5 days~ - Addressed later in issue #508 
  - [x] Apply back-pressure propagation to the streams. 0.5 days
  - [x] Support half open connections by only closing the websocket when both streams have closed.
  - [x] Handle abruptly dropped connections. 0.5 days
  - [x] Refuse any non-websocket communication. trivial
  - [x] Support IPv6
- [x] 2. Create a `ClientClient` for websocket based communication.
  - [x] Create a `ClientClient` class for handling life-cycle of the client. 0.5 days
  - [x] Make the connection secure using TLS. 0.5 days
  - [x] Add custom authentication logic for checking the server's certificate against a NodeId. 0.5 days
  - [x] Apply back-pressure propagation to the streams. 1 days
  - [x] Support half open connections by only closing the websocket when both streams have closed. 0.5 days
  - [x] Handle abruptly dropped connections. 0.5 days
  - [x] Connection timeout logic
  - [x] Support IPv6
  
### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
I'm looking over how to use `uWebsocket` and there are some things that may be issues.

1. SSL certs are provided as files, we can't provide the certificate from memory. 
2. Certs are set when creating the server. I don't see a way to update them without restarting the server.
3. There is no websocket client provided. So if we use this for the server side we still need something else for the client side of the connection.

tegefaulkes:
I think we can use this PR to update the names for the client and server. Before this was referred as the ClientClient and ClientServer. I feel this is a little confusing on a cold read. Maybe we should rename it to `ControlClient` and `ControlServer`? Or anything besides client to refer to the client communication.

tegefaulkes:
I have a really basic server working now. I need to make it much more robust and make the connection secure.

CMCDragonkai:
It's the `ClientService` API. You can expand the name and say `ClientServiceClient`. But `ClientClient` is preferred to be short. Let's not change it right now.

CMCDragonkai:
@tegefaulkes you also need to consider how TLS will work here. Especially with uWebsockets, the TLS may be something that is built in, meaning the native library ships with OpenSSL. I see that it says it can build with openssl or wolfssl. Nodejs has its own TLS library of course, I wonder if that can be integrated, so it can use the existing openssl library in nodejs, but of course on non-node platforms the SSL library has to come from elsewhere.

Are you using the JS wrapper they are supplying or are you creating your own native bridge? If you are doing the native bridge you'll need to use the C++ style I'm using in our js-db which binds to rocksdb.

There is another framework, a bit more advanced called node-addon-api. I mentioned upgrading to that properly here: https://github.com/MatrixAI/js-db/issues/37

For the purposes of speed though, using their existing wrapper will be sufficient, we can improve on it later.

CMCDragonkai:
@tegefaulkes can you expand the tasks here in this PR based on the issue spec.

tegefaulkes:
I've implemented server side writable back-pressure. `ws.send()` returns a number based on the result. the 3 results are 1: success, 0: backpressure and 2: message dropped. So to apply back-pressure I'm using a promise as a semaphore, in the writable stream `write()` method i'm awaiting the back-pressure before writing. There is a drain event that resolves the promise when back-pressure is released.

I'm looking into applying back-pressure to the readable side but I can't find a mechanism for applying it. I'd expect some way to pause any messages from being emitted for a websocket. There is a `cork()` method but that seems to be used for grouping outgoing messages?. Their example only shows writing back-pressure with `ws.send()`. I'm still looking into it.

tegefaulkes:
I can't find any way to implement back-pressure for receiving data. These two issues seem to support that pausing reads is not allowed. 

https://github.com/uNetworking/uWebSockets.js/issues/661
https://github.com/uNetworking/uWebSockets/issues/1020

This means I don't have a way of propagating back-pressure from server reading to the client writing. The best I can do is throw an error if we exceed a buffered amount. 

This would only be a problem if the stream handler can't keep up with reading. This could be likely if we do client stream where we stream something like a file. I'm not sure what to do really.

CMCDragonkai:
With respect to TLS, the encrypted TLS certificate should be using one of the PKCS standards. You should be able to use webcrypto to encrypt the file with a password. The libsodium used in keys domain won't have that supported. In the future we can build our own UWS JS binding and we can compile it in different way, or add in a JS method to the C++ code if it supports TLS certificate structure.

CMCDragonkai:
Actually we have `keys/x509`. You maybe be able to encrypt the x509 cert using the functions there.

CMCDragonkai:
Based on the linked issues, the pasuse/resume does exist in UWS, but perhaps not UWS.js.

There is a comment saying that pausing a node readable stream just does infinite buffering. Not sure if that is true.

But in the QUIC, pausing is a proper pause as it signals the other side to stop sending data, and shouldn't be acknowledging any data being sent.

CMCDragonkai:
You should create an issue upstream on UWSJS, and refer to those issues you already found, and ask for this feature.

CMCDragonkai:
If the underlying UWS can't do it, then all you can do is buffering.

The chrome is introducing `WebSocketStream` (https://developer.chrome.com/en/articles/websocketstream/) which is supposed to have backpressure on both ends.

Reference this discussion: https://github.com/uNetworking/uWebSockets.js/discussions/501

CMCDragonkai:
While the upstream issue is being awaited on, we can go with an unbound buffer for the client service. Clients are expected to be trusted.

CMCDragonkai:
According to https://github.com/uNetworking/uWebSockets/issues/1034 no way to get the remote port right now, but not relevant to client service.

CMCDragonkai:
The design of WebSocketStream might be useful to follow for our adaptation of WS into a valid stream. See the chrome page on it.

CMCDragonkai:
I read through the SO posts on this, by design the WS protocol doesn't have backpressure, at least on the client side. The only thing bringing this in is the `WebSocketStream` that chrome is supporting, but no other platforms atm. Unbounded buffering is the only option right now that will work across the board.

This should be sufficient for now, as secrets are unlikely to be LARGE data anyway.

tegefaulkes:
I'm looking at the byob readable stream. Oddly the API seems to be different from the documentation. I'm expecting to be provided a `controller.byobRequest` that contains a view that I can write into, But the `controller.byobRequest` has the type undefined. It seems like the usual API here isn't supported? I'm not sure what's going on. 

The only available usage is identical to a normal `ReadableStream<Uint8Array>`. En-queuing  data and reading it is identcal. The only difference I can find is when I pass the message to enqueue, the original reference is detached? zeroed? replaced? for the BYOB stream.

That is to say. For a regular stream when you enqueue the data it is not actually copied. So you can modify the message afterwards and have the read result changed. For the BYOB stream the enqueue message can't be modified in this way. I can't say that the BYOB version is zero copy in this case. but I know that the regular readable stream is.

Now not having the expected interface for the BYOB streams is not actually a problem. I don't think they're needed for our use case. It seems that the BYOB stream is used for when you need to read the bytes into a buffer you have. That is to say, the buffer referred to in the BYOB acronym is not internal to the readable stream but provided by whatever is consuming the reader. The usage mimics how you're read a file into a buffer. The expected usage is as follows.

```ts
const readableByteStream = new ReadableStream({...});
const reader = readablebyteStream({mode: 'byob'})'
const buffer = Buffer.alloc(100);
// Our buffer is provided to the stream internals to be written to
const {value, done} = await reader.read(buffer); 
// value is the bytes written
// done is if the stream ended
```

Since we don't consume the streams in this way I think it's unnecessary to implement the readable stream as a BYOB. But like I said. this interface isn't actually provided in our version despite all the docs saying it's available?

Unless I'm reading into this wrong or I've misunderstood something. I'm not going to implement the readable as a BYOB stream as it doesn't fit our usage.

CMCDragonkai:
I believe the purpose of BYOB stream is to be able to have fixed size buffer. This is essential for us.

And yes it's not zero copy, it's supposed to copy whatever was written in to an intermediate fixed size buffer.

A fixed size buffer is necessary as otherwise your data transfer is objects not bytes.

TS API may be out of date. Ensure you're using the latest typescript version. Use like 4.8 or 4.9 or whatever it is.

CMCDragonkai:
Or if you're talking about node, even node types can be upgraded.

tegefaulkes:
That's the thing, I don't think it's used for having a fixed sized buffer for the readable stream. I think it's just to support a BYOB reader so it can write into a provided buffer without extra copying. See the example I gave above.

We can enforce a fixed size buffering with the regular readable stream already. We just need to use the queuing strategy options to set the high water mark and a size function to map chunks to their actual size. I've used this to enforce a maximum buffered bytes for the readable stream in the `ClientServer`.

Besides that. I had a look a the `@types/node` where the interfaces seem to be defined. Even the latest version of this doesn't have the expected interface with the BYOB. The lib.dom types do have the expected interface.

CMCDragonkai:
> That's the thing, I don't think it's used for having a fixed sized buffer for the readable stream. I think it's just to support a BYOB reader so it can write into a provided buffer without extra copying. See the example I gave above.
> 
> We can enforce a fixed size buffering with the regular readable stream already. We just need to use the queuing strategy options to set the high water mark and a size function to map chunks to their actual size. I've used this to enforce a maximum buffered bytes for the readable stream in the `ClientServer`.
> 
> Besides that. I had a look a the `@types/node` where the interfaces seem to be defined. Even the latest version of this doesn't have the expected interface with the BYOB. The lib.dom types do have the expected interface.

I'm confused, every enqueue operation takes an entire buffer object, which can be or arbitrary size. Thus it's not a fixed size buffer queue.

https://github.com/whatwg/streams/blob/main/byte-streams-explainer.md

Still it seems like a good idea to use it anyway. Even if the TS interface is out of date, you can just do whatever the current node runtime supports. Just ignore the types?

tegefaulkes:
Each enqueued buffer can be an arbitrary size. but using the queuing strategy we can enforce an overall queue limit. The strategy defaults to a limit of 1 chunk. But I can tell it to queue up a certain amount of bytes. This is done by specifying the high water mark in bytes and a function that returns the number of bytes each chunk contains. Then inside the stream the `controller.desiredSize` is the amount of free space in the queue in bytes. I can use that to respect the back-pressure or throw an error.

The example you provided doesn't show it using a single internal buffer for buffering the data. It shows it directly writing the data into the buffer provided as a `view`. I can go over this in more detail in a meeting, I'm focusing on other parts right now.

CMCDragonkai:
I see, but there are other benefits to using BYOB mentioned in the link. I haven't fully understood what the implementation complexity with this yet, so we can go through it next meeting.

tegefaulkes:
Just something to keep note of. The `ClientClient` doesn't have a persistent connection between stream creations. Each new stream is technical a new connection. So while I can support providing multiple NodeIds we authenticate the server against. I can guarantee that it's the same node for each created stream.

CMCDragonkai:
Every stream could be a new connection. But on QUIC, every stream is multiplexed onto underlying connections.

For websockets, every stream is indeed a separate connection, unless the underlying websocket protocol is on HTTP2, in which case streams are in fact multiplexed on multiple HTTP2 connections.

The RPC isn't aware of connections anymore, it only considers streams.

tegefaulkes:
Added timeout logic to the `ClientClient`. We can provide the timeout in milliseconds when creating the `ClientClient` or to each connection as a `Timer`. If none is provided we depend on the websocket's timeout behaviour. 

This one is hard to test since the websocket is pretty responsive to bad connections. Giving it a port without anything running gives a connection refused error. Using a random IP address with nothing running gives a different error immediately.  So far as I can tell, if we can't establish a connection we'd end up with an error very quickly with websockets. But we have the option of our own timeout now.

tegefaulkes:
For the secure TLS certs I need to create a encrypted private key PEM. This is pretty similar to the current private key PEM that we can create using the keys utils `privateKeyToPem` utilities. There are two differences though. The PEM has a different header and footer for encrypted files and the private key is encrypted using PKCS#5 standard? The whole thing is a headache...

I need to do more digging.

tegefaulkes:
Ok. so far as I can tell I need to create the encrypted prvate key PEM using PKCS 8 using the `EncryptedPrivateKeyInfo` structure. I need to use PBES2 to encrypt the private key and and identify that as the encryption using the `AlgorithmIdentifier`... Seems like a minor change compared to the current `privateKeyToPem` utility function in the keys utilities.

Do you have any input on the @CMCDragonkai? You've looked into this before?

tegefaulkes:
I'm going to come back to the encrypted private key PEM problem later. It's a much deeper rabbit hole than I expected.

CMCDragonkai:
I think the crypto library has what you need. We wrapped webcrypto already in the keys domain as it used in the x509 module. Look into that.

tegefaulkes:
I have looked at that. Does the x509 provide an encryption method with a passphrase? 

The keys utils `privateKeyToPEM` function does mostly what I need. But I need to use the `EncryptedPrivateKeyInfo` structure and acually encrypt the key. I don't know what encryption methods are allowed but so far as I can tell I need PBES2? But that's not available with the current encryption libraries we have.

tegefaulkes:
I've created tests for the client and server where the connections abruptly fail. This is done by creating a client or server in a separate process and killing them. 

It seems that both the client and server websocket can tell this happens because they both end with a `1006` error code corresponding to `1006 | CLOSE_ABNORMAL | No close code frame has been received`. So i'm not sure we even need to set up ping/pong handlers to detect if the connection has dropped.

Another question is how we handle a connection drop. Given  in both the server and client case we receive an end event with code `1006` immediately, the current implementations just close the streams without error. I feel that we should be throwing an explicit error when the connection fails.

CMCDragonkai:
> Added timeout logic to the `ClientClient`. We can provide the timeout in milliseconds when creating the `ClientClient` or to each connection as a `Timer`. If none is provided we depend on the websocket's timeout behaviour. 
> 
> This one is hard to test since the websocket is pretty responsive to bad connections. Giving it a port without anything running gives a connection refused error. Using a random IP address with nothing running gives a different error immediately.  So far as I can tell, if we can't establish a connection we'd end up with an error very quickly with websockets. But we have the option of our own timeout now.

The timeout is more for long running tasks or when the agent stops responding. You should test for when the agent just takes too long to respond.

CMCDragonkai:
> I have looked at that. Does the x509 provide an encryption method with a passphrase? 
> 
> The keys utils `privateKeyToPEM` function does mostly what I need. But I need to use the `EncryptedPrivateKeyInfo` structure and acually encrypt the key. I don't know what encryption methods are allowed but so far as I can tell I need PBES2? But that's not available with the current encryption libraries we have.

Hmm I don't remember. But PBES2 is just an password based encryption scheme. It's not supported by libsodium (unless you google this in the docs), only webcrypto primitive would have it. The x509 library and associated peculiar ventures libraries would definitely support it.

CMCDragonkai:
See https://github.com/PeculiarVentures/x509/issues/12

CMCDragonkai:
> I've created tests for the client and server where the connections abruptly fail. This is done by creating a client or server in a separate process and killing them. 
> 
> It seems that both the client and server websocket can tell this happens because they both end with a `1006` error code corresponding to `1006 | CLOSE_ABNORMAL | No close code frame has been received`. So i'm not sure we even need to set up ping/pong handlers to detect if the connection has dropped.
> 
> Another question is how we handle a connection drop. Given  in both the server and client case we receive an end event with code `1006` immediately, the current implementations just close the streams without error. I feel that we should be throwing an explicit error when the connection fails.

Explicit connection drop is one possibility but also possible that packets get delayed to infinity. That's why ping/pong is necessary. It should already be part of the libraries since it's part of the websocket spec. You should just enable it.

Also what is 1006, where does this come from?

tegefaulkes:
`1006` is a websocket error code.  https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code

tegefaulkes:
Only thing left is the private key problem. I'm not making much progress on it currently. Everything is functional already so  we can start review and look to merging this. We can defer the private key problem for now since it's not blocking this PR and is mostly just a security concern. I can defer it to a separate issue to be solved later.

tegefaulkes:
What I know and don't know about the private key problem.

So the core of the problem is that the `uWebsocket` library can only load the private key and certs from the file system. This means we need to write the private key in the `PEM` format onto the file system to load it.

This is clearly not ideal because it means we have a secret out in the open. Right now I'm writing it to the temp file system the OS uses. This is usually in memory so it won't remain on the disk. 

So what I need to do for just a little bit more security is to convert the private key to a more secure password encrypted `PEM` format. This is proving to be more difficult than I expected.

For context, we're currently generating the `PEM` format using the peculiar's `ASN1` and `X509` library. it uses the `asn1-schema` library to structure the information and convert it to the `PEM` format. In this case it is using the `asn1Pkcs8.PrivateKeyInfo` with `x509.idEd25519` algorithm identifier and the `BER` encoded private key.

What I need to use is the `EncryptedPrivateKeyInfo` structure and provide it the algorithm identifier and the encrypted private key. From what we already have that is a pretty minor change. There are a few unknowns I've been trying to iron out.

I think the `uWebsocket` is using boring `SSL` under the hood. I think this means that it has the same support for the key file as `openSSL`. The `uSocket` library which `uWebsocket` uses `openssl genrsa` to create the private key `PEM` file in it's examples. 

It supports a few encryption methods but I've narrowed it down to `DES-CBC` if you use the `-des` option for generating the key. `DES-CBC` is supported by the peculiar `webcrypto` library so I think we can use that.

What I don't know is where to find the `AlgorithmIdentifier` that the `EncryptedPrivateKeyInfo` structure needs to specify we're using `DES-CBC`. I'm not certain that `DES-CBC` is the exact encryption we need. I'm not even certain that if I somehow create the encrypted private key format it would even be loaded by `uWebsocket` at the end of the day.

I find i'm spending most of my time just trying to build up the context around this pretty complex and convoluted crypto ecosystem. And to be honest, it's leaving a sour taste in my mouth since the only reason we need to deal with any of this is due to `uWebsocket` arbitrarily not supporting supply certs and keys from within the program.

tegefaulkes:
Cleaned up history and re-based on staging.

tegefaulkes:
Created a new issue for the private key problem at #508. I'll add it to the epic and remove it from the tasks here.

tegefaulkes:
This is ready for review. We'll likely go over it on monday.

CMCDragonkai:
Hey @tegefaulkes what about the final checklist?

Yes we can go through meeting today about this.

src/clientRPC/ClientServer.ts
```diff
@@ -0,0 +1,391 @@
...
+  drain: (ws: WebSocket<any>) => void;
+  close: (ws: WebSocket<any>, code: number, message: ArrayBuffer) => void;
+  pong: (ws: WebSocket<any>, message: ArrayBuffer) => void;
+  logger: Logger;
+  writeBackpressure: boolean;
+};
+
+interface ClientServer extends startStop.StartStop {}
+@startStop.StartStop()
+class ClientServer {
```
CMCDragonkai: This can be renamed `WebSocketServer`. It's more generic than just the client service.

src/clientRPC/ClientClient.ts
```diff
@@ -0,0 +1,345 @@
...
+import { Validator } from 'ip-num';
+import * as clientRpcUtils from './utils';
+import * as clientRPCErrors from './errors';
+import { promise } from '../utils';
+
+const timeoutSymbol = Symbol('TimedOutSymbol');
+
+interface ClientClient extends createDestroy.CreateDestroy {}
+@createDestroy.CreateDestroy()
+class ClientClient {
```
CMCDragonkai: This matches `QUICClient`, but it should be renamed `WebSocketClient`.

Both clients are `CreateDestroy` patterns.

src/clientRPC/ClientServer.ts
CMCDragonkai: This also matches `QUICServer` which is also start stop.

src/clientRPC/ClientClient.ts
```diff
@@ -0,0 +1,345 @@
+import type { ReadableWritablePair } from 'stream/web';
```
CMCDragonkai: We can move all of `clientRPC` files into `src/websockets` as its own domain since it's a generic websocket related work.

src/clientRPC/ClientServer.ts
```diff
@@ -0,0 +1,391 @@
...
+  static async createClientServer({
+    connectionCallback,
```
CMCDragonkai: This is the factory function that makes use of the actual "connection stream" produced by the web socket server.

Most of the time, the pattern is done via an event instead. Most NodeJS servers work like this.

In the `QUICServer`, we make use of the `connection` event like this:

```
quicServer.addEventListener('connection', handleConnection);
```

I think this can be done like:

```
webSocketServer.addEventListener('connection', handleConnection);
```

One difference is that the WSS (websocketserver)'s connection is the stream we use in the RPC system.

Furthermore, the WSS should extend the `EventTarget`, we use this instead of `EventEmitter` because it's generic across JS platforms.

CMCDragonkai: Suppose WSS were to dispatch an event, it would be similar to:

```
this.dispatchEvent(new events.QUICServerConnectionEvent({ detail: connection }));
```

In this case, any listener to the `connection` event will then handle the event, if the function throws a synchronous error, that will be uncaught exception.

There's a paper that argues for this case, basically `handleConnection` is like a "root of the program" that starts a new callstack. Any exceptions must be handled with a root catcher there. Or you guarantee that no exceptions can be thrown in that function.

src/clientRPC/ClientServer.ts
CMCDragonkai: This should therefore extend `EventTarget`. And you document it all the events that could be emitted.

CMCDragonkai: There should be additional events beyond just `connection` for the server.

Examples from `QUICServer` is:

1. `connection` - this is emitted when a connection is READY (after negotiation)
2. `error` - these are emitted during negotiation of a new connection
3. `stop` - emitted at the end of the `stop` function (just before the last logging statement) - this replicates nodejs server behaviour

Perhaps it is wise to have an additional class for just the `WebSocketConnection`? Which should implement the `ReadableWritablePair`?

src/clientRPC/ClientServer.ts
```diff
@@ -0,0 +1,391 @@
...
+            logger,
+          },
+          req.getHeader('sec-websocket-key'),
+          req.getHeader('sec-websocket-protocol'),
+          req.getHeader('sec-websocket-extensions'),
+          context,
+        );
+        count += 1;
+      },
+      open: (ws: WebSocket<Context>) => {
```
CMCDragonkai: I think these handlers should be arrow function properties of your class. We do this across the codebase, like in Tasks system and also js-quic atm.

src/clientRPC/ClientServer.ts
```diff
@@ -0,0 +1,391 @@
...
+        if (this.waitForActive == null) this.waitForActive = promise();
+        this.activeSockets.add(ws);
+        // Set up streams and context
+        this.handleOpen(ws);
```
CMCDragonkai: Try combining this handler with the new arrow function property. And you can just call it `handleUWSOpen`. (If you are specific here, make the `this.server` name also more specific too.).

At that point a single arrow function property should be able to deal with it all, instead of 2 different functions.

One benefit here is that you can then docblock the entire handler itself

CMCDragonkai: It becomes top-level of this class. Furthermore it has to be an arrow function property so that you maintain the `this` context, without requiring rebinding with `x.bind(this)`. Essentially it becomes a lexically bound closure.

src/clientRPC/ClientServer.ts
```diff
@@ -0,0 +1,391 @@
...
+    let count = 0;
+    const tmpDir = await this.fs.promises.mkdtemp(
+      path.join(basePath, 'polykey-'),
+    );
+    // TODO: The key file needs to be in the encrypted format
+    const keyFile = path.join(tmpDir, 'keyFile.pem');
+    const certFile = path.join(tmpDir, 'certFile.pem');
+    await this.fs.promises.writeFile(keyFile, tlsConfig.keyPrivatePem);
+    await this.fs.promises.writeFile(certFile, tlsConfig.certChainPem);
+    this.server = uWebsocket.SSLApp({
```
CMCDragonkai: It's possible that UWS statically compiled openssl into their shared object. We should check this to understand if this ends up with 2 openssl libraries inside our final bundle when running with nodejs. Ideally it should be centralising onto the node's openssl.

CMCDragonkai: In js-quic, which uses quiche, that uses boringssl, and that is statically compiled. So any verification of SSL layer gets complicated, if there is going to be 3 different ssl libraries. Node's own openssl, uws's openssl, and quiche's boringssl. @tegefaulkes can you confirm about uws's using nodejs openssl by opening an issue upstream.

src/clientRPC/ClientServer.ts
```diff
@@ -0,0 +1,391 @@
...
+    this.connectionCallback = connectionCallback;
+    let count = 0;
+    const tmpDir = await this.fs.promises.mkdtemp(
+      path.join(basePath, 'polykey-'),
+    );
+    // TODO: The key file needs to be in the encrypted format
+    const keyFile = path.join(tmpDir, 'keyFile.pem');
+    const certFile = path.join(tmpDir, 'certFile.pem');
+    await this.fs.promises.writeFile(keyFile, tlsConfig.keyPrivatePem);
+    await this.fs.promises.writeFile(certFile, tlsConfig.certChainPem);
```
CMCDragonkai: Because this portion of code is a bit of a hack. Let's move this into a protected function that does explicitly.

It should then be documented what the logic is.

1. Creating a temporary directory
2. Writing encrypted pem files, you need a key, and a cert, the cert can be public/unecrypted
3. Providing the file names to be read by the `SSLApp`.
4. Deleting the directory

As soon as the files have been read by the `SSLApp`, these can be deleted.

In order to do this, you could provide as a callback, or as 2 separate functions.

So it may look like this:

```ts
async start() {
  const uwsServer = this.setupKeyAndCert((keyPath, certPath, password) => {
    return uWebsocket.SSLApp({
      key_file_name: keyPath,
      cert_file_name: certPath,
      passphrase: password
    });
  });
}
```

CMCDragonkai: Make sure our cipher selection is secure and is suitable for modern platforms. So anywhere that we expect a client/thirdparty to call from.

CMCDragonkai: Regarding the encryption...

In order to encrypt the private into an encrypted PEM file, we need to follow the PKCS#5 standard (version 2) to produce an encrypted PKCS#8 pem file.

For example, in openssl, it has a pkcs8 subcommand. https://www.openssl.org/docs/man1.0.2/man1/pkcs8.html

In this subcommand there is a `-v2 alg` option. And in the comments it says:

> The alg argument is the encryption algorithm to use, valid values include des, des3 and rc2. It is recommended that des3 is used.

So this will basically end using an algorithm specified in pkcs#5 v2 RFC. However it's not clear which algorithm this is, and that it's OID is going to be.

One way to solve this is to actually use the `openssl pkcs8` command, and run it with a plaintext PKCS8 key. Using the `-v2 des3` option, and you get back your PKCS8 encrypted PEM.

Subsequently you then just need to interrogate this file. Probably using another `openssl` command.

Actually in the updated version https://www.openssl.org/docs/man1.1.1/man1/openssl-pkcs8.html, it turns out the `-v2 alg` can be omitted, and the default cipher becomes AES256GCM.

So that means, we can just do the encryption using `openssl pkcs8` command, then open up the file, and parse it, to get the algorithm identifier.

This identifier is going to be an OID. The OID is just a string that looks like `1.3.101.110`. But we just need to know which one is for this one.

Use the ASN1 parser to see if you can open up the encrypted file and just console.log out the version identifier string.

Once you get the confirmed OID, we can actually proceed with building this encrypted pem file.

To do this, use https://github.com/PeculiarVentures/asn1-schema/blob/master/packages/pkcs8/src/encrypted_private_key_info.ts.

Then construct the `EncryptedPrivateKeyInfo` object, with the algorithm identifier. Because this identifier may not exist in any of asn1 libraries. The string can be defined in the `utils` if we need to do this. It should go into the `keys/utils`.

The next thing to do is to actually encrypt it. It is explained here.

https://datatracker.ietf.org/doc/html/rfc5208#section-6

We would reuse our own existing code:

```ts
  const pkcs8 = new asn1Pkcs8.PrivateKeyInfo({
    privateKeyAlgorithm: new asn1X509.AlgorithmIdentifier({
      algorithm: x509.idEd25519,
    }),
    privateKey: new asn1Pkcs8.PrivateKey(
      new asn1.OctetString(privateKey).toASN().toBER(),
    ),
  });
```

That provides us the private key info object.

>   The encryption process involves the following two steps:
>
>      1. The private-key information is BER encoded, yielding an octet
>         string.
>
>      2. The result of step 1 is encrypted with the secret key to give
>         an octet string, the result of the encryption process.

But what we actually need is something like:

```ts
new asn1.OctetString(pkcs8).toASN().toBER()
```

That then produces an arraybuffer.

Then that array buffer has to go through the encryption process of AES-256-GCM, whatever is specified as part of the OID we discover through `openssl pkcs8`.

The resulting function can be called `privateKeyToEncryptedPEM`.

src/clientRPC/ClientServer.ts
CMCDragonkai: Once you extend `EventTarget`, you should have an `events.ts` module that defines all the event classes. Similar to in js-quic.

src/clientRPC/ClientClient.ts
```diff
@@ -0,0 +1,345 @@
...
+      throw e;
+    }
+    // Cleaning up connection error
+    ws.removeEventListener('error', openErrorHandler);
+
+    let readableClosed = false;
+    let writableClosed = false;
+    const readableLogger = this.logger.getChild('readable');
+    const writableLogger = this.logger.getChild('writable');
+    const readableStream = new ReadableStream<Uint8Array>(
```
CMCDragonkai: This portion along with shared with WSS, can be extracted out to its own `WebSocketConnection` class that encapsulates the logic of `ReadableStream` and `WritableStream` similar to what I'm doing with `QUICConnection` and `QUICStream`.

src/clientRPC/utils.ts
```diff
@@ -65,201 +60,149 @@ function encodeAuthFromPassword(password: string): string {
...
+ * Verify the server certificate chain when connecting to it from a client
+ * This is a custom verification intended to verify that the server owned
+ * the relevant NodeId.
+ * It is possible that the server has a new NodeId. In that case we will
+ * verify that the new NodeId is the true descendant of the target NodeId.
+ */
+async function verifyServerCertificateChain(
+  nodeIds: Array<NodeId>,
+  certChain: Array<Certificate>,
+): Promise<NodeId> {
```
CMCDragonkai: This looks like it came from the `network` domain. You should use the `grpc` domain instead.

The verification for the client service is actually different. And I believe originally it's under the `client` domain.

CMCDragonkai: For example the client service does not require verifying multiple Node IDs.

CMCDragonkai: Look for where we are verifying the server certificate chain on the GRPC client. (Which is different from how we do this via the TLS inside the `network` domain).

CMCDragonkai (changes requested):
Let's go with a new domain `src/websockets` and `tests/websockets`.

src/clientRPC/ClientServer.ts
tegefaulkes: This has been updated.

src/clientRPC/ClientClient.ts
tegefaulkes: Unfortunately the client and server implementation of the streams are different enough that they can't be shared. But I can have them share an interface or an abstract class so they can implement the same behaviour. 

As for the features the'll have, we'd need...
1. Extend the `ReadableWritablePair<Uint8Array, Uint8Array>` object.
2. Flag for when each and or both streams have ended.
3. Promises that resolve when each and or both streams have ended and reject with errors.
4. Ability to abort the active connection

This class can be set into the active map for both the client and server.

CMCDragonkai: What would be so different between the 2?

In the quic system, it's the same except how it gets constructed/created.

That on the client side, created by connecting. Whereas on server side, created by accepting.

tegefaulkes: I've created an abstract class `WebSocketStream` that standardises tracking the state of the readable, writable and websocket. It also provides an end method that needs to be implemented. It implements the `ReadableWritablePair` and can be used as such. 

It standardises the interface between the client stream and server stream. Bot the client and server extend this for their internal stream representation. The `ActiveConnections` set now contains these objects.

We can now add on to the abstract class when we need to enforce any more features for the streams.

tegefaulkes: > What would be so different between the 2?
> 
> In the quic system, it's the same except how it gets constructed/created.
> 
> That on the client side, created by connecting. Whereas on server side, created by accepting.

The `WebSocketClient` uses the `ws` library which follows nodes's style of the event based server/connections. the `uWebsocket` library we use for the `WebSocketServer` has it's own API going on so we set up the event handlers completely differently. As a result, the readable and writable web streams are structured differently compared to the client.

CMCDragonkai: I see. Well you should still separate into the classes for the connection. Yes you can then share an interface or an abstract class if they have shared implementation.

src/clientRPC/ClientServer.ts
tegefaulkes: This has been done.

src/clientRPC/ClientClient.ts
tegefaulkes: This is done.

src/clientRPC/utils.ts
tegefaulkes: I just doubled checked. This is the `VerifyServerCertificateChain` taken from the GRPC code which is using the method defined in the network utils. 

I'm leaving this as-is.

src/clientRPC/ClientClient.ts
tegefaulkes: This is done.

src/clientRPC/ClientClient.ts
tegefaulkes: It has been moved.

src/clientRPC/ClientServer.ts
tegefaulkes: it has been moved. The  other parts of this discussion are part of #508

src/clientRPC/ClientServer.ts
tegefaulkes: I did some digging and it seems that `uSockets` which `uWebsockets` uses internally uses `boringSSL`. There seems to be an option to compile with openSSL but i haven't looked at the details of that.

https://github.com/uNetworking/uSockets
```
In its minimal, TCP-only, configuration µSockets has no dependencies other than the very OS kernel and compiles down to a tiny binary. In its full configuration it depends on BoringSSL, lsquic and potentially some event-loop library.

Here are some configurations; WITH_LIBUV, WITH_ASIO, WITH_GCD, WITH_ASAN, WITH_QUIC, WITH_BORINGSSL, WITH_OPENSSL, WITH_WOLFSSL.
```

I've created an issue at https://github.com/uNetworking/uWebSockets.js/issues/877.

CMCDragonkai: If we want to change the compilation process it has to be done with a fork.

CMCDragonkai: Also if it uses boringssl we would end up with 2 boringssl in our all when using js-quic quiche.

tegefaulkes:
Should I add events to the `WebSocketStreams` as well?

tegefaulkes:
Using an `EventTarget` is fine and all for internal use. But it doesn't exactly have a nice interface for external use. For example, say we have a connection event. we'd handle it using 

```ts
webSocketServer.addEventListener('connection', (event) => {
  // do things with event
  // `event` has `Event` type, we'd have to convert it to use it properly.
});

// For convenience we want to use it like this with implicit typing.
webSocketServer.addEventListener('connection', (event) => {
  // do things with event
  //`event` has `ConnectionEvent` type
});

// Ideally we have something closer to the EventEmitter interface
webSocketServer.addEventListener('connection', (streamPair, connectionInfo) => {
  // The handler has the args provided directly with the event abstracted away.
  // Would need some complex logic for removing event listeners to work properly.
});
```

The usage is fine but there is no type data along with it. if we're using this there will be no type hints that the `connection` event exists and the `event` object has the `Event` type which hides any internal detail we'll need. There are ways to fix this by wrapping the `EventTarget` methods. But that a bit of logic that will need to be replicated across domains so we'd want to standardise it between the websocket and quic system at a minimum. 

If we want this to work anything like the `EventEmitter` interface node has then it will take a lot of work. If we want the `EventEmitter` interface and the portability then we can make a wrapper class that uses the `EventTarget` internally and implements the `EventEmitter` interface. Is this something we'd want?

I'm looking into using this with the `WebSocketServer` and `WebSocketClient` but i'm not sure I should continue until we have a standard way of using it. I've been referring to the `Quic` implementation of it but I don't think it's usage is set in stone yet. Our ways of implementing the convenience may differ.

 So I think we should at least discuss this before I continue because I am almost certain whatever I implement will have to change.

CMCDragonkai:
> Using an `EventTarget` is fine and all for internal use. But it doesn't exactly have a nice interface for external use. For example, say we have a connection event. we'd handle it using 
> 
> ```ts
> webSocketServer.addEventListener('connection', (event) => {
>   // do things with event
>   // `event` has `Event` type, we'd have to convert it to use it properly.
> });
> 
> // For convenience we want to use it like this with implicit typing.
> webSocketServer.addEventListener('connection', (event) => {
>   // do things with event
>   //`event` has `ConnectionEvent` type
> });
> 
> // Ideally we have something closer to the EventEmitter interface
> webSocketServer.addEventListener('connection', (streamPair, connectionInfo) => {
>   // The handler has the args provided directly with the event abstracted away.
>   // Would need some complex logic for removing event listeners to work properly.
> });
> ```
> 
> The usage is fine but there is no type data along with it. if we're using this there will be no type hints that the `connection` event exists and the `event` object has the `Event` type which hides any internal detail we'll need. There are ways to fix this by wrapping the `EventTarget` methods. But that a bit of logic that will need to be replicated across domains so we'd want to standardise it between the websocket and quic system at a minimum. 
> 
> If we want this to work anything like the `EventEmitter` interface node has then it will take a lot of work. If we want the `EventEmitter` interface and the portability then we can make a wrapper class that uses the `EventTarget` internally and implements the `EventEmitter` interface. Is this something we'd want?
> 
> I'm looking into using this with the `WebSocketServer` and `WebSocketClient` but i'm not sure I should continue until we have a standard way of using it. I've been referring to the `Quic` implementation of it but I don't think it's usage is set in stone yet. Our ways of implementing the convenience may differ.
> 
>  So I think we should at least discuss this before I continue because I am almost certain whatever I implement will have to change.
> 

I'm not using event emitter style at all. It's EventTarget all the way.

About the types. Yes you have to assert type in the handler parameter itself. This is normal. It should work I believe. See the `src/bin/server.ts`.

In the js-quic there's a bunch of event classes, I think if you just use that it will be fine.

CMCDragonkai:
Btw TS is correct not to give you a specific type in this instance. This is because dispatching events is actually fully generic. There's no limitation on dispatching a `connection` event with a completely different subclass.

This is why the resolution of the genericity happens in handler. In Haskell the handler would pattern match the constructors. Here we just assert the type, or a runtime check can be done.

I prefer not to bother with a runtime check when it's obvious (programmer's guarantee) that it would not be dispatched with a different subclass. Kind of like how I originally used `!` in other places.

src/clientRPC/ClientServer.ts
tegefaulkes: I've updated it to extend `EventTarget`. There are 3 events available currently, `start`, `stop` and `connection`. Though the connection really returns a `ReadableWritablePair` implementing class `WebSocketStream`... So maybe it's better to call `connection` -> `stream`?

Also note that the original way of providing the stream creation callback during `createWebSocketServer` is still available but is now optional. Internally it's registering it to the `connection` event.

src/clientRPC/ClientServer.ts
tegefaulkes: This is done.

src/clientRPC/ClientServer.ts
tegefaulkes: I'm considering this done.

PR #509, opened 2023-02-27
Client RPC migration

### LINKED ISSUE #514: fix: node discovery warning message spam
### Specification

When providing no seed nodes to connect to we would end up with logger warning messages `Failed to connect to any seed nodes when syncing node graph`. Since a lot of tests start a `PolykeyAgent` with no seed nodes provided, this ended up with a lot of useless warning messages during tests.

This can be fixed by checking if there are any provided seed nodes and skip discovery if none are provided.

### Tasks

1. Skip discovery if no seeds nodes are provided to the `NodeManager`.

tegefaulkes:
### Description

This PR deals with migrating the clientRPC handlers and code to using the new Agnostic RPC system.

It consists of taking the existing handler code within `client/services` and transplanting it into the new `RPC` system from #249 . 

##### pagination 
Any handlers that stream data need to support the pagination parameters and propagate them. The required parameters are `seek: id`, `number: number`, `order: 'asc'|'desc'`. Most code that streams an output with a generator already support pagination. But in some cases where it is not already supported it needs to be updated. As for the handler parameters, the message should include the following

```ts
type paginationMessage = {
id: string,
order: 'asc' | 'desc',
number: number,
}
```

##### Binary data
Currently there are two cases where sending binary data is applicable. The IDs such as `NodeID`, or `VaultId`. These are converted to an encoded string form just for readability of the message. Using the `NodeIdEncoded` and `VaultIdEncoded` for of the Ids. 

The 2nd case is file contents. Currently these are converted to a binary string when provided within a message using `Buffer.toString('binary')`. This is pretty limiting to small amounts of data so it is subject to change. If we need to send larger files we need to stream it using a raw handler to implement it as a raw binary stream. This has complications such it not supporting any middleware so we need to implement authentication logic withing the raw handler.

##### Standard message structure
To standardise the message structure and types across all messages we are leveraging the typescripts type system. Since it's pretty powerful we can compose messages and message types together effectively. 

For example we have basic type messages that can be composed to make more complex messages

```
type NodeIdMessage = {  
  nodeIdEncoded: NodeIdEncoded;  
};

export type AddressMessage = {
  host: string;
  port: number;
};

type NodeAddressMessage = NodeIdMessage & AddressMessage;

type NodesAddMessage = NodeAddressMessage & {  
  force?: boolean;  
  ping?: boolean;  
};

```

### Issues Fixed

* Related #249 
* Related #249 
* Fixes #514

### Tasks

1. [x] Migrate handlers.
	1. [x] agent
	2. [x] gestalts
	3. [x] identities
	4. [x] keys
	5. [x] nodes
	6. [x] notifications
	7. [x] vaults
2. [x] Migrate tests.
	1. [x] agent
	2. [x] gestalts
	3. [x] identities
	4. [x] keys
	5. [x] nodes
	6. [x] notifications
	7. [x] vaults
4. [x] Update BIN code to use new RPC system.
5. [x] Remove old client GRPC code and tests.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
Re-based on staging after #506 merged. Staging is now the base branch for this PR.

tegefaulkes:
Only two handlers have been converted to server streams so far. 
- Nodes get all
- Notifications read

CMCDragonkai:
> Only two handlers have been converted to server streams so far.
> 
> * Nodes get all
> * Notifications read

Should go into the task list so it's clearer.

tegefaulkes:
While migrating tests I notices that the vaults clone, pull and scan handlers don't have tests.  I'll need to create some for them.

CMCDragonkai:
Those 2 are particularly tricky due to the GRPC hacks we had to do to allow things like `git pull` to work.

With the new RPC system, they have to simulate a stream for the git operations to run over.

Did you manage to simplify the whole architecture? That is a duplex stream that "switches" over to binary data, and then passes Git's HTTP protocol transparently over it?

There was also random hacks for getting exceptions out of isogit so that it would turn into GRPC exceptions, this should also be considered too.

CMCDragonkai:
BTW the above is probably more related to the agent service, and not so much client service.

BUT the performance of vault cloning/sync was so bad previously. Please look into benchmarking this so we can have fast cloning and sync, and it should be much faster now without so many layers of abstraction once QUIC and our new RPC is fully integrated.

CMCDragonkai:
All the fixmes are going to addressed in this PR?

CMCDragonkai:
Hey @tegefaulkes remember address my comments above too. If there needs to be an issue for the agent migration and git tunneling, please remember to create it too.

tegefaulkes:
I've created two new issues.
https://github.com/MatrixAI/Polykey/issues/512
https://github.com/MatrixAI/Polykey/issues/511

tegefaulkes:
As for the `FIXME`'s, there are two that are relevant this PR. one will be addressed and the 2nd one is addressed by #511 now so i'll remove it.

tegefaulkes:
The `ErrorPolykeyRemote` expects to have some metadata about the connection.  This includes nodeId, host port and method name. The problem is we don't have this information in the scope where this error is created. 

So I have two options, I make this data optional so I don't have to fill it in. Or I add some way to provide this information to the `RPCClient`. for each call. 

We should have the information available but it doesn't feel right that the `RPCClient` should have this information since it's an abstraction above any networking. If I had to add it in then the `streamPairCreateCallback` should be the thing providing it.

tegefaulkes:
Just a note, when generating test data that needs to be stringified and parsed as JSON. Any values that are `-0` become `0` and anything `undefined` will be removed. This has the effect of a round trip of stringify and parsing a JSON object will sometimes not be equal. This can result in random test failures.

CMCDragonkai:
I know that fast check generators should have something dedicated to JSON. Not all JS objects is valid JSON.

12 Mar 2023 21:07:25 Brian Botha ***@***.***>:

tegefaulkes:
I think I'll leave removing the GRPC code for the agent migration. It's much easier to remove the whole `GRPC` and `network` domain at that stage. It does mean we have some failing GRPC tests hanging about.

Otherwise this should be good to merge after some clean up.

tegefaulkes:
Cleaned up commit history

tegefaulkes:
This should be good to merge now.

tegefaulkes:
I think #510 is the next thing for me to work on. That is assuming @CMCDragonkai is still working on the Quic stuff otherwise I can start on that. I can make a start on #512 but I can't test it works without Quic working. #511 is low priority, things still work without it. We just miss out on some expected behaviour with key changes.

CMCDragonkai:
I won't be getting to QUIC anytime soon. So you can start on #510.

src/grpc/errors.ts
```diff
@@ -50,7 +50,7 @@ class ErrorGRPCServerVerification<T> extends ErrorGRPC<T> {
   exitCode = sysexits.UNAVAILABLE;
 }
 
-class ErrorPolykeyRemote<T> extends ErrorPolykey<T> {
+class ErrorPolykeyRemoteOLD<T> extends ErrorPolykey<T> {
```
CMCDragonkai, line 53: Why does this still exist? Is it for the agent RPC?

src/RPC/middleware.ts
```diff
@@ -8,6 +8,7 @@ import type {
 import { TransformStream } from 'stream/web';
 import * as rpcErrors from './errors';
 import * as rpcUtils from './utils';
+import { promise } from '../utils';
 const jsonStreamParsers = require('@streamparser/json');
 
 function binaryToJsonMessageStream<T extends JsonRpcMessage>(
```
CMCDragonkai: This function does not have a return type?

src/RPC/middleware.ts
```diff
@@ -8,6 +8,7 @@ import type {
 import { TransformStream } from 'stream/web';
 import * as rpcErrors from './errors';
 import * as rpcUtils from './utils';
+import { promise } from '../utils';
 const jsonStreamParsers = require('@streamparser/json');
```
CMCDragonkai: Do not use `require`. Use `import` or `import()`. This should be deprecated. We only use it in very specific set of cases.

src/RPC/middleware.ts
```diff
@@ -8,6 +8,7 @@ import type {
 import { TransformStream } from 'stream/web';
```
CMCDragonkai: We need to figure out how to conditionally import this later because in react-native it's in the global scope.

src/RPC/middleware.ts
CMCDragonkai, line 1: All the functions in this file require docblock documentation explaining what they do.

Furthermore, why are these functions not just part of `src/rpc/utils.ts`?

src/RPC/RPCClient.ts
CMCDragonkai, line 1: Hey can you change your directory name to be `src/rpc`. We don't use capitalised directory names in any of our projects...

src/RPC/errors.ts
```diff
@@ -51,7 +105,8 @@ export {
   ErrorRpcParse,
   ErrorRpcHandlerFailed,
   ErrorRpcMessageLength,
-  ErrorRpcRemoteError,
+  ErrorRpcMissingResponse,
+  ErrorPolykeyRemote,
   ErrorRpcNoMessageError,
   ErrorRpcPlaceholderConnectionError,
```
CMCDragonkai: Change `Rpc` to `RPC`.

CMCDragonkai: What is `ErrorRpcPlaceholderConnectionError`?

src/RPC/utils.ts
```diff
@@ -371,11 +382,14 @@ function clientInputTransformStream<I extends JSONValue>(method: string) {
   });
 }
 
-function clientOutputTransformStream<O extends JSONValue>() {
+function clientOutputTransformStream<O extends JSONValue>(
```
CMCDragonkai, line 390: I don't get the distinction between `rpc/utils.ts` and `rpc/middleware.ts`. I think they should be unified.

src/RPC/utils.ts
```diff
@@ -371,11 +382,14 @@ function clientInputTransformStream<I extends JSONValue>(method: string) {
...
+  clientMetadata: ClientMetadata,
+) {
```
CMCDragonkai: No return type?

src/bin/agent/CommandLockAll.ts
```diff
@@ -27,7 +26,7 @@ class CommandLockAll extends CommandPolykey {
         this.fs,
         this.logger.getChild(binProcessors.processClientOptions.name),
       );
-      const meta = await binProcessors.processAuthentication(
+      const auth = await binProcessors.processAuthentication(
```
CMCDragonkai, line 35: What is the type of this `auth`?

src/PolykeyClient.ts
```diff
@@ -1,11 +1,16 @@
-import type { FileSystem, Timer } from './types';
+import type { FileSystem } from './types';
```
CMCDragonkai: Did you delete the `Timer` type in `./types`? Does this mean we no longer use that `Timer`? Have you transitioned everything to using `@matrixai/timer`?

src/PolykeyClient.ts
```diff
@@ -57,20 +67,26 @@ class PolykeyClient {
...
-        logger: logger.getChild(GRPCClientClient.name),
-      }));
+    const webSocketClient = await WebSocketClient.createWebSocketClient({
+      host,
+      port,
+      expectedNodeIds: [nodeId],
+      logger: logger.getChild(WebSocketClient.name),
+      connectionTimeout,
+      pingTimeout,
+    });
```
CMCDragonkai: How can this work?

The `PolykeyClient` needs to be generic to the transports. You cannot hardcode the transport inside here.

For agent RPC, it has to use QUIC. You should factor this out and parameterise the transport.

src/PolykeyClient.ts
```diff
@@ -15,20 +20,23 @@ import config from './config';
...
     nodeId,
     host,
     port,
     nodePath = config.defaults.nodePath,
     session,
-    grpcClient,
-    timer,
+    manifest,
+    connectionTimeout,
+    pingTimeout,
```
CMCDragonkai: How can these work? These are "transport" level properties, should not be part of `PolykeyClient`.

src/client/authenticationMiddleware.ts
CMCDragonkai, line 1: How was this done before when using GRPC. Where did we put our authentication middleware?

src/client/index.ts
CMCDragonkai, line 1: Why delete the `index.ts`?

src/client/types.ts
```diff
@@ -1,14 +1,18 @@
...
+} & Omit<T, 'metadata'>;
 
-export type { Authenticate, ClientClientErrors };
+type RPCResponseResult<T extends Record<string, JSONValue> = NoData> = {
+  metadata?: {
+    [Key: string]: JSONValue;
+  } & Partial<{ Authorization: string }>;
+} & Omit<T, 'metadata'>;
+
+export type { RPCRequestParams, RPCResponseResult, NoData };
```
CMCDragonkai: Aren't these generic to `rpc`? Why not have these types in `rpc`?

src/client/utils.ts
CMCDragonkai, line 1: Are all these functions still being used?

CMCDragonkai, line 1: Also why change from `client/utils` to `client/utils.ts`? What was the reason behind this?

src/nodes/NodeManager.ts
```diff
@@ -1190,6 +1190,10 @@ class NodeManager {
         (address) => `${address.host}:${address.port}`,
       )}`,
     );
+    if (seedNodes.length === 0) {
+      logger.debug(`No seed nodes provided, skipping discovery`);
+      return;
+    }
```
CMCDragonkai: What is this about? Why is this here?

src/websockets/WebSocketServer.ts
```diff
@@ -96,7 +96,7 @@ class WebSocketServer extends EventTarget {
 
   protected server: uWebsocket.TemplatedApp;
   protected listenSocket: uWebsocket.us_listen_socket;
-  protected host: string;
+  protected host_: string;
```
CMCDragonkai: Change to `this._host`.

tests/client/service/agentLockAll.test.ts
CMCDragonkai, line 1: Why are all service tests deleted?

CMCDragonkai (changes requested):
Initial review.

tegefaulkes:
For some context, the lion share of the work was done in `clientRPC` directory. When that was done the old `client` directory was removed and `clientRPC` was renamed to `client`. This likely makes the diff a little confusing since changes done in isolation will look like modification to existing stuff.

src/grpc/errors.ts
tegefaulkes: It's the old GRPC stuff, the `agent` RPC still needs it to function. Stuff like this will be fully removed at the end of the agent migration stage.

src/RPC/middleware.ts
tegefaulkes: Added a return type.

src/RPC/middleware.ts
tegefaulkes: Should I make an new issue for this?

src/RPC/errors.ts
tegefaulkes: Updated.

tegefaulkes: That was a placeholder error to test some logic. It's a placeholder error for when the underlying streampair errors out due to a connection error or something. In that case we don't want to try and send any error messages back through the stream. I'll have to look it over, remove it and determine if any logic needs to be updated.

src/RPC/utils.ts
tegefaulkes: They are kind of the same thing. I created `middleware.ts` to hold the middleware transforms  since they are not really used internally by the RPC but used to configure the middleware when constructing the `RPC`. I can re-combine them, but to me it's neater to have a `middleware.ts` for all your middleware needs.

src/RPC/utils.ts
tegefaulkes: Fixed.

src/bin/agent/CommandLockAll.ts
tegefaulkes: It used to be a GRPC metadata object. Now it's `{ Authorization?: string }`. I can use it as the `metadata` for the message, or I can combine it with other metadata.

```ts
{
  metadata: auth,
  something: 123,
}
// OR
{
  metadata: {
  some: 'data',
   ...auth,
  }
  some: 'param',
}
```

src/PolykeyClient.ts
tegefaulkes: Looking at it, it's importing the `@matrixai/timer` now. But looking closely it doesn't seem to be using it. I need to take a closer look at this.

src/PolykeyClient.ts
tegefaulkes: That slipped my mind at the time. At the time I had a problem with providing the `RPCClient` as a parameter, I don't recall exactly why. I'll have the review and fix this because as you mention, this needs to be agnostic to the underlying transport which can be `Quic` or `WebSocket`.

An oversight on my part.

src/PolykeyClient.ts
tegefaulkes: This needs to be fixed as per the above comment.

src/client/authenticationMiddleware.ts
tegefaulkes: I don't know about the GRPC specifics of the middleware. If I had to guess it was the interceptors defined in `src/grpc/utils/FlowCountInterceptor.ts`? which is separate to the the utils.ts but part of the utils directory.

src/client/index.ts
tegefaulkes: That was the old index.ts, Technically I never created an index.ts when doing this, it's just the old one that was removed. I can create an index.ts.

src/client/authenticationMiddleware.ts
CMCDragonkai: That was a hack. That we should get rid of. I'm thinking that either middleware is in its own directory or rolled into utils. No inbetween.

src/PolykeyClient.ts
CMCDragonkai: The `Timer` type can be deleted. If nothing is using it.

src/bin/agent/CommandLockAll.ts
CMCDragonkai: I think you should use `token` or something similar to how http does for their headers. At the very least should all be lowercase.

src/RPC/utils.ts
CMCDragonkai: Like I said it should be either a directory with an `index.ts` or a utils.

src/client/types.ts
tegefaulkes: Yes and no? By design the RPC doesn't care about the parameter structure. This is here to enforce the metadata structure for the `client` domain to be used by the middleware, handler and client implementation.

So it's not used at all in the context of the `rpc` domain, but will be used by both the `client` and `agent` domains, they'll share the middleware code as well. So with that maybe this should be moved to a common place between the two? Not sure.

src/client/utils.ts
tegefaulkes: Yes, these were transplanted from the old authentication logic for client and updated to the new system.

tegefaulkes: No specific reason. When I was working on this in `clientRPC` before removing the old `client` and renaming `clientRPC` to `client` I just didn't have the `utils.ts` in a `utils` directory. It doesn't seem necessary for a single `utils.ts` file.

I could move `utills.ts` back to a `utils` directory and move `authenticationMiddleware.ts` there as well.

src/nodes/NodeManager.ts
tegefaulkes: Technically out of scope of the PR. I added it to remove logger spam during tests. It is technically a fix that is needed.

src/websockets/WebSocketServer.ts
tegefaulkes: Updated.

tests/client/service/agentLockAll.test.ts
tegefaulkes: Those are the old tests, the new ones are under `tests/client/handlers/...`

src/RPC/middleware.ts
tegefaulkes: Import is acting weird with this library. when I use the import semantics or `import()` I get the following error.  `TS2307: Cannot find module '@streamparser/json' or its corresponding type declarations.`

I think it's using module exports. We have problems importing that right?

CMCDragonkai: You can't use `import()` without top level await or being in Async context.

CMCDragonkai: But you should ask upstream and check what you can do

src/RPC/middleware.ts
tegefaulkes: I've added docblocks to these functions now. 

As for why they're not part of `utils.ts` I think I addressed it in one of the other review comments. The're meant to be used when setting up the `RPCServer` or `RPCClient` and not really internally by them. So to keep them from getting lost in the `utils.ts` with the other functions I made them separate. I likely should move both to a `utils` directory though.

I'm resolving this since this discussion can be done in one of the other comments about it.

src/RPC/RPCClient.ts
tegefaulkes: I'll do this, but after all the other changes.

src/RPC/middleware.ts
tegefaulkes: It's not an upstream issue. If I asked they'd likely just point to the first line of their example `import { JSONParser } from '@streamparser/json';`. It's more likely a problem with our configuration, I'll look into it some more.

tegefaulkes:
Ok, since the `PolykeyClient` is agnostic to the transport method now. It needs to support either the `Quic` or `WebSocket`. Given this the `quicConnection` or `WebSocketClient` needs to be created along side the `PolykeyClient` and the stream creation callback and be passed in just like the `RPCClient`.  I guess that simplifies the `PolykeyClient` slightly but forces us to manage the transport client we're using separately. Not the end of the world.

tegefaulkes:
Ok, instead of the `WebSocketClient` being created inside of ` PolykeyClient` and managed by it. It is now created just before and `PolykeyClient` takes the stream creation callback.

```ts
        webSocketClient = await WebSocketClient.createWebSocketClient({
          expectedNodeIds: [clientOptions.nodeId],
          host: clientOptions.clientHost,
          port: clientOptions.clientPort,
        })
          streamPairCreateCallback: () => webSocketClient.startConnection(),
          nodePath: options.nodePath,
          manifest: clientManifest,
          logger: this.logger.getChild(PolykeyClient.name),
        });
```

Every bin command needs to be updated with this change.

src/RPC/middleware.ts
CMCDragonkai: No, this is already in TypeScript-Demo-Lib.

src/RPC/middleware.ts
CMCDragonkai: Why can't we import like that?

src/RPC/middleware.ts
CMCDragonkai: It may be better to have a `middleware` directory and then separate each middleware by files. That way you can import separately. One may want to separately compose middlewares later.

src/RPC/utils.ts
CMCDragonkai: Suggest using `middleware/*` where `*` is name of middleware - use camel cased names.

src/client/types.ts
CMCDragonkai: If they are specific to client/agent "services", one should make that clearer by saying `ClientRPCRequestParams` or `AgentRPCRequestParams`. Right now these look like `rpc` domain types.

src/nodes/NodeManager.ts
CMCDragonkai: Which issue is this for? I thought we had discussed earlier and it already been fixed.

src/websockets/WebSocketServer.ts
CMCDragonkai: Check if anything else is still `x_` and not `_x`.

src/nodes/NodeManager.ts
CMCDragonkai: If this is in fact a proper fix, please create an issue and fix it with this PR, and give demonstration of what exactly this fixes. Like if it is to reduce logger spam. Then post an example of what is expected or not expected during execution by copying the terminal output.

CMCDragonkai:
I think `streamPairCreateCallback` should just be called the `streamFactory`. That's the usual way of referring to factory functions.

src/RPC/middleware.ts
tegefaulkes: Because when I do I get `TS2307: Cannot find module '@streamparser/json' or its corresponding type declarations.`.

Only using `require()` works for reasons I have not worked out yet.

src/client/types.ts
tegefaulkes: They're getting pretty verbose already. Maybe change them to `ClientRPCReq`, `ClientRPCRes`, `AgentRPCReq` and `AgentRPCRes`?

src/nodes/NodeManager.ts
tegefaulkes: Ok, I'll do that.

src/client/types.ts
CMCDragonkai: I don't like shortening type names. Type names can verbose, it's fine.

src/PolykeyClient.ts
tegefaulkes: It has been removed.

src/PolykeyClient.ts
tegefaulkes: This is now fixed, Instead of the transport being internal to the `PolykeyClient`it now takes the `StreamFactory` callback. transport is now created and managed separately.

src/client/index.ts
tegefaulkes: I'll add appropriate `index.ts` files.

src/RPC/errors.ts
tegefaulkes: This should be removed, but I also need to update how we decide to send or not send errors over the RPC. I think the only case so far where we don't send a error message is when we can't due to the connection ending. 

So what I can do here is attempt to send all messages, if sending fails due to a connection failure or whatever then we dispatch the error event contain a compound error of what we tried to send and why it failed.

CMCDragonkai: The usual way to do this is:

1. Handling errors goes to the connection.
2. Connection errors cannot go to the connection, instead it is logged in the actual agent itself, it should not crash the agent
3. On the receiving side, if it gets a handling error, raise it the normal way.
4. If however on the receiving side, the connection is broken (due to keepalive or something else), then raise a special set of errors regarding the reason why the connection is broken

CMCDragonkai: It's important that handling errors AND connection errors does not shutdown the agent. The agent and associated servers should continue to operate.

src/RPC/middleware.ts
CMCDragonkai: We don't have this problem with any of the other libraries. Although I think if they are using the new node module system instead of commonjs, then it can be an issue. But we should be working out a way to be able to import both module systems and not be using `require`.

CMCDragonkai: I think you should indicate this upstream. That way you can point out something about our configuration causing import problems, but that doesn't affect other packages we have been using.

CMCDragonkai: Upstream was quite responsive to our previous problems too involving DoS.

src/PolykeyClient.ts
tegefaulkes: This has been fixed.

CMCDragonkai:
When debugging the RPC, are we still able to use Wireshark to see the JSON messages like we did when debugging grpc?

src/RPC/errors.ts
tegefaulkes: This has been resolved.

src/RPC/middleware.ts
tegefaulkes: I'll make  a new issue and an upstream issue for this. It isn't blocking the PR.

src/bin/agent/CommandLockAll.ts
tegefaulkes: I'll make it lowercase `authorization?: string`.  It can be a token or the raw password so `authorization` is more appropriate than `token`.

tegefaulkes: Fixed.

src/client/types.ts
tegefaulkes: Fixed.

src/RPC/utils.ts
tegefaulkes: Fixed.

src/client/utils.ts
tegefaulkes: fixed.

src/client/index.ts
tegefaulkes: Fixed.

tegefaulkes:
With TLS enabled we won't be able to see the plain text, but otherwise the messages should be readable text in wireshark.

tegefaulkes:
Cleaned up commit history

src/RPC/RPCClient.ts
tegefaulkes: This has been done.

src/RPC/middleware.ts
tegefaulkes: I have created a new issue for this at #516 .

tegefaulkes:
This should be good to go. There are some failing tests but they are part of the domains that will be removed in the agent migration.

CMCDragonkai:
> With TLS enabled we won't be able to see the plain text, but otherwise the messages should be readable text in wireshark.

Wireshark can MITM. I meant if the debugging procedure can be repeated.

src/rpc/errors.ts
```diff
@@ -0,0 +1,107 @@
...
+export {
+  ErrorRPC,
+  ErrorRPCDestroyed,
+  ErrorRPCStopping,
+  ErrorRPCParse,
+  ErrorRPCHandlerFailed,
+  ErrorRPCMessageLength,
+  ErrorRPCMissingResponse,
+  ErrorPolykeyRemote,
+  ErrorSendErrorFailed,
```
CMCDragonkai: This should be `ErrorRPCSendErrorFailed`.

src/rpc/errors.ts
```diff
@@ -0,0 +1,107 @@
...
+  static description = 'RPC Message exceeds maximum size';
+  exitCode = sysexits.DATAERR;
+}
+
+class ErrorRPCMissingResponse<T> extends ErrorRPC<T> {
+  static description = 'Stream ended before response';
+  exitCode = sysexits.UNAVAILABLE;
+}
+
+class ErrorPolykeyRemote<T> extends ErrorPolykey<T> {
```
CMCDragonkai, line 46: I'm moving this to the bottom. Just like in `grpc/errors`.

src/rpc/RPCServer.ts
```diff
@@ -217,22 +217,26 @@ class RPCServer {
...
             } else {
               // These errors are emitted to the event system
+              // This contains the original error from enqueuing
               events.dispatchEvent(
                 new rpcUtils.RPCErrorEvent({
                   detail: {
-                    error: e,
+                    error: new rpcErrors.ErrorRPCSendErrorFailed(undefined, {
+                      cause: [e, reason],
+                    }),
```
CMCDragonkai: Is this expected to bubble up on the agent side, or does this get sent to the remote side?

src/rpc/RPCServer.ts
```diff
@@ -202,7 +202,7 @@ class RPCServer {
             }
             controller.enqueue(value);
           } catch (e) {
-            if (rpcUtils.isReturnableError(e)) {
+            if (reason == null) {
```
CMCDragonkai: What was the reason behind this? That is changing from `isRetunrable` to `null`?

CMCDragonkai:
What were the tests that you disabled for this branch?

src/PolykeyAgent.ts
```diff
@@ -103,7 +107,6 @@ class PolykeyAgent {
     vaultManager,
     notificationsManager,
     sessionManager,
-    grpcServerClient,
```
CMCDragonkai, line 106: Don't just remove this and then establish it in the agent. Preserve the pattern that we are using. Dependency inject the `rpcServerClient`. Or whatever is relevant here.

Plus you didn't update the types to remove `grpcServerClient`, that's still there.

The original thing here was meant to be the RPC server for the client service. That's why there's also the `grpcServerAgent`.

I'm guessing now when we migration the agent service to to this, would we also have a `rpcServerAgent`?

Furthermore you would have 2 RPC servers, one for the client service and one for the agent service.

And I also think the `webSocketServer` should also be DIed like the rest of it.

src/PolykeyAgent.ts
```diff
@@ -401,11 +405,15 @@ class PolykeyAgent {
...
-        });
+      // FIXME: propagate default values
+      webSocketServer = new WebSocketServer(
+        logger.getChild('WebSocketServer'),
+        fs,
+        1_000_000_000, // About 1 GB
+        undefined,
+        1_000,
+        10_000,
+      );
```
CMCDragonkai: This should be DIed like `webSocketServer = webSocketServer ?? new WebSocketServer(...)`.

Also what do you mean by propagating the default values?

src/PolykeyAgent.ts
CMCDragonkai: Note that:

```
    grpcServerClient?: GRPCServer;
```

Still exists in the types here.

src/PolykeyClient.ts
```diff
@@ -57,20 +54,17 @@ class PolykeyClient {
...
-        logger: logger.getChild(GRPCClientClient.name),
-      }));
+    const rpcClient = await RPCClient.createRPCClient<M>({
+      manifest,
+      streamFactory,
+      middleware: middlewareUtils.defaultClientMiddlewareWrapper(
+        authMiddleware.authenticationMiddlewareClient(session),
+      ),
+      logger: logger.getChild(RPCClient.name),
+    });
```
CMCDragonkai: Change this to be DIed optionally like we did with `grpcClient`. The DI enables mock testing.

src/PolykeyClient.ts
```diff
@@ -114,13 +108,13 @@ class PolykeyClient {
...
   public async stop() {
     this.logger.info(`Stopping ${this.constructor.name}`);
-    await this.grpcClient.destroy();
     await this.session.stop();
     this.logger.info(`Stopped ${this.constructor.name}`);
   }
 
   public async destroy() {
     this.logger.info(`Destroying ${this.constructor.name}`);
+    await this.rpcClient.destroy();
```
CMCDragonkai, line 120: So we moved from stopping to destroying for the `RPCClient`?

src/bin/agent/CommandLockAll.ts
```diff
@@ -16,9 +17,11 @@ class CommandLockAll extends CommandPolykey {
     this.addOption(binOptions.clientPort);
     this.action(async (options) => {
       const { default: PolykeyClient } = await import('../../PolykeyClient');
+      const { default: WebSocketClient } = await import(
+        '../../websockets/WebSocketClient'
+      );
       const { default: Session } = await import('../../sessions/Session');
-      const utilsPB = await import('../../proto/js/polykey/v1/utils/utils_pb');
-
+      const { clientManifest } = await import('../../client/handlers');
```
CMCDragonkai: This imports EVERYTHING! The `client/handlers/index.ts` is importing the entire project.

Don't do this. Remember the whole reason we are doing this is `await import` is to specialise to the only the code necessary for the command.

You're making every command dynamically load the entire project source with `await import('../../client/handlers');`

You need to have a more specific import.

The original thing was just `utilsPB` which was just a simple structure for utilities.

Now this ends up slowing the command execution.

I see the way you're creating the `clientManifest`. I think it's too heavy, either we need a different way, or can we do without having a manifest, or can we only import **exactly** what we need without going through the `index.ts` imports.

src/bin/agent/CommandLockAll.ts
```diff
@@ -39,27 +42,37 @@ class CommandLockAll extends CommandPolykey {
...
+        webSocketClient = await WebSocketClient.createWebSocketClient({
+          expectedNodeIds: [clientOptions.nodeId],
           host: clientOptions.clientHost,
           port: clientOptions.clientPort,
+          logger: this.logger.getChild(WebSocketClient.name),
+        });
+        pkClient = await PolykeyClient.createPolykeyClient({
+          streamFactory: () => webSocketClient.startConnection(),
           nodePath: options.nodePath,
+          manifest: clientManifest,
```
CMCDragonkai, line 63: There should be a way of acquiring this information without importing the entire application. I mean also in a transitive sense.

src/bin/polykey-agent.ts
```diff
@@ -114,8 +114,8 @@ async function main(_argv = process.argv): Promise<number> {
     recoveryCode: pkAgent.keyRing.recoveryCode,
     pid: process.pid,
     nodeId: nodesUtils.encodeNodeId(pkAgent.keyRing.getNodeId()),
-    clientHost: pkAgent.grpcServerClient.getHost(),
-    clientPort: pkAgent.grpcServerClient.getPort(),
+    clientHost: pkAgent.webSocketServer.host,
+    clientPort: pkAgent.webSocketServer.port,
```
CMCDragonkai: I'm trying to find where `getHost` and `getPort` methods used to exist. I don't see them on `GRPCClient`. There would be a reason why these getters exist, usually when the property must be mutable, but only internally so `readonly` modifier cannot be used.

src/bin/utils/processors.ts
```diff
@@ -11,10 +11,9 @@ import type {
 import type { SessionToken } from '../../sessions/types';
 import path from 'path';
 import prompts from 'prompts';
-import * as grpc from '@grpc/grpc-js';
 import Logger from '@matrixai/logger';
 import Status from '../../status/Status';
-import * as clientUtils from '../../client/utils';
+import * as clientUtils from '../../client/utils/utils';
```
CMCDragonkai, line 16: This isn't actually what I meant. In many other situations what we do when we have this kind of code is to do this:

```
utils/index.ts
utils/x.ts
utils/y.ts
```

You don't want to do `utils/utils.ts`. That defeats the purpose of putting it into a directory. It's to have file-level separation which can be more efficient for importing or navigation and code review.

CMCDragonkai: Indeed there is another one with:

```
utils.ts
utils/x.ts
```

I don't like this as it is confusing, many other languages have confusing overlaps here, it's better to use `utils/index.ts`.

You can have broad root-level utilities in `utils/index.ts`, and you can specialise certain utilities into other files.

CMCDragonkai: Ok there is an example with `src/bin/utils/utils.ts`.

The idea being that `src/bin/utils/index.ts` does a flat export of `src/bin/utils/utils.ts`.

But the idea of specialisation is still important to have separate files.

src/bin/utils/utils.ts
```diff
@@ -112,7 +112,7 @@ function outputFormatter(msg: OutputObject): string | Uint8Array {
     let currError = msg.data;
     let indent = '  ';
     while (currError != null) {
-      if (currError instanceof grpcErrors.ErrorPolykeyRemote) {
+      if (currError instanceof grpcErrors.ErrorPolykeyRemoteOLD) {
```
CMCDragonkai: Don't we need to have something like this but for the new JSON based RPC?

src/client/errors.ts
```diff
@@ -1,30 +1,24 @@
...
   exitCode = sysexits.NOPERM;
 }
 
 export {
-  ErrorClient,
-  ErrorClientClientDestroyed,
+  ErrorRPCClient,
   ErrorClientAuthMissing,
   ErrorClientAuthFormat,
   ErrorClientAuthDenied,
```
CMCDragonkai, line 24: All these should the be `ErrorRPCClientX`. Always keep to consistent naming.

On the other hand, originally I just called it `ErrorClient` to avoid the `RPC`. You can also change `ErrorRPCClient` to just `ErrorClient`.

CMCDragonkai: This meaning it is an error inside the client service.

CMCDragonkai:
I need to go in deeper into the exact RPC mechanisms, I've been mostly reviewing code quality and standards and integration.

src/rpc/RPCServer.ts
tegefaulkes: This is purely the server side. It wraps the errors that happen when the stream throws an error. So it's only emitted as an error event for the server.

src/rpc/RPCServer.ts
tegefaulkes: After looking over the placeholder error I determined that it wasn't needed and the current method didn't really work. Currently the only time we want to not send an error through the stream is when we can't. So now we're only checking if the stream has already failed. if so we emit the error through the event system otherwise we just send the error message.

src/rpc/RPCServer.ts
CMCDragonkai: Hmm want to rename this a bit.

src/rpc/RPCServer.ts
CMCDragonkai: How come you don't check `e`? You're checking `reason` here.

src/bin/agent/CommandLockAll.ts
tegefaulkes: For the bin commands we can just define the types for the RPC methods we're using in the command. The problem is, it's essential that they match otherwise very annoying to debug bugs happen. To mitigate this the caller and handler are defined side by side in the same location. But as you said this leaves us depending on a bunch of stuff we don't need. 

We can just define the caller in the command and use a truncated manifest, Or separate the client manifest definition to a new file so it only depends on the types it needs.

src/bin/polykey-agent.ts
tegefaulkes: They weren't getters before, just normal methods. I made them getters since the host and port is updated every time you start. A `readonly` can only be set during construction so I couldn't update it every time we started. 

So we have a getter here. The only alternative is to change it to match the`getHost()` style.

src/bin/utils/processors.ts
tegefaulkes: Ok, so you're saying that the methods in `utils/utils.ts` should be moved into the `utils/index.ts`?

src/bin/utils/utils.ts
tegefaulkes: We do have it. it's called `ErrorPolykeyRemote`. I had to rename the old one to avoid conflict until it's fully removed.

src/client/errors.ts
tegefaulkes: I don't think so? They are different domains. These are the Client errors for the `client` domain., the other ones are `rpc` domain errors for the RPCClient. They are completely different.

src/rpc/RPCServer.ts
tegefaulkes: Reason is set by the streams `abort` handler. It lets us know that we can't enqueue the error message if we wanted to. It `reason` could use a more descriptive name in that case.

src/bin/polykey-agent.ts
CMCDragonkai: I think of `getHost` as getters. The `get host` is not a getter. It's a magic method overloading the `.host`.

src/bin/utils/processors.ts
CMCDragonkai: Not necessarily, it's about splitting up the methods into files. I'm looking into the `rpc` now to see what factorisation can occur.

src/bin/utils/utils.ts
CMCDragonkai: I know I meant the line itself. You're checking for the remote old, why not check for the `ErrorPolykeyRemote` too?

src/client/errors.ts
CMCDragonkai: Yea so then `ErrorRPCClient` should be `ErrorClient`?

CMCDragonkai: You're extending `ErrorRPCClient` atm, and the subtypes are just `ErrorClientX`. So instead it should be `ErrorClientX` extending `ErrorClient`.

src/rpc/RPCServer.ts
CMCDragonkai: What would `e` be if `reason` is `null`?

src/PolykeyAgent.ts
tegefaulkes: I missed the `grpcServerClient`, I'll remove it.

As for DI, I can make them both DI-ed.

tegefaulkes: I don't think the rpcServer can be DI-ed. It has a mutual reference to the `PKAgent`.  This means that it needs to be created within the `PolykeyAgent`'s constructor. '

It's cheesy but we can pass the `PKAgent` into it as a promise to be resolved later. that might be the better option here. I think the only problem is, if you don't resolve the promise, the agent stop handler will hang until timeout.

src/PolykeyAgent.ts
tegefaulkes: There are config values that are just hard-coded here. They need to be added to the `PolykeyAgent` config and defaults.

tegefaulkes: This has been changed.

src/PolykeyAgent.ts
tegefaulkes: I went with the promise method. Seems like the easiest solution to the problem.

src/rpc/RPCServer.ts
tegefaulkes: `e` would be an error thrown in the handler if it failed to do something.

src/PolykeyAgent.ts
tegefaulkes: This has been fixed, `rpcServerClient` and `webSocketServerClient` can now be dependency injected.

src/PolykeyClient.ts
tegefaulkes: Yeah, the `RPCClient` doesn't really have any state anymore. It's just a container for some data and an interface for making calls. It doesn't really need to be destroyed even. it only prevents it from being used from then on.

src/PolykeyClient.ts
tegefaulkes: I'm thinking about how to set the manifest or DI the `RPCClient`. Technically the options are mutually exclusive. Ideally this would be enforced with the type.

I can think of two options. 
1. the `manifest` and `rpcClient` are both optional but creation will fail if if neither are provided. Providing both would default to the DI-ed `rpcClient`.
2. The other option is to have the `manifest` be the `RPCClient|ClientManifest`, if the manifest is provided then we create the `rpcClient`. 

I'm not sure which one is preferred here.

tegefaulkes: I'm going with option 2, it's much cleaner.

tegefaulkes: This is fixed.

src/bin/agent/CommandLockAll.ts
tegefaulkes: This is the same problem as the above comment. I'll do the discussion there.

src/bin/agent/CommandLockAll.ts
tegefaulkes: The core of this is we need a way to define the manifest without importing everything. The reason everything is being imported because the client caller definitions are defined in the same file of the handler.  So we're implicitly importing the handler dependencies.

The only thing the callers need are the types involved. but we need to keep them in a logical place close to the handlers because they need to be created and updated side by side. The reason for this is that the client and agent definitions are closely coupled by type, not enforced by any type system and if they don't match exactly we end up with type bugs that are hard to debug without intimate knowledge of the system. 

As for a fix for this. We need to define the server and client manifest in seperate files and we can't defined the callers in the handlers file. I need to make a `clientManifest.ts` file and possibly a server variant for this. I'm not sure if these files should be kept in the `handlers` directory or the `client` directory. 

I'll keep them in the handlers directory for now.

src/rpc/RPCServer.ts
```diff
@@ -27,9 +26,9 @@ import {
   ServerHandler,
   UnaryHandler,
 } from './handlers';
-import * as rpcUtils from './utils';
+import * as rpcUtils from './utils/utils';
 import * as rpcErrors from './errors';
-import * as middlewareUtils from './middleware';
+import * as middlewareUtils from './utils/middleware';
 import { never } from '../utils/utils';
 import { sysexits } from '../errors';
```
CMCDragonkai, line 36: We need some docblock for `sensitive`. Well docblock in `createRPCServer` and explain what does `sensitive` mean.

src/rpc/types.ts
```diff
@@ -228,6 +229,13 @@ type MapCallers<T extends ClientManifest> = {
...
+type ClientMetadata = {
+  nodeId: NodeId;
+  host: string;
+  port: number;
+  command: string;
+} & POJO;
+
 export type {
   JsonRpcRequestMessage,
   JsonRpcRequestNotification,
```
CMCDragonkai: I'm changing all of this to be `JSONRPC...`.

src/rpc/RPCServer.ts
```diff
@@ -44,10 +43,10 @@ class RPCServer {
   }: {
     manifest: ServerManifest;
     middleware?: MiddlewareFactory<
```
CMCDragonkai: Why is `middleware` a `MiddlewareFactory`? I think it should be called `middlewareFactory` since you later call it.

Also please add a docblock to the `middleware` property explaining its types and how it is intended to work.

src/rpc/RPCServer.ts
CMCDragonkai, line 1: Are all the `manifestItem.handle` methods or can they be arrow functions? If they are arrow functions, there's no need to `bind` them.

tegefaulkes: They need to access this for the context of the handler. So I don't think they can be arrow functions here.

CMCDragonkai: The `this` context can exist in arrow function properties.

```ts
class C {
  f = () => {
    console.log(this);
  }
}
```

CMCDragonkai: But I can see you made each `handle` an abstract method. I wonder if it could be done as an abstract property.

src/rpc/types.ts
```diff
@@ -12,11 +12,12 @@ import type {
...
 import type { NodeId } from '../nodes/types';
 import type { Certificate } from '../keys/types';
+import type { POJO } from '../types';
 
 /**
  * This is the JSON RPC request object. this is the generic message type used for the RPC.
  */
-type JsonRpcRequestMessage<T extends JSONValue = JSONValue> = {
+type JSONRPCRequestMessage<T extends JSONValue = JSONValue> = {
   // A String specifying the version of the JSON-RPC protocol. MUST be exactly "2.0"
```
CMCDragonkai: I think if you want these comments to show up in the typedoc generated comments, all of these must be docblocks, not `//` comments.

We only use line comments for inside functions/procedures.

src/rpc/types.ts
```diff
@@ -84,7 +85,7 @@ type JsonRpcResponseError = {
 //  -32603	Internal error	Internal JSON-RPC error.
 //  -32000 to -32099
```
CMCDragonkai: I don't think we should be using these codes.

Instead you should just have unique exception classes for each possible error.

The reason is because in an OOP language like ourselves, the class name itself is sufficient identification, code is unnecessary.

We do use `code` for sysexit codes, but that's only because a number is necessary when exiting a program.

When sending the error over the RPC, we don't need it, we only need the serialised version of the exception object which is then later deserialised on the remote side.

CMCDragonkai: So you should just literally have:

```
class ErrorRPCParseError { }
class ErrorRPCInvalidRequest { }
class ErrorRPCMethodNotFound { }
class ErrorRPCInvalidParams { }
class ErrorRPCInternalError { }
```

The name itself is unique and more descriptive than an arbitrary number.

CMCDragonkai: If the JSON RPC spec requires a `code: number`, we can fill it with our sysexit code. But the really we should be replacing that with our "serialised" error structure. The one we use with `fromJSON` and `toJSON`.

src/rpc/RPCServer.ts
CMCDragonkai, line 1: When you do:

```ts
  protected registerUnaryHandler<I extends JSONValue, O extends JSONValue>(
    method: string,
    handler: UnaryHandlerImplementation<I, O>,
  ) {
    const wrapperDuplex: DuplexHandlerImplementation<I, O> = async function* (
      input,
      connectionInfo,
      ctx,
    ) {
      for await (const inputVal of input) {
        yield handler(inputVal, connectionInfo, ctx);
        break;
      }
    };
    this.registerDuplexStreamHandler(method, wrapperDuplex);
  }
```

The `yield handler(...)`, should there be an await there? Like `yield await handler(...)`?

src/rpc/types.ts
tegefaulkes: They're mostly for reference and explaining what each field is for. It doesn't need to show up in the IDE docs.

src/rpc/RPCServer.ts
CMCDragonkai: The whole point of having an async generator function is so that you can use `await` inside. But the `handler` here returns a promise. So why not await the promise and yield back the result?

src/rpc/RPCServer.ts
```diff
@@ -357,19 +360,19 @@ class RPCServer {
...
-    callback: (event: RPCErrorEvent) => void,
+    callback: (event: rpcUtils.RPCErrorEvent) => void,
     options?: boolean | AddEventListenerOptions | undefined,
   ) {
     this.events.addEventListener(type, callback, options);
   }
 
-  @ready(new rpcErrors.ErrorRpcDestroyed())
+  @ready(new rpcErrors.ErrorRPCDestroyed())
   public removeEventListener(
```
CMCDragonkai: Question... why didn't you just extend the `EventTarget`?

src/rpc/types.ts
tegefaulkes: We're not using these codes anywhere. They were for reference. They can be removed.

src/rpc/RPCServer.ts
tegefaulkes: It's not strictly needed, the only difference is how the call stack looks if any error was thrown. The handler will be awaited when awaiting the value from the `AsyncGenerator`.  There is no specific reason it was done this way, I can add an await in.

src/rpc/RPCServer.ts
tegefaulkes: I can, but i'd lose the type information for the handler. dealing with the raw event target with custom events with added data can be confusing and annoying.  This also allows for the `@ready` decorator which I'm not sure is strictly needed.

src/bin/agent/CommandLockAll.ts
tegefaulkes: The `clientManifest` and `serverManifest` are now in their own respective files, `clientManifest.ts` and `serverManifest.ts`. They are both still exported by the index. 

We are now importing just the `clientManifest` for use with the `rpcClient`. This should only depend on type imports now. 

To get any more specific than this then we'd need separate files for each caller definition. That would be too much to maintain for little benifit.

src/bin/polykey-agent.ts
tegefaulkes: Sure, but the technical term for the magic methods overriding `.host` are getters and setters aren't they?

Is there a change to be made here?

src/bin/utils/processors.ts
tegefaulkes: I'm confused then. I created the utils directory in the client and `rpc` domain to store the `utils.ts` and `middleware.ts` files together. I can maybe split the `utils.ts` files up more? haven't I done what you are saying here?

src/bin/utils/utils.ts
tegefaulkes: Ah I see, I misunderstood. It should be checking for both until the old one is removed. I'll fix this up.

src/client/errors.ts
tegefaulkes: Ah, I didn't see that in the context of the comment. I'll fix it.

tegefaulkes:
There is something I've just noticed. We generally don't docblock functions with deconstructed objects as the parameters. I just looked into it and the usual way doesn't quite work. The proper way of documenting these parameters according to the docs is...

```
  /**
   * @param obj - name of the object parameter.
   * @param obj.a - a parameter.
   * @param obj.b - b parameter.
   * @protected
   */
  protected someFunction({
    a,
    b,
  }: {
    a: number,
    b: string
  }) {}
```

This gives us a much cleaner documentation in the IDE

```
  Params: { a, b, }: {a: number, b: string} – name of the object parameter.
    Config options:
        a – a parameter.
        b – b parameter.
```


the normal way  results in

```
  /**
   * @param a - a parameter.
   * @param b - b parameter.
   * @protected
   */
  protected otherFunction({
    a,
    b,
  }: {
    a: number,
    b: string
  }) {}
```

```
Params: { a, b, }: {a: number, b: string} – a parameter. @param b - b parameter.
```


Just something to keep in mind.

tegefaulkes:
I don't recall disabling any tests on this branch.

src/rpc/RPCServer.ts
tegefaulkes: I've added docblocks to both `RPCClient` and `RPCServer` filling in some details.

src/rpc/RPCServer.ts
tegefaulkes: Done.

src/bin/polykey-agent.ts
CMCDragonkai: The reason to use `.getHost()`...

```ts
class X {
  // this means x is readable and immutable
  readonly x;
  // this means y is readable and mutable
  y;
  _z
  // this means z is readable, but internally mutable, but externally immutable
  get z () {
    return this._z;
  }
  // why use getA over the above?
  // supposed to mean that the A property is initialised after construction, or has specialised construction
  getA() {

  }
}
```

The `host` and `port` are not directly set before construction, they occur after an asynchronous start. That's why I always make them `getHost` and `getPort` to distinguish them from static properties. We've also talked about this distinction before. Then `@ready` can be applied too here, where one can say that `host` and `port` are only available after the server is started. Prior to this, you may do something like `await Server.createServer({ host: '0.0.0.0', port: 0 })`. But then afterwards it is "resolved", with `server.getPort() // returns 1235`.

The point is `host` and `port` are dynamic properties, and I reserve `.a` or `.b` properties to be static properties.

src/rpc/RPCServer.ts
CMCDragonkai: > Reason is set by the streams `abort` handler. It lets us know that we can't enqueue the error message if we wanted to. It `reason` could use a more descriptive name in that case.

You mean the `cancel` handler?

src/rpc/types.ts
CMCDragonkai: Not the IDE docs, the typedocs. If it's switched to docblocks, you should be able to see it too.

src/rpc/RPCServer.ts
CMCDragonkai: Yea I think ti should be changed to `await`. That seems like the most common way of doing it.

CMCDragonkai: Plus it is better for debugging purposes here.

I can understand using such things for function wrappers. But in these situations, it's better safe than sorry.

src/rpc/RPCServer.ts
CMCDragonkai: I'm extending it in the QUIC code base. I'm sure it's possible to do this without it. Also possible to have an overloaded signature in the subclass I think. I'll try.

CMCDragonkai:
> There is something I've just noticed. We generally don't docblock functions with deconstructed objects as the parameters. I just looked into it and the usual way doesn't quite work. The proper way of documenting these parameters according to the docs is...
> 
> ```
>   /**
>    * @param obj - name of the object parameter.
>    * @param obj.a - a parameter.
>    * @param obj.b - b parameter.
>    * @protected
>    */
>   protected someFunction({
>     a,
>     b,
>   }: {
>     a: number,
>     b: string
>   }) {}
> ```
> 
> This gives us a much cleaner documentation in the IDE
> 
> ```
>   Params: { a, b, }: {a: number, b: string} – name of the object parameter.
>     Config options:
>         a – a parameter.
>         b – b parameter.
> ```
> 
> the normal way results in
> 
> ```
>   /**
>    * @param a - a parameter.
>    * @param b - b parameter.
>    * @protected
>    */
>   protected otherFunction({
>     a,
>     b,
>   }: {
>     a: number,
>     b: string
>   }) {}
> ```
> 
> ```
> Params: { a, b, }: {a: number, b: string} – a parameter. @param b - b parameter.
> ```
> 
> Just something to keep in mind.

Are you updating the docs regarding this?

src/rpc/RPCServer.ts
CMCDragonkai: I've changed to extension here and also simplified the event class itself. Alignment with js-quic is more important here. However subclasses cannot provide a more specialised signature unfortunately. Maybe there's a hack, but not important.

CMCDragonkai:
Did you test if the TLS system is working for the websocket server?

And currently what happens if the root certificate is rotated? Is there an automatic restart of the client service RPC server?

CMCDragonkai:
I just noticed that when running `npm run docs`, it doesn't generate any docs for the rpc module, or at least `RPCServer` and `RPCClient`. Any ideas why?

src/rpc/RPCServer.ts
```diff
@@ -253,7 +282,7 @@ class RPCServer {
       ctx,
     ) {
       for await (const inputVal of input) {
-        yield handler(inputVal, connectionInfo, ctx);
+        yield await handler(inputVal, connectionInfo, ctx);
         break;
       }
```
CMCDragonkai, line 309: Given that this is a unary handler, why are we doing a `for await` on the `input`?

Shouldn't we expect just one piece of data from the `input`?

That is instead of doing `for await`, it would be clearer to do:

```
const inputVal = input.next();
```

---

Oh wait, because it is an `AsyncIterable`, you have to use `for await`.

I am adding a comment here.

CMCDragonkai: It is possible to do this:

```
const inputVal = (await input[Symbol.asyncIterator]().next()).value;
```

However it's a bit wordy.

src/rpc/RPCServer.ts
tegefaulkes: Yes.

src/bin/polykey-agent.ts
tegefaulkes: Ok, This has been fixed.

src/rpc/types.ts
tegefaulkes: I've made them doc comments and some to the message types.

tegefaulkes:
Addressing some questions.

1.  for If I'm updating docs in the style I mentioned. I've applied it to the RPC domain for the create methods. There are not many places where that style is needed and the other domains are out of scope for this PR.
3. There are tests for the websocket that check if the cert authentication is working. We don't explicitly test if TLS is not set, it's on by default and I know if it was off the client would be rejected when connecting with WSS. 
4. As for certificate rotation, I think that has an issue already? #511 
5. As for `rpc` domain not generating docs. I have no idea. I think the only docs are the createX ones in the style I mentioned above.

CMCDragonkai:
In the `WebSocketServer.ts` there's still a:

```
          this.signalReadableEnd(Error('TMP READABLE CANCELLED'));
```

This shouldn't be there.

CMCDragonkai:
The usage of PK agent promise in the handlers should reference the issue https://github.com/MatrixAI/Polykey/issues/442 discussing lazy dependencies.

CMCDragonkai:
We've been looking at the `RPCServer.registerDuplexStreamHandler`.

There's some confusion here about how exceptions/cancellation is being used on the streams.

If you use:

```
const gF = async function *() {
  yield 1;
  yield 2;
  yield 3;
};

const g = gF();
```

And you use `await g.throw(new Error());` this line will throw an exception at this point if the generator above does not catch "thrown-into" exceptions at `yield 1;`.

```
const gF = async function *() {
  try {
    yield 1;
    yield 2;
    yield 3;
  } catch (e) {
    yield 4;
    yield 5;
  }
};
```

So it's bad idea to do `await g.throw(new Error());` if you're not expecting the generator to actually catch the exception. Because the method itself will return a rejected promise.

In our case, the `cancel` method of the `reverseMiddlewareStream` is receiving a "reason" to cancel the reverse stream. This can come from anywhere that controls the stream. It could be the web socket server cancelling both the forward and reverse streams in the event a transport connection error.

So therefore, any caller of `cancel(reason)` would not be expecting it to throw an exception.

Therefore, the `cancel` method of the reverse stream AND the forward stream, should not actually be throwing an exception into the generator abstractions we're using. Instead they should just "end" the generators. This can be done with `await g.return(undefined);`

However this causes another problem, the RPC handlers are not given any notice/signal about the fact that their streams are now cancelled. The only thing that would happen is if they end up calling `await g.next()` and getting `done` equal to `true`. This does not distinguish from the streams ending, vs the streams being cancelled.

So inside the handler, it may just stop operating. There's no way to "cleanup" itself.

Now we have a `ctx` thing which provides an `AbortSignal`. Can we use this to indicate that the stream has been cancelled?

It may not be a problem... because it is possible to do this:

```
const gF = async function *() {
  try {
    yield 1;
    yield 2;
    yield 3;
  } finally {
    // do final clean up
  }
};
```

CMCDragonkai:
The `cancel` method right now does the `dispatchEvent` for the reason. This may not really be necessary. If we propagate the `reason` into the handler. Then you don't really need to dispatch event here. It's not really an error.

CMCDragonkai:
1. Need to test the above commit about propagating forward and reverse stream cancellation reasons into the handler.
2. Fix up this https://github.com/MatrixAI/Polykey/pull/509#issuecomment-1480514684 weird temporary error constructor.
3. The docs still don't generate for rpc... investigate why.
4. So we know that TLS rotation doesn't work at this point.
5. Other review comments...

src/rpc/RPCServer.ts
CMCDragonkai: It actually does work if you do this:

```
  abstract handle: (
    input: AsyncIterable<Input>,
    connectionInfo: ConnectionInfo,
    ctx: ContextCancellable,
  ) => Promise<Output>;
```

That's an abstract arrow function property.

We can deal with this later.

src/rpc/RPCServer.ts
CMCDragonkai: See discussion https://github.com/MatrixAI/Polykey/pull/509#issuecomment-1480552627

src/bin/utils/processors.ts
CMCDragonkai: Will leave this to later, this requires analysis of the middleware functions and how best to organise them.

src/rpc/RPCServer.ts
CMCDragonkai: See the discussion: https://github.com/MatrixAI/Polykey/pull/509#issuecomment-1480565073

src/rpc/RPCServer.ts
CMCDragonkai: Can create new issue for this.

CMCDragonkai:
Later I can look into the middleware and separate it into files if necessary. There's no common generic middlewares yet. But the `middleware.ts` is just the default base functions for downstream to create other middleware. So all middleware functions and utils functions can just be put into `rpc/utils.ts`.

But this is not important to do, it's just code organisation.

CMCDragonkai:
Please create a new issue, or update the larger epic with all the code that needs to be removed during the agent migration. This should include `grpc` and `network` domains which would be replaced by `rpc` and `js-quic` respectively. But also mention the removal of the old exceptions.

When you do end up doing the agent migration, you should be able to change to using abstract async arrow property instead.

I still want you to have a read of the lazy dependency injection issue, because the usage of `Promise` there is basically very similar.

CMCDragonkai:
Once you get all tasks done, clean up the commits. Then proceed to merge.

Drop your docs commit, because that can be done in staging. We need to investigate why docs is not generating properly afterwards.

tegefaulkes:
I'm going back to the `generator.return()` solution for the stream fail handling. Trying to propagate the error through `generator.throw()` is not working as expected and trying to fix it is adding more complexity than it's worth right now.

CMCDragonkai:
Does that mean the handler doesn't get the signal for the exception?

23 Mar 2023 18:27:52 Brian Botha ***@***.***>:

tegefaulkes:
Yeah.

CMCDragonkai:
> Yeah.

We could send it via the abort signal.

CMCDragonkai:
> I'm going back to the `generator.return()` solution for the stream fail handling. Trying to propagate the error through `generator.throw()` is not working as expected and trying to fix it is adding more complexity than it's worth right now.

What was the behaviour that you saw?

tegefaulkes:
When aborting the output stream the test would just hang. The `generator.throw` never resolved. I think the generators was stuck waiting for input before reaching a yield point.

tegefaulkes:
Looks like `WebSocketClient` has a bunch of `TMP` errors, Looks like I didn't go through them creating proper errors. I'll do that now.

CMCDragonkai:
> When aborting the output stream the test would just hang. The `generator.throw` never resolved. I think the generators was stuck waiting for input before reaching a yield point.

Ok well so now readable is cancelled we get an error when we read.

But if writable is cancelled, then we are just finishing the yield.

Therefore any cleanup has to occur as part of a finally block in each handler.

So I think that is the right way to do it.

You should add a docbloc about using the `finally` to cleanup. Can you add a test for `finally`?

tegefaulkes:
I'll add the docblocks. I already have two tests checking that the handler ends with a finally block.

tegefaulkes:
Cleaning up history now. Almost ready for merge.

tegefaulkes:
This is ready to merge. All that is left to do is house keeping.

CMCDragonkai:
What housekeeping?

tegefaulkes:
I just had to update or create issues before I could resolve the last review comment. it's done now.

I checked and there is no docs changes commited.

I've re-enabled CI so this should be good to merge any time now.

src/rpc/RPCServer.ts
tegefaulkes: Created.

Issue #512, opened 2023-03-09
Agent RPC migration


tegefaulkes:
### Specification

Similar to MatrixAI/Polykey#509 but separate enough to be it's own issue/PR. We need to migrate the agent-agent communication to make use of the new agnostic RPC and quic system being developed.  Thankfully there are much fewer agent-agent handlers we need to migrate but they tend to be more complicated.

Some complications that need to be addresses is the the git pull and clone logic. They need to be updated to make use of the raw streams. The raw stream feature may need to be expanded to support this.

This is currently blocked by the quic system. 

Part of migrating the agent, the `network` and `GRPC` domains should be removed. Any code still depending on them need to be updated to the new systems.

We need to spec out how we want to handle the common elements between the client and agent RPC. They will end up sharing middleware, message data structure and metadata structure. These things should be kept in a common location but the `RPC` domain isn't appropriate.

### Additional context

* Related MatrixAI/Polykey#509 
* Related MatrixAI/Polykey#249 
* Related MatrixAI/Polykey#495 
* Related MatrixAI/js-rpc#5 
* Related MatrixAI/Polykey#510 

### Tasks

1. migrate handlers to the new RPC system
2. migrate tests
3. update RPC usage within the code
4. Remove any old GRPC or network code. GRPC domain should be fully removed at this point. Not sure about network.

tegefaulkes:
This one will take much less time compared to the client migration since there are much less handlers to migrate. At the same time we need to work out some things with the git handlers. They need to be implemented with the raw handlers and possibly need some features added to it to support middlware.

For timing this is blocked by the quic changes but doesn't strictly require them to get started, just to run it.

tegefaulkes:
There are going to be common elements used between the agent RPC and the client RPC. This stuff should be kept to a common domain. The specifics of this need to be worked out during this issue.

https://github.com/MatrixAI/Polykey/pull/513#discussion_r1156720526

tegefaulkes:
When migrating the agent handlers, we need to update them to use the CTX and timeouts.

CMCDragonkai:
Since closing MatrixAI/Polykey#466:

Very important to review also the client RPC to ensure deadlines are also applied too.

On the server side, handlers that are asynchronous should have a deadline propagated into asynchronous subcalls.

tegefaulkes:
There are a few stages to this one. 

1. Migrate the agent handlers to the agnostic RPC.
[...]

PR #513, opened 2023-03-14
RPC timeout implementation

### LINKED ISSUE #510: Add timeout support to the Agnostic RPC system
### Specification

We need to support default and client supplied timeouts for the RPCServer handlers.

We need to support some features with this.
* User can provide a timeout when making a client call
* Client calls have default timeouts,
* When a client times out the connection is immediately ended with a timeout code.
* Handlers are provided with a timeout timer as a `TimedCancellable` context
* Handlers have a default timeout provided 
* Handler timeout can be overridden by the client supplied value but only if it's less than the default.
* When a handler times out is the abort is advisory and then the connection is forced closed after a small delay.

I need to do some prototyping with how I want to implement this. We have two options. 
1. The RPC system supports timeouts directly.
2. RPC system allows timeouts to be supplied via the middleware.

Note that the raw handlers don't support the middleware, It's pretty low level in that respect. So raw handlers wouldn't support timeouts with option 2. Conversely Option 1 means adding a timeout field to the RPC messages since by design the RPC system doesn't interpret or apply constraints to the params.

Given some constraints the first option would mean the timeout feature is always available and enforced. The second option means it's up to the implementation of the middleware and handlers to enforce it.  I need to weigh both options after some prototyping. I'm leaning towards option 1, but It may require appending a field to the JSONRPC  request message format.

##### `RPCServer` changes

1. Two parameters have been added to the `RPCServer` when creating it. These are `defaultTimeout` and `forceEndDelay`. `defaultTimeout` sets the default amount of time before a handler times out and an abort signal is sent. Since this signal is advisory the `forceEndDelay` (name subject to change) sets the time between the signal and the streamPair being directly aborted. 
2. When handling a stream a timeout timer is created sends a timeout abort to the abort controller. When created it is started with a delay of `defaultTimeout`. When a handler is selected the timer is `reset` with the provided timeout for the handler if is is less than the default. If none is set then the timer is just `refresh`-ed. 
3. The duplex handler and by extension the other 3 basic handler types will refresh the timeout timer every time a message is sent or received. This is not done by the raw handler for now. A raw handler can directly affect the timeout timer by refreshing or resetting it.
4. The handlers have a timeout parameter that can be set when implementing the handler. This will be used as the timeout if it defined and less than the `RPCServer` `defaultTimeout` that was provided.
5. The `signal` and `timer` are provided to the handlers as a `ctx: ContextTimed`. The timer can be `refresh`-ed or `reset(delay)` or even `cancel(reason)`. This gives us the freedom to override the timeout behaviour from the handlers if we want that. If that is not desired, we can reduce the functionality by casting it with a more restrictive interface.

### Additional context

* Related #466
* Related #450
* Related #243

### Tasks

1. `RPCServer` handle timeouts and provide them to the handlers
3. `RPCClient` takes a timeout value or default and force closes the connection when exceeded.
4. `RPCClient` can communicate it's timeout to the `RPCServer`.

### LINKED ISSUE #516: Importing `@streamparser/json` results in error when using `tsc` or `jest` relying on `tsc`
### Specification

As part of #249 we're using a JSON stream parser to parse the JSONRPC messages. There seems to be a problem with this library specifically where we can't import it with the `import` keyword. So far a work around has been to import it with `const jsonStreamParsers = require('@streamparser/json');`. 

We need to import it with `import { JSONParser } from '@streamparser/json';`. but currently doing so results in a `TS2307: Cannot find module '@streamparser/json' or its corresponding type declarations.` error.

So far I haven't been able to determine why the usual import isn't working. From what I can tell it's using conditional exports within it's `package.json` file for specifying the exports. This may be related?

### Additional context

* Related #509 

### Tasks

1. determine the cause of the import error.
2. find a fix to import the stream parser using the `import` keyword.

tegefaulkes:
### Description

This PR handles adding the timeout features to the agnostic RPC system

### Issues Fixed

* Fixes #510 
* Fixes #516 

### Tasks

- [x] 1. Update `RPCServer` to support timeouts.
- [x] 2. Update `RPCClient` to support timeouts.
- [x] 3. Update client handlers to make use of the new cancellation/timeout `ctx`.
- ~5. Update agentHandlers to make use of the new cancellation/timeout `ctx`.~ This is blocked by #512 
- [x] 6. Support communicating timeouts between `RPCClient` and `RPCServer`. support updating the timeouts based on these communicated values.
- [x] 7. Update `WebSocketClient` to take a `ctx: ContextTime` and allow it to abort the current connection.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -57,9 +57,9 @@ type NetworkConfig = {
   clientHost?: Host;
   clientPort?: Port;
   maxReadBufferBytes?: number;
-  idleTimeout?: number;
-  pingInterval?: number;
-  pingTimeout?: number;
+  idleTimeoutTime?: number;
+  pingIntervalTime?: number;
+  pingTimeoutTime?: number;
 };
 
 interface PolykeyAgent extends CreateDestroyStartStop {}
@@ -193,7 +193,7 @@ class PolykeyAgent {
       ...config.defaults.nodeConnectionManagerConfig,
       ...utils.filterEmptyObject(nodeConnectionManagerConfig),
     };
-    const _networkConfig = {
+    const networkConfig_ = {
       ...config.defaults.networkConfig,
       ...utils.filterEmptyObject(networkConfig),
     };
@@ -454,13 +454,13 @@ class PolykeyAgent {
           connectionCallback: (streamPair) =>
             rpcServerClient!.handleStream(streamPair, {}),
           fs,
-          host: _networkConfig.clientHost,
-          port: _networkConfig.clientPort,
+          host: networkConfig_.clientHost,
+          port: networkConfig_.clientPort,
           tlsConfig,
-          maxReadBufferBytes: _networkConfig.maxReadBufferBytes,
-          idleTimeout: _networkConfig.idleTimeout,
-          pingInterval: _networkConfig.pingInterval,
-          pingTimeout: _networkConfig.pingTimeout,
+          maxReadableStreamBytes: networkConfig_.maxReadBufferBytes,
+          connectionIdleTimeoutTime: networkConfig_.idleTimeout,
+          pingIntervalTime: networkConfig_.pingIntervalTime,
+          pingTimeoutTime: networkConfig_.pingTimeoutTime,
           logger: logger.getChild('WebSocketServer'),
         }));
       grpcServerAgent =
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -56,10 +56,13 @@ type NetworkConfig = {
   // RPCServer for client service
   clientHost?: Host;
   clientPort?: Port;
-  maxReadBufferBytes?: number;
-  idleTimeoutTime?: number;
+  maxReadableStreamBytes?: number;
+  connectionIdleTimeoutTime?: number;
   pingIntervalTime?: number;
   pingTimeoutTime?: number;
+  handlerTimeoutTime?: number;
+  handlerTimeoutGraceTime?: number;
+  clientParserBufferByteLimit?: number;
 };
 
 interface PolykeyAgent extends CreateDestroyStartStop {}
@@ -439,8 +442,11 @@ class PolykeyAgent {
           }),
           middlewareFactory: rpcUtilsMiddleware.defaultServerMiddlewareWrapper(
             clientUtilsMiddleware.middlewareServer(sessionManager, keyRing),
+            networkConfig_.clientParserBufferByteLimit,
           ),
           sensitive: false,
+          handlerTimeoutTime: networkConfig_.handlerTimeoutTime,
+          handlerTimeoutGraceTime: networkConfig_.handlerTimeoutGraceTime,
           logger: logger.getChild('RPCServerClient'),
         });
       }
@@ -451,14 +457,14 @@ class PolykeyAgent {
       webSocketServerClient =
         webSocketServerClient ??
         (await WebSocketServer.createWebSocketServer({
-          connectionCallback: (streamPair) =>
-            rpcServerClient!.handleStream(streamPair),
+          connectionCallback: (rpcStream) =>
+            rpcServerClient!.handleStream(rpcStream),
           fs,
           host: networkConfig_.clientHost,
           port: networkConfig_.clientPort,
           tlsConfig,
-          maxReadableStreamBytes: networkConfig_.maxReadBufferBytes,
-          connectionIdleTimeoutTime: networkConfig_.idleTimeout,
+          maxReadableStreamBytes: networkConfig_.maxReadableStreamBytes,
+          connectionIdleTimeoutTime: networkConfig_.connectionIdleTimeoutTime,
           pingIntervalTime: networkConfig_.pingIntervalTime,
           pingTimeoutTime: networkConfig_.pingTimeoutTime,
           logger: logger.getChild('WebSocketServer'),
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -452,7 +452,7 @@ class PolykeyAgent {
         webSocketServerClient ??
         (await WebSocketServer.createWebSocketServer({
           connectionCallback: (streamPair) =>
-            rpcServerClient!.handleStream(streamPair, {}),
+            rpcServerClient!.handleStream(streamPair),
           fs,
           host: networkConfig_.clientHost,
           port: networkConfig_.clientPort,
@@ -775,7 +775,7 @@ class PolykeyAgent {
         host: _networkConfig.clientHost,
         port: _networkConfig.clientPort,
         connectionCallback: (streamPair) =>
-          this.rpcServerClient.handleStream(streamPair, {}),
+          this.rpcServerClient.handleStream(streamPair),
       });
       // Agent server
       await this.grpcServerAgent.start({
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -12,8 +12,8 @@ import { DB } from '@matrixai/db';
 import { CreateDestroyStartStop } from '@matrixai/async-init/dist/CreateDestroyStartStop';
 import RPCServer from './rpc/RPCServer';
 import WebSocketServer from './websockets/WebSocketServer';
-import * as rpcMiddlewareUtils from './rpc/utils/middleware';
-import * as clientMiddleware from './client/utils/middleware';
+import * as rpcUtilsMiddleware from './rpc/utils/middleware';
+import * as clientUtilsMiddleware from './client/utils/middleware';
 import { WorkerManager } from './workers';
 import * as networkUtils from './network/utils';
 import KeyRing from './keys/KeyRing';
@@ -437,8 +437,8 @@ class PolykeyAgent {
             sessionManager: sessionManager,
             vaultManager: vaultManager,
           }),
-          middlewareFactory: rpcMiddlewareUtils.defaultServerMiddlewareWrapper(
-            clientMiddleware.middlewareServer(sessionManager, keyRing),
+          middlewareFactory: rpcUtilsMiddleware.defaultServerMiddlewareWrapper(
+            clientUtilsMiddleware.middlewareServer(sessionManager, keyRing),
           ),
           sensitive: false,
           logger: logger.getChild('RPCServerClient'),
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -12,8 +12,8 @@ import { DB } from '@matrixai/db';
 import { CreateDestroyStartStop } from '@matrixai/async-init/dist/CreateDestroyStartStop';
 import RPCServer from './rpc/RPCServer';
 import WebSocketServer from './websockets/WebSocketServer';
-import * as middlewareUtils from './rpc/utils/middleware';
-import * as authMiddleware from './client/utils/authenticationMiddleware';
+import * as rpcMiddlewareUtils from './rpc/utils/middleware';
+import * as clientMiddleware from './client/utils/middleware';
 import { WorkerManager } from './workers';
 import * as networkUtils from './network/utils';
 import KeyRing from './keys/KeyRing';
@@ -437,11 +437,8 @@ class PolykeyAgent {
             sessionManager: sessionManager,
             vaultManager: vaultManager,
           }),
-          middlewareFactory: middlewareUtils.defaultServerMiddlewareWrapper(
-            authMiddleware.authenticationMiddlewareServer(
-              sessionManager,
-              keyRing,
-            ),
+          middlewareFactory: rpcMiddlewareUtils.defaultServerMiddlewareWrapper(
+            clientMiddleware.middlewareServer(sessionManager, keyRing),
           ),
           sensitive: false,
           logger: logger.getChild('RPCServerClient'),
--- a/src/bin/agent/CommandLockAll.ts
+++ b/src/bin/agent/CommandLockAll.ts
@@ -58,7 +58,7 @@ class CommandLockAll extends CommandPolykey {
           logger: this.logger.getChild(WebSocketClient.name),
         });
         pkClient = await PolykeyClient.createPolykeyClient({
-          streamFactory: () => webSocketClient.startConnection(),
+          streamFactory: (ctx) => webSocketClient.startConnection(ctx),
           nodePath: options.nodePath,
           manifest: clientManifest,
           logger: this.logger.getChild(PolykeyClient.name),
--- a/src/bin/identities/CommandPermissions.ts
+++ b/src/bin/identities/CommandPermissions.ts
@@ -56,7 +56,7 @@ class CommandPermissions extends CommandPolykey {
           logger: this.logger.getChild(WebSocketClient.name),
         });
         pkClient = await PolykeyClient.createPolykeyClient({
-          streamFactory: () => webSocketClient.startConnection(),
+          streamFactory: (ctx) => webSocketClient.startConnection(ctx),
           nodePath: options.nodePath,
           manifest: clientManifest,
           logger: this.logger.getChild(PolykeyClient.name),
--- a/src/client/handlers/gestaltsGestaltList.ts
+++ b/src/client/handlers/gestaltsGestaltList.ts
@@ -22,8 +22,8 @@ class GestaltsGestaltListHandler extends ServerHandler<
     yield* db.withTransactionG(async function* (
       tran,
     ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
+      if (ctx.signal.aborted) throw ctx.signal.reason;
       for await (const gestalt of gestaltGraph.getGestalts(tran)) {
-        if (ctx.signal.aborted) throw ctx.signal.reason;
         const gestaltMessage: GestaltMessage = {
           gestalt: {
             matrix: {},
@@ -35,13 +35,11 @@ class GestaltsGestaltListHandler extends ServerHandler<
         const newGestalt = gestaltMessage.gestalt;
         newGestalt.identities = gestalt.identities;
         for (const [key, value] of Object.entries(gestalt.nodes)) {
-          if (ctx.signal.aborted) throw ctx.signal.reason;
           newGestalt.nodes[key] = {
             nodeId: nodesUtils.encodeNodeId(value.nodeId),
           };
         }
         for (const keyA of Object.keys(gestalt.matrix)) {
-          if (ctx.signal.aborted) throw ctx.signal.reason;
           let record = newGestalt.matrix[keyA];
           if (record == null) {
             record = {};
@@ -51,6 +49,7 @@ class GestaltsGestaltListHandler extends ServerHandler<
             record[keyB] = null;
           }
         }
+        if (ctx.signal.aborted) throw ctx.signal.reason;
         yield gestaltMessage;
       }
     });
--- a/src/client/handlers/gestaltsGestaltList.ts
+++ b/src/client/handlers/gestaltsGestaltList.ts
@@ -13,14 +13,17 @@ class GestaltsGestaltListHandler extends ServerHandler<
   ClientRPCRequestParams,
   ClientRPCResponseResult<GestaltMessage>
 > {
-  public async *handle(): AsyncGenerator<
-    ClientRPCResponseResult<GestaltMessage>
-  > {
+  public async *handle(
+    input_,
+    connectionInfo_,
+    ctx,
+  ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
     const { db, gestaltGraph } = this.container;
     yield* db.withTransactionG(async function* (
       tran,
     ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
       for await (const gestalt of gestaltGraph.getGestalts(tran)) {
+        if (ctx.signal.aborted) throw ctx.signal.reason;
         const gestaltMessage: GestaltMessage = {
           gestalt: {
             matrix: {},
@@ -32,11 +35,13 @@ class GestaltsGestaltListHandler extends ServerHandler<
         const newGestalt = gestaltMessage.gestalt;
         newGestalt.identities = gestalt.identities;
         for (const [key, value] of Object.entries(gestalt.nodes)) {
+          if (ctx.signal.aborted) throw ctx.signal.reason;
           newGestalt.nodes[key] = {
             nodeId: nodesUtils.encodeNodeId(value.nodeId),
           };
         }
         for (const keyA of Object.keys(gestalt.matrix)) {
+          if (ctx.signal.aborted) throw ctx.signal.reason;
           let record = newGestalt.matrix[keyA];
           if (record == null) {
             record = {};
--- a/src/client/handlers/gestaltsGestaltList.ts
+++ b/src/client/handlers/gestaltsGestaltList.ts
@@ -14,8 +14,8 @@ class GestaltsGestaltListHandler extends ServerHandler<
   ClientRPCResponseResult<GestaltMessage>
 > {
   public async *handle(
-    input_,
-    connectionInfo_,
+    _input,
+    _connectionInfo,
     ctx,
   ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
     const { db, gestaltGraph } = this.container;
--- a/src/client/handlers/gestaltsGestaltList.ts
+++ b/src/client/handlers/gestaltsGestaltList.ts
@@ -15,7 +15,8 @@ class GestaltsGestaltListHandler extends ServerHandler<
 > {
   public async *handle(
     _input,
-    _connectionInfo,
+    _cancel,
+    _meta,
     ctx,
   ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
     const { db, gestaltGraph } = this.container;
--- a/src/client/handlers/identitiesAuthenticate.ts
+++ b/src/client/handlers/identitiesAuthenticate.ts
@@ -24,6 +24,7 @@ class IdentitiesAuthenticateHandler extends ServerHandler<
     _,
     ctx,
   ): AsyncGenerator<ClientRPCResponseResult<AuthProcessMessage>> {
+    if (ctx.signal.aborted) throw ctx.signal.reason;
     const { identitiesManager } = this.container;
     const {
       providerId,
@@ -44,24 +45,23 @@ class IdentitiesAuthenticateHandler extends ServerHandler<
     if (provider == null) {
       throw new identitiesErrors.ErrorProviderMissing();
     }
-    if (ctx.signal.aborted) throw ctx.signal.reason;
     const authFlow = provider.authenticate(ctx.timer.getTimeout());
-    if (ctx.signal.aborted) throw ctx.signal.reason;
     let authFlowResult = await authFlow.next();
     if (authFlowResult.done) {
       never();
     }
+    if (ctx.signal.aborted) throw ctx.signal.reason;
     yield {
       request: {
         url: authFlowResult.value.url,
         dataMap: authFlowResult.value.data,
       },
     };
-    if (ctx.signal.aborted) throw ctx.signal.reason;
     authFlowResult = await authFlow.next();
     if (!authFlowResult.done) {
       never();
     }
+    if (ctx.signal.aborted) throw ctx.signal.reason;
     yield {
       response: {
         identityId: authFlowResult.value,
--- a/src/client/handlers/identitiesAuthenticate.ts
+++ b/src/client/handlers/identitiesAuthenticate.ts
@@ -21,7 +21,8 @@ class IdentitiesAuthenticateHandler extends ServerHandler<
     input: ClientRPCRequestParams<{
       providerId: string;
     }>,
-    _,
+    _cancel,
+    _meta,
     ctx,
   ): AsyncGenerator<ClientRPCResponseResult<AuthProcessMessage>> {
     if (ctx.signal.aborted) throw ctx.signal.reason;
--- a/src/client/handlers/identitiesAuthenticate.ts
+++ b/src/client/handlers/identitiesAuthenticate.ts
@@ -21,6 +21,8 @@ class IdentitiesAuthenticateHandler extends ServerHandler<
     input: ClientRPCRequestParams<{
       providerId: string;
     }>,
+    _,
+    ctx,
   ): AsyncGenerator<ClientRPCResponseResult<AuthProcessMessage>> {
     const { identitiesManager } = this.container;
     const {
@@ -42,7 +44,9 @@ class IdentitiesAuthenticateHandler extends ServerHandler<
     if (provider == null) {
       throw new identitiesErrors.ErrorProviderMissing();
     }
-    const authFlow = provider.authenticate();
+    if (ctx.signal.aborted) throw ctx.signal.reason;
+    const authFlow = provider.authenticate(ctx.timer.getTimeout());
+    if (ctx.signal.aborted) throw ctx.signal.reason;
     let authFlowResult = await authFlow.next();
     if (authFlowResult.done) {
       never();
@@ -53,6 +57,7 @@ class IdentitiesAuthenticateHandler extends ServerHandler<
         dataMap: authFlowResult.value.data,
       },
     };
+    if (ctx.signal.aborted) throw ctx.signal.reason;
     authFlowResult = await authFlow.next();
     if (!authFlowResult.done) {
       never();
--- a/src/client/handlers/vaultsCreate.ts
+++ b/src/client/handlers/vaultsCreate.ts
@@ -5,7 +5,7 @@ import type { VaultIdMessage, VaultNameMessage } from './types';
 import * as vaultsUtils from '../../vaults/utils';
 import { UnaryHandler } from '../../rpc/handlers';
 
-class VaultsCreatehandler extends UnaryHandler<
+class VaultsCreateHandler extends UnaryHandler<
   {
     db: DB;
     vaultManager: VaultManager;
@@ -28,4 +28,4 @@ class VaultsCreatehandler extends UnaryHandler<
   }
 }
 
-export { VaultsCreatehandler };
+export { VaultsCreateHandler };
--- a/src/client/utils/middleware.ts
+++ b/src/client/utils/middleware.ts
@@ -24,10 +24,13 @@ function middlewareServer(
       sessionManager,
       keyRing,
     );
-  return (ctx) => {
-    const authMiddleware = authMiddlewareFactory(ctx);
-    const timeoutMiddleware =
-      timeoutMiddlewareUtils.timeoutMiddlewareServer(ctx);
+  return (ctx, cancel, meta) => {
+    const authMiddleware = authMiddlewareFactory(ctx, cancel, meta);
+    const timeoutMiddleware = timeoutMiddlewareUtils.timeoutMiddlewareServer(
+      ctx,
+      cancel,
+      meta,
+    );
     // Order is auth -> timeout
     return {
       forward: {
@@ -56,10 +59,13 @@ function middlewareClient(
 > {
   const authMiddlewareFactory =
     authenticationMiddlewareUtils.authenticationMiddlewareClient(session);
-  return (ctx) => {
-    const authMiddleware = authMiddlewareFactory(ctx);
-    const timeoutMiddleware =
-      timeoutMiddlewareUtils.timeoutMiddlewareClient(ctx);
+  return (ctx, cancel, meta) => {
+    const authMiddleware = authMiddlewareFactory(ctx, cancel, meta);
+    const timeoutMiddleware = timeoutMiddlewareUtils.timeoutMiddlewareClient(
+      ctx,
+      cancel,
+      meta,
+    );
     // Order is timeout -> auth
     return {
       forward: {
--- a/src/client/utils/middleware.ts
+++ b/src/client/utils/middleware.ts
@@ -1,0 +1,81 @@
+import type { ClientRPCRequestParams, ClientRPCResponseResult } from '../types';
+import type { Session } from '../../sessions';
+import type {
+  JSONRPCRequest,
+  JSONRPCResponse,
+  MiddlewareFactory,
+} from '../../rpc/types';
+import type SessionManager from '../../sessions/SessionManager';
+import type KeyRing from '../../keys/KeyRing';
+import * as authenticationMiddlewareUtils from './authenticationMiddleware';
+import * as timeoutMiddlewareUtils from './timeoutMiddleware';
+
+function middlewareServer(
+  sessionManager: SessionManager,
+  keyRing: KeyRing,
+): MiddlewareFactory<
+  JSONRPCRequest<ClientRPCRequestParams>,
+  JSONRPCRequest<ClientRPCRequestParams>,
+  JSONRPCResponse<ClientRPCResponseResult>,
+  JSONRPCResponse<ClientRPCResponseResult>
+> {
+  const authMiddlewareFactory =
+    authenticationMiddlewareUtils.authenticationMiddlewareServer(
+      sessionManager,
+      keyRing,
+    );
+  return (ctx) => {
+    const authMiddleware = authMiddlewareFactory(ctx);
+    const timeoutMiddleware =
+      timeoutMiddlewareUtils.timeoutMiddlewareServer(ctx);
+    // Order is auth -> timeout
+    return {
+      forward: {
+        writable: authMiddleware.forward.writable,
+        readable: authMiddleware.forward.readable.pipeThrough(
+          timeoutMiddleware.forward,
+        ),
+      },
+      reverse: {
+        writable: timeoutMiddleware.reverse.writable,
+        readable: timeoutMiddleware.reverse.readable.pipeThrough(
+          authMiddleware.reverse,
+        ),
+      },
+    };
+  };
+}
+
+function middlewareClient(
+  session: Session,
+): MiddlewareFactory<
+  JSONRPCRequest<ClientRPCRequestParams>,
+  JSONRPCRequest<ClientRPCRequestParams>,
+  JSONRPCResponse<ClientRPCResponseResult>,
+  JSONRPCResponse<ClientRPCResponseResult>
+> {
+  const authMiddlewareFactory =
+    authenticationMiddlewareUtils.authenticationMiddlewareClient(session);
+  return (ctx) => {
+    const authMiddleware = authMiddlewareFactory(ctx);
+    const timeoutMiddleware =
+      timeoutMiddlewareUtils.timeoutMiddlewareClient(ctx);
+    // Order is timeout -> auth
+    return {
+      forward: {
+        writable: timeoutMiddleware.forward.writable,
+        readable: timeoutMiddleware.forward.readable.pipeThrough(
+          authMiddleware.forward,
+        ),
+      },
+      reverse: {
+        writable: authMiddleware.reverse.writable,
+        readable: authMiddleware.reverse.readable.pipeThrough(
+          timeoutMiddleware.reverse,
+        ),
+      },
+    };
+  };
+}
+
+export { middlewareServer, middlewareClient };
--- a/src/client/utils/timeoutMiddleware.ts
+++ b/src/client/utils/timeoutMiddleware.ts
@@ -9,6 +9,7 @@ import { TransformStream } from 'stream/web';
  */
 function timeoutMiddlewareServer(ctx: ContextTimed) {
   const currentTimeout = ctx.timer.delay;
+  // Flags for tracking if the first message has been processed
   let forwardFirst = true;
   let reverseFirst = true;
   return {
@@ -52,6 +53,7 @@ function timeoutMiddlewareServer(ctx: ContextTimed) {
  */
 function timeoutMiddlewareClient(ctx: ContextTimed) {
   const currentTimeout = ctx.timer.delay;
+  // Flags for tracking if the first message has been processed
   let forwardFirst = true;
   let reverseFirst = true;
   return {
--- a/src/client/utils/timeoutMiddleware.ts
+++ b/src/client/utils/timeoutMiddleware.ts
@@ -1,13 +1,18 @@
 import type { JSONRPCRequest, JSONRPCResponse } from '../../rpc/types';
 import type { ClientRPCRequestParams, ClientRPCResponseResult } from '../types';
 import type { ContextTimed } from '../../contexts/types';
+import type { JSONValue } from '../../types';
 import { TransformStream } from 'stream/web';
 
 /**
  * This adds its timeout to the reverse metadata and updates it's timeout based
  * on the forward metadata.
  */
-function timeoutMiddlewareServer(ctx: ContextTimed) {
+function timeoutMiddlewareServer(
+  ctx: ContextTimed,
+  _cancel: (reason?: any) => void,
+  _meta: Record<string, JSONValue> | undefined,
+) {
   const currentTimeout = ctx.timer.delay;
   // Flags for tracking if the first message has been processed
   let forwardFirst = true;
@@ -50,8 +55,14 @@ function timeoutMiddlewareServer(ctx: ContextTimed) {
  * This adds its own timeout to the forward metadata and updates it's timeout
  * based on the reverse metadata.
  * @param ctx
+ * @param _cancel
+ * @param _meta
  */
-function timeoutMiddlewareClient(ctx: ContextTimed) {
+function timeoutMiddlewareClient(
+  ctx: ContextTimed,
+  _cancel: (reason?: any) => void,
+  _meta: Record<string, JSONValue> | undefined,
+) {
   const currentTimeout = ctx.timer.delay;
   // Flags for tracking if the first message has been processed
   let forwardFirst = true;
--- a/src/client/utils/timeoutMiddleware.ts
+++ b/src/client/utils/timeoutMiddleware.ts
@@ -1,0 +1,91 @@
+import type { JSONRPCRequest, JSONRPCResponse } from '../../rpc/types';
+import type { ClientRPCRequestParams, ClientRPCResponseResult } from '../types';
+import type { ContextTimed } from '../../contexts/types';
+import { TransformStream } from 'stream/web';
+
+/**
+ * This adds its timeout to the reverse metadata and updates it's timeout based
+ * on the forward metadata.
+ */
+function timeoutMiddlewareServer(ctx: ContextTimed) {
+  const currentTimeout = ctx.timer.delay;
+  let forwardFirst = true;
+  let reverseFirst = true;
+  return {
+    forward: new TransformStream<
+      JSONRPCRequest<ClientRPCRequestParams>,
+      JSONRPCRequest<ClientRPCRequestParams>
+    >({
+      transform: (chunk, controller) => {
+        controller.enqueue(chunk);
+        if (forwardFirst) {
+          forwardFirst = false;
+          const clientTimeout = chunk.params?.metadata?.timeout;
+
+          if (clientTimeout == null) return;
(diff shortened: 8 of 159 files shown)


CMCDragonkai:
When this is integrated into the handlers, does this need to be rebased on top of #509?

CMCDragonkai:
@tegefaulkes can you also flesh out the remaining tasks here, and what tests you have.

tegefaulkes:
Re-based on #509

tegefaulkes:
This needs to be rebased on staging after the recent merge. Doing that now.

tegefaulkes:
Cleaned up the commit history and re-based on staging. I'll check for any problems that the re-base may have introduced and then work on the `RPCClient` timeout support.

tegefaulkes:
`RPCClient` needs the following features to support timeouts.
1. A default timeout value to  be used when none is provided during a client call.
2. Client calls need an optional `ContextedTimed` ctx. We should use the `timedCancellable` decorator for the caller methods.
3. A timeout timer needs to be created for each call if none was provided.  
4. The timer needs to be refreshed when data is received or sent.
5. The timer needs to be cleaned up when call ends or is cancelled.

tegefaulkes:
I can't use the `TimedCancellable` decorator for most of the caller methods. The decorator cleans up any `AbortController` or timer it creates when the method call has ended. The problem is, since we're returning streams most of the time. The context needs to exist for the duration of the streams and not just the method creating them.

tegefaulkes:
It is possible to abort a webstream. I've been doing it with `readable.pipeTo(writable, {signal})`. This is the only way to cancel a stream if you don't control either side of it due to locking. 

This comes with a small problem. It is possible to provide a reason for abortion. But this reason isn't being propagated through the stream. Aborting the stream this way results in the error  `AbortError: The operation was aborted` for the stream. This is the same as the default `DOMException` error for abortion that we saw when fixing the `promiseCancellable` problem.

Ultimately this means our abort or timeout reason won't be thrown if the stream times out like this. Not through the streams at least.

CMCDragonkai:
> I can't use the `TimedCancellable` decorator for most of the caller methods. The decorator cleans up any `AbortController` or timer it creates when the method call has ended. The problem is, since we're returning streams most of the time. The context needs to exist for the duration of the streams and not just the method creating them.

You don't need to use the decorators. Use the HOF variant OR you can just construct the object yourself and maintain lifecycle somehow. Those were intended for convenience of thing it to the function scope context. But your timer object should be tied to the stream life cycle.

CMCDragonkai:
1. Timeout property is optional, and should be managed like any other metadata param just like authentication. Which means middleware.
2. Rather than using signal reason from `toPipe`, maintain a timer and abort controller attached to the lifecycle of every stream at the `RPCClient` and `RPCServer` level. Then you have more flexibility here, and can inject these control points to whatever needs it.
3. It is possible that the `streamFactory` can receive an abort controller, or even a timer really. It's an async function can be aborted. Right now js-quic doesn't take this. But you could update your websocket transport layer to deal with it accordingly.

tegefaulkes:
Few notes.
1. It's possible to refresh the timer when sending or receiving a message. To do this we need to make use of a pass-through `TransformStream`. We should do this to match the watchdog style timeout of the server.
2. In the absence of  using a signal with a `pipeTo`, the caller has no way of ending a stream directly when it times out. This leaves us two options. The stream creator, this case the `WebSocketServer` or the quic system, needs to respond to an abort signal by cancelling and cleaning up the stream. The other alternative is to pass in a timer and refresh it directly or race it when interacting with the client

To emphasise point 2. the `ctx: ContextTimed` being passed into the stream factory is expected to do two things. 
1. Race the stream creation. The `WebSocket` domain does create a new connection for each stream. The quic system does not, if there is an active connection  then a stream is instant with possibly some negotiation?
2. Terminate a running stream. Say if a connection is lagging, the other side isn't responding but the connection is good, the connection is failed but the stream hasn't ended yet.

Point 2. is needed for the timeout mechanism to work properly. From the `RPCClient`'s perspective a stream has 3 parts worth mentioning, the end that the user interacts with, the middle that the caller method can see and the end that the transport layer is interacting with.

Due to how stream locking works and composing, you can't really interact with a stream from the middle. And from the `RPCClients` perspective once we construct the transformation and pass either end to the user and transport layer, we have no control over that stream. So if we want to timeout and close the stream automatically, the only options are tell the transport layer to do it via the `ctx`. Or interact with the middle with the `pipeTo` which as mentioned before doesn't give clean results. 

The only option left is to have the transport layer respect a signal and terminate the stream.

tegefaulkes:
When creating the `RPCClient`, at the time I opted not to include a map of active connections and the ability to await all of the/ force end them. At the time it didn't seem necessary due to the life cycle of a connection being tied closely to the life cycle of the the calling method or stream.

based on previous discussion I'll look into adding it in if only for debug usage.

tegefaulkes:
From the client caller method's perspective, there are two kinds of timers, one that was provided and a default timer. 

If we provide a timer when making a call, I'm not sure we want to be refreshing when messages are sent. If we are composing multiple actions under a single timeout, we don't want an rpc call to mutate it. We may also want it to.

So the question is, do we want to enable refreshing implicitly based on if the timer was provided, or based on an option or both?

Right now I'm thinking, it will do refreshing if it's using the default timer, or not do refreshing if a timer was provided. This can be overridden with an option.

tegefaulkes:
Quick note, if a `Timer` is left running then it holds the process open. This is caused by the underlying timer from `setTimeout()` and `setInterval()`. They must be cancelled to prevent the process from being held open.

So if you create a `Timer` for a timeout when making an RPC call, it must be cancelled. Otherwise it's essentially a resource leak. 

`Timers` might benefit from a `withTimer` context function.

CMCDragonkai:
> Quick note, if a `Timer` is left running then it holds the process open. This is caused by the underlying timer from `setTimeout()` and `setInterval()`. They must be cancelled to prevent the process from being held open.
> 
> So if you create a `Timer` for a timeout when making an RPC call, it must be cancelled. Otherwise it's essentially a resource leak.
> 
> `Timers` might benefit from a `withTimer` context function.

Its the job of the caller when passing in the `Timer` to ensure deallocation when appropriate. So rpc call can't do anything. RPC call should close it if the timer was created by the rpc call and not externally. Basically ownership encompasses lifecycle.

tegefaulkes:
`WebSocketClient` needs to be updated to take a `ctx: ContextTimed` when for `startConnection()`. The timeout logic can also use improvements.

CMCDragonkai:
Ready for review?

Should we skip the agent migration or should just be merged?

I think after this you should go straight to QUIC so we can start on it.

tegefaulkes:
Somewhat ready for a review. Just dealing with a bug right now.

Right now the problem is, If a client RPC call is made and an error is returned via the reverse path, the forward path isn't cleaned up. For a duplex call, this meand that the writable isn't closed when an  error is received on the readable side. 

So I need to clean up the streams if an error is received for the client. For symmetry I should make sure that the sever side cleans up and ends the streams if an error message is sent. This cleaning up is an absolute must for the unary handler, but I'm not sure about the duplex or the streaming handlers? 

Do we expect the user to handle the clean up if an error is received? Ideally The cleaning up is handled automatically. But we're running into cancelling the stream from the middle problem again. 

I may not handle closing the stream after an error for the duplex, the other callers I can handle it. Implementing it for the duplex would be best though.

tegefaulkes:
Thinking about it, anything where the user is interacting with the writable stream, we can't control that nicely. The easiest solution here is to have any use of client side streaming use a finally block the clean up the streams. For server and unary calls we can automatically clean up the writable side since we internally control that stream. 

I'll add information to the docblocks about requiring the clean up of streams during usage. 

On the server side, all kinds of handlers can clean up both the streams if an error happens. I need to check why if it's not doing that already.

CMCDragonkai:
> Do we expect the user to handle the clean up if an error is received? Ideally The cleaning up is handled automatically. But we're running into cancelling the stream from the middle problem again.

If the resource is created automatically, it must be cleaned up automatically. That's the symmetry.

If the resource is created manually within the function scope, then it must be cleaned up manually within the function scope.

CMCDragonkai:
The https://github.com/juanjoDiaz/streamparser-json/issues/25 might be fixed in time for this PR.

src/PolykeyAgent.ts
```diff
@@ -12,8 +12,8 @@ import { DB } from '@matrixai/db';
 import { CreateDestroyStartStop } from '@matrixai/async-init/dist/CreateDestroyStartStop';
 import RPCServer from './rpc/RPCServer';
 import WebSocketServer from './websockets/WebSocketServer';
-import * as middlewareUtils from './rpc/utils/middleware';
-import * as authMiddleware from './client/utils/authenticationMiddleware';
+import * as rpcMiddlewareUtils from './rpc/utils/middleware';
```
CMCDragonkai: This should be called `rpcMiddleswareUtils`. You should follow the directory structure if you're renaming. But I think we will re-org the middleware structure later.

src/bin/agent/CommandLockAll.ts
```diff
@@ -58,7 +58,7 @@ class CommandLockAll extends CommandPolykey {
           logger: this.logger.getChild(WebSocketClient.name),
         });
         pkClient = await PolykeyClient.createPolykeyClient({
-          streamFactory: () => webSocketClient.startConnection(),
+          streamFactory: (ctx) => webSocketClient.startConnection(ctx),
```
CMCDragonkai, line 61: Note that this `ctx` primarily contains only a `signal` right? Or does it also container a timer?

```
ctx.timer
ctx.signal
```

If it contains both, then we could use both to know what to do. I think the stream factory will only deal with the `signal`.

The only time it would also use the `ctx.timer` is when the timer is `ctx` is passed to down to any timed operations.

Just trying to remember how we intended to make this work.

CMCDragonkai: Also note that in the js-quic this will need to be updated.

Specifically:

```
QUICClient.createQUICClient
QUICConnection.connectQUICConnection
QUICConnections.streamNew
```

src/bin/identities/CommandPermissions.ts
```diff
@@ -56,7 +56,7 @@ class CommandPermissions extends CommandPolykey {
           logger: this.logger.getChild(WebSocketClient.name),
         });
         pkClient = await PolykeyClient.createPolykeyClient({
-          streamFactory: () => webSocketClient.startConnection(),
+          streamFactory: (ctx) => webSocketClient.startConnection(ctx),
```
CMCDragonkai, line 59: Note that in QUIC, creating a new stream involves potentially creating a new connection and stream.

However here, you already have the `webSocketClient` and you are calling `startConnection`.

Does that mean... when using this in a QUIC situation, then you'd be using `quicConnection.streamNew`?

src/client/handlers/gestaltsGestaltList.ts
```diff
@@ -13,14 +13,17 @@ class GestaltsGestaltListHandler extends ServerHandler<
   ClientRPCRequestParams,
   ClientRPCResponseResult<GestaltMessage>
 > {
-  public async *handle(): AsyncGenerator<
-    ClientRPCResponseResult<GestaltMessage>
-  > {
+  public async *handle(
+    input_,
+    connectionInfo_,
```
CMCDragonkai: Why do these both have trailing underscores?

src/client/handlers/gestaltsGestaltList.ts
```diff
@@ -13,14 +13,17 @@ class GestaltsGestaltListHandler extends ServerHandler<
...
+    ctx,
+  ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
     const { db, gestaltGraph } = this.container;
     yield* db.withTransactionG(async function* (
       tran,
     ): AsyncGenerator<ClientRPCResponseResult<GestaltMessage>> {
       for await (const gestalt of gestaltGraph.getGestalts(tran)) {
+        if (ctx.signal.aborted) throw ctx.signal.reason;
```
CMCDragonkai: I wonder if it's better to do it a bit differently.

This:

```
if (ctx.signal.aborted) throw ctx.signal.reason;
```

Should be done at the beginning of the function before any `await`.

Then it can also be checked at the end of each loop operation.

However that's not them most ideal.

The most ideal, is that the `ctx` is passed directly into the operation itself that is here would be `gestaltGraph.getGestalts(tran, ctx)`.

Note that you should also then check before every `yield`.

But it is not necessary to check on each synchronous loop.

src/client/handlers/gestaltsGestaltList.ts
```diff
@@ -32,11 +35,13 @@ class GestaltsGestaltListHandler extends ServerHandler<
         const newGestalt = gestaltMessage.gestalt;
         newGestalt.identities = gestalt.identities;
         for (const [key, value] of Object.entries(gestalt.nodes)) {
+          if (ctx.signal.aborted) throw ctx.signal.reason;
```
CMCDragonkai: Like here, this is inside a synchronous loop, it's not necessary to check here, since it is not possible for the signal to be aborted while inside a loop.

Remember JS is singlethreaded here, so the signal can only be aborted IF there's an async await or yield.

src/client/handlers/gestaltsGestaltList.ts
```diff
@@ -32,11 +35,13 @@ class GestaltsGestaltListHandler extends ServerHandler<
...
           newGestalt.nodes[key] = {
             nodeId: nodesUtils.encodeNodeId(value.nodeId),
           };
         }
         for (const keyA of Object.keys(gestalt.matrix)) {
+          if (ctx.signal.aborted) throw ctx.signal.reason;
```
CMCDragonkai: Same here, remove this.

CMCDragonkai: Look for any situation BEFORE any `await` or `yield`.

Ideally if that `await op()` can take the `ctx` directly, then you don't need to check it.

src/client/handlers/identitiesAuthenticate.ts
CMCDragonkai, line 1: Apply the principles as in `src/client/handlers/gestaltsGestaltList.ts` regarding the checking of the context signal.

src/client/handlers/vaultsCreate.ts
```diff
@@ -28,4 +28,4 @@ class VaultsCreatehandler extends UnaryHandler<
   }
 }
 
-export { VaultsCreatehandler };
+export { VaultsCreateHandler };
```
CMCDragonkai, line 31: If this is the only class, it should be default exported. Then you can just do `import VaultsCreateHandler from './vaultsCreate`.

Actually I think it should be `import VaultsCreateHandler from ./VaultsCreateHandler`.

src/client/utils/middleware.ts
CMCDragonkai, line 1: So we have generic middleware inside `src/rpc` right? Then client specific middleware and agent specific middleware?

I reckon any given middleware should have its own file.

src/client/utils/timeoutMiddleware.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * This adds its own timeout to the forward metadata and updates it's timeout
+ * based on the reverse metadata.
+ * @param ctx
+ */
+function timeoutMiddlewareClient(ctx: ContextTimed) {
+  const currentTimeout = ctx.timer.delay;
+  let forwardFirst = true;
+  let reverseFirst = true;
```
CMCDragonkai, line 69: What are these for? Should have comments describing the base case.

src/rpc/RPCClient.ts
```diff
@@ -39,12 +42,15 @@ class RPCClient<M extends ClientManifest> {
...
    * path and `Uint8Array` to `JSONRPCResponse` on the reverse path.
+   * @param obj.defaultTimeout - default timeout to be used if none is provided
+   * for a client call.
    * @param obj.logger
    */
   static async createRPCClient<M extends ClientManifest>({
     manifest,
     streamFactory,
     middlewareFactory = middlewareUtils.defaultClientMiddlewareWrapper(),
+    defaultTimeout = 60_000, // 1 minuet
```
CMCDragonkai: Typo here.

But this should be called `xTime`.

Note `defaultTimeout`.

This is because `timeout` as a word should be reserved for an actual timeout object. At least in our code here. See `src/network/` for examples for this.

This is technically the amount of time.

Also we can specific what is `x`.

Is it `streamKeepAliveTime`?

Because remember this time is different depending on unary call vs stream call. During stream call it's a "keep alive". During unary, it's applied once for the entire unary response.

Therefore you should call it `streamKeepAliveTimeoutTime`, we have something similar in `Proxy` as `connKeepAliveTimeoutTime`.

We can be more specific here, in case in the future we have other "times".

No need to specify default as that would be obvious here.

Also note that you don't document what happens when `defaultTimeout` is just `undefined`. You might want to indicate that having undefined means there's no default timeout at all.

src/rpc/RPCClient.ts
```diff
@@ -78,22 +86,24 @@ class RPCClient<M extends ClientManifest> {
...
+              this.serverStreamCaller(method, params, ctx);
           case 'CLIENT':
-            return () => this.clientStreamCaller(method);
+            return (ctx) => this.clientStreamCaller(method, ctx);
           case 'DUPLEX':
-            return () => this.duplexStreamCaller(method);
+            return (ctx) => this.duplexStreamCaller(method, ctx);
           case 'RAW':
-            return (header) => this.rawStreamCaller(method, header);
+            return (header, ctx) => this.rawStreamCaller(method, header, ctx);
```
CMCDragonkai, line 103: Can you confirm in your spec that timeout when used in streams is per-message? Specifically it should be per-chunk, since a message doesn't yet exist.

For any given chunk of data, the timeout should be reset.

src/rpc/RPCClient.ts
```diff
@@ -133,45 +146,87 @@ class RPCClient<M extends ClientManifest> {
...
-    await writer.close();
-
-    return callerInterface.readable;
+    try {
+      await writer.write(parameters);
+      await writer.close();
+      return callerInterface.readable;
+    } finally {
+      // Clean up if any problems, ignore errors if already closed
+      await writer.close().catch(() => {});
```
CMCDragonkai: Is it possible other exceptions may exist? If we can identify a specific type, it would be better to check that type before discarding it, otherwise throw upwards.

src/rpc/RPCClient.ts
```diff
@@ -185,31 +240,98 @@ class RPCClient<M extends ClientManifest> {
...
-        command: method,
-      });
-    const inputMessageTransformStream =
-      rpcUtils.clientInputTransformStream<I>(method);
-    const middleware = this.middlewareFactory();
+      rpcUtils.clientOutputTransformStream<O>(
+        {
+          nodeId: IdInternal.fromBuffer<NodeId>(Buffer.alloc(32, 0)), // FIXME
+          host: '',
+          port: 0,
```
CMCDragonkai: Why are these hardcoded?

src/rpc/RPCServer.ts
```diff
@@ -72,13 +76,17 @@ class RPCServer extends EventTarget {
       JSONRPCResponse
     >;
     sensitive?: boolean;
+    defaultTimeout?: number;
+    forceEndDelay?: number;
```
CMCDragonkai: Again here naming should end with `Time`. So it can be `endTime` or whatever.

src/rpc/RPCServer.ts
```diff
@@ -100,6 +111,8 @@ class RPCServer extends EventTarget {
     manifest,
     middlewareFactory,
     sensitive,
+    defaultTimeout = 60_000, // 1 minuet
+    forceEndDelay = 2_000, // 2 seconds
```
CMCDragonkai: Note that these defaults should be also reset inside `src/config.ts`...?

See the existing `proxyConfig`.

src/rpc/RPCServer.ts
```diff
@@ -351,56 +402,116 @@ class RPCServer extends EventTarget {
...
+      if (timeout != null && timeout < this.defaultTimeout) {
+        // Reset timeout with new delay if it is less than the default
+        timer.reset(timeout);
+      } else {
+        // Otherwise refresh
+        timer.refresh();
+      }
       const outputStream = handler(
-        [leadingMetadataMessage, inputStream],
+        [headerMessage.value, inputStream],
```
CMCDragonkai, line 518: Did we settle on `head, tail` as terminology?

src/rpc/errors.ts
```diff
@@ -94,6 +94,16 @@ class ErrorPolykeyRemote<T> extends ErrorPolykey<T> {
...
 
+class ErrorRpcStreamEnded<T> extends ErrorRPC<T> {
+  static description = 'Handled stream has ended';
+  exitCode = sysexits.NOINPUT;
+}
+
+class ErrorRpcTimedOut<T> extends ErrorRPC<T> {
+  static description = 'RPC handler has timed out';
+  exitCode = sysexits.UNAVAILABLE;
+}
```
CMCDragonkai: Should be `ErrorRPC...`.

Also note `exitCode`. If these are intended to be thrown to the CLI, then do they make sense here?

The `UNAVAILABLE` is pretty generic.

src/rpc/utils/utils.ts
```diff
@@ -487,6 +442,5 @@ export {
   toError,
   clientInputTransformStream,
   clientOutputTransformStream,
-  extractFirstMessageTransform,
```
CMCDragonkai, line 490: Was this moved somewhere?

src/websockets/WebSocketClient.ts
```diff
@@ -42,7 +45,7 @@ class WebSocketClient {
       port,
       maxReadableStreamBytes,
       expectedNodeIds,
-      connectionTimeout,
+      connectionTimeout ?? Infinity,
```
CMCDragonkai: So if we can default to `Infinity`? Perhaps then instead of `undefined`, you force it to be a number, but default it to `Infinity` as well for the RPC code.

And naming here should be consistent, again should be called `xTime`.

If it's a "keepalive" thing, we should consistently name it `xKeepAliveTimeoutTime`.

src/websockets/WebSocketClient.ts
```diff
@@ -368,8 +414,23 @@ class WebSocketStreamClientInternal extends WebSocketStream {
...
+    if (!this.readableEnded_) {
+      this.readableController?.error(err);
+      this.signalReadableEnd(err);
+    }
+    if (!this.writableEnded_) {
+      this.writableController?.error(err);
+      this.signalWritableEnd(err);
+    }
+    // Then close the websocket
+    if (!this.webSocketEnded_) {
```
CMCDragonkai: What is with these trailing underscores? Don't we use `this._x` for readable externally immutable properties? Make sure you're keeping the standard so I don't have to keep reminding here.

tests/rpc/RPCServer.test.ts
```diff
@@ -741,8 +741,341 @@ describe(`${RPCServer.name}`, () => {
       await rpcServer.destroy();
     },
   );
-  // TODO:
-  //  - Test odd conditions for handlers, like extra messages where 1 is expected.
-  //  - Expectations can't be inside the handlers otherwise they're caught.
-  //  - get the tap transform stream working
+  test('default timeout', async () => {
```
CMCDragonkai: Can we be more specific about this test description?

tests/rpc/RPCServer.test.ts
```diff
@@ -741,8 +741,341 @@ describe(`${RPCServer.name}`, () => {
...
+        timeout = 25;
+        public async handle(input: JSONValue, _, _ctx): Promise<JSONValue> {
+          ctxShortProm.resolveP(_ctx);
+          await waitProm.p;
+          return input;
+        }
+      }
+      const ctxLongProm = promise<ContextTimed>();
+      class TestMethodLongTimeout extends UnaryHandler {
+        timeout = 100;
```
CMCDragonkai, line 866: Overriding is setting the property?

tests/rpc/RPCServer.test.ts
```diff
@@ -741,8 +741,341 @@ describe(`${RPCServer.name}`, () => {
...
+    const then = Date.now();
+    expect(ctx.signal.reason).toBeInstanceOf(rpcErrors.ErrorRpcTimedOut);
+    // Should end after grace period
+    await expect(outputResult).rejects.toThrow();
+    expect(Date.now() - then).toBeGreaterThanOrEqual(100);
+    await expect(outputResult).toReject();
+    await rpcServer.destroy();
+  });
+  testProp(
+    'Middleware can update timeout timer',
```
CMCDragonkai: All test descriptions should be lowercase.

tests/rpc/utils/utils.test.ts
```diff
@@ -14,29 +12,6 @@ describe('utils tests', () => {
...
-      expect(await firstMessageProm).toStrictEqual(messages[0]);
-      expect(await AsyncIterable.as(parsedStream).toArray()).toStrictEqual(
-        messages.slice(1),
-      );
-    },
-    { numRuns: 1000 },
-  );
-
   // TODO:
   //  - Test for badly structured data
```
CMCDragonkai: Arbitrary todos here. All todos should be tickable tasks on the PR!

CMCDragonkai (changes requested):
First level review. Still need to go deeper in your logic that's in `rpc` and `websockets`.

src/bin/agent/CommandLockAll.ts
tegefaulkes: For the `streamFactory` we're passing a `ContextTimed`, the timer is used for timeout for the connection here since the websocket makes a new connection per stream. The signal is used to abort an active connection.

For the quic system only the signal will be used to abort an active stream. A timer could possibly be used to timeout the stream negotiation if the connection is not responding at this point but I'm unsure of the internal details here.

CMCDragonkai: But when timer is exhausted, the signal is aborted. So we only need to watch the signal.

But it seems you're also saying to watch it across the lifecycle of the stream too.

src/bin/identities/CommandPermissions.ts
tegefaulkes: We call whatever is creating a new stream, so yes, for the quic system we'd call the `streamNew` method. The life-cycle of the connection itself would be maintained separately here.

src/client/handlers/gestaltsGestaltList.ts
tegefaulkes: They're not being used here. I may have gotten the type of underscore mixed up.

Leading means unused and trailing means alternative/internal? I'll fix them up.

CMCDragonkai: So we have 3 naming conventions:

1. `this._x` means it is externally immutable, but internally mutable, this is usually combined with `get x(){ return this._x; }`.
2. `f(_a)` means un-used.
3. `a_` is alternative, because this is the most similar to `a'`, which is how haskell conventionally indicates an "updated" variable with the same meaning

src/client/handlers/vaultsCreate.ts
tegefaulkes: I intentionally avoided the default export here to enforce the name of the class when importing. Keep in mind, with default exports you can import them with any name, E.G. `import SomeRandomName from './VaultsCreate'` is valid. Since these handlers are imported in large blocks I wanted to avoid a situation where the import names could detach from the actual class name if we're not being careful.

src/client/utils/middleware.ts
tegefaulkes: That's what I've been doing, There are two main middlewares now, authentication and timeout. This `middleware` file just combines the two and is used as the single import for most use cases.

src/rpc/RPCClient.ts
tegefaulkes: I'm refreshing the timeout per message by design. It's too cheap for a bad actor to hold a connection open if it refreshes per chunk.

If I wanted to, even for a unary call I could send a single byte every 50ish seconds to hold a connection open. I wouldn't have to send a valid message at all to do this so long as it's valid JSON for parsing.  I could do this for a very long time until I hit the message parsing buffer limit.

Having it per message means you'd have to at least send valid messages and conform to the protocol of the RPC method you're using.

src/client/handlers/vaultsCreate.ts
CMCDragonkai: It was just convention when there's single class files. It's only lower case names when you intend to export other things. We use this standard elsewhere.

src/rpc/RPCClient.ts
CMCDragonkai: What about raw streams?

src/rpc/RPCClient.ts
tegefaulkes: It's in a finally block so I wanted to avoid duplicating the error if there was one. In fact on reviewing it, i'm being dumb here, I should be cleaning up the readable if there is an error with the writable here. I'll fix it.

src/rpc/RPCClient.ts
tegefaulkes: It's placeholder for some metadata used then creating errors. I haven't found a good way to get the information we want here. It's something I need to review and come back to.

src/rpc/RPCServer.ts
tegefaulkes: If I recall, you swapped the order so I propagated that change here. I don't think we decided to change the naming though.

src/rpc/utils/utils.ts
tegefaulkes: It was removed. I simplified the logic this was used in.

src/rpc/RPCClient.ts
CMCDragonkai: Not all errors should require a host and port then?

src/websockets/WebSocketClient.ts
tegefaulkes: You have me doubting myself now. I just checked other parts of the code and it looks like variables used internally that share a name use a postfix `something_`. Unused variables use a prefixed `_something`.

I'll review and fix any places that have this wrong.

tests/rpc/RPCServer.test.ts
tegefaulkes: Yes, the alternative is to set it during construction.

src/rpc/RPCClient.ts
tegefaulkes: I've fixed this up for the server streaming. Unary caller does the same thing however. 

Inside the try block we're closing the streams, these may throw errors but it's unlikely. In the finally we try the clean up but these will usually throw an error since the should already be closed in the try block. Any errors should come from inside the try block ,these are just here to double check that they closed in case of an error.

Maybe it could be done better?

src/PolykeyAgent.ts
tegefaulkes: I'm confused, it is called `rpcMiddlewareUtils`, do you mean the `clientMiddleware` below it should be called `clientMiddlewareUtils`?

src/client/handlers/gestaltsGestaltList.ts
tegefaulkes: I wasn't aware that 1. was a convention we were using but looking at the code again I can see that this is the case. I'll fix the usage.

tegefaulkes: This has been fixed.

src/client/handlers/gestaltsGestaltList.ts
tegefaulkes: I'll make the changes.

tegefaulkes: I've updated where relevant, The handlers will check for abortion just before yielding  and at the beginning of the handler.

src/client/handlers/gestaltsGestaltList.ts
tegefaulkes: Fixed.

src/client/handlers/identitiesAuthenticate.ts
tegefaulkes: Fixed.

src/PolykeyAgent.ts
CMCDragonkai: Oh I mean it should be `rpcUtilsMiddleware`. But actually the middleware should have separate structure later.

The name should match the directory path.

src/client/utils/timeoutMiddleware.ts
tegefaulkes: We only want to process the first message for the forward and reverse path. Theses are flags for that. I'll add comments.

src/client/handlers/gestaltsGestaltList.ts
CMCDragonkai: Why not propagate the `ctx` directly into `getGestalts` too? Or is that too much work atm?

src/client/utils/timeoutMiddleware.ts
tegefaulkes: Fixed.

src/rpc/RPCClient.ts
tegefaulkes: - We have a bunch of timeout related names in the domains `rpc` and `websocket`.
- `RPCClient` has `defaultTimeout` - This is the default timeout used if no timeout was provided for a client call.
- `RPCServer` has `defaultTimeout` - This is the default used if no timeout was provided for the handler definition. and `forceEndDelay` - Is the time between a handler timing out with signal which is advisory and when the stream is forced closed.
- `WebSocketClient` has `connectionTimeout`, `pingTimeout` and `pingInterval`.
- `WebSocketServer` has `idleTimeout`, `pingInterval` and `pingTimeout`. 

Changes
- `defaultTimeout` => `streamKeepAliveTimeoutTime`
- `forceEndDelay` => `timeoutForceCloseTime`
- `connectionTimeout` => `connectionTimeoutTime`
- `pingTimeout` => `pingTimeoutTime`
- `pingInterval` => pingIntervalTime`
- `idleTimeout` => `connectionIdleTimeoutTime`

CMCDragonkai: What's the difference between `connectionIdleTimeoutTime` and `pingTimeoutTime`?

CMCDragonkai: Also if `connectionTimeoutTime` exists on `WebSocketClient`, is the equivalent `connectionIdleTimeoutTime` on the `WebSocketServer`? If so, why not make these the same name?

tegefaulkes: This has been updated. I also clarified some of the doc comments.

src/rpc/RPCClient.ts
tegefaulkes: Raw streams doesn't support the refreshing. It will be up to the handlers to implement the refreshing logic. Mostly for the same reason above.

src/rpc/RPCClient.ts
tegefaulkes: All remote errors should, we'd want to know where the error came from. 

The problem here is, the information is supplied from the client side, but given that the `RPCClient` is agnostic to the transport. We don't have this information within the `RPCClient`, we shouldn't have it. so I don't have a nice way of supplying this information right now. 

Two ways of solving this come to mind.
1. The error revival could be done as a middleware, the metadata can be passed to the middleware and bypass the `RPCClient` internals that way.
2. The metadata can be supplied by the server side, but we'd have a similar problem, So it would have to be part of the middleware as well I think.

src/rpc/RPCServer.ts
tegefaulkes: This has been fixed.

src/rpc/errors.ts
tegefaulkes: Fixed. These are generic to client and server, So they could be throw on the CLI.

src/websockets/WebSocketClient.ts
tegefaulkes: This has been updated.

src/rpc/errors.ts
CMCDragonkai: What do you mean thrown on the CLI?

src/rpc/RPCServer.ts
CMCDragonkai: You should add these into `src/config.ts`.

src/rpc/RPCClient.ts
CMCDragonkai: > If I wanted to, even for a unary call I could send a single byte every 50ish seconds to hold a connection open. I wouldn't have to send a valid message at all to do this so long as it's valid JSON for parsing. I could do this for a very long time until I hit the message parsing buffer limit.

This isn't really an issue.

The purpose of timeouts are is not to deal with DOS. It's to deal with idle clients/idle servers.

What you're describing is perfectly fine.

src/PolykeyAgent.ts
tegefaulkes: This has been updated.

src/rpc/RPCClient.ts
tegefaulkes: Ok, I'll have to make some changes then.

src/websockets/WebSocketClient.ts
tegefaulkes: I've fixed this up.

src/config.ts
```diff
@@ -102,8 +102,8 @@ const config = {
       clientHost: '127.0.0.1' as Host,
       clientPort: 0 as Port,
       maxReadBufferBytes: 1_000_000_000, // About 1 GB
-      pingInterval: 1_000, // 1 second
-      pingTimeout: 10_000, // 10 seconds
+      pingIntervalTime: 1_000, // 1 second
+      pingTimeoutTime: 10_000, // 10 seconds
```
CMCDragonkai: Let's say that `networkConfig` is generic across transport options (websocket, quic).

While `rpcConfig` is specific to `src/rpc` domain.

Then `proxyConfig` gets removed once we have fully migrated.

src/PolykeyAgent.ts
```diff
@@ -457,13 +454,13 @@ class PolykeyAgent {
...
-          host: _networkConfig.clientHost,
-          port: _networkConfig.clientPort,
+          host: networkConfig_.clientHost,
+          port: networkConfig_.clientPort,
           tlsConfig,
-          maxReadBufferBytes: _networkConfig.maxReadBufferBytes,
-          idleTimeout: _networkConfig.idleTimeout,
-          pingInterval: _networkConfig.pingInterval,
-          pingTimeout: _networkConfig.pingTimeout,
+          maxReadableStreamBytes: networkConfig_.maxReadBufferBytes,
```
CMCDragonkai: This stuff should exist inside `src/config.ts`.

src/config.ts
CMCDragonkai: So the difference between having values here and the domain specific creators having default values, is that these defaults for the entire application, whereas the domains may have defaults specific to the domain.

The `PolykeyAgent` and `PolykeyClient` should correspondingly not have any default values.

src/bin/identities/CommandPermissions.ts
CMCDragonkai: The problem here is:
1. We want to be able to `readable.cancel` or `writable.abort` due to timer deadline which means the stream is destroyed due to idleness.
2. You cannot just use `transformStream.readable.cancel()` without ownership, if called it results in a locked error.
3. The `pipeTo` and `pipeThrough` does support the abort signal, however attempting to use this does not propagate the original reason across the composed streams, instead it would use an abortion error.

Therefore the solution is to have a way to abort the original stream directly.

2 options:

1. Put an `cancel` method into the a derived `ReadableWritablePair` type. We can call this `RPCStream` interface.
2. Demand that the `streamFactory` returns a `PromiseCancellable<[ReadableWritablePair, cancel]>`.

Option 1 is actually the same way we did this with the `Task` type in `src/tasks` domain.

Let's go with option 1.

Create a `RPCStream` interface type with the `cancel` method.

Update all the RPC usages of `ReadableWritablePair` to instead use `RPCStream`.

Then update the transport layer stream factories to produce `RPCStream`.

That also means that `streamFactory(ctx)` now can preserve the original meaning of `timedCancellable`.

That means you can use the `timedCancellable` HOF wrapper around `streamFactory`.

Furthermore the transport layer has to understand `ctx.signal` is used for the lifetime of the asynchronous creation, not the lifetime of the stream itself.

CMCDragonkai: Also because the context decorators are not a separate package. And I think there was a reason why this was the case, we cannot use the decorators, instead the HOF has to be used.

But `timedCancellable` HOF should give you back `PromiseCancellable`.

Meaning the `StreamFactory: (ctx: ContextTimed) => Promise<RPCStream<...>>`, because the wrapping happens at the RPC domain.

Note that using the `timedCancellable` allows you to configure the defaults during the wrapping.

src/bin/agent/CommandLockAll.ts
CMCDragonkai: See: https://github.com/MatrixAI/Polykey/pull/513#discussion_r1156712794

src/client/utils/timeoutMiddleware.ts
```diff
@@ -0,0 +1,93 @@
+import type { JSONRPCRequest, JSONRPCResponse } from '../../rpc/types';
+import type { ClientRPCRequestParams, ClientRPCResponseResult } from '../types';
+import type { ContextTimed } from '../../contexts/types';
+import { TransformStream } from 'stream/web';
+
+/**
+ * This adds its timeout to the reverse metadata and updates it's timeout based
+ * on the forward metadata.
+ */
+function timeoutMiddlewareServer(ctx: ContextTimed) {
```
CMCDragonkai: This is going to need to be abstracted out to a common middleware location. That could be `src/rpc/middleware`, that could be `src/middleware`.

I can see here you are using `ClientRPCRequestParams` here, but this could be a generic type parameter.

```
timeoutMiddlewareServer<Req, Res> {
  ...
}
```

Furthermore the generic timeout middleware should work even for raw streams. This is why I mentioned you should be able to reset the timer upon any data chunk.

CMCDragonkai: So because the transport layer already has timeout for idleness that tracks bytes being transferred, there's no need to have a timeout for raw streams.

Therefore in this scenario, this is a "RPC" level timeout which has to involve JSON because you have the `timeout` property in the params.

CMCDragonkai: So that means we have 2 layers of timeout.

The bottom layer is concerned with idleness at the networking packet level.

The top layer is concerned with application execution of the handlers.

CMCDragonkai: So the only thing needed here is to figure out how to abstract these generic middlewares between `client` and `agent` domains.

src/rpc/RPCClient.ts
CMCDragonkai: We can reuse the `RPCStream` to do this:

```ts
interface RPCStream extends ReadableWritablePair {
  cancel(reason?: any) => void;
  meta?: POJO;
}
```

Then this `streamFactory` which produces the `RPCStream` interface, is the only bridge between the transport domain and the RPC domain.

That's the only place to provide any data about the stream.

src/rpc/RPCServer.ts
```diff
@@ -56,12 +58,21 @@ class RPCServer extends EventTarget {
    * path and `JSONRPCResponse` to `Uint8Array` on the reverse path.
    * @param obj.sensitive - If true, sanitises any rpc error messages of any
    * sensitive information.
+   * @param obj.streamKeepAliveTimeoutTime - Time before a connection is cleaned up due to no activity. This is the
+   * value used if the handler doesn't specify its own timeout time. This timeout is advisory and only results in a
+   * signal sent to the handler. Stream is forced to end after the timeoutForceCloseTime. Defaults to 60,000
+   * milliseconds.
+   * @param obj.timeoutForceCloseTime - Time before the stream is forced to end after the initial timeout time.
+   * The stream will be forced to close after this amount of time after the initial timeout. This is a grace period for
+   * the handler to handle timeout before it is forced to end. Defaults to 2,000 milliseconds.
```
CMCDragonkai, line 66: The client is tracking whether the server is idle. The server is tracking whether the handler is idle.

So these 2 properties in the `RPCServer` is really about the handler.

You want to call these:

1. `handlerTimeoutTime` - when this is exhausted we notify the handler by calling the abort controller
2. `handlerTimeoutGraceTime` - this is about if the handler is not aborting, then we force "abort" by closing the streams

The idea here is to make it clear that these times are about the handler.

In both cases, you could have `Infinity` as a value.

src/client/handlers/vaultsCreate.ts
CMCDragonkai: We can deal with this later. It was because of legacy from GRPC.

src/rpc/RPCClient.ts
CMCDragonkai: Raw streams cannot have timeouts... because the transport layer is already doing this. So this is not relevant anymore. But we need comments inside explaining why.

CMCDragonkai: If you are resetting the timer every time you receive a message and every time you respond. In this case resetting before the response is sent.

You have to deal with the possibility that the other side is not consuming any of the data you send.

From the client's perspective the clock starts as soon as you attempt a call. Meaning that if the server is "SLOW" then eventually it will be aborted.

From the server's perspective, if you are responding with data, the client may be SLOW. Then in that case, you want to make sure that the time spent responding is also tracked.

CMCDragonkai: We should do some tests that test race conditions here.. just to see how client timer and server timer interacts with large request and large responses. Imagine the request/response messages are the size of 1 MiB.

Message size should be considered here.

What we want to prevent is memory overflow here. Even in the case of raw streams.

Can you make sure the buffer size for JSON parsing is part of the `src/config.ts` as well.

src/rpc/RPCClient.ts
CMCDragonkai: Every time there is an empty catch, we must have a comment explaining why. And it should point to later parts of the code which does either handle the exception for some reason.

tests/rpc/RPCServer.test.ts
CMCDragonkai: The `timeout` property for `abstract class Handler` should have a docblock explaining what it is used for when used with downstream classes.

tests/rpc/utils/utils.test.ts
CMCDragonkai: These should be issues.

CMCDragonkai:
Once all comments resolved you can proceed to merge then we start on QUIC.

src/bin/identities/CommandPermissions.ts
tegefaulkes: This change should work. Only concern now is that we have two timeouts, One handled by the TimedCancelable logic for creating the stream, and a separate one for the refresh.

I'm not sure how to get the keep alive timeout. Do I provide a separate option for a keep alive timeout timer separate from the CTX?

CMCDragonkai: The first timeout is for acquiring the stream.

The second timeout is to keep the stream alive.

Note that when a RPC call is made, you have a timeout for the entire operation. You pass the same timer down to acquisition of the stream... then the second timeout for the entire keepalive. So from the client's perspective, these are just 2 operations that could happen.

On the server side, you only have the second timeout, the first timeout does not ever exist.

tegefaulkes: As discussed. The timers will be shared for keep alive and stream timeout. So same as currently designed.

tegefaulkes: This has been resolved.

src/rpc/RPCClient.ts
tegefaulkes: I've added comments for the `.catch()` usage.

src/rpc/RPCClient.ts
tegefaulkes: Metadata such as this is provided by the `RPCStream` metadata now. This is done for both the server and client.

tests/rpc/RPCServer.test.ts
tegefaulkes: A comment has been added.

src/client/utils/timeoutMiddleware.ts
tegefaulkes: I'll add some details to the agent migration issue. this will be addressed there.

src/rpc/RPCServer.ts
tegefaulkes: This has been fixed.

src/PolykeyAgent.ts
tegefaulkes: Any missing config has been added and names have been updated.

src/rpc/RPCServer.ts
tegefaulkes: They have been added.

src/rpc/RPCClient.ts
tegefaulkes: Buffer limit for parsing has been added to the config.

As for testing race conditions, I don't see the need right now. The difference between the client timing itself out or the server doing it is pretty minor. If the client times itself out then the stream ends with a timeout error.  If the server does it then the client sees the connection ending early and the streams end with a different error, connection ended early I think.

I could add more informative websocket stream ending with error codes. But I don't see the need right now. I can create issues for both timeout race testing and more informative error codes for websockets ending.

tests/rpc/utils/utils.test.ts
tegefaulkes: Fixed.

src/client/utils/timeoutMiddleware.ts
tegefaulkes: I mentioned some details in the agent migration issue #512

tegefaulkes:
One review comment left. I don't feel it needs to be addressed in this PR. It is non-blocking and non-critical.

I'll start on merge prep.

src/rpc/RPCClient.ts
CMCDragonkai: > As for testing race conditions, I don't see the need right now. The difference between the client timing itself out or the server doing it is pretty minor. If the client times itself out then the stream ends with a timeout error. If the server does it then the client sees the connection ending early and the streams end with a different error, connection ended early I think.

Can you make sure there are tests covering when the client timesout VS when the server timesout. Please give the name of the test here.

> I could add more informative websocket stream ending with error codes. But I don't see the need right now. I can create issues for both timeout race testing and more informative error codes for websockets ending.

This needs to be designed along with the mapping functions use in js-quic. Can you point out where in our spec that error codes is being specified?

tegefaulkes: RPCServer has two tests, if the stream fails and plenty of timeout tests.
* `'should emit stream error if input stream fails'`
* `'should emit stream error if output stream fails'`

The `RPCClient` doesn't have any tests. But it doesn't do anything complex here. if it's timing out then it cancels the stream, if the stream ends early then the stream throws. Both cases is the stream throwing an error. 

I don't expect any odd interactions happening depending on race conditions with timeouts. So I didn't write any integration tests with that in mind. 

As for the WS error codes. It's not in the spec. There is only 1 code I'm using `4001` as a generic catch all. We don't use the closing frames to convey any errors. the only time it would contain messages is if we ended the connection abruptly and we only use that information for debug messages.

We could expand this feature out but I don't think it's needed.  It's also out of scope for this PR.

tegefaulkes:
This is ready for merge.

src/rpc/RPCClient.ts
CMCDragonkai: Can you change to 4000 as the generic error code. 4001 and above should be specific codes. This matches the `0` being used in QUIC.

CMCDragonkai:
Last commit is `fix: test fixes`, is that something that can be better descriptive or squashed?

CMCDragonkai:
There's a few wip commits. Can these be squashed? And also a `fixin` commit which I don't understand what that is.

tegefaulkes:
Oh I missed those. let me fix them. 

A `fixin` commit is just a name I use for commits that need to be squashed with the commit next to it.

tegefaulkes:
Fixed up and ready to merge.

CMCDragonkai:
Nice!

CMCDragonkai:
Hey I think `fix: test fixes` should have a test scope in the future.

PR #430, opened 2022-07-26
Fixing `swc` specific bug with decorators and constructors

tegefaulkes:
### Description

Due to a difference with swc on how decorators are handled. Classes were constructed without decorators applied when called inside of `createX` static functions. This was solved by using `new this()` to construct instead.

Fixed up some linting errors as well. Some  `// @ts-ignore` lines were added without descriptions.

### Tasks

- [x] 1. Use `new this()` to construct the class inside of `createX` static methods.

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [x] Updated inline-comment documentation
* [x] Lint fixed
* [x] Squash and rebased
* [ ] Sanity check the final build

PR #525, opened 2023-05-23
`js-quic` integration and Agent migration

tegefaulkes:
### Description

This PR Does 3 things.
1. Migrates agent service to the agnostic RPC system.
2. integrates `@matrixai/js-quic` into `Polykey`.
3. Removes all old agent and network code. This includes GRPC and network domain

* Related MatrixAI/Polykey#512
* Related MatrixAI/Polykey#495
* Related MatrixAI/Polykey#527 

### Issues Fixed

* Fixes MatrixAI/Polykey#512
* Fixes MatrixAI/Polykey#234

### Tasks

* [x] 1. nodes domain needs changes
    * [x] 1. `NodeConnection` needs to be gutted and replaced with `RPCClient` and `QUICClient` usage. Besides this, usage of the `NodeConnection` is mostly the same.
    * [x] 2. Tests need to be updated
* [x] 2. Verification logic needs to be transplanted for use with quic.
    * [x] 1. This needs to be tested.
* [x] 3. Ensure that the proper connection information is provided by the streams from the quic system.
    * [x] 1. This needs to be tested.
* [x] 4. `PolykeyAgent` needs to be updated
   * [x] 1. GRPC agent server needs to be replaced with a `RPCServer` and `QUICServer` combo
   * [x] 2. `Proxy` needs to be removed.
   * [x] 3. Configs need to be updated and fully propagated.
* [x] 5. `Agent` domain needs to be migrated to using the agnostic RPC code.
   * [x] 1. Tests need to be migrated
* [x] 6. Old code needs to be removed
    * [x] 1. network domain gutted
    * [x] 2. GRPC domain gutted
    * [x] 3. Remove protobuf? and other package dependencies that are not used anymore.
* [x] 7. tests!
    * [x] 1. `network` domain tests need to be removed, any tests still needed should be transplanted.
    * [x] 2. `grpc` domain tests need to be removed, any tests still needed should be transplanted.
* ~[ ] 8. Update relevant handlers with pagination~ - Differed to stage 2 PR
* ~[ ] 9. Update agent handlers to be timed cancellable, implement cancellation.~ - Differed to stage 2 PR

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Updated inline-comment documentation
* [x] Lint fixed
* [x] Squash and rebased
* [x] Sanity check the final build

Issue #268, opened 2021-10-26
Publishing the PK CLI for General Use


CMCDragonkai:
### Requirements of this design

The new structure of our projects is:

Polykey 
 - js-polykey renames to Polykey
 - keeps all issues in js-polykey
 - license: GPL3 (possible move to APL3)
 - npm: polykey
 - not hosted on nix (it's not an application)
 - supports network API
 - embeddable as a library
 - bin/ code is extracted to Polykey-CLI
 - maintains wiki: theory, reference, guides, tutorials
 - maintains community discussions
 - main README.md (open source entry point)
 - requires cross-platform gitlab runner for building and QA of possibly-native addons

Polykey-CLI
 - license: GPL3
 - npm: polykey-cli
 - nix: polykey-cli
 - extracts bin/ code out of js-polykey
 - requires cross-platform gitlab runner for building and QA

Polykey-Desktop - electron/tauri wrapper around library code
 - license: GPL3
 - not hosted on npm (it's not a JS library or application)
 - nix: polykey-desktop
 - requires cross-platform gitlab runner for building & QA

Polykey-Mobile - nativescript wrapper around library code
 - license: unknown
 - not hosted on npm (it's not a JS library or application)
 - not hosted on nix (it's not a desktop application)
 - requires cross-platform gitlab runner for building & QA

We may not need to split out to Polykey-CLI just yet, but it can be done later.

### Additional context

* #390 - this should all be done automatically in the CI/CD
* #426 - which setup the repository structure
* https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

### Specification

1. [x] - Undeprecate the NPM `polykey`, as Polykey Core library now uses `polykey`
2. [x] - Publish to `polykey`
3. [x] - Deprecate `@matrixai/polykey` and make it point to `polykey` and make the deprecated `js-polykey` and make it point to `polykey`
   * https://www.npmjs.com/package/@matrixai/polykey
   * https://www.npmjs.com/package/polykey
   * https://www.npmjs.com/package/js-polykey

CMCDragonkai:
### Review CLI exit codes in relation to sysexit standards

Exceptions are currently using sysexit codes from this stanard: https://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html

It's a good idea to review these exit codes and see how to ensure we all of our CLI commands are returning the proper exit codes.

[...]

PR #529, opened 2023-07-04
Swapping to SWC, Updating to M1 MacOS, and updating dependencies

### LINKED ISSUE #528: Update MacOS GitLab CI/CD configuration by 31st May
The beta macos machines will be deprecated by 31st May and from then it will be running on the M1 architecture, and will require us to have premium subscription to GitLab again.

See: 

* https://github.com/MatrixAI/js-quic/issues/7#issuecomment-1551451917
* https://gitlab.com/gitlab-com/runner-saas-macos-access-requests/-/issues/233#note_1394334519

Note that if we were to go towards full cross-compilation we may avoid this. But since we have other code like js-db and later that may still use node-gyp we're going to need it anyway.

Do note that M1 and x64 macos is capable of cross compiling architectures. We don't bother with the universal binary cause it's not needed, we are able to publish separate optional packages atm. That way `lipo` is avoided. That does mean though... when distributing PK, we have to distribute an M1 version and a x64 version. The x64 versions are likely to be deprecated in the following years.

### Tasks

- [x] 1. Use `saas-macos-medium-m1` as the tag.
- [x] 2. Use `macos-12-xcode-14` as the image.
- [x] 3. Make sure to update the `systeminformation` to `5.18.5` if it is being used in benchmarking
- [ ] 4. Test that build still works fine on M1

CMCDragonkai:
### Description

This PR is for switching to using `@swc/jest`.

I think this was a dangling branch, but has a change that we want to make use of now that everything is now using `swc`.

However there are some things in this change that still need testing like the removal of jose and babel-jest.

### Issues Fixed

* Related #430 
* Fixes #528

### Tasks

- [x] 1. Update all MatrixAI dependencies to the latest that has been released
- [ ] 2. Test swc tests work
- [ ] 3. Remove or simplify tests that are not required

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
@tegefaulkes this should be merged first... after completing the tasks.

CMCDragonkai:
The SWC may still have some weird decorator problem but we don't actually use it, so if we have a test for that, we can just disable for those.

This PR can be used to sort out #528 too.

Although atm due to problems compiling js-quic on Windows, we may have some problems releasing a build for Windows.

At any case we will focus on just Linux for now.

CMCDragonkai:
@tegefaulkes focus on this one first. Get the SWC into it, and update all relevant dependencies before `js-quic`. Then merge this into staging.

Then in #525, you can rebase that on top of this and bring in the 0.0.10 js-quic and start integrating that.

We will then address #268 and move all the of the CLI related things to Polykey-CLI so we can get the CI/CD to pass here since the pkg has problems building with the latest nodejs and we can focus in on packing and bundling in Polykey-CLI.

tegefaulkes:
I'm going to leave updating dependencies to later. They include breaking API changes and I'd rather have fixes included in the the commit. But at the same time I don't want to re-base on top of any fixes that need to be done.

CMCDragonkai:
The only breaking change should be the js-db and js-async-locks when it comes to timeouts...

CMCDragonkai:
Also this will break the CI build due to https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/94#issuecomment-1611154454.

We have to accept the breaking for now until all CLI stuff moves to Polykey-CLI.

tegefaulkes:
I can't fully tests things are working since the state of the staging branch is between the client and agent migration. I'll have to do any testing and fixes in the agent migration branch after rebasing it on top of this.

CMCDragonkai:
Yes that's fine, it just has to be merged to staging, you should do some sanity check though.

tegefaulkes:
Alright, I'm merging this now. given the incomplete state of staging I can't run any tests to fully check if all the changes work. That will be done in #525 after rebasing it.

PR #535, opened 2023-07-20
Agent migration stage 2

### LINKED ISSUE #2: RPC Raw Streams for Binary Data
### Specification

RPC methods that need to send binary data such as secrets data for the client RPC or GIT pack data for the agent RPC. This needs to make use of the raw streams in the RPC system. 

Raw streams take a header message and then act as a normal binary stream. The header can include any metadata you need. The binary data can then be streamed. To ensure there are no problems with Endianness we can chunk the data into protobuf messages.

Some things to note
1. Raw streams don't have a return header message/metadata currently. It could be implemented.
2. Raw streams don't support middleware.
3. If we use protobuf messages then we need a standard way of delimiting messages in the stream.

See https://github.com/MatrixAI/js-rpc/issues/2 for clarification on the usage of binary data formats like CBOR or Protobuf. It's only needed in limited usecases, and can be done as part of the RPC caller and handler.

When dropping to raw streams there should still be an exchange protocol between client and server.

The client sends a "header" message, and then should expect the server to send back a "header" message. Both should be in JSON. Only then the client can drop down to binary. However this can be just one of the ways of using it. This primarily applies to streaming: client streaming, server streaming and duplex streaming.

For example:

* In client streaming, the client sends 1 header message, then starts sending binary data. The server then sends back a JSON message (possibly at the end or at any time?)
* In server streaming, the client sends 1 header message, the server send backs 1 header message, then drops down to binary.
* In duplex streaming. The client sends a header message. The server responds with a header message. Both client and server can start sending binary data (at any time?).

It seems we should enforce that there be a header message sent, even if it is empty like a `{}`.

### Additional context

* Related https://github.com/MatrixAI/Polykey/issues/495#issuecomment-1498472431
* Related MatrixAI/Polykey#495 
* Related MatrixAI/Polykey#509 
* Related MatrixAI/Polykey#512 

### Tasks

1. Spec out a standard way of sending raw file/binary data.
2. Update all relevant handlers to use raw streaming for sending this data.

### LINKED ISSUE #511: Updating `TLSconfig` for `WebSocketServer`
### Specification

When the root key pair changes we need to update anything that depends on this. This change event is propagated through the events system. One of the changes is updating the TLS config the RPC servers are using. Previously this could be done dynamically but now with the `WebSocketServer` this can only be set when starting it.

This means that the `WebSocketServer` needs to be restarted when this event happens. This means any active connections are potentially interrupted. 

To support this we need to make the following changes.
1. When the root keypair is updated we need to restart the `WebSocketServer`.
2. When restarting the `WebSocketServer` we need to gracefully end any active connections.
3. We need to propagate an abort signal to the stream handler to signal ending.
4. expand the force stop options to allow for forcing connections to end, Signalling connections to end gracefully or just wait for all active connections to end.
5. Add a feature to block all new connections so when waiting for connections to end new ones can't start.

### Additional context

* Related #509 

### Tasks

1. When the root keypair is updated we need to restart the `WebSocketServer`.
2. When restarting the `WebSocketServer` we need to gracefully end any active connections.
3. We need to propagate an abort signal to the stream handler to signal ending.
4. expand the force stop options to allow for forcing connections to end, Signalling connections to end gracefully or just wait for all active connections to end.
5. Add a feature to block all new connections so when waiting for connections to end new ones can't star

### LINKED ISSUE #527: `NodeConnectionManager` manages remote initiated connections
### Specification

With the creation and integration of `@matrixai/js-quic`, the concept of a connection is now generic to `forward` and `reverse` initiated connections. This means, no matter who initiated the connection, streams can be created in either direction.

Previously the `NodeConnectionManager` only managed `forward` initiated connections. Now it should be able to handle reverse connections and add them to the connection map. This is essential for re-using connections efficiently.

The `NodeConnection` needs to be updated to be able to wrap `QUICConnection`s coming out of the `QUICServer`.  This will be a new async creation method. The cleanup logic will need to switch based on if it was a forward connection with a `QUICClient` or a reverse connection.

The `QUICServer` needs to be moved into and managed by the `NodeConnectionManager`. It's lifecycle will be fully managed. I don't know if we want to be able to DI it.

Reverse connections coming out of the `QUICServer` needs to be handled, wrapped in a `NodeConnection` and added into the connection map. There may be some complexities with locking here, that still needs to be prototyped.

All reverse stream events coming out of the `QUICConnectons` need to be handled by the `RPCServer` via a handleStream callback.

### Additional context

* Related #525

### Tasks

* [x] 1. Update `NodeConnection` to wrap reverse created `QUICStreams` and handle clean up in all cases. `QUICClient` can be optional and needs to be handled.
* [x] 2. `QUICServer` needs to be moved and managed by `NodeConnectionManager`
* [x] 3. Reverse connections coming from the `QUICServer` needs to be handled and put into the connection map.
* [x] 4. Reverse streams from all `QUICConnection`s needs to be handled by a `RPCServer` via a `handleStream` callback. 
* [x] 5. Tests need to be expanded and updated to handle new behaviour

### LINKED ISSUE #540: Replace `uws` with `ws`
### Specification

After some discussion we've concluded that `uws` is not fit for use. We should look into replacing the server side websocket code with `ws`.

This needs to be further specced out.

### Additional context

* Related MatrixAI/Polykey#511 
* Related MatrixAI/js-ws#4 

### Tasks

1. Replace `uws` with `ws`.

tegefaulkes:
### Description

The following tests were disabled in MatrixAI/Polykey#534 and need to be fixed and re-enabled here.
* validation/index.test.ts:372  'manipulate `this` when using function expressions'
* vaults/VaultInternal.test.ts:187  'can change commits and preserve the log with no intermediate vault mutation'
* vaults/VaultInternal.test.ts:236  'can change to the latest commit'
* vaults/VaultInternal.test.ts:427  'concurrent read operations allowed'
* vaults/VaultManager.test.ts:473  'with remote agents'

Related issues 

- Related: MatrixAI/js-rpc#2 
- Related: MatrixAI/js-rpc#5 
- Related: MatrixAI/Polykey#237 
- Related: MatrixAI/Polykey#527 
- Related: MatrixAI/Polykey#400
- Need to complete the vaults changes after issues have been addressed.
- Need to complete final fixes and polish to tests.

### Issues Fixed

* Fixes MatrixAI/Polykey#527
* Fixes MatrixAI/Polykey#540 
* Fixes MatrixAI/Polykey#511 
* Fixes https://github.com/MatrixAI/js-rpc/issues/2

### Tasks

- ~1. MatrixAI/js-rpc#5~
- ~2. MatrixAI/Polykey#237~
- [x] 3. MatrixAI/Polykey#527 
- [x] 4. MatrixAI/Polykey#540 
- [x] 5. MatrixAI/Polykey#511 
- ~6. MatrixAI/Polykey#461~
- [x] 7. Remove `commander` dependency as it is no longer used.
- [x] 8. vault RPC error management and streaming, Adding leading response metadata to RPC streams. and cleaning handlers
- ~9. F&F signalling. Don't wait for signalling to complete before ending RPC call. Also upgrade signalling to coalesce signalling attempts. Also have the relaying node fill in all connection information.~
- ~10. MatrixAI/Polykey#537 Local discovery with MDNS and node graph expansion to handle extra metadata.~

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
https://github.com/MatrixAI/Polykey/pull/525#discussion_r1272025906 needs to be addressed here.

tegefaulkes:
ICE still needs to be re-implemented here. I've already marked out in the code where it needs to be.

CMCDragonkai:
What happened to the previous ICE?

tegefaulkes:
There was some refactoring of the connection code. Mostly moving multi connection logic inside of the `NodeConnectioManager`. I just didn't implement it yet and left it for stage 2 PR.

tegefaulkes:
I'm back onto this now. First step is to address MatrixAI/Polykey#527. Then to remove `uWebsockets` in MatrixAI/Polykey#540.

tegefaulkes:
MatrixAI/Polykey#527 is done.

Looking at either removing uws or fixing vaults cloning/pulling next. I think removing `uws` may be more critical to getting CLI and testnet up and running.

CMCDragonkai:
I suggest when working on ws to factor out to js-ws.

CMCDragonkai:
Just wanted to mention this. There's no issue tracking this atm, because https://github.com/MatrixAI/js-rpc/issues/2 is actually a much larger issue addressing all forms of binary usage including error management.

Instead, the only thing that needs to be done here is ensuring that there is leading message on both ends even when dropping to raw streams.

This should be done as part of fixing up the vault/git RPC calls.

I don't think it's necessary to create an issue for this... since it's already part of fixing up vaults work.

tegefaulkes:
The path


1. vault RPC error management and streaming, Adding leading response metadata to RPC streams. and cleaning handlers
2. FAF signalling. Don't wait for signalling to complete before ending RPC call. Also upgrade signalling to coalesce signalling attempts. Also have the relaying node fill in all connection information.
3. Local discovery with MDNS and node graph expansion to handle extra metadata.
4. Getting CLI building. Minimally need the docker image working, then PKG working, @CMCDragonkai is working on this.
5. initial testnet deployment, deploying the test seednodes and stability testing
6. Running NAT punch through tests, and general network integration testing.

CMCDragonkai:
I'm putting ESM on the backburner for now due to problems with workers. It's quite likely we will need to entirely replace the threadsjs with our own implementation. So for now there's quite a few libraries that are fully ESM native, you can't really upgrade to them atm (actually I'm not entirely sure what happens if a CJS project tries to import it, the goal was to be fully ESM).

Instead I'm going to focus on CLI review. The priorities will be local traversal, solidifying the design of the signalling, CLI CI, testnet testing.

We can come back to the ESM stuff soon, and hopefully get a better build.

tegefaulkes:
Added a reverse response message to RPC raw streams. This is provided as part of the return for the raw handler. with the return type as `[JSONValue, ReadableStream<Uint8Array>]`. The JSONValue provided is the leading response message for that stream. The handler can now also throw, in that case the error will be converted to an RPC error message which will be thrown on the caller side.

CMCDragonkai:
Can you abstract around the `QUICConfig` type? https://github.com/MatrixAI/Polykey-CLI/pull/4#issuecomment-1678490277

PK CLI shouldn't be depending on that directly.

CMCDragonkai:
PK should be exporting any necessary functions for creating NodeIds: https://github.com/MatrixAI/Polykey-CLI/pull/4#issuecomment-1678493116

Currently the PK CLI does this:

```ts
const invalidNodeId = IdInternal.fromString<NodeId>('INVALIDID');

function generateRandomNodeId(): NodeId {
  const random = keysUtils.getRandomBytes(16).toString('hex');
  return IdInternal.fromString<NodeId>(random);
}
```

This sort of code seems like something that PK can provide. PK CLI then doesn't need `@matrixai/id` directly.

The generation of NodeID is used in these tests:

```
»» ~/Projects/Polykey-CLI/tests
 ♖ ag 'generateRandomNodeId'                   ⚡(staging) pts/5 16:45:35
vaults/vaults.test.ts
26:  const nodeId1 = testUtils.generateRandomNodeId();
27:  const nodeId2 = testUtils.generateRandomNodeId();
28:  const nodeId3 = testUtils.generateRandomNodeId();
411:          const targetNodeId = testUtils.generateRandomNodeId();
457:        const targetNodeId = testUtils.generateRandomNodeId();
536:        const targetNodeId = testUtils.generateRandomNodeId();

utils.test.ts
94:      const nodeId = testUtils.generateRandomNodeId();

nodes/add.test.ts
17:  const validNodeId = testUtils.generateRandomNodeId();
```

I suggest that...

1. PK should expose a function that creates a `NodeId`.
2. It should be able to take in string data.
3. Maybe `toNodeId`? Is there an existing parser for this? See the validation standard https://github.com/MatrixAI/MatrixAI-Graph/issues/44
4. It should be available in `nodes/utils`, so I can use it in the tests in PK CLI.

CMCDragonkai:
The `commander` package is no longer used in PK. It can be removed too.

The `@types/nexpect` is no longer used in PK, it can be removed.

CMCDragonkai:
I'll make some changes to `nodes/utils` to get the right function for `@matrixai/id`. 

And see what I can do for the QUIC type... and you can rebase on top.

But you should remove `commander` since it will change the `package-lock.json`.

CMCDragonkai:
Functions that is in PK that produce `NodeId`:

* `parseNodeId` - `ids/index.ts` - calls `decodeNodeId`
* `decodeNodeId` - `validation/utils.ts` - this decodes a multibase base32 hex encoding public key
* `publicKeyToNodeId` - `keys/utils/asymmetric.ts` - this takes a Ed25519 Public Key buffer to the `NodeId`

Whereas... the function used in PK CLI tests is basically just:

```
IdInternal.fromString<NodeId>('...somestring...');
```

This `string` is expected to be the random binary string. The opposite of `String.fromCharCode`. The binary version of some buffer.

It looks like this:

```
> x = new Uint8Array([0, 0, 0])
Uint8Array(3) [ 0, 0, 0 ]
> String.fromCharCode(...x)
'\x00\x00\x00'
```

If the string is logged out, it is not necessarily printable characters.

---

I could try changing the tests so that instead of doing that, it would use the `decodeNodeId` which then expects a multibase base 32 hex encoded public key.

However I think it is more efficient to provide a function that simply takes a random 32 byte string and produces a `NodeId` from that.

We can easily do something like:

* `toNodeIdString`
* `fromNodeIdString`

To address MatrixAI/Polykey#532, these 2 functions should exist in `ids/index.ts`. In fact `ids/index.ts` already has `createXIdGenerator` variants, which can take over exactly what we expect to our tests.

CMCDragonkai:
I'm going to do `encodeNodeIdString` and `decodeNodeIdString`, because we actually do have a type for this called `NodeIdString`.

CMCDragonkai:
Ok regarding network configuration, I'm going to need to do on this branch.

I'm proposing something like this:

```ts
    networkConfig: {
      /**
       * Agent host defaults to `::` dual stack.
       * This is because the agent service is supposed to be public.
       */
      agentHost: '::',
      agentPort: 0,
      /**
       * Client host defaults to `localhost`.
       * This will depend on the OS configuration.
       * Usually it will be IPv4 `127.0.0.1` or IPv6 `::1`.
       * This is because the client service is private most of the time.
       */
      clientHost: 'localhost',
      clientPort: 0,
      /**
       * If using dual stack `::`, then this forces only IPv6 bindings.
       */
      ipv6Only: false,

      /**
       * Agent service transport keep alive interval time.
       * This the maxmum time between keep alive messages.
       * See the transport layer for further details.
       */
      agentKeepAliveIntervalTime: 10_000, // 10 seconds

      /**
       * Agent service transport max idle timeout.
       * This is the maximum time that a connection can be idle.
       * See the transport layer for further details.
       */
      agentMaxIdleTimeout: 60_000, // 1 minute

      /**
       * Client service transport parameters.
       */
      clientMaxReadableStreamBytes: 1_000_000_000, // About 1 GB
      clientMaxIdleTimeout: 120, // 2 minutes
      clientPingIntervalTime: 1_000, // 1 second
      clientPingTimeoutTimeTime: 10_000, // 10 seconds

      /**
       * Controls the stream parser buffer limit.
       * This is the maximum number of bytes that the stream parser
       * will buffer before rejecting the RPC call.
       */
      clientParserBufferByteLimit: 1_000_000, // About 1MB
      clientHandlerTimeoutTime: 60_000, // 1 minute
      clientHandlerTimeoutGraceTime: 2_000, // 2 seconds
    },
```

The client parameters aren't entirely properly namespaced, and I want to remove the `clientMaxReadableStreamBytes` since MatrixAI/js-ws#3 is also about introducing backpressure to our websocket system.

Right now until `js-ws` has been factored out, and since we're just using `ws`, it's fine to enable it. Once browser `WebSocket` is needed for 7th testnet deployment, then we will figure out how that works in the browser.

CMCDragonkai:
Rebased on top. Remember to sync up @tegefaulkes tomorrow.

CMCDragonkai:
There are 2 types in `nodes/types.ts` that is `QuicConfig` and `QUICClientConfig`.

The `QUICClientConfig` isn't used at all, and I'm removing it right now.

The `QuicConfig` type is being used... but it should not be.

Now that `NodeConnectionManager` is encapsulating the QUIC transport layer (which I wasn't really sure if that was the best idea), the `NodeConnectionManager` must present its own configuration self-standing, it cannot be propagating implementation details. Encapsulating QUIC means that QUIC is now an implementation detail.

CMCDragonkai:
I've pushed WIP commits for the above removal of configuration. It needs further work in `NodeConnectionManager`.

The way I think about it, if the transport layer of QUIC is inside `nodes` domain then the same thing has to happen to the WS for the client service. But if there isn't an equivalent, it sounds like this is something that should be maintained separately and have a separate lifecycle.

Not sure, need to examine the rest of this.

CMCDragonkai:
1. Change NCM to wrap over quic configuration. QUIC configuration is an implementation detail due to encapsulation.
2. Try spreading QUIC config required parameters (and give them renaming to meet the standard if necessary) into the NCM creation POJO.
3. Refer to my changes in the `config.ts` network changes for default parameters. I've changed `clientHost` to be `localhost` and `agentHost` to `::`. Can you please test this to see if this works properly upon startup?
4. Tests should use `encodeNodeIdString` or `decodeNodeIdString` rather than directly using `IdInternal`.
5. Please do a prepatch release as soon as possible.

CMCDragonkai:
I've updated the `src/index.ts` to export everything. Except `src/websockets` since @amydevs will be factoring that out soon.

The `src/rpc` might be factored out but I suspect it will still have stuff like middleware specific to Polykey.

CMCDragonkai:
I've removed some vestigial env variables that are all migrated to `Polykey-CLI`.

I noticed that there is some weird env variables being used in the websocket tests.

```
[nix-shell:~/Projects/Polykey]$ ag 'PK_TEST'
tests/websockets/WebSocket.test.ts
332:          PK_TEST_HOST: host,
333:          PK_TEST_PORT: `${webSocketServer.getPort()}`,
334:          PK_TEST_NODE_ID: nodesUtils.encodeNodeId(nodeId),
375:          PK_TEST_KEY_PRIVATE_PEM: tlsConfig.keyPrivatePem,
376:          PK_TEST_CERT_CHAIN_PEM: tlsConfig.certChainPem,
377:          PK_TEST_HOST: host,

tests/websockets/testClient.ts
17:    expectedNodeIds: [nodesUtils.decodeNodeId(process.env.PK_TEST_NODE_ID!)!],
18:    host: process.env.PK_TEST_HOST ?? '127.0.0.1',
19:    port: parseInt(process.env.PK_TEST_PORT!),

tests/websockets/testServer.ts
18:    keyPrivatePem: process.env.PK_TEST_KEY_PRIVATE_PEM as PrivateKeyPEM,
19:    certChainPem: process.env.PK_TEST_CERT_CHAIN_PEM as CertificatePEMChain,
25:    host: process.env.PK_TEST_HOST ?? '127.0.0.1',
```

Never use env variables in this way, it creates alot of problems down the line.

@amydevs when extracting the websocket tests to js-ws, remove any usage of env variables like this. Env variables must always be pushed in through the `jest.config.js` gateway.

tegefaulkes:
Almost done now. the vaults handlers have been converted. the `vaultsGitPackGet` has some weird stuff with the git utils but that's out of scope for this.

Things are mostly working but the newly enabled tests are revealing a bug with `js-quic` and cancelling streams so i'll need to investigate that. Cancelling results in a `never()` error in some cases.

After that I can merge this PR and move the remaining issues to a new PR.

CMCDragonkai:
Tasks 10, 9, 2, 1 could all be separate PRs at your convenience.

CMCDragonkai:
MatrixAI/Polykey#552 is branched off from this because it's quite a major change. It's necessary to be able to complete https://github.com/MatrixAI/Polykey-CLI/pull/4#issuecomment-1680131511.

So this PR can focus on finishing `8.` to merge. 6. and 10. should be done together in sequence as they are part of the same domain.

Merging this will help @amydevs start the factoring out on a clean base.

tegefaulkes:
Tasks 1, 2, 6, 9 and 10 are to be done in new PRS. Once 8 is complete this will be merged and a pre-release done.

tegefaulkes:
Pending CI but ready to merge.

Issue #365, opened 2022-03-28
Decentralised NAT Signalling


tegefaulkes:
### Requirements of this design

1. Maintaining active connections between nodes to form a 'strongly connected' network.
2. Decentralised connection forming and `NAT` punching.
3. Decentralised message relaying.
4. Possibly caching known connection chains to speed up connection forming.

### Additional context
- https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1079550302
- https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1079552148
- https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1079552739
- https://github.com/MatrixAI/js-polykey/pull/326#issuecomment-1079553642

### Specification
We need to be able to hole punch a connection to the target node in a decentralised way. This can happen in stages.
1. Holding connections to form a strongly connected graph.
2. Discovery of a connected path.
3. Forming a direct connection using a connected path.

#### Holding connections
To ensure that a connection path is available we need to have an environment where a chain of connections to the target is likely to exist. To do this each node needs to hold active connections to other nodes. the distribution of these connections needs to ensure that the overall network is strongly connected. Bonus points for minimising the number of hops. 

Using the kademlia structure we can maintain active connections to the N closest nodes.  but we can also hold connections between other nodes across the network.  The closest nodes should maintain strong connections locally in the network while the extra connections will form shortcuts between parts of the network that are not closely related. Note I mean close the kademlia sense of closeness.

#### Discovery
The discovery process is very similar to the Kademlia `FIND_NODE` process. the difference here is that we only search over the active connections `Proxy` connections. Using this process we should be able to find a chain of connections that reaches our target Node.

Generally the idea is that we are walking the graph of active connections to find a chain of connections that leads to our target. Naively this would be a pretty expensive and exhaustive search but utilising the Kademlia structure we can ensure that we only walk closer to our target at each step of the search.

#### Forming direct connections
Given a chain of connections A>B>C>D we should be able to form a direct connection of A>B. We can form incrementally closer connections by using each step of the chain as a signaller to punch a connection to the next part of the chain. I will refer to this increasingly closer connection as a `Feeler`.

With the Chain A>B>C>D we can start by requesting B to signal a hole punch between A and C. If the hole-punch is successful then we now have a connection chain of A>C>D. We can then repeat the step use C as the new signaller to form a connection between A and D thus completing the desired direct connection.

```

    
[...]

Issue #182, opened 2021-06-04
Decentralised NAT Relaying


joshuakarp:
### Specification

Recall that a network structure that uses symmetric NAT cannot establish a connection between two nodes via hole-punching. For this, we require a relay connection. Suppose we have a keynode `A` that wants to connect to keynode `B`. However, `B` is behind a symmetric NAT structure. Keynode `C` is assumed to already have a direct connection to `B`:

![image](https://uploads.linear.app/e2a70dcf-d4d4-49a0-8551-1f7b7c87efa2/8064f1bd-0cf9-4837-a8fb-57e451bc6cf6/62cf39e9-10c1-460e-9947-fdfc11914371?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2UyYTcwZGNmLWQ0ZDQtNDlhMC04NTUxLTFmN2I3Yzg3ZWZhMi84MDY0ZjFiZC0wY2Y5LTQ4MzctYThmYi01N2U0NTFiYzZjZjYvNjJjZjM5ZTktMTBjMS00NjBlLTk5NDctZmRmYzExOTE0MzcxIiwiaWF0IjoxNzEyNTM5NzkxLCJleHAiOjMzMjgzMDk5NzkxfQ.jnmFL-n-rPZaJM2Ah5P5N7OrSDb5f1NAI6Nk34bF2jo)

In order for `A` to send a request to `B`, it can send this request to `C`, to then relay to node `B`. To implement this, we need some mechanism from the `ReverseProxy` to the `ForwardProxy` (as seen in keynode `C`).

Currently we will rely on the seed/bootstrap node cluster #194 as the designated nodes to do P2P kademlia seeding, hole punch relaying AND also proxy relaying.

Hole punch relaying is needed to traverse restricted-cone NATs and below. We implemented a bidirectional hole punching protocol, which means this is still needed even in the case of full cone NAT. Although I'm not entirely sure if this is in fact true, and may need to be proven on #159.

Proxy relaying is needed to traverse symmetric NATs.

[...]

PR #555, opened 2023-08-21
NAT Signalling (Hole Punching) should be Fire & Forget, Coalesced, and Secured with Signatures

### LINKED ISSUE #148: Authenticate the sender of a hole-punching signalling message
Upon receiving a hole-punching message (`agentService:sendHolePunchMessage()`), there is no reasonable means of ensuring that the message has come from the node ID stated in the message. Verification of a provided signature requires the public key of the source node.

In the future, we plan to move to Ed25519 keys (once this is supported in node-forge). In this event, the node ID will be the public key (instead of the public key fingerprint, as we require from RSA keys being 4096 bits). As such, authentication will be trivial, as we can simply verify the node ID (as the public key) against the signature.

tegefaulkes:
### Description

Currently the way NAT signalling is implemented is pretty simplistic. Some optimisations and security improvements needs to be made. There are 3 main improvements that needs to be made.

1. fire and forget
2. Signal attempt coalescing.
3. Secure signal requests

##### fire and forget

Currently, when a signalling request is made, the RPC call remains active for the duration of the signalling request. This forms a chain of active RPC requests between the requester and the relay, as well as between the relay and the target. We want to refactor this so that the RPC request is only active while initiating the relay request.

To this end, signalling  and hole punching attempts need to be back-grounded.  any RPC request to relay a signal needs to return immediately and attempt the signal relay in the background.

In the same vein, when the target receives the request, it needs to return and do the hole punch attempts in the background.

##### Signal attempt coalescing

If multiple requests are made for the same hole punch attempt, then these attempts need to be coalesced together. This can be applied to both the signalling  and hole punching attempts. 

For each stage we need to track the signalling and punching attempts in a map. The map can be uniquely keyed with the nodeIds of the requester and requestee node concatinated together as a string. If the domain is shut down then these need to be cancelled.

##### Secure signal requests

Related: #148 

We need a way to securely verify that the information of the request is correct and that the target of the hole punch messages made the request in the first place. There has been some discussion on this in #148.

### Issues Fixed

* Fixes #148 

### Tasks

* [x] 1. Make RPC requests for signalling and hole punch requests non-blocking
* [x] 2. Coalesce signalling and hole punching requests to prevent duplication of work
* [x] 3. Manage back-grounded requests, Allow cancellation
* [x] 4. Apply authentication to requests. #148 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
I'm thinking of breaking up the `NodesHolePunchMessageSendHandler` into two separate handlers. A signal request and a hole punch request. Right now both kinds of request are done in the same handler. The action it takes switches based on the message contents.

Right now these requests are only made through a single hop on a seed node. Later on with the #182 plan a hole punch will only be done with a single hop. So splitting the logic into two kinds of requests should simplify things.

tegefaulkes:
Mostly done, Just need to fix some tests and then work on the authentication logic. I think this may need to be speced out a little more.

tegefaulkes:
Looking into authentication now.

tegefaulkes:
New issue at https://github.com/MatrixAI/Polykey/issues/556 for proof of work limiting.

tegefaulkes:
Pretty much done now, I'm going to do a review.

src/agent/handlers/nodesHolePunchRequest.ts
```diff
@@ -0,0 +1,39 @@
...
+    _cancel,
+    meta,
+  ): Promise<AgentRPCResponseResult> {
+    const { nodeConnectionManager, logger } = this.container;
+    // Connections should always be validated
+    const requestingNodeId = agentUtils.nodeIdFromMeta(meta);
+    if (requestingNodeId == null) {
+      throw new agentErrors.ErrorAgentNodeIdMissing();
+    }
+    // Const srcNodeId = nodesUtils.encodeNodeId(requestingNodeId);
```
tegefaulkes: Remove this

src/agent/handlers/serverManifest.ts
```diff
@@ -9,13 +9,14 @@ import type NodeConnectionManager from '../../nodes/NodeConnectionManager';
...
 import { NodesClosestLocalNodesGetHandler } from './nodesClosestLocalNodesGet';
-import { NodesHolePunchMessageSendHandler } from './nodesHolePunchMessageSend';
+import { NodesHolePunchSignalHandler } from './nodesHolePunchSignal';
 import { NodesCrossSignClaimHandler } from './nodesCrossSignClaim';
 import { NotificationsSendHandler } from './notificationsSend';
 import { NodesClaimsGetHandler } from './nodesClaimsGet';
 import { VaultsScanHandler } from './vaultsScan';
 import { VaultsGitInfoGetHandler } from './vaultsGitInfoGet';
 import { VaultsGitPackGetHandler } from './vaultsGitPackGet';
+import { NodesHolePunchRequestHandler } from './nodesHolePunchRequest';
```
tegefaulkes: Let's move  this next to the other one

src/nodes/NodeConnectionManager.ts
```diff
@@ -104,6 +103,12 @@ class NodeConnectionManager {
    * NodeIds can't be used to properly retrieve a value from the map.
    */
   protected connections: Map<NodeIdString, ConnectionAndTimer> = new Map();
+  // Used to track active hole punching attempts
+  protected activePunchMap = new Map<string, PromiseCancellable<void>>();
+  // Used to rate limit punch attempts per IP Address
+  protected activeAddressMap = new Map<string, Semaphore>();
+  // Used track active signalling attempts
+  protected activeSignalSet = new Set<PromiseCancellable<void>>();
```
tegefaulkes: Needs better commentary and should be using docblocks

src/nodes/NodeConnectionManager.ts
```diff
@@ -122,6 +127,7 @@ class NodeConnectionManager {
     const quicConnection = connectionEvent.detail;
     await this.handleConnectionReverse(quicConnection);
   };
+  protected rateLimiter = new RateLimiter(60000, 20, 10, 1);
```
tegefaulkes: Move next to the other maps for tracking punching

src/nodes/NodeConnectionManager.ts
```diff
@@ -1140,107 +1158,84 @@ class NodeConnectionManager {
...
+      await client.methods.nodesHolePunchRequest({
+        address,
+      });
+    }).finally(() => {
+      this.activeSignalSet.delete(connProm);
+    });
+    this.activeSignalSet.add(connProm);
+  }
+
+  // TODO: make timedCancellable
```
tegefaulkes: Needs to be addressed

src/nodes/NodeManager.ts
tegefaulkes, line 1: The changes to this file can be reverted, not needed anymore.

src/utils/RateLimiter.ts
tegefaulkes, line 1: @CMCDragonkai should have a look at this. It was pretty quick and dirty for now. 

Should I make a new issue for extracting this out somewhere? Where else will we use it?

tests/agent/handlers/nodesHolePunchRequest.test.ts
tegefaulkes, line 1: I think a few of the dependencies need to be cut out here. Need to check.

tests/agent/handlers/nodesHolePunchSignal.test.ts
tegefaulkes, line 1: Same here, need to review imports and clean up.

tests/nodes/NodeConnectionManager.general.test.ts
```diff
@@ -76,10 +77,10 @@ describe(`${NodeConnectionManager.name} general test`, () => {
 
   let dataDir: string;
 
-  let remotePolykeyAgent: PolykeyAgent;
+  let remotePolykeyAgentA: PolykeyAgent;
```
tegefaulkes, line 80: Creating an agent is comparatively expensive. I should check where they're being used and reduce the amount of times I need to start an agent.

tegefaulkes (commented):
General review for reference. Just added some notes for changes.

tegefaulkes:
There are some tmp errors that need to be cleaned up.

tegefaulkes:
Pending CI and merge. I'm considering this done now.

I'll move on to the configuration PR next. Then MDNS, node graph changes and final codebase cleanup. From there its testnet testing time.

src/utils/RateLimiter.ts
CMCDragonkai: Where's the TokenBucket class?

CMCDragonkai:
Tests pass but there's a resource leak in PK. This is probably the problem that occurred before. Resource leaks is a critical problem you need to have a look into this at a deeper level.

image

tests/agent/handlers/nodesHolePunchRequest.test.ts
```diff
@@ -0,0 +1,147 @@
+import type * as quicEvents from '@matrixai/quic/dist/events';
+import type { Host as QUICHost } from '@matrixai/quic';
```
CMCDragonkai: This type is no longer necessary. We changed js-quic to take `string` for `host` and `number` for `port`.

This avoids needing all these type conversions.

src/utils/RateLimiter.ts
CMCDragonkai, line 1: This should be a directory. Make it so that it can actually be exported and factored out.

src/agent/handlers/nodesHolePunchRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+    input: AgentRPCRequestParams<HolePunchRequestMessage>,
+    _cancel,
+    meta,
+  ): Promise<AgentRPCResponseResult> {
+    const { nodeConnectionManager, logger } = this.container;
+    // Connections should always be validated
+    const requestingNodeId = agentUtils.nodeIdFromMeta(meta);
+    if (requestingNodeId == null) {
+      throw new agentErrors.ErrorAgentNodeIdMissing();
+    }
```
CMCDragonkai: I don't see how this accomplishes #148. The receiver is supposed to verify that the message is signed by the original sending node?

src/agent/handlers/nodesHolePunchSignal.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { NodeAddress } from '../../nodes/types';
+import type { HolePunchSignalMessage } from './types';
+import type { JSONValue } from '../../types';
+import { never } from '../../utils';
+import * as agentErrors from '../errors';
+import * as agentUtils from '../utils';
+import { UnaryHandler } from '../../rpc/handlers';
+import * as validationUtils from '../../validation/utils';
+
+class NodesHolePunchSignalHandler extends UnaryHandler<
```
CMCDragonkai: It is not clear what this separation is doing?

You need 2 handlers right?

1. One for receiving a signalling request.
2. One for receiving the hole punching request.

The reason why these calls were the same was so that it would eventually lead to decentralised signalling.

I'm not sure how these 2 calls make sense...

The naming of these 2 is problematic, because we should name them as RPC calls that are called by the origin node.

You could say:

* A -nodesConnect -> S
* S -nodesConnect -> B
* A -> B

It's supposed to be the same call, so that this may be possible:

* A -nodesConnect -> S1
* S1 -nodesConnect -> S2
* S2 -nodesConnect -> B
* A -*> B

Furthermore I don't see any authentication being done here. I don't see how this achieves #148.

CMCDragonkai:
I think this PR requires more work.

1. I don't understand how your fire and forget is achieved here. Where is this background processing happening if even required?
2. I think the resource leak needs to solved before committing any more code to staging.
3. I don't see any authentication happening right now so #148 is not being solved.
4. Make the rate limiter a directory with 2 files so it can be used generically.

src/utils/RateLimiter.ts
tegefaulkes: It's in there.

tests/agent/handlers/nodesHolePunchRequest.test.ts
tegefaulkes: I missed them here, removed.

src/agent/handlers/nodesHolePunchRequest.ts
tegefaulkes: I opted not to do that in the end, I concluded it didn't add any real functionality or security. I can add it it, but it won't prevent other nodes from requesting punches to a target node. Mostly because we can verify the the information has been signed by someone but we can't verify that the information is correct.

A bad actor can just generate a signed request to some other address.

CMCDragonkai: As discussed, it's still better to have this extra layer of protection, using the `Token` class should be sufficient.

And I want to leave the possibility of extensibility here with further crypto protection in the future.

CMCDragonkai:
Names of the 2 separate handlers:

* `nodesConnectionSignalingInitial`
* `nodesConnectionSignalingFinal`

Best names we could come up for now, and fits the SOV order.

CMCDragonkai:
Changed to:

* `nodesConnectionSignalInitial`
* `nodesConnectionSignalFinal`

More verby.

src/utils/RateLimiter.ts
tegefaulkes: It has been moved and refactored a little.

tegefaulkes:
I just need to fix one of the tests files not exiting the process. Besides that the rest that needs to be done like the message signing, authentication and handling background promises can be addressed later. We're putting this PR on hold until we work some things out.

tegefaulkes:
Ok, this PR is on hold for now.

CMCDragonkai:
This needs to be rebased and work should start on this.

I'm getting wary of all the stop-gap solutions that are introducing resource leaks and memory leaks here. I think it's important any stop-gap solution isn't resulting a resource leak for quic and ws.

Also here too if there are background processes running. Make sure if you're not using the `tasks` domain that you have proper tracking of these promises plus signal cancellation.

We should make sure to have tests that test the signal cancellations/abrupt aborts of the code.

CMCDragonkai:
Needs rebase. What's the status of this?

tegefaulkes:
Rebased on staging now. It was tricky given the number of changes to staging since this was worked on. I had to branch off of staging and pull the changes needed from a diff.

I haven't fully tested it yet and it likely still needs some polish.

Currently it isn't actually doing any authentication because at the time I concluded that it's too easy to work around if you're a bad actor. But if we still want that I need to apply signing and verification to the requests.

CMCDragonkai:
> Currently it isn't actually doing any authentication because at the time I concluded that it's too easy to work around if you're a bad actor. But if we still want that I need to apply signing and verification to the requests.

We still need auditability here so you should add in the signed requests. The signed requests is a quick add, and should be important for our metrics and observability.

tegefaulkes:
Finished cleaning up and testing. I'll add signing and verification now.

tegefaulkes:
Everything is done now. I need to go over the existing review comments to see if they're still relevant.

src/agent/handlers/nodesHolePunchRequest.ts
tegefaulkes: I've added in message signing and verification.

src/agent/handlers/nodesHolePunchSignal.ts
tegefaulkes: Authentication is being addressed now by the latest changes.

The names have also be updated.

As for how it's designed now. We will only every send a request to an single relaying node. There will be no multi-hop requests. It's not needed now and won't be needed after #365 is done. SO the steps are clearly request -> relay -> punch. It simplifies the logic.

tegefaulkes:
Ok, this is ready to merge.

src/rateLimiter/RateLimiter.ts
CMCDragonkai, line 1: I think this can put into `src/utils/ratelimiter`.

src/nodes/NodeConnectionManager.ts
```diff
@@ -129,6 +128,22 @@ class NodeConnectionManager {
    * Default timeout for RPC handlers
    */
   public readonly rpcCallTimeoutTime: number;
+  /**
+   * Used to track active hole punching attempts
+   */
+  protected activePunchMap = new Map<string, PromiseCancellable<void>>();
```
CMCDragonkai: This should be `activePunchPMap`. Do we use `Map` and `Set` suffixes commonly across the codebase? I don't remember it.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1074,11 +1096,11 @@ class NodeConnectionManager {
...
+  public holePunchReverse(
     host: Host,
     port: Port,
-    ctx?: Partial<ContextTimed>,
-  ): Promise<void>;
+    ctx?: Partial<ContextTimedInput>,
+  ): PromiseCancellable<void>;
   @ready(new nodesErrors.ErrorNodeConnectionManagerNotRunning())
   @timedCancellable(
     true,
```
CMCDragonkai: If there's no "forward" version of this, there should not be a "reverse" version of this.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1657,19 +1702,15 @@ class NodeConnectionManager {
...
-              undefined,
-              ctx,
-            )
+            this.holePunchSignalRequest(targetNodeId, seedNodeId, ctx)
               // Ignore results
               .then(
                 () => {},
-                () => {},
+                (e) =>
+                  this.logger.debug(`signal request failed with ${e.message}`),
```
CMCDragonkai: What's this?

src/nodes/NodeConnectionManager.ts
```diff
@@ -1378,106 +1400,129 @@ class NodeConnectionManager {
...
     true,
     (nodeConnectionManager: NodeConnectionManager) =>
       nodeConnectionManager.connectionConnectTimeoutTime,
   )
-  public async relaySignalingMessage(
-    message: HolePunchRelayMessage,
-    sourceAddress: NodeAddress,
+  public async holePunchSignalRequest(
+    targetNodeId: NodeId,
+    signallingNodeId: NodeId,
```
CMCDragonkai: I think the naming of this should correlate to "initial" and "final". Otherwise it is hard to understand and correlate them to request/forward/reverse.

src/nodes/NodeConnectionManager.ts
CMCDragonkai: Nothing is being done with this exception.

src/nodes/agent/errors.ts
```diff
@@ -8,4 +8,22 @@ class ErrorAgentNodeIdMissing<T> extends ErrorAgent<T> {
...
+  static description = 'Failed to verify request message signature';
+  exitCode = sysexits.TEMPFAIL;
+}
+
+class ErrorNodesConnectionSignalRelayVerificationFailed<
+  T,
+> extends ErrorAgent<T> {
+  static description = 'Failed to verify relay message signature';
+  exitCode = sysexits.TEMPFAIL;
+}
```
CMCDragonkai: Are these really `TEMPFAIL`? I would say `UNAVAILABLE` right?

src/rateLimiter/RateLimiter.ts
CMCDragonkai: Changing this requires changing the test paths too.

CMCDragonkai:
We need an explanation preferably a diagram on how all of the rate limiting, semaphores, active promise tracking actually work together, this code is quite tricky and will easily tech debt if we don't understand it.

src/nodes/agent/errors.ts
tegefaulkes:
```

 /**
   * Temporary failure, indicating something that is not really an error.
   * In sendmail, this means that a mailer (e.g.) could not create a connection,
   * and the request should be reattempted later.
   */
  TEMPFAIL: 75,

// VS
/**
   * A service is unavailable. This can occur if a support program or file
   * does not exist. This can also be used as a catchall message when
   * something you wanted to do does not work, but you do not know why.
   */
  UNAVAILABLE: 69,
```

I'm using `TEMPFAIL` because it makes more sense for a network failure. That said, it's not a network problem but a verification one. So Yeah, maybne `UNAVALIABLE` makes more sense.

src/rateLimiter/RateLimiter.ts
tegefaulkes: This has been moved.

src/nodes/NodeConnectionManager.ts
tegefaulkes: No, usually we don't include them.

I'll rename this to `activePunchPs`?

src/nodes/NodeConnectionManager.ts
tegefaulkes: I'll rename it to `holePunch`.

src/nodes/NodeConnectionManager.ts
tegefaulkes: It was a debugging message from testing left in. All errors are just ignored here. It would either be a cancellation of the promise or a failure of the socket to send. In either case we don't care in this part of the code as Socket failures are handled elsewhere.

I'll remove the debugger message.

src/nodes/NodeConnectionManager.ts
tegefaulkes: I've updated the names and commentary.

src/nodes/agent/errors.ts
tegefaulkes: Updated.

src/nodes/NodeConnectionManager.ts
tegefaulkes: Changed the name and updated the docblock.

CMCDragonkai:
@tegefaulkes diagram?

tegefaulkes:
I've been meaning to get to this. Just been side-tracked with other issues at the moment.

PR #552, opened 2023-08-16
Consolidating Configuration Entropy and Nodes and Client Services Refactoring

CMCDragonkai:
### Description

The configuration of Polykey has suffered from alot of entropy. This PR attempts to consolidate our configuration structure.

It's based on this comment: https://github.com/MatrixAI/Polykey-CLI/pull/4#issuecomment-1680131511

A note about task 5. It's important to understand that in `PolykeyAgent`, we have a bunch of dependencies that can be injected, but their injection is really intended for mocking, which is only for testing, not intended for production usage. This is different from optional dependencies which has legitimate usecases when injecting or not injecting. If we were to follow task 5. strictly, this would add alot of boilerplate to managing optional injected dependencies for everything in `PolykeyAgent`. And right now, there's no easy way to manage this without manually setting up booleans and conditional checks for all these optional dependencies. Therefore... task 5. should only be applied to situations where injecting or not-injecting is a legitimate usecase such as between `QUICSocket` and `QUICServer` and `QUICClient` and between `EncryptedFS` and `DB`. But optional dependencies in `PolykeyAgent` are the exception. IN FACT, because JS allows module mocking... we may even remove the ability to inject these dependencies and use `jest.mock` instead.

### Issues Fixed

* Related https://github.com/MatrixAI/Polykey-CLI/pull/4#issuecomment-1680131511
* Related #444
* Related #386 

### Tasks

- [x] 1. Move `*base` to `config.paths`. As these are path constants. They are not defaults to be configured.
- [x] 2. Separate "user config" from "system config". There is a bunch of constants that control how the network works, these are not supposed to be changed by the user. These are constants that we developers of Polykey tune empirically for optimal operation. Remember at this point of the abstraction, PK is an application, no longer a library.
  - `config.defaultsUser`
  - `config.defaultsSystem`
- [x] 3. Create a type based on `config.defaultsUser` and have it be referenced everywhere there needs to be some spread of the parameters that will be defaulted by `config.defaultsUser`. This prevents entropy where parameter names get changed and we forget that they need to be consistent.
  - What happens if you only use a subset of the default parameters? Do you end up creating a subtype?
  - All relevant types should be exposed in `src/config.ts` so that way it is always consistent, even if it is a bit weird. It gives us one place to ensure compliance between parameters that should have the same name and same meaning.
- [x] 4. Restructure configuration properties into a special `config` parameter, and do a deep merge
  - `PolykeyAgent.createPolykeyAgent` - create a `PolykeyAgentConfig` type
[...]

PR #560, opened 2023-09-13
Consolidating `ClientService` and js-ws and js-rpc integration

### LINKED ISSUE #559: Consolidating client service logic and lifecycle
### Specification

We need to consolidate the client service into the client domain. This includes moving all of that logic into the client domain.

Currently we create a `WebsocketServer`, `RPCServer` and client manifest separately where needed, namely the `PolykeyAgent`. The creation and life-cycle of theses need to be consolidated into a `ClientService` class for handling all of this.

While we're here we can use the extracted `RPC` library once it's ready. So this can include removal of the internal `RPC` implementation from `Polykey`.

Requirements:
1. `ClientSevice` will be a `CreateDestroyStartStop`.
2. The `ClientService` will handle the life-cycle and setting up of the relevant `RPCServer` and `WebsocketServer`.
3. `ClientService` will be generic on the manifest. This is required to reduce the complexity in tests.

### Additional context

* Related #552 

### Tasks

1. Create a `ClientService` class within the client domain. 
2. Move `RPC` and `WebsocketServer` creation and life-cycle into `ClientService`.
3. `PolykeyAgent` will be updated to use this.
4. tests will be updated to use this. Should reduce setup logic.

tegefaulkes:
### Description

In this PR we're consolidating the client service logic into a `ClientService` class that handles its life cycle.

### Issues Fixed

* Fixes #559 

### Tasks

- [x] 1. Create a ClientService class within the client domain.
- [x] 2. Move RPC and WebsocketServer creation and life-cycle into ClientService.
- [x] 3. PolykeyAgent will be updated to use this.
- [x] 4. Tests will be updated to use this. Should reduce setup logic.
- [x] 5. Add `PolykeyAgent.clientServiceHost`, `PolykeyAgent.clientServicePort`, `PolykeyAgent.agentServiceHost`, `PolykeyAgent.agentServicePort` as getter property so that they can be used directly instead of referring to encapsulated objects for the host and port.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* ~Squash and rebased~


CMCDragonkai:
I'm reassigning this to @amydevs and @addievo. You can take over this PR or join this PR for integrating js-ws and js-rpc.

Make sure to do a release before integrating here.

This PR could be rebased on top of #552 if #552 merges first.

If this merges first, then #552 can be rebased on top.

CMCDragonkai:
Remember to review #559.

However if this is consolidating, @tegefaulkes how come there isn't any deleted files in the diff right? I thought files would be moved from somewhere else.

tegefaulkes:
> Remember to review #559.
> 
> However if this is consolidating, @tegefaulkes how come there isn't any deleted files in the diff right? I thought files would be moved from somewhere else.

No deletions yet, it's mostly consolidating the client into a `ClientService` that handles all of the client stuff. So it was just adding that and modifying other places to use it. No deletions needed there.

However if we're using the `js-rpc` and `js-ws` libraries now then we need to remove the `rpc` and `websocket` domain and replace it with the libs.

CMCDragonkai:
flow of ClientService

CMCDragonkai:
@addievo add yourself to the `constributors` in `package.json`.

CMCDragonkai:
@amydevs make sure to push often here even for intermediate commits.

amydevs:
Client service will require an empty stream handler for the ClientServiceClient, as this behaviour is currently unexpected and will cause creation of unneeded resources. When we decide to add push notifs, etc. We can replace this with handlers for that.

CMCDragonkai:
Just immediately close the stream as soon as you get the event. For now server shouldn't be creating streams. But we will make use of this feature later in the future.

CMCDragonkai:
Has this been rebased on #560?

CMCDragonkai:
This is the main PR to work on now.

Goals:

* Get js-ws and js-rpc fully integrated and tested
* Fix up any tests that relate to using the client service and/or agent service
* Make sure this is working with Polykey-CLI

I can see that there are branch conflicts in this PR... I thought this was rebased?

CMCDragonkai:
@amydevs please rebase this on top of staging too.

tegefaulkes:
It was rebased, but a linting commit in staging brought in a bunch of conflicts. They'll be mostly whitespace but git doesn't fully understand that.

CMCDragonkai:
Who's rebasing this atm?

CMCDragonkai:
I'm going to rebase this on staging now. Don't touch it.

CMCDragonkai:
In order to do this cleanly, I need to squash the relevant commits PR into 1 commit and then rebase on top.

CMCDragonkai:
Lots of breaking conflicts here. Probably as a result of linting changes. To avoid in this future @tegefaulkes mentioned always linting our commits, we can try this to avoid linting conflicts.

Another thing is that we should really only have 1 large PR at a time, working across multiple PRs requires alot more experience. So let's limit staging control to myself or @tegefaulkes for now.

src/vaults/VaultManager.ts
```diff
@@ -36,6 +36,7 @@ import * as vaultsEvents from './events';
 import * as utils from '../utils';
 import * as vaultsUtils from '../vaults/utils';
 import * as vaultsErrors from '../vaults/errors';
+import * as vaultsEvents from './events';
```
CMCDragonkai: This should be `./utils` and `./errors`.

tests/agent/handlers/notificationsSend.test.ts
```diff
@@ -9,8 +9,8 @@ import os from 'os';
 import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
 import { QUICClient, QUICServer } from '@matrixai/quic';
 import { DB } from '@matrixai/db';
-import RPCClient from '@/rpc/RPCClient';
-import RPCServer from '@/rpc/RPCServer';
+import RPCClient from '@matrixai/rpc/dist/RPCClient';
+import RPCServer from '@matrixai/rpc/dist/RPCServer';
```
CMCDragonkai: @addievo

```ts
import { RPCClient, RPCServer } from '@matrixai/rpc';
```

tests/agent/handlers/nodesClaimsGet.test.ts
```diff
@@ -7,10 +7,10 @@ import os from 'os';
 import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
 import { QUICClient, QUICServer } from '@matrixai/quic';
 import { DB } from '@matrixai/db';
+import RPCClient from '@matrixai/rpc/dist/RPCClient';
+import RPCServer from '@matrixai/rpc/dist/RPCServer';
 import * as nodesUtils from '@/nodes/utils';
 import { encodeProviderIdentityId } from '@/identities/utils';
-import RPCClient from '@/rpc/RPCClient';
-import RPCServer from '@/rpc/RPCServer';
```
CMCDragonkai: Same here.

tests/agent/handlers/nodesClosestLocalNode.test.ts
```diff
@@ -7,8 +7,8 @@ import os from 'os';
 import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
 import { QUICClient, QUICServer } from '@matrixai/quic';
 import { DB } from '@matrixai/db';
-import RPCClient from '@/rpc/RPCClient';
-import RPCServer from '@/rpc/RPCServer';
+import RPCClient from '@matrixai/rpc/dist/RPCClient';
+import RPCServer from '@matrixai/rpc/dist/RPCServer';
```
CMCDragonkai: All imports of `@matrixai/rpc`.

CMCDragonkai:
Ok I've had to squash down everything here, and then use `git checkout --ours` to prefer the `staging` version of things.

This means any code that was conflicting between staging and this branch has been lost.

@amydevs when you are comparing your local branch. You can push up as `feature-client-consolidation-2`. Then compare any changes that you want to bring over here.

I'm also noticing alot of spurious changes introduced into this PR that shouldn't have been in this PR so I'm not sure why.

@tegefaulkes use `feature-client-consolidation-3`.

CMCDragonkai:
The only person who should be working on this is @amydevs and apply changes that you have locally on top of this single commit PR atm.

CMCDragonkai:
So @amydevs has finished the commit surgery now. And has separated all the commits into 3 commits.

1. The first commit is a basic change for all the handlers to use arrow function properties (or equivalent for the CS, SS, DS)
2. The second commit just brings in `@matrixai/ws` and removes `src/websockets` and related tests. It does not actually test if this is working.
3. The third commit is a WIP commit because it's not in a "working state" - this one introduces `ClientService` as an encapsulation of `RPCServer` and `WebSocketServer` and it also brings in `@matrixai/rpc` and this last thing has a lot of disparate changes, it's spread out everywhere because many parts of the codebase uses RPC imports.

For this third commit there are several things that is still pending and SHOULD be squashed into this third commit when done.

1. The `@matrixai/rpc` still has errors at this point - things like the `toError` and `fromError`. It should be fixed here, and squashed into the third commit.
2. The indexed exports `src/index.ts` of `@matrixai/rpc` should be exporting `RPCServer` and `RPCClient` at minimum but also the relevant utils, events... etc. Imports should like this `import { RPCServer, ... } from '@matrixai/rpc';` when inside PK.
   - The reason why you sometimes see a `dist/...` is only because of `@matrixai/async-init`.
3. Import ordering. There's a bunch of places where import order isn't correct, it's always roughly speaking: types, impls, then within each group, its native imports, externals, relative, relative downwards, relative upwards.
4. Testing - make sure that `ClientService` and all the related behaviour is actually working with tests that test the RPC calls for the client service.

With those 4 things done, that can squash into the 3rd commit, and you can rename it to be `feat: ...` @amydevs @addievo.

Then this can be merged. (Thus at the end this PR should only have 3 commits).

@tegefaulkes has fixes necessary for the integration of Polykey CLI, and any further testing of the agent service.

src/nodes/utils.ts
```diff
@@ -9,7 +9,7 @@ import * as nodesErrors from './errors';
 import * as keysUtils from '../keys/utils';
 import { encodeNodeId, decodeNodeId } from '../ids';
 import { bytes2BigInt, never } from '../utils';
-import * as rpcErrors from '../rpc/errors';
+import * as rpcErrors from '@matrixai/rpc/dist/errors';
```
CMCDragonkai: `import { errors as rpcErrors } from '@matrixai/rpc';`.

src/nodes/agent/handlers/VaultsGitInfoGet.ts
```diff
@@ -3,7 +3,7 @@ import type Logger from '@matrixai/logger';
 import type { ContextTimed } from '@matrixai/contexts';
 import type ACL from '../../../acl/ACL';
 import type VaultManager from '../../../vaults/VaultManager';
-import type { JSONRPCRequest } from '../../../rpc/types';
+import { JSONRPCRequest } from '@matrixai/rpc/dist/types';
```
CMCDragonkai: `import type { JSONRPCRequest } from '@matrixai/rpc';`.

This is only possible if `@matrixai/rpc` does this:

```
export * from './types';
```

This is different from:

```
export * as types from './types';
```

src/nodes/NodeConnectionManager.ts
```diff
@@ -48,8 +48,8 @@ import * as validationUtils from '../validation/utils';
 import * as networkUtils from '../network/utils';
 import * as utils from '../utils';
 import config from '../config';
-import RPCServer from '../rpc/RPCServer';
-import * as rpcUtilsMiddleware from '../rpc/utils/middleware';
+import RPCServer from '@matrixai/rpc/dist/RPCServer';
+import * as rpcUtilsMiddleware from '@matrixai/rpc/dist/middleware';
```
CMCDragonkai: `import { middleware as rpcMiddleware } from '@matrixai/rpc';`.

CMCDragonkai:
If 1., 2., 3. is all done, then start testing the client service. This tests the entire ws plus RPC usage.

To make sure this is structured right, please expand the task list above with all test fixes and new tests or test removals and test changes that is needed or has already been done in this PR. @addievo @amydevs

CMCDragonkai:
@addievo when you resolve changes, make a comment about how you resolved it.

CMCDragonkai:
I expect to merge Monday.

addievo:
1. was resolved in https://github.com/MatrixAI/js-rpc/pull/33
2. Changed exports in rpc to 

```typescript
export { default as RPCClient } from './RPCClient';
export { default as RPCServer } from './RPCServer';
export * as utils from './utils';
export * from './types';
export * as errors from './errors';
export * as events from './events';
export * from './handlers';
export * from './callers';
export * as middleware from './middleware';
``` 

Consequently all imports were standardised to .... from ```from '@matrixai/rpc```

3. Lintfix for fixing import order

addievo:
> If 1., 2., 3. is all done, then start testing the client service. This tests the entire ws plus RPC usage.
> 
> To make sure this is structured right, please expand the task list above with all test fixes and new tests or test removals and test changes that is needed or has already been done in this PR. @addievo @amydevs

I'll look over ws+rpc usage tests present with Amy and decide if more are needed

CMCDragonkai:
Ok great.

tegefaulkes:
I'm merging #562 into here and doing the final work in this PR. Given this merge we shouldn't do any history editing to this PR if we can avoid it.

tegefaulkes:
Few things I noticed while cleaning up
1. Need to update client handlers to do the default export. Could skip if we make an issue
2. Authentication middleware needs a look over. Seems to be subject to change based on from and to error changes.
3. Might be a bug with the websocket streams, I'm seeing a stream being recreated and closed due to error constantly.

I've cleaned up but I haven't run all the tests again yet.

tegefaulkes:
It occurs to me that the `serverManifest` is a parameter of the `NodeConnectionManager` but the `clientManifest` the `NodeConnection` uses is hard coded. This may be something we'd want to make consistent.

tegefaulkes:
There's a bug with the handles using arrow functions. This just applies to the `Serverhandler`, and `DuplexHandler` since they are generators.

Since we switched to arrow functions, the `this.container` doesn't exist inside the generators because `this` doesn't exist since we don't bind `this` to the arrow functions anymore. It's a small fix.

CMCDragonkai:
> There's a bug with the handles using arrow functions. This just applies to the `Serverhandler`, and `DuplexHandler` since they are generators.
> 
> Since we switched to arrow functions, the `this.container` doesn't exist inside the generators because `this` doesn't exist since we don't bind `this` to the arrow functions anymore. It's a small fix.

Arrow functions get the `this` bound lexcially when defined inside a class.

But you're talking about the generator functions which aren't arrow functions. In that case they should get their `this` bound.

How did you solve this?

CMCDragonkai:
> It occurs to me that the `serverManifest` is a parameter of the `NodeConnectionManager` but the `clientManifest` the `NodeConnection` uses is hard coded. This may be something we'd want to make consistent.

Make them both hardcoded or both paramaterised but defaulted?

tegefaulkes:
Easiest way is to bind this back for the generator functions.

tests/agent/handlers/nodesClosestLocalNode.test.ts
tegefaulkes: Fixed up import of rpc.

tests/agent/handlers/nodesClaimsGet.test.ts
tegefaulkes: Also fixed.

tests/agent/handlers/notificationsSend.test.ts
tegefaulkes: Also fixed.

src/vaults/VaultManager.ts
tegefaulkes: Fixed.

CMCDragonkai:
Did you make both `clientManifest` and `serverManifest` hardcoded?

src/PolykeyAgent.ts
```diff
@@ -40,6 +40,7 @@ import * as workersUtils from './workers/utils';
 import TaskManager from './tasks/TaskManager';
 import { serverManifest as clientServerManifest } from './client/handlers';
 import agentServerManifest from './nodes/agent/handlers';
+import ClientService from './client/ClientService';
```
CMCDragonkai: Is this renaming of `import { serverManifest as clientServerManifest } from './client/handlers';` necessary?

I feel like the imports here can be tidied up too.

src/PolykeyClient.ts
```diff
@@ -60,15 +62,17 @@ class PolykeyClient {
...
     });
-    const rpcClientClient = await RPCClient.createRPCClient({
+    const rpcClientClient = new RPCClient({
       manifest: clientManifest,
       streamFactory,
       middlewareFactory: rpcUtilsMiddleware.defaultClientMiddlewareWrapper(
         clientUtilsMiddleware.middlewareClient(session),
         parserBufferByteLimit,
       ),
+      toError: networkUtils.toError,
```
CMCDragonkai, line 71: I thought `toError` here would be specific to client domain? Or are we using `network/utils` for both client service and agent service?

src/PolykeyClient.ts
```diff
@@ -1,18 +1,20 @@
...
 import * as clientUtilsMiddleware from './client/utils/middleware';
 import { Session } from './sessions';
 import * as utils from './utils';
 import * as errors from './errors';
 import * as events from './events';
 import config from './config';
 import { clientManifest } from './client/handlers/clientManifest';
+import * as networkUtils from './network/utils';
 
+let idGen: IdGen;
```
CMCDragonkai: Why is `idGen` defined outside the entire class?

src/PolykeyClient.ts
tegefaulkes: network domain is for stuff shared between client and agent network stuff.  One thing that still needs to be done is the `fromError` used for the clientService  needs to be a sensitive version that strips out the callstack.

CMCDragonkai: The sensitive version is just whatever the `replacer` is. So I don't think `fromError` does any stripping. You define the `replacer` and that should be one stripping out the call stack and any other relevant information.

src/PolykeyAgent.ts
tegefaulkes: It used to be, but now it's not needed.

src/PolykeyClient.ts
tegefaulkes: I have no idea, that should not be done.

tegefaulkes: It's not even defined anywhere. What is `IdGen` used for?

CMCDragonkai: Yea, also I'm pretty the `idGen` is now optional. So for now just get rid of it.

src/PolykeyAgent.ts
CMCDragonkai: Also I'm pretty sure that `client/handlers/index.ts` is supposed to default export the manifest.

Just like in `src/nodes/agent/handlers`.

The client manifest and server manifest are supposed to be default exports. Can you check?

src/PolykeyClient.ts
tegefaulkes: `fromError` can do the stripping. Why do we even need the replacer? In any case, the `fromError` and `toError` needs to be fixed up. I didn't want to spend much more time on it with other things needing fixes at the time.

I think @amydevs should have a look over it.

CMCDragonkai:
> Easiest way is to bind this back for the generator functions.

I'm not sure this makes sense to me. I just asked chatgpt...

The constructor binding approach generally won't work if you define the async generator function as a class field, like `handle = async *function () { }`. Class fields are essentially syntactic sugar that allows you to declare and initialize instance properties directly within the class body. When you define a function as a class field, it gets attached to the instance, not the prototype. Therefore, binding it again inside the constructor doesn't work as expected, because the class field has already captured the instance context.

Here's an example to demonstrate:

```javascript
class MyClass {
  constructor() {
    this.handle = this.handle.bind(this);  // This line is essentially a no-op
  }

  handle = async function*() {
    // `this` is already bound to the class instance because of the class field initialization
  }
}
```

In this case, the `handle` function is already bound to the instance of the class due to the class field definition. The constructor's explicit `bind()` becomes a no-op in this context because `handle` is already an instance method, not a prototype method. Therefore, binding it again in the constructor doesn't change the behavior of `this` within `handle`.

So, if you're using class fields to define your async generator function, you don't need to worry about binding `this` in the constructor—it's automatically bound to the instance of the class.

src/PolykeyAgent.ts
tegefaulkes: Yeah, that needs to be fixed as well. I made a note before somewhere. Ah, here https://github.com/MatrixAI/Polykey/pull/560#issuecomment-1752649971. 2. and 3. in that comment are no longer relevant.

There are something like 60 handlers, so it'll take too much time right now.

CMCDragonkai:
On second thought... chatgpt might be wrong here.

You are correct, and I apologize for the earlier error. The constructor binding is indeed effective when using class field syntax with `async function*` or regular functions. It's not a no-op in this case; it would bind `this` to the class instance.

When you define a method as a class field like `handle = async function*() { }`, it doesn't automatically bind `this`. The method is initialized with each new instance of the class, but its `this` context is determined by how it's invoked, just like any regular function.

Binding the function in the constructor would ensure that `this` is correctly set to the instance of the class. Here's an example:

```javascript
class MyClass {
  constructor() {
    this.handle = this.handle.bind(this);
  }

  handle = async function*() {
    // `this` is now correctly bound to the class instance
  }
}
```

In this example, explicitly binding `this.handle` in the constructor ensures that it will always be bound to an instance of `MyClass`, regardless of how or where it is invoked. Thank you for bringing this to my attention, and I appreciate the opportunity to correct this information.

src/PolykeyClient.ts
CMCDragonkai: By default removing `idGen` should be making it so that `id` field is always just `null`. As that is allowed in the JSON RPC spec.

CMCDragonkai: In the future we can add an auto-increment. Or... even something useful for tracing.

src/PolykeyClient.ts
CMCDragonkai: It's cause `fromError` returns a JSON object to be stringified. The `replacer` can only be used when you do `JSON.stringify`. It's 2 steps.

tegefaulkes:
I fixed that by doing the binding when registering the handler inside the `RPCServer`. Pretty much how it worked before but I only changed it back for the generators.

src/PolykeyAgent.ts
CMCDragonkai: @addievo can do this, if you create a new issue, assign it to him.

src/PolykeyClient.ts
tegefaulkes: Removed.

src/PolykeyAgent.ts
tegefaulkes: Import was fixed.

The handlers need to be addressed later. I'll make an issue in a moment.

tegefaulkes:
`ClientService` still needs to be converted to `StartStop` rather than `CreateDestroyStartStop`. I'll make an issue in a moment.

tegefaulkes:
I'm skipping squashing here. This has been merged into and I don't want any potential problems.

src/PolykeyAgent.ts
tegefaulkes: new issue at #563

src/PolykeyClient.ts
tegefaulkes: New issue at #564 to address this.

tegefaulkes:
> Did you make both `clientManifest` and `serverManifest` hardcoded?

The `clientManifest` is hard coded inside the `NodeConnection`. The `serverManifest` is passed into the `NCM` prevent it from depending on everything during testing.

tegefaulkes:
New issue #565.

PR #567, opened 2023-10-11
Client Domain Structure of Callers and Handlers should match Agent Callers and Handlers

### LINKED ISSUE #520: Use default exports for handler classes
### Specification

As a standard we want to use default exports if a file exports a single class. The client handlers used to export the caller and handler definition. Now it's just the handler so it needs to be updated to a default export.

### Additional context

* Related #509 
* Related #513 

### Tasks

1. update all of the client handlers to use default exports.

### LINKED ISSUE #563: Update `client` domain structure
### Specification

During #552 the `agent` domain was rolled into the `nodes` domain at `src/nodes/agent`. The agent domain was restructured during this time as well.

1. All the handlers were updated to use default exports.
2. All the handlers were updated to use arrow functions.
3. The `ServerManifest` was moved into the `index.ts`.
4. The caller definitions were moved into `nodes/agent/callers`.
5. The caller definitions were split into separate files. (I don't like this change, it means that each caller which can be defined in a single line need to be spread across 60+ files.) I should check with @CMCDragonkai to see if it's required.
6. `clientManifest` was moved into `nodes/agent/callers/index.ts`.
7. `errors.ts`, `types.ts` and `utils.ts` are in `nodes/agent` directory now.

Pretty much try to structure the client domain like `nodes/agent` directory.  If anything is unclear, make a comment here asking about it.

### Additional context

* Related #522
* Related #562 
* Related #560

### Tasks

* [x] 1. All the handlers were updated to use default exports.
* [x] 2. All the handlers were updated to use arrow functions.
* [x] 3. The `ServerManifest` was moved into the `index.ts`.
* [x] 4. The caller definitions were moved into `nodes/agent/callers`.
* [x] 5. The caller definitions were split into separate files. (I don't like this change, it means that each caller which can be defined in a single line need to be spread across 60+ files.) I should check with @CMCDragonkai to see if it's required.
* [x] 6. `clientManifest` was moved into `nodes/agent/callers/index.ts`.
* [x] 7. `errors.ts`, `types.ts` and `utils.ts` are in `nodes/agent` directory now.

### LINKED ISSUE #565: Convert `ClientService` to `StartStop`
### Specification

The `ClientService` is currently `CreateDestroyStartStop` with how the `RPCServer` and `WebSocketServer` works it makes more sense to be `StartStop`.

### Additional context

* Related #560 

### Tasks

1. Convert `ClientService` to `StartStop`.
2. Update usage of it where needed.
3. Update tests using it.

addievo:
### Description
The client domain is out of spec with the updated node domain. The goal of this PR is to match spec of client domain with nodes domain.

### Issues Fixed

* Fixes #563
* Fixes #565
* Fixes #520 

### Tasks
* [x] 1. All the handlers were updated to use default exports.
* [x] 2. All the handlers were updated to use arrow functions.
* [x] 3. The `ServerManifest` was moved into the `index.ts`.
* [x] 4. The caller definitions were moved into `client/callers`.
* [x] 5. The caller definitions were split into separate files. 
* [x] 6. `clientManifest` was moved into `client/callers/index.ts`.
* [x] 7. `errors.ts`, `types.ts` and `utils.ts` are in `client` directory now.
* [x] 8. Convert `ClientService` to `startStop`.
* [x] 9. Update everything in `tests/client`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
Yea make sure that `src/client/callers` and `src/client/handlers` **look exactly the same** as `src/nodes/agent`.

Default exports, and utility function structure, and update all imports elsewhere, and have a look at tests too.

Make sure to rebase on staging when ready for review too.

CMCDragonkai:
Name of the PR should be more descriptive. I'm changing it.

CMCDragonkai:
I think #565 should also be fixed here.

src/client/callers/agentLockAll.ts
```diff
@@ -0,0 +1,8 @@
+import {UnaryCaller} from "@matrixai/rpc";
+import {ClientRPCRequestParams, ClientRPCResponseResult} from "@/client/types";
+
+const agentLockAll = new UnaryCaller<
+  ClientRPCRequestParams,
+  ClientRPCResponseResult
+>();
```
tegefaulkes: The agent callers does something like this for the types.

``` ts
type CallerTypes = HandlerTypes<NodesClaimsGetHandler>;

const nodesClaimsGet = new ServerCaller<
  CallerTypes['input'],
  CallerTypes['output']
>();
```

This needs to be replicated across all the callers as well.

src/client/handlers/agentLockAll.ts
```diff
@@ -18,4 +18,4 @@ class AgentLockAllHandler extends UnaryHandler<
   };
 }
 
-export { AgentLockAllHandler };
+export default AgentLockAllHandler;
```
tegefaulkes: The names of the handlers was also updated. So `AgentLockAllHandler` should become `AgentLockAll`. This applies to all handlers

src/client/types.ts
```diff
@@ -24,3 +30,220 @@ type ClientRPCResponseResult<
...
+  identityId: string;
+};
+export type ProviderSearchMessage = {
+  authIdentityId?: string;
+  identityId: string;
+  disconnected: boolean;
+  limit?: number;
+  searchTermList?: Array<string>;
+  providerIdList: Array<string>;
+};
```
tegefaulkes: In stead of exporting where the type is defined, they need to be exported at the bottom. Refer to how the `nodes/agent/types.ts` did this.

addievo:
@tegefaulkes go over ClientService, there might be a little things wrong about it, I couldn't quite figure out how to implement StartStop that well

src/client/callers/agentLockAll.ts
addievo: Could you elaborate on this?

src/client/handlers/agentLockAll.ts
addievo: ok i'll update the 63 handlers yet again

src/client/types.ts
addievo: Resolved

src/client/callers/agentLockAll.ts
tegefaulkes: Look at how the `nodes/agent/callers` do this and do the same thing.

src/client/handlers/agentLockAll.ts
tegefaulkes: You can do a find and replace to just remove the `Handler` from the names.

src/client/types.ts
tegefaulkes: Is it? I don't see any changes for it.

addievo: Not pushed yet

src/client/ClientService.ts
```diff
@@ -16,34 +16,32 @@ import * as errors from './errors';
...
+                         middlewareFactory,
+                         host = config.defaultsUser.clientServiceHost,
+                         port = config.defaultsUser.clientServicePort,
+                         keepAliveTimeoutTime = config.defaultsSystem.clientKeepAliveTimeoutTime,
+                         keepAliveIntervalTime = config.defaultsSystem.clientKeepAliveIntervalTime,
+                         rpcCallTimeoutTime = config.defaultsSystem.rpcCallTimeoutTime,
+                         rpcParserBufferSize = config.defaultsSystem.rpcParserBufferSize,
+                       },
+                       logger = new Logger(),
+                     }: {
```
tegefaulkes: New logger needs a name.

src/client/handlers/agentLockAll.ts
addievo: > You can do a find and replace to just remove the `Handler` from the names.

That messes up all the imports, I need to refactor each one manually by renaming it

src/client/ClientService.ts
```diff
@@ -87,48 +81,40 @@ class ClientService {
...
+                         host = config.defaultsUser.clientServiceHost,
+                         port = config.defaultsUser.clientServicePort,
+                       },
+                     }: {
+    manifest: ServerManifest;
+    options: {
+      host?: string;
+      port?: number;
+    };
+  }): Promise<void> {
```
tegefaulkes: Did you move the start method? It shouldn't be moved.

tegefaulkes (commented):
Two small things stand out, otherwise seems good.

src/client/callers/agentLockAll.ts
CMCDragonkai: Get the CallerTypes.

src/client/ClientService.ts
addievo: Wdym by moved?

src/client/ClientService.ts
addievo: Gotcha

addievo: Its not lintfixed yet so stuff might look a lil jank, I'll lint it before merge

src/client/ClientService.ts
tegefaulkes: The start method has been moved to just after the constructor? the order should be something like

1. properties
2. handlers
3. constructor
4. getters and setters
5.  async init functions like start and stop
6. public methods
7. private methods

Refer to how other start stop classes are structured.

tegefaulkes: Did you fix this?

src/client/callers/agentLockAll.ts
```diff
@@ -0,0 +1,11 @@
+import { HandlerTypes, UnaryCaller } from "@matrixai/rpc";
+import NodesClosestLocalNodesGetHandler from "../../nodes/agent/handlers/NodesClosestLocalNodesGet";
+
+type CallerTypes = HandlerTypes<NodesClosestLocalNodesGetHandler>;
+
```
tegefaulkes: Why are all the callers importing `NodesClosestLocalNodesGetHandler`?

They should import their respective handler as a type import `import type AgentLockAll from '../handlers/AgentLockAll`

Importing as a type is critical here. We're importing the handler types here so we can derive the caller types. That's why we're making this change here.

src/client/callers/agentLockAll.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const agentLockAll = new UnaryCaller<
+  CallerTypes['input'],
+  CallerTypes['output']
+>();
+export default agentLockAll;
```
tegefaulkes: Don't forget proper spacing, linter won't fix this.

src/client/callers/agentUnlock.ts
```diff
@@ -0,0 +1,10 @@
+import { HandlerTypes, UnaryCaller } from "@matrixai/rpc";
+import {ClientRPCRequestParams, ClientRPCResponseResult} from "../../client/types";
```
tegefaulkes: we're in the client directory, the path should be `../types`

tegefaulkes: Actually, this should be removed. we don't use these types directly anymore. Linter should've complained about that so you still need to lint.

CMCDragonkai:
I'm taking over this. Will note all things missing.

1. `src/client/index.ts` isn't aligned.
2. `src/client/middleware.ts` and `src/client/timeoutMiddleware.ts` and `src/client/authenticationMiddleware.ts` is really awkward. This might be better put together.
3. Code isn't linted.
4. Probably not tested.
5. Imports isn't aligned.
6. Spacing isn't aligned.
7. Indeed the `import type XHandler` should have been `import type X`.
8. Dependencies should be type imports for `src/client/handlers/index.ts`

@addievo don't touch anything.

src/client/callers/agentLockAll.ts
CMCDragonkai: Doing this everywhere.

CMCDragonkai:
Client service I think is a bit broken. Will need to fix it.

src/client/ClientService.ts
CMCDragonkai: Done.

CMCDragonkai:
Rebased on top of staging.

There are test failures here... could it be due to import changes?

CMCDragonkai:
Agent handlers were not aligned properly... just noticed this. File naming wasn't done correctly.

CMCDragonkai:
Ok so now we are just in the tedious process of updating all the `tests/client` imports and `ClientService` construction. Furthermore, alot of the tests is repeating the setup alot, they were originally separate files originally but @tegefaulkes gathered them together to reduce memory usage.

tests/client/handlers/keys.test.ts
```diff
@@ -882,51 +879,50 @@ describe('keysPublicKey', () => {
...
+    });
+  });
+  test('gets the public key', async () => {
+    const response = await rpcClient.methods.keysPublicKey({
+      password,
     });
-
-    // Doing the test
-    const response = await rpcClient.methods.keysPublicKey({});
     expect(response.publicKeyJwk).toEqual({
```
CMCDragonkai: @tegefaulkes see this. Before no `password`, now it is required by the type because that's what the handler actually requires.

However if you force it to take no `password` the test passes. How is this possible? Why is there no authentication of the password? If I give it a wrong password, it still accepts it.

CMCDragonkai:
Should be finished in a next hr or so, just need confirmation on why that password behaviour is happening. @tegefaulkes

tests/client/handlers/keys.test.ts
tegefaulkes: I'm a little lost here. What password are you talking about?

tegefaulkes: Oh I think I see, The handler is wrong, It shouldn't take a password as the input. The input type should be `ClientRPCRequestParams`

CMCDragonkai: There's a few other commands taking password too. Are they also wrong?

tegefaulkes: It might be but depends on the command. Getting the private key required a password from what I recall.

CMCDragonkai: From my understanding, password shouldn't be transferred like this if it is for authentication purposes. Auth relies on the metadata which is supposed to be the session token. But the root password could be used.

So authentication wise, it shouldn't be in the input message. The auth middleware is supposed to take over this.

Only in cases like changing passwords would this make sense.

CMCDragonkai: Only `keysPublicKey` had to be changed. This is fixed. This also means in you need to update `Polykey-CLI` @tegefaulkes

CMCDragonkai:
All tests in `tests/client` is passing now. Some things of note:

1. Import order and import deduplication has not been maintained, this resulted in lots of messy imports that was very confusing. This needs discipline here.
2. The `keysPublicKey` command does not take a password message.
3. The `notificationsSend` handler had the wrong type.
4. All the cases where we are repeating test setup should be moved into describe.
5. There's a lack of care taken when writing the order of declarations, the order should be natural following the order of usage.
6. I think overall there's too much reliance on the IDE happening here, here's a simple trick, go get the `ag` command and always run it against your directory to see if things are being mis-used, and always check your imports and exports before you finish. This can save alot of time at the end.
gs
Most importantly! The vault tests have todos for some of the most important commands!!! These are the commands that we are actually demoing, and these require actual tests.

```
✎ todo clones a vault
✎ todo pulls from a vault
✎ todo scans a vault
```

Why are we testing small things, and not the big things? This needs to be done as we are deploying and reviewing the UX.

CMCDragonkai:
BTW @amydevs when sanity checking final build, what I do is `npm run build` then I do a `node` and I do `require('./dist/index.js')`.

PR #579, opened 2023-10-13
Improve search iteration for Github identity search

### LINKED ISSUE #578: GitHub Identity Search is slow.
### Describe the bug
GitHub Identity Search is slow.

### To Reproduce

1. Search for followers/following on a github acc.

### Expected behavior

Should be faster

addievo:
### Description

PR aims to improve github identity search since it was very slow iterating over every user followed / following. Now it does a rate search and Promise.all if current rate limit allows it.

### Issues Fixed

* Fixes #578

### Tasks

- [x] 1. Introduce rate limit
- [x] 2. Introduce Promise.all for followers/following.
- [x] 3. Default back to old search if exceeds rate limit.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/identities/providers/github/GitHubProvider.ts
+++ b/src/identities/providers/github/GitHubProvider.ts
@@ -280,7 +280,6 @@ class GitHubProvider extends Provider {
       );
     }
     providerToken = await this.checkToken(providerToken, authIdentityId);
-
     const foundIdentityIds: Set<IdentityId> = new Set();
     for (const identityGroup of ['following', 'followers'] as const) {
       let cursor: string | undefined;
@@ -289,7 +288,7 @@ class GitHubProvider extends Provider {
           `${this.apiUrl}/graphql`,
           {
             method: 'POST',
-            body: this.getConnectedIdentityDatasRequestBody(authIdentityId, identityGroup, cursor)
+            body: this.getConnectedIdentityDatasGraphQLBody(authIdentityId, identityGroup, cursor)
           },
           providerToken,
         );
@@ -322,7 +321,10 @@ class GitHubProvider extends Provider {
             }
           );
         }
-        const foundIdentityData: any[] = data?.data?.user?.[identityGroup]?.nodes ?? [];
+        // FollowerConnection and FollowingConnection always exists on User
+        const foundIdentityGroupData = data.data.user[identityGroup];
+        // Array<User> always exists on FollowerConnection and FollowingConnection
+        const foundIdentityData: IdentityData[] = foundIdentityGroupData.nodes;
         for (const identityData of foundIdentityData) {
           identityData.providerId = this.id;
           if (!foundIdentityIds.has(identityData.identityId) && identitiesUtils.matchIdentityData(identityData, searchTerms)) {
@@ -333,7 +335,8 @@ class GitHubProvider extends Provider {
         if (foundIdentityData.length === 0) {
           break;
         } else {
-          const endCursor: string | undefined = data?.data?.user?.followers?.pageInfo.endCursor;
+          // endCursor may be nullish if this is the last page
+          const endCursor: string | null = foundIdentityGroupData.pageInfo.endCursor;
           if (endCursor == null) break;
           cursor = endCursor;
         }
@@ -341,7 +344,22 @@ class GitHubProvider extends Provider {
     }
   }
 
-  protected getConnectedIdentityDatasRequestBody(
+  /**
+   * Returns a string suitable for use as the request body to the GitHub GraphQL endpoint.
+   * This is used to construct a query that returns either the `followers` or the `following` of a user.
+   *
+   * Schemas Used:
+   * - https://docs.github.com/en/graphql/reference/queries#user
+   * - https://docs.github.com/en/graphql/reference/objects#user
+   * - https://docs.github.com/en/graphql/reference/objects#followerconnection
+   *
+   * @param authIdentityId - The GitHub authentication token to use when getting user data
+   * @param identityGroup - Specify whether the GraphQL query requests the `followers` or the `following` of a user
+   * @param cursor - cursor for pagination,
+   * this can be retrieved from `.data.user[identityGroup].pageinfo.endCursor`
+   * of the JSON body on a response for a request made with the return value of this method as the body.
+   */
+  protected getConnectedIdentityDatasGraphQLBody(
     authIdentityId: IdentityId,
     identityGroup: 'following' | 'followers',
     cursor?: string
--- a/src/identities/providers/github/GitHubProvider.ts
+++ b/src/identities/providers/github/GitHubProvider.ts
@@ -266,48 +266,6 @@ class GitHubProvider extends Provider {
     };
   }
 
-  public async fetchAllFollowing(
-    authIdentityId: IdentityId,
-    token: ProviderToken,
-  ): Promise<any[]> {
-    let following = [];
-    let page = 1;
-    while (true) {
-      const request = this.createRequest(
-        `${this.apiUrl}/user/following?per_page=100&page=${page}`,
-        { method: 'GET' },
-        token,
-      );
-      const response = await fetch(request);
-      const data = await response.json();
-      following = following.concat(data);
-      if (data.length < 100) break;
-      page++;
-    }
-    return following;
-  }
-
-  public async fetchAllFollowers(
-    authIdentityId: IdentityId,
-    token: ProviderToken,
-  ): Promise<any[]> {
-    let followers = [];
-    let page = 1;
-    while (true) {
-      const request = this.createRequest(
-        `${this.apiUrl}/user/followers?per_page=100&page=${page}`,
-        { method: 'GET' },
-        token,
-      );
-      const response = await fetch(request);
-      const data = await response.json();
-      followers = followers.concat(data);
-      if (data.length < 100) break;
-      page++;
-    }
-    return followers;
-  }
-
   /**
    * Gets connected IdentityData from following and follower connections.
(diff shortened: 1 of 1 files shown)


src/identities/providers/github/GitHubProvider.ts
```diff
@@ -280,102 +280,92 @@ class GitHubProvider extends Provider {
...
+          pageInfo {
+            endCursor
+            startCursor
+          }
+          totalCount
+        }
+      }
+    }`;
+    return JSON.stringify({ query });
+  }
```
CMCDragonkai, line 385: This is a GraphQL request?

src/identities/providers/github/GitHubProvider.ts
```diff
@@ -280,102 +280,92 @@ class GitHubProvider extends Provider {
...
-          method: 'GET',
-        },
-        providerToken,
-      );
-      const response = await fetch(request);
-      if (!response.ok) {
-        if (response.status === 401) {
-          throw new identitiesErrors.ErrorProviderUnauthenticated(
-            `Invalid access token`,
+
```
CMCDragonkai: Get rid of un-used newlines.

src/identities/providers/github/GitHubProvider.ts
```diff
@@ -280,102 +280,92 @@ class GitHubProvider extends Provider {
...
+    const foundIdentityIds: Set<IdentityId> = new Set();
+    for (const identityGroup of ['following', 'followers'] as const) {
+      let cursor: string | undefined;
+      while (true) {
+        const request = this.createRequest(
+          `${this.apiUrl}/graphql`,
+          {
+            method: 'POST',
+            body: this.getConnectedIdentityDatasRequestBody(authIdentityId, identityGroup, cursor)
```
CMCDragonkai: Since it's a special request constructor, can you indicate the name of it as some sort of graphql request?

Also I need docblock comments explaining how it works, and which part of GitHub API documentation you're referring to.

src/identities/providers/github/GitHubProvider.ts
```diff
@@ -280,102 +280,92 @@ class GitHubProvider extends Provider {
...
-        const identityData = await this.getIdentityData(
-          authIdentityId,
-          item.login,
-        );
-        if (
-          identityData &&
-          identitiesUtils.matchIdentityData(identityData, searchTerms)
-        ) {
-          yield identityData;
+        const foundIdentityData: any[] = data?.data?.user?.[identityGroup]?.nodes ?? [];
```
CMCDragonkai: How robust is this? And why `any` type?

src/identities/providers/github/GitHubProvider.ts
```diff
@@ -280,102 +280,92 @@ class GitHubProvider extends Provider {
...
+          identityData.providerId = this.id;
+          if (!foundIdentityIds.has(identityData.identityId) && identitiesUtils.matchIdentityData(identityData, searchTerms)) {
+            foundIdentityIds.add(identityData.identityId);
+            yield identityData;
+          }
+        }
+        if (foundIdentityData.length === 0) {
+          break;
+        } else {
+          const endCursor: string | undefined = data?.data?.user?.followers?.pageInfo.endCursor;
```
CMCDragonkai: Same here, how robust is this?

src/identities/providers/github/GitHubProvider.ts
CMCDragonkai, line 1: Did you test possible edge cases? What happens when request fails? Gets rate limited, unauthenticated... etc?

CMCDragonkai:
Needs to be squashed and commentary resolved.

src/identities/providers/github/GitHubProvider.ts
addievo: We plan to add rate check before polling requests here.

amydevs:
```
const response = await fetch(request);
        if (!response.ok) {
          if (response.status === 401) {
            throw new identitiesErrors.ErrorProviderUnauthenticated(
              `Invalid access token`,
            );
          }
          throw new identitiesErrors.ErrorProviderCall(
            `Provider responded with ${response.status} ${response.statusText}`,
          );
        }
        let data;
        try {
          data = await response.json();
        } catch (e) {
          throw new identitiesErrors.ErrorProviderCall(
            `Provider response body is not valid JSON`,
            { cause: e },
          );
        }
        const error = data?.errors?.at?.(0);
        if (error != null) {
          throw new identitiesErrors.ErrorProviderCall(
            `Provider response body contains an error: ${error.message}`,
            {
              data: error
            }
          );
        }
```

The first if statement and the try catch were carried over from the original, and they continue to handle JSON parse failures and Unauthenticated errors. The last if statement handles any errors that are returned by GraphQL. I've decided to only throw the first error for now, so that what is thrown from the asyncgenerator is consistent. But this can be made an aggregateerror instead.

amydevs: I've also decided to remove rate limiting for now, GraphQL API rates are limited by the hour, and it would be unrealistic to wait an hour for a `polykey identities search` if we were ever rate limited.

src/identities/providers/github/GitHubProvider.ts
amydevs: endCursor will always either not exist or be a string, never anything else. If pageInfo exists, then endCursor may or may not exist depending on the current page is the last page.

src/identities/providers/github/GitHubProvider.ts
amydevs: I decided to not use the IdentityData type here, because it doesn't contain all the required properties, which I add in when I iterate over the array. The ternery operators are used in case the server returns an error data structure, but realistically, as errors are caught before this line, `nodes` should always be an array and not undefined.

src/identities/providers/github/GitHubProvider.ts
amydevs: 1. user Query Gets a User with an auth token: https://docs.github.com/en/graphql/reference/queries#user
2. The User object contains the FollowerConnection object: https://docs.github.com/en/graphql/reference/objects#user
3. The FollowerConnection object Contains User objects: https://docs.github.com/en/graphql/reference/objects#followerconnection
4. The User object contains the following fields:
    - login
    - name
    - email
    - name

src/identities/providers/github/GitHubProvider.ts
amydevs: yes, the server expects the client to send a POJO containing the `query` property, which is a graphql query string.

src/identities/providers/github/GitHubProvider.ts
amydevs: fixed:
```ts
// FollowerConnection and FollowingConnection always exists on User
        const foundIdentityGroupData = data.data.user[identityGroup];
        // Array<User> always exists on FollowerConnection and FollowingConnection
        const foundIdentityData: IdentityData[] = foundIdentityGroupData.nodes;
        for (const identityData of foundIdentityData) {
          identityData.providerId = this.id;
          if (!foundIdentityIds.has(identityData.identityId) && identitiesUtils.matchIdentityData(identityData, searchTerms)) {
            foundIdentityIds.add(identityData.identityId);
            yield identityData;
          }
        }
        if (foundIdentityData.length === 0) {
          break;
        } else {
          // endCursor may be nullish if this is the last page
          const endCursor: string | null = foundIdentityGroupData.pageInfo.endCursor;
          if (endCursor == null) break;
          cursor = endCursor;
        }
```

src/identities/providers/github/GitHubProvider.ts
amydevs: fixed
```ts
 // FollowerConnection and FollowingConnection always exists on User
        const foundIdentityGroupData = data.data.user[identityGroup];
        // Array<User> always exists on FollowerConnection and FollowingConnection
        const foundIdentityData: IdentityData[] = foundIdentityGroupData.nodes;
```

src/identities/providers/github/GitHubProvider.ts
amydevs: Fixed:
```ts
/**
   * Returns a string suitable for use as the request body to the GitHub GraphQL endpoint.
   * This is used to construct a query that returns either the `followers` or the `following` of a user.
   *
   * Schemas Used:
   * - https://docs.github.com/en/graphql/reference/queries#user
   * - https://docs.github.com/en/graphql/reference/objects#user
   * - https://docs.github.com/en/graphql/reference/objects#followerconnection
   *
   * @param authIdentityId - The GitHub authentication token to use when getting user data
   * @param identityGroup - Specify whether the GraphQL query requests the `followers` or the `following` of a user
   * @param cursor - cursor for pagination,
   * this can be retrieved from `.data.user[identityGroup].pageinfo.endCursor`
   * of the JSON body on a response for a request made with the return value of this method as the body.
   */
  protected getConnectedIdentityDatasGraphQLBody(
    authIdentityId: IdentityId,
    identityGroup: 'following' | 'followers',
    cursor?: string
  ): string {
```

src/identities/providers/github/GitHubProvider.ts
CMCDragonkai: Ok so it will just be:

```
          throw new identitiesErrors.ErrorProviderCall(
            `Provider responded with ${response.status} ${response.statusText}`,
          );
```

CMCDragonkai:
It's now instant, great!

Originally it was 2 requests, one to the page, another to get the user detail for each user. Now it is just 1 request to the page, which also contains the additional information.

PR #582, opened 2023-10-13
Encapsulate WebSocketClient in PolykeyClient

addievo:
### Description

This PR aims to encapsulate WebSocketClient in PolykeyClient.

At the same time `Options` usage has to be limited to only `PolykeyAgent` and `PolykeyClient`.

Because:

1. Subdomains aren't really meant to have "options". They are just groups of classes closely related, that may or may not be factored out in a separate package one day.
2. Classes themselves have options - more correctly as parameters. As parameters, they should be top level, not nested within a `options`. Why? Because you can just do `X.createX({ param1, param2 });`, no need to do `X.createX({ options: { param1, param2 });` as the `options` is superfluous.
3. But why does `PolykeyAgent` and `PolykeyClient` have `options`? The reason is simply because there is so many parameters, that it is useful to have a separate type to keep track of them all. But this is exceptional to `PolykeyAgent` and `PolykeyClient`, it should not be considered the default way of creating classes in our codebase. And this type `PolykeyAgentOptions` and `PolykeyClientOptions` will be used in various places, in particular tests that need to run many different variants of the Polykey program.
4. There should necessarily be an exact relationship between the parameters in each domain class, and the options of `PolykeyAgent` and `PolykeyClient`. We want to retain some flexibility here.

Thus we will design which will be put into `src/types.ts`:

```
type PolykeyAgentOptions
type PolykeyClientOptions
```

But all other domains will not do such a thing.

### Tasks

- [x] 1. Align `PolykeyClientOptions` with `PolykeyAgentOptions`
- [x] 2. Ensure that subdomains are just flattened parameters, not `options` which is unique only to `PolykeyClient` and `PolykeyAgent`.
- [x] 3. Encapsulate `WebSocketClient` into `PolykeyClient` and brought back `nodeId` as a necessary parameter, verification of the client service certificate is now done within `PolykeyClient`.
- [x] 4. Made `PolykeyClient` a `@timedCancellable` operation with the timeout defaulted to `config.defaultsSystem.clientConnectTimeoutTime`. This propagates down to the `await WebSocketClient.createWebSocketClient`.
- [x] 5. Created password ops limits and mem limits defaults in the config using the raw value and added tests to ensure that the raw values is maintained. It is now configured at the top level.
- [x] 6. Update all tests using `PolykeyAgent` and `PolykeyClient` and subdomains which was relying upon `options`, these call signatures are all changed.
- [x] 7. Test that the timed cancellable works for `PolykeyClient`.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -1,8 +1,8 @@
 import type { DeepPartial, FileSystem } from './types';
 import type { PolykeyWorkerManagerInterface } from './workers/types';
 import type { TLSConfig } from './network/types';
-import type { SeedNodes, NodesOptions } from './nodes/types';
-import type { Key, KeysOptions } from './keys/types';
+import type { SeedNodes } from './nodes/types';
+import type { Key, PasswordOpsLimit, PasswordMemLimit } from './keys/types';
 import path from 'path';
 import process from 'process';
 import Logger from '@matrixai/logger';
@@ -54,26 +54,30 @@ type PolykeyAgentOptions = {
   seedNodes: SeedNodes;
   workers: number;
   ipv6Only: boolean;
-  keys: KeysOptions;
-  rpc: {
-    callTimeoutTime: number;
-    parserBufferSize: number;
+  keys: {
+    passwordOpsLimit: PasswordOpsLimit;
+    passwordMemLimit: PasswordMemLimit;
+    strictMemoryLock: boolean;
+    certDuration: number;
+    certRenewLeadTime: number;
+    recoveryCode: string;
   };
   client: {
-    connectTimeoutTime: number;
     keepAliveTimeoutTime: number;
     keepAliveIntervalTime: number;
+    rpcCallTimeoutTime: number;
+    rpcParserBufferSize: number;
   };
-  nodes: NodesOptions;
-};
-
-type PolykeyAgentStartOptions = {
-  clientServiceHost: string;
-  clientServicePort: number;
-  agentServiceHost: string;
-  agentServicePort: number;
-  ipv6Only: boolean;
-  workers: number;
+  nodes: {
+    connectionIdleTimeoutTime: number;
+    connectionFindConcurrencyLimit: number;
+    connectionConnectTimeoutTime: number;
+    connectionKeepAliveTimeoutTime: number;
+    connectionKeepAliveIntervalTime: number;
+    connectionHolePunchIntervalTime: number;
+    rpcCallTimeoutTime: number;
+    rpcParserBufferSize: number;
+  };
 };
 
 interface PolykeyAgent extends CreateDestroyStartStop {}
@@ -113,9 +117,9 @@ class PolykeyAgent {
   }: {
     password: string;
     options?: DeepPartial<PolykeyAgentOptions>;
+    fresh?: boolean;
     fs?: FileSystem;
     logger?: Logger;
-    fresh?: boolean;
   }): Promise<PolykeyAgent> {
     logger.info(`Creating ${this.name}`);
     const umask = 0o077;
@@ -131,18 +135,18 @@ class PolykeyAgent {
       workers: config.defaultsUser.workers,
       ipv6Only: config.defaultsUser.ipv6Only,
       keys: {
+        passwordOpsLimit: config.defaultsUser.passwordOpsLimit,
+        passwordMemLimit: config.defaultsUser.passwordMemLimit,
+        strictMemoryLock: config.defaultsUser.strictMemoryLock,
         certDuration: config.defaultsUser.certDuration,
         certRenewLeadTime: config.defaultsUser.certRenewLeadTime,
       },
-      rpc: {
-        callTimeoutTime: config.defaultsSystem.rpcCallTimeoutTime,
-        parserBufferSize: config.defaultsSystem.rpcParserBufferSize,
-      },
       client: {
-        connectTimoutTime: config.defaultsSystem.clientConnectTimeoutTime,
         keepAliveTimeoutTime: config.defaultsSystem.clientKeepAliveTimeoutTime,
         keepAliveIntervalTime:
           config.defaultsSystem.clientKeepAliveIntervalTime,
+        rpcCallTimeoutTime: config.defaultsSystem.rpcCallTimeoutTime,
+        rpcParserBufferSize: config.defaultsSystem.rpcParserBufferSize,
       },
       nodes: {
         connectionIdleTimeoutTime:
@@ -158,7 +162,7 @@ class PolykeyAgent {
         connectionHolePunchIntervalTime:
           config.defaultsSystem.nodesConnectionHolePunchIntervalTime,
       },
-    });
+    }) as PolykeyAgentOptions;
     // This can only happen if the caller didn't specify the node path and the
     // automatic detection failed
     if (optionsDefaulted.nodePath == null) {
@@ -181,7 +185,6 @@ class PolykeyAgent {
     const dbPath = path.join(statePath, config.paths.dbBase);
     const keysPath = path.join(statePath, config.paths.keysBase);
     const vaultsPath = path.join(statePath, config.paths.vaultsBase);
-
     let status: Status | undefined;
     let schema: Schema | undefined;
     let keyRing: KeyRing | undefined;
@@ -217,7 +220,9 @@ class PolykeyAgent {
       });
       keyRing = await KeyRing.createKeyRing({
         keysPath,
-        options: optionsDefaulted.keys,
+        passwordOpsLimit: optionsDefaulted.keys.passwordOpsLimit,
+        passwordMemLimit: optionsDefaulted.keys.passwordMemLimit,
+        strictMemoryLock: optionsDefaulted.keys.strictMemoryLock,
         fs,
         fresh,
         password,
@@ -256,7 +261,8 @@ class PolykeyAgent {
         db,
         keyRing,
         taskManager,
-        options: optionsDefaulted.keys,
+        certDuration: optionsDefaulted.keys.certDuration,
+        certRenewLeadTime: optionsDefaulted.keys.certRenewLeadTime,
         logger: logger.getChild(CertManager.name),
         fresh,
       });
@@ -310,7 +316,20 @@ class PolykeyAgent {
         nodeGraph,
         tlsConfig,
         seedNodes: optionsDefaulted.seedNodes,
-        options: optionsDefaulted.nodes,
+        connectionFindConcurrencyLimit:
+          optionsDefaulted.nodes.connectionFindConcurrencyLimit,
+        connectionIdleTimeoutTime:
+          optionsDefaulted.nodes.connectionIdleTimeoutTime,
+        connectionConnectTimeoutTime:
+          optionsDefaulted.nodes.connectionConnectTimeoutTime,
+        connectionKeepAliveTimeoutTime:
+          optionsDefaulted.nodes.connectionKeepAliveTimeoutTime,
+        connectionKeepAliveIntervalTime:
+          optionsDefaulted.nodes.connectionKeepAliveIntervalTime,
+        connectionHolePunchIntervalTime:
+          optionsDefaulted.nodes.connectionHolePunchIntervalTime,
+        rpcParserBufferSize: optionsDefaulted.nodes.rpcParserBufferSize,
+        rpcCallTimeoutTime: optionsDefaulted.nodes.rpcCallTimeoutTime,
         logger: logger.getChild(NodeConnectionManager.name),
       });
       nodeManager = new NodeManager({
@@ -387,8 +406,8 @@ class PolykeyAgent {
         ),
         keepAliveTimeoutTime: optionsDefaulted.client.keepAliveTimeoutTime,
         keepAliveIntervalTime: optionsDefaulted.client.keepAliveIntervalTime,
-        rpcCallTimeoutTime: optionsDefaulted.rpc.callTimeoutTime,
-        rpcParserBufferSize: optionsDefaulted.rpc.parserBufferSize,
+        rpcCallTimeoutTime: optionsDefaulted.client.rpcCallTimeoutTime,
+        rpcParserBufferSize: optionsDefaulted.client.rpcParserBufferSize,
         logger: logger.getChild(ClientService.name),
       });
     } catch (e) {
@@ -585,7 +604,14 @@ class PolykeyAgent {
     fresh = false,
   }: {
     password: string;
-    options?: Partial<PolykeyAgentStartOptions>;
+    options?: Partial<{
+      clientServiceHost: string;
+      clientServicePort: number;
+      agentServiceHost: string;
+      agentServicePort: number;
+      ipv6Only: boolean;
+      workers: number;
+    }>;
     workers?: number;
     fresh?: boolean;
   }) {
--- a/src/PolykeyClient.ts
+++ b/src/PolykeyClient.ts
@@ -1,24 +1,39 @@
-import type { FileSystem } from './types';
-import type { StreamFactory } from '@matrixai/rpc';
+import type { PromiseCancellable } from '@matrixai/async-cancellable';
+import type { ContextTimed, ContextTimedInput } from '@matrixai/contexts';
+import type { DeepPartial, FileSystem } from './types';
+import type { NodeId } from './ids/types';
 import path from 'path';
 import Logger from '@matrixai/logger';
-import { CreateDestroyStartStop } from '@matrixai/async-init/dist/CreateDestroyStartStop';
-import { RPCClient } from '@matrixai/rpc';
-import { middleware as rpcMiddleware } from '@matrixai/rpc';
-import * as clientMiddleware from './client/middleware';
+import {
+  CreateDestroyStartStop,
+  ready,
+} from '@matrixai/async-init/dist/CreateDestroyStartStop';
+import { timedCancellable, context } from '@matrixai/contexts/dist/decorators';
+import { WebSocketClient } from '@matrixai/ws';
+import { RPCClient, middleware as rpcMiddleware } from '@matrixai/rpc';
 import { Session } from './sessions';
+import * as ids from './ids';
 import * as utils from './utils';
 import * as errors from './errors';
 import * as events from './events';
-import config from './config';
 import * as networkUtils from './network/utils';
+import * as validationErrors from './validation/errors';
+import * as clientUtils from './client/utils';
+import * as clientMiddleware from './client/middleware';
 import clientClientManifest from './client/callers';
+import config from './config';
 
 /**
- * This PolykeyClient would create a new PolykeyClient object that constructs
- * a new RPCClient which attempts to connect to an existing PolykeyAgent's
- * RPC server.
+ * Optional configuration for`PolykeyClient`.
  */
+type PolykeyClientOptions = {
+  nodePath: string;
+  keepAliveTimeoutTime: number;
+  keepAliveIntervalTime: number;
+  rpcCallTimeoutTime: number;
+  rpcParserBufferSize: number;
+};
+
 interface PolykeyClient extends CreateDestroyStartStop {}
 @CreateDestroyStartStop(
   new errors.ErrorPolykeyClientRunning(),
@@ -33,48 +48,127 @@ interface PolykeyClient extends CreateDestroyStartStop {}
   },
 )
 class PolykeyClient {
-  static async createPolykeyClient({
-    nodePath = config.defaultsUser.nodePath,
-    streamFactory,
-    streamKeepAliveTimeoutTime,
-    parserBufferByteLimit,
-    fs = require('fs'),
-    logger = new Logger(this.name),
-    fresh = false,
-  }: {
-    nodePath?: string;
-    streamFactory: StreamFactory;
-    streamKeepAliveTimeoutTime?: number;
-    parserBufferByteLimit?: number;
-    fs?: FileSystem;
-    logger?: Logger;
-    fresh?: boolean;
-  }): Promise<PolykeyClient> {
+  /**
+   * Creates a Polykey Client.
+   *
+   * @param opts
+   * @param opts.nodeId
+   * @param opts.host
+   * @param opts.port
+   * @param ctx
+   */
+  public static createPolykeyClient(
+    opts: {
+      nodeId: string | NodeId;
+      port: number;
+      host: string;
+      options?: DeepPartial<PolykeyClientOptions>;
+      fresh?: boolean;
+      fs?: FileSystem;
+      logger?: Logger;
+    },
+    ctx?: Partial<ContextTimedInput>,
+  ): PromiseCancellable<PolykeyClient>;
+  @timedCancellable(
+    true,
+    config.defaultsSystem.clientConnectTimeoutTime,
+    errors.ErrorPolykeyClientCreateTimeout,
+  )
+  public static async createPolykeyClient(
+    {
+      // Required parameters
+      nodeId,
+      host,
+      port,
+      // Options
+      options = {},
+      fresh = false,
+      // Optional dependencies
+      fs = require('fs'),
+      logger = new Logger(this.name),
+    }: {
+      nodeId: string | NodeId;
+      host: string;
+      port: number;
+      options?: DeepPartial<PolykeyClientOptions>;
+      fresh?: boolean;
+      fs?: FileSystem;
+      logger?: Logger;
+    },
+    @context ctx: ContextTimed,
+  ): Promise<PolykeyClient> {
     logger.info(`Creating ${this.name}`);
-    if (nodePath == null) {
+    let nodeId_: NodeId;
+    if (typeof nodeId === 'string') {
+      try {
+        nodeId_ = ids.parseNodeId(nodeId);
+      } catch (e) {
+        if (e instanceof validationErrors.ErrorParse) {
+          throw new errors.ErrorPolykeyClientNodeIdInvalid(
+            'Encoded node ID must be a multibase base32hex encoded public-key',
+            {
+              cause: e,
+              data: { nodeId },
+            },
+          );
+        }
+        throw e;
+      }
+    } else {
+      nodeId_ = nodeId;
+    }
+    const optionsDefaulted = utils.mergeObjects(options, {
+      nodePath: config.defaultsUser.nodePath,
+      connectTimeoutTime: config.defaultsSystem.clientConnectTimeoutTime,
+      keepAliveTimeoutTime: config.defaultsSystem.clientKeepAliveTimeoutTime,
+      keepAliveIntervalTime: config.defaultsSystem.clientKeepAliveIntervalTime,
+      callTimeoutTime: config.defaultsSystem.rpcCallTimeoutTime,
+      parserBufferSize: config.defaultsSystem.rpcParserBufferSize,
+    }) as PolykeyClientOptions;
+    if (optionsDefaulted.nodePath == null) {
       throw new errors.ErrorUtilsNodePath();
     }
-    await utils.mkdirExists(fs, nodePath);
-    const sessionTokenPath = path.join(nodePath, config.paths.tokenBase);
+    await utils.mkdirExists(fs, optionsDefaulted.nodePath);
+    const sessionTokenPath = path.join(
+      optionsDefaulted.nodePath,
+      config.paths.tokenBase,
+    );
     const session = await Session.createSession({
       sessionTokenPath,
       logger: logger.getChild(Session.name),
       fresh,
     });
-    const rpcClientClient = new RPCClient({
+    const webSocketClient = await WebSocketClient.createWebSocketClient(
+      {
+        host,
+        port,
+        config: {
+          verifyPeer: true,
+          verifyCallback: async (certs) => {
+            await clientUtils.verifyServerCertificateChain([nodeId_], certs);
+          },
+          keepAliveTimeoutTime: optionsDefaulted.keepAliveTimeoutTime,
+          keepAliveIntervalTime: optionsDefaulted.keepAliveIntervalTime,
+        },
+        logger: logger.getChild(WebSocketClient.name),
+      },
+      ctx,
+    );
+    const rpcClient = new RPCClient({
       manifest: clientClientManifest,
-      streamFactory,
+      streamFactory: () => webSocketClient.connection.newStream(),
       middlewareFactory: rpcMiddleware.defaultClientMiddlewareWrapper(
         clientMiddleware.middlewareClient(session),
-        parserBufferByteLimit,
+        optionsDefaulted.rpcParserBufferSize,
       ),
       toError: networkUtils.toError,
-      streamKeepAliveTimeoutTime,
+      streamKeepAliveTimeoutTime: optionsDefaulted.rpcCallTimeoutTime,
       logger: logger.getChild(RPCClient.name),
     });
     const pkClient = new this({
-      nodePath,
-      rpcClientClient: rpcClientClient,
+      nodePath: optionsDefaulted.nodePath,
+      webSocketClient,
+      rpcClient,
       session,
       fs,
       logger,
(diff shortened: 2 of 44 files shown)


src/identities/providers/github/GitHubProvider.ts
CMCDragonkai, line 1: Use `Infinity`?

src/PolykeyClient.ts
CMCDragonkai, line 1: Do we even bother with injection? We already removed alot of optional dependencies from `PolykeyAgent`.

tegefaulkes:
Don't forget to update all code and test that depend on the `PolykeyClient`. That may just be the `PolykeyClient` tests in `tests/PolykeyClient.test.ts` but check for all usages in the code base.

CMCDragonkai:
Make 4. a separate PR, keep this PR to only PK client encapsulating WS client.

CMCDragonkai:
Can you address all comments.

src/identities/providers/github/GitHubProvider.ts
addievo: Using now

src/PolykeyClient.ts
addievo: Removed injectable

CMCDragonkai:
Rebase, squash down to 1 commit, and test what you're doing here, fix all tests.

CMCDragonkai:
Note in the future, we don't use reverts in feature branches. You're just supposed to move ahead, and eventually squash down.

tegefaulkes:
```
Summary of all failing tests
 FAIL  tests/PolykeyClient.test.ts (8.94 s)
  ● PolykeyClient › preserving and destroying session state

    TypeError: Cannot destructure property 'address' of 'undefined' as it is undefined.

      64 |       fresh,
      65 |     });
    > 66 |     const ws: WebSocketClient = new WebSocketClient();
         |                                 ^
      67 |     const rpcClientClient = new RPCClient({
      68 |       manifest: clientClientManifest,
      69 |       streamFactory,

      at new WebSocketClient (node_modules/@matrixai/ws/src/WebSocketClient.ts:292:5)
      at new WebSocketClient (node_modules/@matrixai/events/src/Evented.ts:55:9)
      at new WebSocketClient (node_modules/@matrixai/async-init/src/CreateDestroy.ts:49:26)
      at Function.createPolykeyClient (src/PolykeyClient.ts:66:33)
      at Object.<anonymous> (tests/PolykeyClient.test.ts:67:22)

  ● PolykeyClient › end to end with authentication logic

    TypeError: Cannot destructure property 'address' of 'undefined' as it is undefined.

      64 |       fresh,
      65 |     });
    > 66 |     const ws: WebSocketClient = new WebSocketClient();
         |                                 ^
      67 |     const rpcClientClient = new RPCClient({
      68 |       manifest: clientClientManifest,
      69 |       streamFactory,

      at new WebSocketClient (node_modules/@matrixai/ws/src/WebSocketClient.ts:292:5)
      at new WebSocketClient (node_modules/@matrixai/events/src/Evented.ts:55:9)
      at new WebSocketClient (node_modules/@matrixai/async-init/src/CreateDestroy.ts:49:26)
      at Function.createPolykeyClient (src/PolykeyClient.ts:66:33)
      at Object.<anonymous> (tests/PolykeyClient.test.ts:97:22)


Test Suites: 1 failed, 70 passed, 71 total
Tests:       2 failed, 5 skipped, 10 todo, 580 passed, 597 total
Snapshots:   0 total
Time:        193.513 s

```

Only the `PolykeyClient.test.ts` test is failing. Not surprising since...
1. You haven't actually updated the tests using `PolykeyClient`. 
2. You change is actually broken. `WebSocketClient` is `CreateDestroy`, so you shouldn't be constructing it directly with `new WebSocketClient()`. You should be using the `async-init` methods that are relevant.

CMCDragonkai:
Needs rebase.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "polykey",
-  "version": "1.2.1-alpha.8",
+  "version": "1.2.1-alpha.9",
   "homepage": "https://polykey.com",
   "author": "Matrix AI",
```
CMCDragonkai: What is this? Don't do this.

package-lock.json
CMCDragonkai, line 1: Don't do this.

src/PolykeyClient.ts
```diff
@@ -35,15 +37,19 @@ interface PolykeyClient extends CreateDestroyStartStop {}
...
     logger = new Logger(this.name),
     fresh = false,
   }: {
     nodePath?: string;
-    streamFactory: StreamFactory;
+    wsPort: number;
+    wsHost: string;
+    wsConfig?: Partial<WebSocketConfig>;
     streamKeepAliveTimeoutTime?: number;
     parserBufferByteLimit?: number;
```
CMCDragonkai: Change the signature so:

1. All required parameters up top
2. All optional parameters as `options = {}`
3. See `PolykeyAgent` - `options?: DeepPartial<PolykeyAgentOptions>;`

addievo:
image

src/PolykeyClient.ts
tegefaulkes: the `wsConfig` is wholesale optional here. We shouldn't do this. `Polykeyclient` should be providing the custom verification parameters. It should be up to the `PolykeyClient` to set up all of the verification logic.

CMCDragonkai: Any options can be spread into a general `options?: DeepPartial<PolykeyClientOptions>`. We're lifting the abstraction to the PK Client.

CMCDragonkai: @addievo declare a type above `PolykeyClientOptions` with the relevant options of PolyKey client, again, refer to `PolykeyAgent.ts`.

addievo: Done

package-lock.json
addievo: Removed

package.json
addievo: Removed

src/PolykeyClient.ts
CMCDragonkai: It is not done, you must provide a spec first and then we can review it.

CMCDragonkai:
Don't resolve unless you've pushed.

tegefaulkes:
I was going to look into this but it's actually pretty relevant to this PR.

The `src/client/utils.ts` `verifyServerCertificateChain` may have a problem. When verifying with an old `NodeId`, the custom verification is failing even though the old NodeId and new NodeId exist in the cert chain. 

There needs to be some tests added to test the custom verification logic for certain cases.
1. Verification success.
2. Verification failure. 
3. Verification succeeds using old node ID that is still valid in the cert chain.

CMCDragonkai:
This PR is not done.

1. The `PolykeyClientOptions` lacks a clear spec.
2. It needs to be considered in line with `PolykeyAgentOptions`
3. It needs to be defaulted using the `utils.mergeObject`
4. The rpc option naming between client and server needs to be aligned.
5. The ws options need to be filled automatically except for a few transport level defaults.

src/PolykeyAgent.ts
```diff
@@ -585,7 +576,14 @@ class PolykeyAgent {
...
     password: string;
-    options?: Partial<PolykeyAgentStartOptions>;
+    options?: Partial<{
+      clientServiceHost: string;
+      clientServicePort: number;
+      agentServiceHost: string;
+      agentServicePort: number;
+      ipv6Only: boolean;
+      workers: number;
+    }>;
```
tegefaulkes, line 614: Don't bother using `Partial<>` if it's not specified as a separate type.
```suggestion
    options?: {
      clientServiceHost?: string;
      clientServicePort?: number;
      agentServiceHost?: string;
      agentServicePort?: number;
      ipv6Only?: boolean;
      workers?: number;
    };
```

src/PolykeyClient.ts
```diff
@@ -1,10 +1,11 @@
-import type { FileSystem } from './types';
-import type { StreamFactory } from '@matrixai/rpc';
+import type { DeepPartial, FileSystem } from './types';
+import type { WebSocketConfig } from '@matrixai/ws/dist/types';
```
tegefaulkes: Don't import from `dist`, that's only special cases. If the types can't be imported like this then that needs to be fixed on `@matrixai/ws`.

```suggestion
import type { WebSocketConfig } from '@matrixai/ws';
```

src/PolykeyClient.ts
```diff
@@ -13,6 +14,21 @@ import * as events from './events';
 import config from './config';
 import * as networkUtils from './network/utils';
 import clientClientManifest from './client/callers';
+import {option} from "fast-check";
```
tegefaulkes: Importing `fast-check`? in a `src` file?

```suggestion
```

src/PolykeyClient.ts
```diff
@@ -13,6 +14,21 @@ import * as events from './events';
...
+type PolykeyClientOptions = {
+  nodePath?: string;
+  tokenBase: string;
+  // wsConfig?: Partial<WebSocketConfig>;
+  rpc: {
+    callTimeoutTime: number;
+    parserBufferSize: number;
+  };
+  wsConfig?: Partial<WebSocketConfig>;
+};
```
tegefaulkes: I think generally if an option is not defaulted then it can be  and `thing?: Type` optional type here.

Also @CMCDragonkai do we want this to be flatter? As i the `rpc` and `wsConfig` options flattened into the `PolykeyClientOptions`?

Also, remove the commented out `wsConfig` line.

src/PolykeyClient.ts
```diff
@@ -34,51 +50,66 @@ interface PolykeyClient extends CreateDestroyStartStop {}
...
     logger = new Logger(this.name),
     fresh = false,
   }: {
-    nodePath?: string;
-    streamFactory: StreamFactory;
-    streamKeepAliveTimeoutTime?: number;
-    parserBufferByteLimit?: number;
+    port: number;
+    host: string;
+    options: DeepPartial<PolykeyClientOptions>;
```
tegefaulkes: Even when defaulted with `options = {},` above you still need to specifiy this asn optional with `?` otherwise the types will complain.

```suggestion
    options?: DeepPartial<PolykeyClientOptions>;
```

src/PolykeyClient.ts
```diff
@@ -34,51 +50,66 @@ interface PolykeyClient extends CreateDestroyStartStop {}
...
     fresh?: boolean;
   }): Promise<PolykeyClient> {
     logger.info(`Creating ${this.name}`);
-    if (nodePath == null) {
+    const optionsDefaulted = utils.mergeObjects(options, {
+      nodePath: config.defaultsUser.nodePath,
+      tokenBase: config.paths.tokenBase,
+      callTimeoutTime: config.defaultsSystem.rpcCallTimeoutTime,
+      parserBufferSize: config.defaultsSystem.rpcParserBufferSize,
+    });
```
tegefaulkes: Why is `tokenBase` an option? we don't configure this.

Also the defaults structure doesn't match the structure of `options` The resulting `optionsDefaulted` will look like
```
{
   nodePath?: string;
   tokenBase: string;
   rpc: {
     callTimeoutTime: number;
     parserBufferSize: number;
   };
   callTimeoutTime: number;
   parserBufferSize: number;
   wsConfig?: Partial<WebSocketConfig>;
}
```

Which is definitely not correct.

src/PolykeyClient.ts
```diff
@@ -34,51 +50,66 @@ interface PolykeyClient extends CreateDestroyStartStop {}
...
+    );
     const session = await Session.createSession({
       sessionTokenPath,
       logger: logger.getChild(Session.name),
       fresh,
     });
+    const ws = await WebSocketClient.createWebSocketClient({
+      host: host,
+      port: port,
+      config: optionsDefaulted.wsConfig,
```
tegefaulkes: I recall @CMCDragonkai mentioning this. We don't want to pass in the whole `wsConfig`. The `PolykeyClient` needs to provide some of these options without allowing the user to set them via the parameters. The main one is the custom validation.

The `NodeConnectionManger` has a pretty clear example of this when creating the `QUICServer`.

src/PolykeyClient.ts
```diff
@@ -34,51 +50,66 @@ interface PolykeyClient extends CreateDestroyStartStop {}
...
+      logger: logger.getChild('WebSocketClient'),
```
tegefaulkes: Get the logger name directly from the class. Be do this by convention.

```suggestion
      logger: logger.getChild(WebSocketClient.name),
```

src/PolykeyClient.ts
```diff
@@ -34,51 +50,66 @@ interface PolykeyClient extends CreateDestroyStartStop {}
...
     });
     const pkClient = new this({
-      nodePath,
+      nodePath: optionsDefaulted.nodePath,
       rpcClientClient: rpcClientClient,
       session,
       fs,
       logger,
     });
+    pkClient.ws = ws;
```
tegefaulkes: Don't do this, It should be provided to the constructor, not set after constructing.

```suggestion
```

src/PolykeyClient.ts
```diff
@@ -88,6 +119,7 @@ class PolykeyClient {
   public readonly session: Session;
   public readonly rpcClientClient: RPCClient<typeof clientClientManifest>;
```
tegefaulkes: Probably should be called just `rpcClient` now.

src/PolykeyClient.ts
```diff
@@ -88,6 +119,7 @@ class PolykeyClient {
...
 
+  protected ws: WebSocketClient;
```
tegefaulkes: Let's keep names consistent with the class. 
```suggestion
  protected webSocketClient: WebSocketClient;
```

src/PolykeyClient.ts
```diff
@@ -124,9 +156,12 @@ class PolykeyClient {
 
   public async destroy() {
     this.logger.info(`Destroying ${this.constructor.name}`);
+    await this.ws.destroy({ force: true });
     await this.session.destroy();
     this.logger.info(`Destroyed ${this.constructor.name}`);
   }
 }
```
tegefaulkes: Try not to hard code force here. If it's an option, it should be an option on the `PolykeyClient.destroy` as well.

```suggestion
  public async destroy({ force = true }: { force?: boolean }) {
    this.logger.info(`Destroying ${this.constructor.name}`);
    await this.ws.destroy({ force });
    await this.session.destroy();
    this.logger.info(`Destroyed ${this.constructor.name}`);
  }
}
```

tests/PolykeyClient.test.ts
```diff
@@ -77,11 +81,16 @@ describe('PolykeyClient', () => {
...
+      nodePath: nodePath,
+      tokenBase: config.paths.tokenBase,
+      rpc: {
+        callTimeoutTime: config.defaultsSystem.rpcCallTimeoutTime,
+        parserBufferSize: config.defaultsSystem.rpcParserBufferSize,
+      },
+      wsConfig: {
         verifyPeer: true,
         verifyCallback: async (certs) => {
           await clientUtils.verifyServerCertificateChain(
```
tegefaulkes: The custom verification options should be set inside the `PolykeyClient`.

tests/PolykeyClient.test.ts
```diff
@@ -77,11 +81,16 @@ describe('PolykeyClient', () => {
...
   test('end to end with authentication logic', async () => {
-    const webSocketClient = await WebSocketClient.createWebSocketClient({
-      config: {
+    const options: PolykeyClientOptions = {
+      nodePath: nodePath,
+      tokenBase: config.paths.tokenBase,
+      rpc: {
+        callTimeoutTime: config.defaultsSystem.rpcCallTimeoutTime,
+        parserBufferSize: config.defaultsSystem.rpcParserBufferSize,
+      },
```
tegefaulkes: Unless they need to be set for the test to work, don't bother setting them. Less noise for the test.

tegefaulkes (changes requested):
Few things to address.

src/PolykeyClient.ts
addievo: lmao, ide auto import, not lintfixed

src/PolykeyClient.ts
addievo: @amydevs you need to fix this, types currently can't be imported directly

src/PolykeyAgent.ts
addievo: That is how it was originally, not a change I made, you want me to remove the Partial?

src/PolykeyClient.ts
addievo: Done

src/PolykeyAgent.ts
tegefaulkes: Yeah.

src/PolykeyClient.ts
addievo: uh idk how to go about this other than optionsDefaulted, wherein I have set it to false, if you have any other more concrete solutions, feel free to share

src/PolykeyClient.ts
addievo: Wdym by why is tokenBase an option? In the jests the tokenBase is used as config.paths.tokenBase, should it not be an option?

src/PolykeyClient.ts
CMCDragonkai: Do not leave this in the codebase. It does not matter what IDE you use, you must maintain correctness as much as possible on every commit you're submitting, otherwise it wastes time for people reviewing.

src/PolykeyAgent.ts
```diff
@@ -67,15 +67,6 @@ type PolykeyAgentOptions = {
   nodes: NodesOptions;
```
CMCDragonkai: I don't know if `NodesOptions` makes sense here. The main reason is because options at `PolykeyAgent` isn't a composition of options of subdomains. It's its own thing. I don't want to tie the API of `PolykeyAgent` to the underlying domains. So I'm removing this and replacing it with embedded options. The `NCM` do its own thing though.

CMCDragonkai:
Taking over this PR. I'm going to standardise the option names between the agent and client. And it's important to know that domain options are not composed into the agent/client options. They are abstracted from each other.

CMCDragonkai:
All options must be top-level to ensure it's easy to compare between `PolykeyAgent` and `PolykeyClient`.

CMCDragonkai:
The reason is like this.

1. Subdomains aren't really meant to have "options". They are just groups of classes closely related, that may or may not be factored out in a separate package one day.
2. Classes themselves have options - more correctly as parameters. As parameters, they should be top level, not nested within a `options`. Why? Because you can just do `X.createX({ param1, param2 });`, no need to do `X.createX({ options: { param1, param2 });` as the `options` is superfluous.
3. But why does `PolykeyAgent` and `PolykeyClient` have `options`? The reason is simply because there is so many parameters, that it is useful to have a separate type to keep track of them all. But this is exceptional to `PolykeyAgent` and `PolykeyClient`, it should not be considered the default way of creating classes in our codebase. And this type `PolykeyAgentOptions` and `PolykeyClientOptions` will be used in various places, in particular tests that need to run many different variants of the Polykey program.
4. There should necessarily be an exact relationship between the parameters in each domain class, and the options of `PolykeyAgent` and `PolykeyClient`. We want to retain some flexibility here.

CMCDragonkai:
I'm reverting: 0f81b34

src/PolykeyAgent.ts
CMCDragonkai: Rearchitecting this.

src/PolykeyClient.ts
CMCDragonkai: We're going to use flattened options for EVERYTHING except PKAgent and PKClient. So I'm reverting some old changes. Monitor this @addievo I'll need you to take over after I'm finished to update all test call signatures.

src/PolykeyClient.ts
CMCDragonkai: Not going to use that type.

src/PolykeyClient.ts
CMCDragonkai: Subdomains will all be grouped.

Because websocket is just encapsulated as part of PKClient, and it's not a subdomain, I might just flatten it.

CMCDragonkai:
Reverting my old `BootstrapOptions` change too.

```
[nix-shell:~/Projects/Polykey/src]$ ag 'Options'
PolykeyAgent.ts
48:type PolykeyAgentOptions = {
111:    // Options
119:    options?: DeepPartial<PolykeyAgentOptions>;
847:export type { PolykeyAgentOptions };

keys/KeyRing.ts
53:    ...startOptions
88:    await keyRing.start(startOptions);
161:    const { fresh = false, ...setupKeyPairOptions } = options;
171:      await this.setupKeyPair(setupKeyPairOptions);

PolykeyClient.ts
21:type PolykeyClientOptions = {
60:    options?: DeepPartial<PolykeyClientOptions>;
173:export type { PolykeyClientOptions };

vaults/VaultOps.ts
11:type FileOptions = {
145:  fileOptions?: FileOptions,
150:      await efs.rmdir(secretName, fileOptions);
171:  fileOptions?: FileOptions,
174:  const recursive = !!fileOptions?.recursive;
178:      await efs.mkdir(dirPath, fileOptions);

sigchain/Sigchain.ts
299:    const orderOptions =
301:    let seekOptions:
306:      seekOptions =
317:      ...orderOptions,
318:      ...seekOptions,
347:    const orderOptions =
349:    let seekOptions:
354:      seekOptions =
365:      ...orderOptions,
366:      ...seekOptions,
```

All removed now except of course `PolykeyAgentOptions`.

Oh yea and this is another reason to use `ag`, you can actually copy the result as a comment, and you can't do that in your IDE. Use your terminal more @addievo @tegefaulkes!

CMCDragonkai:
I noticed that in `PolykeyAgent`, the `connectTimeoutTime` isn't being used.

The `connectTimeoutTime` is supposed to be the timeout to connect to the server, and separate from the keep alive.

Now `PolykeyAgent` runs the server side of the client service, so it doesn't actually connect to other clients. However when establishing the reverse side of connections you may still want a timeout in case of protocol logic.

In QUIC:

```
      await connection.start(
        {
          data,
          remoteInfo,
        },
        { timer: this.minIdleTimeout },
      );
```

Is used on the server side.

@amydevs how is this done in WS?

CMCDragonkai:
@amydevs Ok so:

1. `config.clientConnectTimeout` - is used as the default of `PolykeyClientOptions.connectTimeoutTime`
2. This is now removed from `PolykeyAgentOptions` and instead moved into `PolykeyClientOptions`.
3. This option is used as part of the `ctx` `{ timer: optionsDefaulted.connectTimeoutTime }` we creating the `WebSocketClient`. This makes sense, as it is the timeout for the transport connecting to the client service.
4. The `rpcCallTimeoutTime` however is instead being set to `streamKeepAliveTimeoutTime` for `RPCClient`. I think this option needs to be renamed. It's too confusing what it means. I think it can be updated https://github.com/MatrixAI/js-rpc/pull/47

CMCDragonkai:
```ts
    const webSocketClient = await WebSocketClient.createWebSocketClient(
      {
        host: host,
        port: port,
        config: {
          verifyPeer: true,
          keepAliveTimeoutTime: optionsDefaulted.keepAliveTimeoutTime,
          keepAliveIntervalTime: optionsDefaulted.keepAliveIntervalTime,
        },
        logger: logger.getChild(WebSocketClient.name),
      },
      {
        // this makes sense
        timer: optionsDefaulted.connectTimeoutTime,
      },
    );
    const rpcClient = new RPCClient({
      manifest: clientClientManifest,
      streamFactory: () => webSocketClient.connection.newStream(),
      middlewareFactory: rpcMiddleware.defaultClientMiddlewareWrapper(
        clientMiddleware.middlewareClient(session),
        optionsDefaulted.rpcParserBufferSize,
      ),
      toError: networkUtils.toError,
      // this should be renamed
      streamKeepAliveTimeoutTime: optionsDefaulted.rpcCallTimeoutTime,
      logger: logger.getChild(RPCClient.name),
    });
```

CMCDragonkai:
The ops limits numbers are:

```
[nix-shell:~/Projects/Polykey]$ node
Welcome to Node.js v20.5.1.
Type ".help" for more information.
> const sodium = require('sodium-native')
undefined
> sodium.crypto_pwhash_OPSLIMIT_MIN
1
> sodium.crypto_pwhash_OPSLIMIT_MAX
4294967295
> sodium.crypto_pwhash_OPSLIMIT_INTERACTIVE
2
> sodium.crypto_pwhash_OPSLIMIT_MODERATE
3
> sodium.crypto_pwhash_OPSLIMIT_SENSITIVE
4
```

So actually instead of loading `keys/utils/password` into `config.ts`, I'll just set the raw number. I need to put some tests into `tests/keys/utils/password.test.ts` to ensure these numbers are maintained.

CMCDragonkai:
Actually we can just the strings:

```
type PasswordOpsLimitChoice = 'min' | 'max' | 'interactive' | 'moderate' | 'sensitive';

type PasswordMemLimitChoice = 'min' | 'max' | 'interactive' | 'moderate' | 'sensitive';
```

Can derive them from the password option keys.

CMCDragonkai:
Ok time to fix up the tests.

In one of the tests `PolykeyClient.test.ts` we have `end to end with authentication logic`.

Here it is written:

```
    const options: PolykeyClientOptions = {
      nodePath: nodePath,
      wsConfig: {
        verifyPeer: true,
        verifyCallback: async (certs) => {
          await clientUtils.verifyServerCertificateChain(
            [pkAgent.keyRing.getNodeId()],
            certs,
          );
        },
      },
    };
```

This says it is verifying the server certificate of the client service.

I'd expect that this would be done automatically, since the `PolykeyClient` doesn't actually know what the node ID is.

Right now because the `WebSocketClient` is encapsulated, I've just switched on `verifyPeer`. I suspect this would not work because the verification of the client service needs to do it without referring the traditional x509 verification since we're dealing with self-signed node certificates.

So shouldn't this `verifyCallback` be embedded into `PolykeyClient` then?

src/PolykeyClient.ts
CMCDragonkai: Done.

src/PolykeyClient.ts
CMCDragonkai: Done.

src/PolykeyClient.ts
CMCDragonkai: We should discuss whether `force` should be defaulted to true here... since we never do it anywhere else except for transport. So we should understand what the effects here.

src/PolykeyAgent.ts
CMCDragonkai: Removed.

CMCDragonkai:
The `clientConnectTimeoutTime` is now the default value for `@timedCancellable`, so `PolykeyClient` is now a timed cancellable thing. The main thing is `WebSocketClient.createWebSocketClient`.

I've also recovered the `nodeId: NodeId` that was removed way before. I don't remember why you did this commit @tegefaulkes https://github.com/MatrixAI/Polykey/commit/7e050e8062bbceaa7abb16e9489cfebc16a82fd0.

tests/PolykeyClient.test.ts
CMCDragonkai: It is now done, we are taking `nodeId` inside `PolykeyClient`.

CMCDragonkai:
The `PolykeyClient.test.ts` now passes.

One thing I noticed was that it is possible to establish a connection to an agent without application-level authentication, because authentication occurs at the RPC level.

This is similar to being able to establish a TCP connection without authentication. Which is sort of what is happening here in this case.

It's interesting that agent to agent connections P2P wise, this is not possible, we do mutual TLS, thus ensuring the connection is authenticated from the beginning.

But for the client service, we cannot use MTLS because clients may have access to the root certificate of the agent. Thus they are using the root password or recovery code or a session token in their RPC requests.

I'm just wondering if we can move down this auth to the connection layer itself, to avoid malicious connections just hanging on to the client service. One quick way to deal with this is with a timeout on the PK client that has no RPC request activity, but this can be complicated because who knows, maybe I do want a long-running PK client connection.

If the auth were to be pushed down to the connection, this could work since websocket connections are just HTTP 1.1 connections atm. However transport layer wise, we don't have a defined interface for such an interaction. It would fit into the HTTP 1.1 basic authentication since you could set a basic token as part of the initial HTTP request.

@amydevs can you track this to create an issue.

CMCDragonkai:
@addievo make sure to review this PR, so you can see the real scope of this particular issue, it was definitely more complex than originally thought. Because this is a the root of the entire project, you now have to consider global integrity of the entire application, which means you have to consider all sorts of trade offs here. So this is a good learning exercise.

src/PolykeyClient.ts
```diff
@@ -122,11 +240,14 @@ class PolykeyClient {
     this.logger.info(`Stopped ${this.constructor.name}`);
   }
 
-  public async destroy() {
+  public async destroy({ force = false }: { force?: boolean }) {
```
CMCDragonkai, line 243: I decided to make this default to `false`. I don't actually think `force` should be true by default. There is a special case in our transports that does this, but at an application level we should know why we are forcing it.

CMCDragonkai: @tegefaulkes @amydevs @addievo Still need to consider what happens for `PolykeyAgent`.

amydevs:
> The `PolykeyClient.test.ts` now passes.
> 
> One thing I noticed was that it is possible to establish a connection to an agent without application-level authentication, because authentication occurs at the RPC level.
> 
> This is similar to being able to establish a TCP connection without authentication. Which is sort of what is happening here in this case.
> 
> It's interesting that agent to agent connections P2P wise, this is not possible, we do mutual TLS, thus ensuring the connection is authenticated from the beginning.
> 
> But for the client service, we cannot use MTLS because clients may have access to the root certificate of the agent. Thus they are using the root password or recovery code or a session token in their RPC requests.
> 
> I'm just wondering if we can move down this auth to the connection layer itself, to avoid malicious connections just hanging on to the client service. One quick way to deal with this is with a timeout on the PK client that has no RPC request activity, but this can be complicated because who knows, maybe I do want a long-running PK client connection.
> 
> If the auth were to be pushed down to the connection, this could work since websocket connections are just HTTP 1.1 connections atm. However transport layer wise, we don't have a defined interface for such an interaction. It would fit into the HTTP 1.1 basic authentication since you could set a basic token as part of the initial HTTP request.
> 
> @amydevs can you track this to create an issue.

i feel like verifyCallback should be changed to also pass through the entire request if HTTP headers are going to be used for auth rather than MTLS @CMCDragonkai

addievo:
So theres a prob with CommandStop and CommandStatus, if status is "NULL", then nodeID/Host/Port can be undefined, but PKC expects them regardless, causing issues. What do I do here?

```ts
async function processClientStatus(
  nodePath: string,
  nodeId?: NodeId,
  clientHost?: string,
  clientPort?: number,
  fs = require('fs'),
  logger = new Logger(processClientStatus.name),
): Promise<
  | {
      statusInfo: StatusStarting | StatusStopping | StatusDead;
      status: Status;
      nodeId: NodeId | undefined;
      clientHost: string | undefined;
      clientPort: number | undefined;
    }
  | {
      statusInfo: StatusLive;
      status: Status;
      nodeId: NodeId;
      clientHost: string;
      clientPort: number;
    }
  | {
      statusInfo: undefined;
      status: undefined;
      nodeId: NodeId;
      clientHost: string;
      clientPort: number;
    }
> {
  if (nodeId != null && clientHost != null && clientPort != null) {
    return {
      statusInfo: undefined,
      status: undefined,
      nodeId,
      clientHost,
      clientPort,
    };
  } else if (nodeId == null && clientHost == null && clientPort == null) {
    const statusPath = path.join(nodePath, config.paths.statusBase);
    const statusLockPath = path.join(nodePath, config.paths.statusLockBase);
    const status = new Status({
      statusPath,
      statusLockPath,
      fs,
      logger: logger.getChild(Status.name),
    });
    const statusInfo = await status.readStatus();
    if (statusInfo == null) {
      return {
        statusInfo: { status: 'DEAD', data: {} },
        status,
        nodeId: undefined,
        clientHost: undefined,
        clientPort: undefined,
      };
    } else if (statusInfo.status === 'LIVE') {
      nodeId = statusInfo.data.nodeId;
      clientHost = statusInfo.data.clientHost;
      clientPort = statusInfo.data.clientPort;
      return {
        statusInfo,
        status,
        nodeId,
        clientHost,
        clientPort,
      };
    } else {
      return {
        statusInfo,
        status,
        nodeId: undefined,
        clientHost: undefined,
        clientPort: undefined,
      };
    }
  } else {
    const errorMsg = arrayZip(
      [nodeId, clientHost, clientPort],
      [
        'missing node ID, provide it with --node-id or PK_NODE_ID',
        'missing client host, provide it with --client-host or PK_CLIENT_HOST',
        'missing client port, provide it with --client-port or PK_CLIENT_PORT',
      ],
    )
      .flatMap(([option, msg]) => {
        if (option == null) {
          return [msg];
        } else {
          return [];
        }
      })
      .join('; ');
    throw new errors.ErrorPolykeyCLIClientOptions(errorMsg);
  }
}
```

Problem domain

In CommandStop

```ts
    try {
          nodeId: clientStatus.nodeId,
          host: clientStatus.clientHost,
          port: clientStatus.clientPort,
          options: {
            nodePath: options.nodePath,
          },
          logger: this.logger.getChild(PolykeyClient.name),
        });
```

x | undefined is not assignable to x

PR #584, opened 2023-10-16
Local Polykey Node Discovery

### LINKED ISSUE #537: `MDNS` integration and `NodeGraph` structure expansion
### Specification

With the creation of `MDNS` we need to integrate it's usage into Polykey. With that comes some considerations surrounding
what connection information needs to be stored in the `NodeGraph` and what information needs to be shared between nodes. 

Right now the `NodeGraph` is based on the Kademlia spec and is pretty simple. It only stores a single Address for each node. A There are only 20 nodes per bucket and only the oldest active ones are stored. This is all based on the Kademlia spec.

The `NodeGraph` logic needs to be expanded to allow for more information relating to each node to be stored. We need to store `IPv4` as well as `IPv6` addresses. We need to track information about if the node is publicly contactable or needs to be hole punched. We need sticky nodes such as the seed nodes or user defined static information. 

#### Step 1 - MDNS Sidecar onto NodeConnectionManager and NodeManager

I've decided as the simplest solution, MDNS can be integrated into methods of NodeConnectionManager and NodeManager related to discovery, like `findNode`, etc. This essentially allows NodeConnectionManager and NodeManager use both MDNS and NodeGraph as discovery mechanisms independent of each other.

Note: MDNS should not be integrated into `findNode` directly, but as a separate method to be called by the callers of `findNode` in non-kademllia contexts.

The NodeGraph related data structures will still have to be edited like so to facilitate the separation of MDNS addresses and NodeGraph addresses:

```ts
type NodeAddressScope = 'local' | 'external';
type NodeAddress = {
  host: Host | Hostname;
  port: Port;
  scopes: Array<NodeAddressScope>;
};
```

This will mean that there will be overlapping NodeData that exists on both the NodeGraph and MDNS, this will have to be dealt with.

The methods on NCM: `getConnectionWithAddress` and `getConnection` will need to be amended to support multiple `NodeAddress`es as input. This isn't the ideal way to do this, but it will have to be like this until the NodeGraph itself is made to support storing multiple IP addresses in Step 2. Much of the flow will need to be rewritten to support this change.

##### Scopes

Currently in Step 1, `scopes` will only be used for disabling holepunching on `local` addresses. An address can be either `local`, `external`, or both. Local means that the address is locally routable, so holepunching is not required when connecting to that address. External means the that the address is gliobally routable, but will require holepunching to establish a connection.

Scopes aren't just for disabling signaling or not, they are also used for the purpose of classifying IP addresses into different levels. This means that in future, when scopes are incorporated as part of the nodegraph, we can specify for only external addresses to be shared on the public mainnet.

Being able to define scopes as an array also means that i address could have multiple classifications, which would be useful in the future if we ever needed it expanded. Also addresses with both `local` and `external` will be understood to be shared on the mainnet, but not require holepunching to be connected to.

##### Custom Group Address and Port

We're going to need to select a port and ipv4 + ipv6 groups for now, as to not interfere with existing MDNS stacks. However, as native integration of MDNS improves, this may eventually no longer be needed.

I've decided to take inspiration from our slogan `fait accompli`, to choose these addresses:

| | IPv4 | IPv6 | Port |
| -| - | - | - |
| Value | `224.0.0.250` | `ff02::fa17` | `64023` |
| Reasoning | `250` is decimal representation for `fa` | `fa17` looks like `fait` in `fait accompli` | `64023` is the decimal representation of `fa17` |

All these Values have been checked against IANA to be not of conflict with any existing reservations.

#### Step 2 - Full Integration of MDNS to Supply Data to NodeGraph

Untitled-2023-10-23-0424 excalidraw(1)

The NodeGraph is expanded to support multiple IPv4 and IPv6 addresses.

This will allow us to integrate MDNS into the NodeGraph, so that it is used as the central Node address book.

```ts
type NodeData = {
  addresses: Array<NodeAddress>;
  lastUpdated: number;
};
```

This might require alot of API changes. We'll see if we can get there in time.

### Additional context

* Related: #525
* Related: #234 
* Related: #400
* Related: #537
* Related: #605

### Tasks

- [x] 1. Step 1 - Integrate `MDNS` and it's usage into initial network entry and discovery.
  - [x] Integrate MDNS querying into `NodeConnectionManager.findNode`
- [x] 2. Step 2 - `NodeGraph` functionality needs to be expanded to handle connection metadata, support IPv6 and general policy for garbage collection, data updating and data sharing policy.
  - [x] The initial sync node graph proceedure should be extracted out of `NodeManager.start` and be made a public function. It's usage should be called in `PolykeyAgent` when starting. We should make it optional for testing.
  - ~[ ] Integrate MDNS querying into `NodeManager.syncNodeGraph`~
  - ~[ ] Integrate MDNS querying into `NodeManager.getNodeAddress`~
  - ~[ ] Integrate MDNS querying into `NodeManager.setNode`~
  - [x] Integrate MDNS into `NodeManager.findNode` - this was moved out from NCM into NM now.

amydevs:
### Description

This PR integrates MDNS into the NodeConnectionManager to allow for the discovery of nodes on the local network, as well as disable holepunching for connections to local network addresses.

This PR should only focus on the Step 1 of #537 spec, as modifying the node graph to accept multiple IP addresses will require ACL changes that default disable the sharing of local addresses with seed nodes, as well as expansion to support multiple IP addresses.

This PR will encapsulate MDNS in the NodeConnectionManager, as well as make methods such as `getConnection` to be able to query MDNS for known hosts through the newly added `findNodesAll` and `findNodesLocal` methods. `findNodesAll` is a method that calls both the original `findNode` and `findNodesLocal` to aggregate a list of all NodeAddresses. This is notably only used in connection-related methods rather than those related to Kademlia (NodeGraph).

When the NodeConnectionManager is started, MDNS will be started, with a Polykey service registered, and a query for Polykey services started.

When the NodeConnectionManager is stopped, MDNS will be stopped alongside all pending queries, and deregistering all registered services.

Current scopes: `external` and `local`.

### Issues Related

* Related #537 

### Tasks

- [x] 1. Integrate MDNS into NodeConnectionManager
- [x] 2. Separate MDNS and Kademlia Behaviour
- [x] 3. Update NodeConnectionManager Methods to Support Multiple IP Addresses
  - [x]  `getConnectionWithAddress`
  - [x] `getConnection` 
- [x] 4. Select a well-known port and multicast address for Polykey usage.
- [x] 5. Modify Network Domain to Support Link-Local Addresses with a Interface ID Suffix.
- [x] 6. Creating of connections should happen in stages, where local addresses should be tried for first before external addresses 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -7,6 +7,7 @@ import path from 'path';
 import process from 'process';
 import Logger from '@matrixai/logger';
 import { DB } from '@matrixai/db';
+import { MDNS } from '@matrixai/mdns';
 import {
   CreateDestroyStartStop,
   ready,
@@ -82,6 +83,10 @@ type PolykeyAgentOptions = {
     connectionHolePunchIntervalTime: number;
     rpcCallTimeoutTime: number;
     rpcParserBufferSize: number;
+  };
+  mdns: {
+    groups: Array<string>;
+    port: number;
   };
 };
 
@@ -167,6 +172,10 @@ class PolykeyAgent {
         connectionHolePunchIntervalTime:
           config.defaultsSystem.nodesConnectionHolePunchIntervalTime,
       },
+      mdns: {
+        groups: config.defaultsSystem.agentMdnsGroups,
+        port: config.defaultsSystem.agentMdnsPort,
+      }
     });
     // This can only happen if the caller didn't specify the node path and the
     // automatic detection failed
@@ -201,6 +210,7 @@ class PolykeyAgent {
     let gestaltGraph: GestaltGraph | undefined;
     let identitiesManager: IdentitiesManager | undefined;
     let nodeGraph: NodeGraph | undefined;
+    let mdns: MDNS | undefined;
     let nodeConnectionManager: NodeConnectionManager | undefined;
     let nodeManager: NodeManager | undefined;
     let discovery: Discovery | undefined;
@@ -316,6 +326,15 @@ class PolykeyAgent {
         keyRing,
         logger: logger.getChild(NodeGraph.name),
       });
+      mdns = new MDNS({
+        logger: logger.getChild(MDNS.name)
+      });
+      await mdns.start({
+        id: keyRing.getNodeId().toBuffer().readUint16BE(),
+        hostname: nodesUtils.encodeNodeId(keyRing.getNodeId()),
+        groups: optionsDefaulted.mdns.groups,
+        port: optionsDefaulted.mdns.port,
+      })
       // Remove your own node ID if provided as a seed node
       const nodeIdOwnEncoded = nodesUtils.encodeNodeId(keyRing.getNodeId());
       delete optionsDefaulted.seedNodes[nodeIdOwnEncoded];
@@ -323,6 +342,7 @@ class PolykeyAgent {
         keyRing,
         nodeGraph,
         tlsConfig,
+        mdns,
         seedNodes: optionsDefaulted.seedNodes,
         connectionFindConcurrencyLimit:
           optionsDefaulted.nodes.connectionFindConcurrencyLimit,
@@ -426,6 +446,7 @@ class PolykeyAgent {
       await discovery?.stop();
       await identitiesManager?.stop();
       await gestaltGraph?.stop();
+      await mdns?.stop();
       await acl?.stop();
       await sigchain?.stop();
       await certManager?.stop();
@@ -448,6 +469,7 @@ class PolykeyAgent {
       acl,
       gestaltGraph,
       nodeGraph,
+      mdns,
       taskManager,
       nodeConnectionManager,
       nodeManager,
@@ -486,6 +508,7 @@ class PolykeyAgent {
   public readonly acl: ACL;
   public readonly gestaltGraph: GestaltGraph;
   public readonly nodeGraph: NodeGraph;
+  public readonly mdns: MDNS;
   public readonly taskManager: TaskManager;
   public readonly nodeConnectionManager: NodeConnectionManager;
   public readonly nodeManager: NodeManager;
@@ -530,6 +553,7 @@ class PolykeyAgent {
     acl,
     gestaltGraph,
     nodeGraph,
+    mdns,
     taskManager,
     nodeConnectionManager,
     nodeManager,
@@ -552,6 +576,7 @@ class PolykeyAgent {
     acl: ACL;
     gestaltGraph: GestaltGraph;
     nodeGraph: NodeGraph;
+    mdns: MDNS;
     taskManager: TaskManager;
     nodeConnectionManager: NodeConnectionManager;
     nodeManager: NodeManager;
@@ -576,6 +601,7 @@ class PolykeyAgent {
     this.gestaltGraph = gestaltGraph;
     this.discovery = discovery;
     this.nodeGraph = nodeGraph;
+    this.mdns = mdns;
     this.taskManager = taskManager;
     this.nodeConnectionManager = nodeConnectionManager;
     this.nodeManager = nodeManager;
@@ -706,14 +732,17 @@ class PolykeyAgent {
         host: optionsDefaulted.clientServiceHost,
         port: optionsDefaulted.clientServicePort,
       });
+      await this.mdns.start({
+        id: this.keyRing.getNodeId().toBuffer().readUint16BE(),
+        hostname: nodesUtils.encodeNodeId(this.keyRing.getNodeId()),
+        groups: optionsDefaulted.mdns.groups,
+        port: optionsDefaulted.mdns.port,
+      })
       await this.nodeManager.start();
       await this.nodeConnectionManager.start({
         host: optionsDefaulted.agentServiceHost,
         port: optionsDefaulted.agentServicePort,
         ipv6Only: optionsDefaulted.ipv6Only,
-        enableMdns: true,
-        mdnsGroups: optionsDefaulted.agentMdnsGroups,
-        mdnsPort: optionsDefaulted.agentMdnsPort,
         manifest: agentServerManifest({
           acl: this.acl,
           db: this.db,
@@ -771,6 +800,7 @@ class PolykeyAgent {
       await this.vaultManager?.stop();
       await this.discovery?.stop();
       await this.nodeGraph?.stop();
+      await this.mdns?.stop();
       await this.nodeConnectionManager?.stop();
       await this.nodeManager?.stop();
       await this.clientService.stop({ force: true });
@@ -813,6 +843,7 @@ class PolykeyAgent {
     await this.clientService.stop({ force: true });
     await this.identitiesManager.stop();
     await this.gestaltGraph.stop();
+    await this.mdns.stop();
     await this.acl.stop();
     await this.sigchain.stop();
     await this.certManager.stop();
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -175,7 +175,7 @@ class PolykeyAgent {
       mdns: {
         groups: config.defaultsSystem.mdnsGroups,
         port: config.defaultsSystem.mdnsPort,
-      }
+      },
     });
     // This can only happen if the caller didn't specify the node path and the
     // automatic detection failed
@@ -327,14 +327,14 @@ class PolykeyAgent {
         logger: logger.getChild(NodeGraph.name),
       });
       mdns = new MDNS({
-        logger: logger.getChild(MDNS.name)
+        logger: logger.getChild(MDNS.name),
       });
       await mdns.start({
         id: keyRing.getNodeId().toBuffer().readUint16BE(),
         hostname: nodesUtils.encodeNodeId(keyRing.getNodeId()),
         groups: optionsDefaulted.mdns.groups,
         port: optionsDefaulted.mdns.port,
-      })
+      });
       // Remove your own node ID if provided as a seed node
       const nodeIdOwnEncoded = nodesUtils.encodeNodeId(keyRing.getNodeId());
       delete optionsDefaulted.seedNodes[nodeIdOwnEncoded];
@@ -670,7 +670,7 @@ class PolykeyAgent {
         mdns: {
           groups: config.defaultsSystem.mdnsGroups,
           port: config.defaultsSystem.mdnsPort,
-        }
+        },
       });
       // Register event handlers
       this.certManager.addEventListener(
@@ -745,7 +745,7 @@ class PolykeyAgent {
         hostname: nodesUtils.encodeNodeId(this.keyRing.getNodeId()),
         groups: optionsDefaulted.mdns.groups,
         port: optionsDefaulted.mdns.port,
-      })
+      });
       await this.nodeManager.start();
       await this.nodeConnectionManager.start({
         host: optionsDefaulted.agentServiceHost,
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -173,8 +173,8 @@ class PolykeyAgent {
           config.defaultsSystem.nodesConnectionHolePunchIntervalTime,
       },
       mdns: {
-        groups: config.defaultsSystem.agentMdnsGroups,
-        port: config.defaultsSystem.agentMdnsPort,
+        groups: config.defaultsSystem.mdnsGroups,
+        port: config.defaultsSystem.mdnsPort,
       }
     });
     // This can only happen if the caller didn't specify the node path and the
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -711,6 +711,9 @@ class PolykeyAgent {
         host: optionsDefaulted.agentServiceHost,
         port: optionsDefaulted.agentServicePort,
         ipv6Only: optionsDefaulted.ipv6Only,
+        enableMdns: true,
+        mdnsGroups: optionsDefaulted.agentMdnsGroups,
+        mdnsPort: optionsDefaulted.agentMdnsPort,
         manifest: agentServerManifest({
           acl: this.acl,
           db: this.db,
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -650,6 +650,10 @@ class PolykeyAgent {
         | { recoveryCode: string }
         | { privateKey: Buffer }
         | { privateKeyPath: string };
+      mdns: {
+        groups: Array<string>;
+        port: number;
+      };
     }>;
     workers?: number;
     fresh?: boolean;
@@ -663,6 +667,10 @@ class PolykeyAgent {
         agentServicePort: config.defaultsUser.agentServicePort,
         workers: config.defaultsUser.workers,
         ipv6Only: config.defaultsUser.ipv6Only,
+        mdns: {
+          groups: config.defaultsSystem.mdnsGroups,
+          port: config.defaultsSystem.mdnsPort,
+        }
       });
       // Register event handlers
       this.certManager.addEventListener(
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -113,7 +113,7 @@ type NodesFindMessage = {
   addresses: Array<AddressMessage>;
 };
 
-type NodesGetMessage = NodeAddressMessage &  { bucketIndex: number };
+type NodesGetMessage = NodeAddressMessage & { bucketIndex: number };
 
 type NodesAddMessage = NodeAddressMessage & {
   force?: boolean;
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -109,7 +109,11 @@ type AddressMessage = {
 
 type NodeAddressMessage = NodeIdMessage & AddressMessage;
 
-type NodesGetMessage = NodeAddressMessage & { bucketIndex: number };
+type NodesFindMessage = {
+  addresses: Array<AddressMessage>;
+};
+
+type NodesGetMessage = NodeAddressMessage &  { bucketIndex: number };
 
 type NodesAddMessage = NodeAddressMessage & {
   force?: boolean;
@@ -327,6 +331,7 @@ export type {
   NodeIdMessage,
   AddressMessage,
   NodeAddressMessage,
+  NodesFindMessage,
   NodeConnectionMessage,
   ActionsListMessage,
   SetIdentityActionMessage,
--- a/src/config.ts
+++ b/src/config.ts
@@ -13,12 +13,12 @@ const testnet: Record<string, NodeAddress> = {
   v7v9ptvcdbdf8p4upok3prpmu3938ns8v4g45dib7sm5hqvvehv70: {
     host: 'testnet.polykey.com' as Host,
     port: 1314 as Port,
-    scopes: ['external']
+    scopes: ['external'],
   },
   v270ktdd3cs3mp1r3q3dkmick92bn927mii9or4sgroeogd1peqb0: {
     host: 'testnet.polykey.com' as Host,
     port: 1314 as Port,
-    scopes: ['external']
+    scopes: ['external'],
   },
 };
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -257,7 +257,7 @@ const config = {
      * The default values of `224.0.0.250` and `ff02::fa17` have been selected due to
      * the resemblance of `fa17` to the latin word `fait`, and `250` being the decimal representation of that.
      */
-    agentMdnsGroups: ['224.0.0.250', 'ff02::fa17'],
+    mdnsGroups: ['224.0.0.250', 'ff02::fa17'],
     /**
      * The port that the MDNS stack will operate on.
      *
@@ -266,7 +266,7 @@ const config = {
      * The default value has been selected as the decimal decimal representation of `fa17`,
      * which resembles the latin word `fait`.
      */
-    agentMdnsPort: 64023,
+    mdnsPort: 64023,
   },
   /**
    * Default user configuration.
--- a/src/config.ts
+++ b/src/config.ts
@@ -13,10 +13,12 @@ const testnet: Record<string, NodeAddress> = {
   v7v9ptvcdbdf8p4upok3prpmu3938ns8v4g45dib7sm5hqvvehv70: {
     host: 'testnet.polykey.com' as Host,
     port: 1314 as Port,
+    scopes: ['external']
   },
   v270ktdd3cs3mp1r3q3dkmick92bn927mii9or4sgroeogd1peqb0: {
     host: 'testnet.polykey.com' as Host,
     port: 1314 as Port,
+    scopes: ['external']
   },
 };
 
@@ -246,6 +248,25 @@ const config = {
      * Interval for hole punching reverse node connections.
      */
     nodesConnectionHolePunchIntervalTime: 1_000, // 1 second
+    /**
+     * Multicast group addresses that the MDNS stack will operate on.
+     *
+     * These values are well-known, and hence must not be changed by the user.
+     *
+     * The default values of these groups must start with either `224.0` (IPv4) or `ff02` (IPv6).
+     * The default values of `224.0.0.250` and `ff02::fa17` have been selected due to
+     * the resemblance of `fa17` to the latin word `fait`, and `250` being the decimal representation of that.
+     */
+    agentMdnsGroups: ['224.0.0.250', 'ff02::fa17'],
+    /**
+     * The port that the MDNS stack will operate on.
+     *
+     * This is well-known, and hence must not be changed by the user.
+     *
+     * The default value has been selected as the decimal decimal representation of `fa17`,
+     * which resembles the latin word `fait`.
+     */
+    agentMdnsPort: 64023,
   },
   /**
    * Default user configuration.
--- a/src/nodes/NodeConnectionManager.ts
+++ b/src/nodes/NodeConnectionManager.ts
@@ -159,6 +159,7 @@ class NodeConnectionManager {
   protected logger: Logger;
   protected keyRing: KeyRing;
   protected nodeGraph: NodeGraph;
+  protected mdns: MDNS | undefined;
   protected tlsConfig: TLSConfig;
   protected seedNodes: SeedNodes;
 
@@ -182,8 +183,6 @@ class NodeConnectionManager {
   protected connectionLocks: LockBox<Lock> = new LockBox();
 
   protected rpcServer: RPCServer;
-
-  protected mdns: MDNS;
 
   /**
    * Dispatches a `EventNodeConnectionManagerClose` in response to any `NodeConnectionManager`
@@ -290,6 +289,7 @@ class NodeConnectionManager {
   public constructor({
     keyRing,
     nodeGraph,
+    mdns,
     tlsConfig,
     seedNodes = {},
     connectionFindConcurrencyLimit = config.defaultsSystem
@@ -310,6 +310,7 @@ class NodeConnectionManager {
   }: {
     keyRing: KeyRing;
     nodeGraph: NodeGraph;
+    mdns?: MDNS;
     tlsConfig: TLSConfig;
     seedNodes?: SeedNodes;
     connectionFindConcurrencyLimit?: number;
@@ -406,10 +407,6 @@ class NodeConnectionManager {
       logger: this.logger.getChild(RPCServer.name),
     });
 
-    const mdns = new MDNS({
-      logger: this.logger.getChild(MDNS.name)
-    });
-
     this.quicClientCrypto = quicClientCrypto;
     this.quicSocket = quicSocket;
     this.quicServer = quicServer;
@@ -438,18 +435,12 @@ class NodeConnectionManager {
     port = 0 as Port,
     reuseAddr = false,
     ipv6Only = false,
-    enableMdns = true,
-    mdnsGroups = ['224.0.0.250', 'ff02::fa17'] as Array<Host>,
-    mdnsPort = 64023 as Port,
     manifest = {},
   }: {
     host?: Host;
     port?: Port;
     reuseAddr?: boolean;
     ipv6Only?: boolean;
-    enableMdns?: boolean;
-    mdnsGroups?: Array<Host>;
-    mdnsPort?: Port;
     manifest?: ServerManifest;
   }) {
     const address = networkUtils.buildAddress(host, port);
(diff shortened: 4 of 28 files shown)


amydevs:
<s>When a Polykey agent has it's host bound to `::0` and is trying to connect to a node at some ipv6 address, it is failing due to the formatting of my MDNS supplied ipv6 addresses.

When i encapsulate my IPv6 addresses in brackets (i.e [::1]), they are not recognized as hosts by `resolveHostnames` in src/network/utils.ts of (calls `isHost` to check if the provided address is a valid host). `resolveHostnames` will hence return no addresses at all, and will cause no QUICConnections to be made.

When i don't encapsulate my IPv6 addresses in brackets, they are recognized as hosts by `resolveHostnames` in src/network/utils.ts. However, the `std::net` `IPv6Addr` `parse` function fails only on link-local IPv6 addresses that aren't encapsulated in brackets (i.e fe80::260:8ff:fe52:f9d8) fail to be parsed as ipv6 addresses.

So, we're going to be ignoring ipv6 for now regarding MDNS.</s>

See https://github.com/MatrixAI/Polykey/pull/584#issuecomment-1763898632

CMCDragonkai:
If it only fixes a portion of #537, don't say `Fixes #537`, say `Related #537` in the OP.

CMCDragonkai:
Is the bracket encapsulated format the correct format?

If so, just align on that, and then just update `isHost` to make it work.

CMCDragonkai:
QUIC definitely supports ipv6. Try testing with non-link local IPv6 addresses first.

CMCDragonkai:
The square braces should only come into play with URIs, generally speaking none of internal network code should be using square braced hosts. So it should not be expected to work.

CMCDragonkai:
The only time we use square braces is on input and output, like for example the seed node configuration.

CMCDragonkai:
Can you check if maybe a scoped ipv6 address is needed, and if so, perhaps that needs to be passed all the way down. In that case the `isHost` or `isIPv6` check needs to support scoped ipv6 address.

amydevs:
> Can you check if maybe a scoped ipv6 address is needed, and if so, perhaps that needs to be passed all the way down. In that case the `isHost` or `isIPv6` check needs to support scoped ipv6 address.

it seems that `ip-num` does not correctly parse nor validate ipv6 link-local addresses, this will have to be addressed

amydevs:
it turns out there were two issues with MDNS integration that showed up on attempting to connect to a node via ipv6.

1. Link-local ip addresses will fail to be parsed by `std::net` Rust library if they do not contain the network interface name or scope id.
2. Promise.all([Promise.race(connProms)]) was causing getMultiConnection to fail where there is more than one address provided. In the case of MDNS discovery, there were more than one address discovered on my machine due IPv4 + IPv6 interfaces.

CMCDragonkai:
> it turns out there were two issues with MDNS integration that showed up on attempting to connect to a node via ipv6.
> 
> 1. Link-local ip addresses will fail to be parsed by `std::net` Rust library if they do not contain the network interface name or scope id.
> 2. Promise.all([Promise.race(connProms)]) was causing getMultiConnection to fail where there is more than one address provided. In the case of MDNS discovery, there were more than one address discovered on my machine due IPv4 + IPv6 interfaces.

See #585, @tegefaulkes any fixes relating to multi-connection?

So how are you working around ip-num support?

tests/nodes/NodeConnectionManager.general.test.ts
```diff
@@ -206,6 +207,7 @@ describe(`${NodeConnectionManager.name} general test`, () => {
     await nodeManager.start();
     await nodeConnectionManager.start({
       host: localHost as Host,
+      enableMdns: false,
     });
```
CMCDragonkai: Why have it disabled ever? Why not always have it on and fully encapsulated?

CMCDragonkai:
Can you explain the result of this PR? What changes to the behaviour to agent startup? You should expand the spec, since this is a rather large major feature.

tegefaulkes:
I'll do a proper review tomorrow.

tests/nodes/NodeConnectionManager.general.test.ts
amydevs: i disabled it for tests that didn't need MDNS, because some of those tests would fail expecting for a certain node to not be found.

amydevs:
> > it turns out there were two issues with MDNS integration that showed up on attempting to connect to a node via ipv6.
> > 
> > 1. Link-local ip addresses will fail to be parsed by `std::net` Rust library if they do not contain the network interface name or scope id.
> > 2. Promise.all([Promise.race(connProms)]) was causing getMultiConnection to fail where there is more than one address provided. In the case of MDNS discovery, there were more than one address discovered on my machine due IPv4 + IPv6 interfaces.
> 
> See #585, @tegefaulkes any fixes relating to multi-connection?
> 
> So how are you working around ip-num support?

i copied over the fixes yeah

tests/nodes/NodeConnectionManager.general.test.ts
CMCDragonkai: Can we make this a bit more robust? So that it works regardless, so less manual configuration. How would I know when mdns should be enabled or not, remember less configuration is better.

CMCDragonkai:
You can rebase now, since it has been merged to staging.

CMCDragonkai:
Are all the tasks there still todo, and are there more tasks to do? When can @tegefaulkes review this?

amydevs:
> Are all the tasks there still todo, and are there more tasks to do? When can @tegefaulkes review this?

I think review should be ready, although I have to make the MDNS tests to be more concise, and haven't tested it in a local network over multiple devices yet.

CMCDragonkai:
Are you going to tick off the tasks then?

CMCDragonkai:
Can you elaborate on the current problem atm.

CMCDragonkai:
https://github.com/MatrixAI/js-mdns/issues/28

tests/nodes/NodeConnectionManager.general.test.ts
CMCDragonkai: Ideas? Can we make it just work no matter the situation? If it really is indeed a sort of toggle, it should be `mdns: true/false` I think. But better to just always be enabled, we don't have a situation where it doesn't need it.

Unless dependency injection makes sense here? Thus requiring an external thing to create it first and then inject into it, that would make it a required dependency.

src/client/types.ts
```diff
@@ -103,20 +103,31 @@ type NodeIdMessage = {
...
   ping?: boolean;
 };
 
-type NodeConnectionMessage = NodeAddressMessage & {
+type NodeConnectionMessage = NodeIdMessage & {
+  host: string;
+  port: number;
   hostname: string;
   usageCount: number;
   timeout: number;
```
tegefaulkes: This can be simplified, keep the `AddressMesage` type as it was, just whatever needs a bunch of them can include an `Array<AddressMessage>`.

src/nodes/NodeConnectionManager.ts
```diff
@@ -407,12 +416,18 @@ class NodeConnectionManager {
     port = 0 as Port,
     reuseAddr = false,
     ipv6Only = false,
+    enableMdns = true,
+    mdnsGroups = ['224.0.0.250', 'ff02::fa17'] as Array<Host>,
+    mdnsPort = 64023 as Port,
```
tegefaulkes: Defaults here should come from the config if it's defined there. I think all the start parameters need to be fixed.

src/nodes/NodeConnectionManager.ts
```diff
@@ -508,6 +546,9 @@ class NodeConnectionManager {
     await this.quicServer.stop({ force: true });
     await this.quicSocket.stop({ force: true });
     await this.rpcServer.stop({ force: true });
+    if (this.mdns[running]) {
+      await this.mdns.stop();
+    }
```
tegefaulkes: Since stop is idempotent I think you can just call stop without checking if its running. 

Usually we only check if something is running prior to stopping if its happening in the background. But in this case the `mdns` could just not be running. So I'm not sure, maybe @CMCDragonkai can weigh in.

src/nodes/NodeConnectionManager.ts
```diff
@@ -893,8 +935,10 @@ class NodeConnectionManager {
           throw e;
         })
         .finally(async () => {
+          if (iceProm != null) {
           iceProm.cancel('Connection was established');
           await iceProm;
+          }
```
tegefaulkes: This can simplify.

```suggestion
          iceProm?.cancel('Connection was established');
          await iceProm;
```

src/nodes/NodeConnectionManager.ts
```diff
@@ -651,21 +692,21 @@ class NodeConnectionManager {
...
     @context ctx: ContextTimed,
   ): Promise<ConnectionAndTimer> {
     // If the connection already exists then we need to return it.
     const existingConnection = await this.getExistingConnection(targetNodeId);
     if (existingConnection != null) return existingConnection;
 
-    // If there was no address provided then we need to find it.
-    if (address == null) {
+    // If there were no addresses provided then we need to find them.
+    if (addresses == null) {
```
tegefaulkes: Since `addresses` is an array now, you need to check if it's an empty array and handle that. It should throw an error as invalid usage. 

It's probably checked somewhere else so likely won't cause a problem. But as the saying goes, fail early and fail often.

src/nodes/NodeConnectionManager.ts
```diff
@@ -700,19 +741,19 @@ class NodeConnectionManager {
...
       nodeConnectionManager.connectionConnectTimeoutTime,
   )
-  protected async getConnectionWithAddress(
+  protected async getConnectionWithAddresses(
     targetNodeId: NodeId,
-    address: NodeAddress,
+    addresses: Array<NodeAddress>,
     @context ctx: ContextTimed,
   ): Promise<ConnectionAndTimer> {
     const targetNodeIdString = targetNodeId.toString() as NodeIdString;
```
tegefaulkes: Here too, empty array shouldn't be allowed so check and throw.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1173,6 +1217,99 @@ class NodeConnectionManager {
...
+          const scopes: Array<'local' | 'external'> = ['local'];
+          if (externalAddress?.host === host && externalAddress.port === service.port as any) {
+            scopes.push('external');
+            concatExternal = false;
+          }
+          addresses.push({
+            host: host as any,
+            port: service.port as any,
+            scopes
+          });
```
tegefaulkes: A lot of `as any` here, This is a bit of a pet peve for me so try not to do it. If you have to cast then cast it to the type you need.

src/nodes/types.ts
```diff
@@ -6,9 +6,12 @@ import type { Host, Hostname, Port } from '../network/types';
  */
 type NodeGraphSpace = '0' | '1';
 
+type NodeAddressScope = 'local' | 'external';
+
 type NodeAddress = {
   host: Host | Hostname;
   port: Port;
+  scopes: Array<NodeAddressScope>;
 };
```
tegefaulkes, line 15: Adding in the scopes field makes this a little clunky. But this seems fine for now. We can go over this again with the node graph expansion.

tests/nodes/NodeConnectionManager.mdns.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      nodeGraph,
+      options: {
+        connectionConnectTimeoutTime: 1000,
+      },
+      logger: logger.getChild(`${NodeConnectionManager.name}Local`),
+      tlsConfig: clientTlsConfig,
+      seedNodes: undefined,
+    });
+
+    const serviceProm = promise();
```
tegefaulkes: We need to fully deconstruct the promise now.

tegefaulkes (changes requested):

tests/nodes/NodeConnectionManager.general.test.ts
amydevs: i think it would make sense for seednodes to have it disabled, maybe injected dep would make sense

src/nodes/NodeConnectionManager.ts
amydevs: i need to make the types in mdns itself to not use Opaue types for the service detail i think

src/nodes/NodeConnectionManager.ts
amydevs:
```ts
// If there were no addresses provided then we need to find them.
    if (addresses == null || addresses.length === 0) {
      // Find the node
      addresses = await this.findNodeAll(targetNodeId, undefined, ctx);
      if (addresses.length === 0) throw new nodesErrors.ErrorNodeGraphNodeIdNotFound();
    }
```

src/nodes/NodeConnectionManager.ts
amydevs:
```ts
if (addresses.length === 0) {
      throw new nodesErrors.ErrorNodeConnectionManagerNodeAddressRequired();
    }
```

tests/nodes/NodeConnectionManager.general.test.ts
amydevs: this is no longer relevant, due to MDNS being an injected optional dependency

src/client/types.ts
amydevs: Resolved:
```ts
// Nodes messages

type NodeIdMessage = {
  nodeIdEncoded: NodeIdEncoded;
};

type AddressMessage = {
  host: string;
  port: number;
};

type NodeAddressMessage = NodeIdMessage & AddressMessage;

type NodesFindMessage = {
  addresses: Array<AddressMessage>;
};

type NodesGetMessage = NodeAddressMessage &  { bucketIndex: number };

type NodesAddMessage = NodeAddressMessage & {
  force?: boolean;
  ping?: boolean;
};

type NodeConnectionMessage = NodeAddressMessage & {
  hostname: string;
  usageCount: number;
  timeout: number;
};
```

src/nodes/NodeConnectionManager.ts
amydevs: This is no longer relevant due to MDNS being an injected optional dep

src/nodes/NodeConnectionManager.ts
amydevs: This is no longer relevant due to MDNS being an injected optional dep

src/nodes/NodeConnectionManager.ts
amydevs:
```ts
if (this.mdns != null) {
      const service = this.mdns.networkServices.get(mdnsUtils.toFqdn({ name: nodesUtils.encodeNodeId(targetNodeId), type: "polykey", protocol: "udp" }));
      if (service != null) {
        for (const host_ of service.hosts) {
          let host: string;
          switch (this.quicSocket.type) {
            case 'ipv4':
              if (quicUtils.isIPv4(host_)) host = host_;
              else if (quicUtils.isIPv4MappedIPv6(host_)) host = quicUtils.fromIPv4MappedIPv6(host_);
              else continue;
              break;
            case 'ipv6':
              if (quicUtils.isIPv6(host_)) host = host_;
              else continue;
              break;
            case 'ipv4&ipv6':
              host = host_;
              break;
            default:
              continue;
          }
          const scopes: Array<'local' | 'external'> = ['local'];
          if (externalAddress?.host === host && externalAddress.port === service.port) {
            scopes.push('external');
            concatExternal = false;
          }
          addresses.push({
            host: host as Host,
            port: service.port as Port,
            scopes
          });
          this.logger.debug(
            `found address for ${nodesUtils.encodeNodeId(targetNodeId)} at ${
              host
            }:${service.port}`,
          );
        }
      }
    }
```

src/config.ts
```diff
@@ -246,6 +248,25 @@ const config = {
...
+    agentMdnsGroups: ['224.0.0.250', 'ff02::fa17'],
+    /**
+     * The port that the MDNS stack will operate on.
+     *
+     * This is well-known, and hence must not be changed by the user.
+     *
+     * The default value has been selected as the decimal decimal representation of `fa17`,
+     * which resembles the latin word `fait`.
+     */
+    agentMdnsPort: 64023,
```
CMCDragonkai: The prefix for this should just be `mdnsGroups` and `mdnsPort` because you're setting them up in `mdns` group in the `PolykeyAgentOptions`.

CMCDragonkai:
We need to elaborate a summary about the future implications of this design. @tegefaulkes Please review this and write up here what are the implications of the changes here in this codebase.

@amydevs can you draw up excalidraw diagram indicating what is changing here structurally.

src/nodes/agent/handlers/NodesHolePunchMessageSend.ts
```diff
@@ -103,6 +103,7 @@ class NodesHolePunchMessageSend extends UnaryHandler<
         await nodeConnectionManager.relaySignalingMessage(input, {
           host: meta.remoteHost,
           port: meta.remotePort,
+          scopes: ['external'],
         });
       }
     });
```
CMCDragonkai: Is this affected by #555 @tegefaulkes?

CMCDragonkai:
Also requires a comment on #537 indicating how that was partially resolved by this PR. What still needs to be done, and what are the impacts.

CMCDragonkai:
Does this prefer the local the network over the global network address?

It would need to do this in case we have multiple addresses...

So currently the local addresses don't get put into the node graph, that means they don't shared across kademlia.

Then if you have access to the local addresses, you shouldn't even bother to try the global addresses at all? If we simultaneously try the global addresses we may end up with wasted signalling or relaying requests on the seed nodes.

CMCDragonkai:
@tegefaulkes 

> Then if you have access to the local addresses, you shouldn't even bother to try the global addresses at all? If we simultaneously try the global addresses we may end up with wasted signalling or relaying requests on the seed nodes.

What would happen if #555 ends up resulting in the signalling being done in the background.

CMCDragonkai:
If local addresses exist, it could try the local addresses first, and then try global addresses if we want to optimise for less resource usage.

CMCDragonkai:
@amydevs is going to make this mergeable today, @tegefaulkes please review this tomorrow for merging.

amydevs:
The latest commit makes it so that starting a connection (`getConnectionWithAddresses`) will first try the local addresses, and then the external addresses.
This is currently slow and requires an O(n) operation where n is the number of addresses. A solution would be to be able to pass in multiple arrays of addresses for each address scope. if this currently isn't a concern, commit can be dropped.

```ts
const addressGroups: { local: Array<NodeAddress>, external: Array<NodeAddress> } = { local: [], external: [] };
    for (const address of addresses) {
      if (address.scopes.includes('local')) {
        addressGroups.local.push(address);
      }
      else {
        addressGroups.external.push(address);
      }
    }
    let timeoutDivisions = 0;
    for (const addressGroup of Object.values(addressGroups)) {
      if (addressGroup.length !== 0) {
        timeoutDivisions++;
      }
    }
    this.logger.debug(`Getting NodeConnection for ${targetNodeIdEncoded}`);
    return await this.connectionLocks
      .withF([targetNodeIdString, Lock, ctx], async () => {
        this.logger.debug(`acquired lock for ${targetNodeIdEncoded}`);
        // Attempting a multi-connection for the target node
        const timeout = ctx.timer.getTimeout() / timeoutDivisions;
        for (const addressGroup of Object.values(addressGroups)) {
          if (addressGroup.length === 0) {
            continue;
          }
          let results = await this.establishMultiConnection(
            [targetNodeId],
            addressGroup,
            {
              signal: ctx.signal,
              timer: timeout
            },
          );
          // Should be a single result.
          for (const [, connAndTimer] of results) {
            return connAndTimer;
          }
        }
        // Should throw before reaching here
        utils.never();
      })
```

src/nodes/agent/handlers/NodesHolePunchMessageSend.ts
tegefaulkes: It might be, we do send host and port, but I'm not sure the scope is relevant to our use.

CMCDragonkai:
Will require rebase, there's some conflicts.

CMCDragonkai:
> The latest commit makes it so that starting a connection (`getConnectionWithAddresses`) will first try the local addresses, and then the external addresses. This is currently slow and requires an O(n) operation where n is the number of addresses. A solution would be to be able to pass in multiple arrays of addresses for each address scope. if this currently isn't a concern, commit can be dropped.
> 
> ```ts
> const addressGroups: { local: Array<NodeAddress>, external: Array<NodeAddress> } = { local: [], external: [] };
>     for (const address of addresses) {
>       if (address.scopes.includes('local')) {
>         addressGroups.local.push(address);
>       }
>       else {
>         addressGroups.external.push(address);
>       }
>     }
>     let timeoutDivisions = 0;
>     for (const addressGroup of Object.values(addressGroups)) {
>       if (addressGroup.length !== 0) {
>         timeoutDivisions++;
>       }
>     }
>     this.logger.debug(`Getting NodeConnection for ${targetNodeIdEncoded}`);
>     return await this.connectionLocks
>       .withF([targetNodeIdString, Lock, ctx], async () => {
>         this.logger.debug(`acquired lock for ${targetNodeIdEncoded}`);
>         // Attempting a multi-connection for the target node
>         const timeout = ctx.timer.getTimeout() / timeoutDivisions;
>         for (const addressGroup of Object.values(addressGroups)) {
>           if (addressGroup.length === 0) {
>             continue;
>           }
>           let results = await this.establishMultiConnection(
>             [targetNodeId],
>             addressGroup,
>             {
>               signal: ctx.signal,
>               timer: timeout
>             },
>           );
>           // Should be a single result.
>           for (const [, connAndTimer] of results) {
>             return connAndTimer;
>           }
>         }
>         // Should throw before reaching here
>         utils.never();
>       })
> ```

I'm thinking you should do something like this?

* Try all local
* Try all global

That way local set is prioritised first.

Timeout wise, it would be shared between the 2... it might not be a problem, but maybe it needs to increase. Needs empirical test, tell me what the result will be.

CMCDragonkai:
I need to have some empirical results on what you mean by "slow". @amydevs

amydevs:
> I need to have some empirical results on what you mean by "slow". @amydevs

i mean iterating over every address is a bit inefficient, especially because i have to do it twice, once when turning the queried MDNS services into NodeAddresses, and the second time here

amydevs:
> > The latest commit makes it so that starting a connection (`getConnectionWithAddresses`) will first try the local addresses, and then the external addresses. This is currently slow and requires an O(n) operation where n is the number of addresses. A solution would be to be able to pass in multiple arrays of addresses for each address scope. if this currently isn't a concern, commit can be dropped.
> > ```ts
> > const addressGroups: { local: Array<NodeAddress>, external: Array<NodeAddress> } = { local: [], external: [] };
> >     for (const address of addresses) {
> >       if (address.scopes.includes('local')) {
> >         addressGroups.local.push(address);
> >       }
> >       else {
> >         addressGroups.external.push(address);
> >       }
> >     }
> >     let timeoutDivisions = 0;
> >     for (const addressGroup of Object.values(addressGroups)) {
> >       if (addressGroup.length !== 0) {
> >         timeoutDivisions++;
> >       }
> >     }
> >     this.logger.debug(`Getting NodeConnection for ${targetNodeIdEncoded}`);
> >     return await this.connectionLocks
> >       .withF([targetNodeIdString, Lock, ctx], async () => {
> >         this.logger.debug(`acquired lock for ${targetNodeIdEncoded}`);
> >         // Attempting a multi-connection for the target node
> >         const timeout = ctx.timer.getTimeout() / timeoutDivisions;
> >         for (const addressGroup of Object.values(addressGroups)) {
> >           if (addressGroup.length === 0) {
> >             continue;
> >           }
> >           let results = await this.establishMultiConnection(
> >             [targetNodeId],
> >             addressGroup,
> >             {
> >               signal: ctx.signal,
> >               timer: timeout
> >             },
> >           );
> >           // Should be a single result.
> >           for (const [, connAndTimer] of results) {
> >             return connAndTimer;
> >           }
> >         }
> >         // Should throw before reaching here
> >         utils.never();
> >       })
> > ```
> 
> I'm thinking you should do something like this?
> 
>     * Try all local
> 
>     * Try all global
> 
> 
> That way local set is prioritised first.
> 
> Timeout wise, it would be shared between the 2... it might not be a problem, but maybe it needs to increase. Needs empirical test, tell me what the result will be.

```ts
let timeoutDivisions = 0;
    const addressGroups: { local: Array<NodeAddress>, external: Array<NodeAddress> } = { local: [], external: [] };
    for (const address of addresses) {
      let scope = address.scopes.includes('local') ? 'local' : 'external';
      // If this is the first time an addressGroup has had an address added, the timeout divisions must be incremented.
      if (addressGroups[scope].length === 0) {
        timeoutDivisions++;
      }
      addressGroups[scope].push(address);
    }
    this.logger.debug(`Getting NodeConnection for ${targetNodeIdEncoded}`);
    return await this.connectionLocks
      .withF([targetNodeIdString, Lock, ctx], async () => {
        this.logger.debug(`acquired lock for ${targetNodeIdEncoded}`);
        // Attempting a multi-connection for the target node using local addresses
        const timeout = ctx.timer.getTimeout() / timeoutDivisions;
        let results: Map<NodeIdString, ConnectionAndTimer> | undefined;
        if (addressGroups.local.length !== 0) {
          results = await this.establishMultiConnection(
            [targetNodeId],
            addressGroups.local,
            {
              signal: ctx.signal,
              timer: timeout
            },
          );
        }
        // If there are no results from the attempted local connections, attempt a multi-connection for the target node using external addresses
        if (results == null || results.size === 0) {
          results = await this.establishMultiConnection(
            [targetNodeId],
            addressGroups.external,
            {
              signal: ctx.signal,
              timer: timeout
            },
          );
        }
        // Should be a single result.
        for (const [, connAndTimer] of results) {
          return connAndTimer;
        }
        // Should throw before reaching here
        utils.never();
      })
```
i've fixed this

src/PolykeyAgent.ts
```diff
@@ -316,6 +326,15 @@ class PolykeyAgent {
         keyRing,
         logger: logger.getChild(NodeGraph.name),
       });
+      mdns = new MDNS({
+        logger: logger.getChild(MDNS.name)
+      });
+      mdns.start({
+        id: keyRing.getNodeId().at(0),
```
tegefaulkes: What is the `id` for exactly? And you're betting it from a byte in the `NodeId`?

I'm not saying it's a problem, just curious.

src/PolykeyAgent.ts
```diff
@@ -316,6 +326,15 @@ class PolykeyAgent {
         keyRing,
         logger: logger.getChild(NodeGraph.name),
       });
+      mdns = new MDNS({
+        logger: logger.getChild(MDNS.name)
+      });
```
tegefaulkes: I don't see the `mdns` being passed into anything that uses it. The `NodeConnectionManager` needs it right?

src/nodes/NodeConnectionManager.ts
```diff
@@ -783,16 +829,29 @@ class NodeConnectionManager {
...
+  @timedCancellable(
+    true,
+    (nodeConnectionManager: NodeConnectionManager) =>
+      nodeConnectionManager.connectionConnectTimeoutTime,
+  )
   protected async establishMultiConnection(
     nodeIds: Array<NodeId>,
     addresses: Array<NodeAddress>,
-    ctx: ContextTimed,
+    @context ctx,
```
tegefaulkes: Should have the type annotation still. `@context ctx: ContextTimed`

Generally I didn't apply the `@timedCancellable` decoration to protected methods,because they should always be called within something that is already `timedCancellable`.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1205,6 +1265,124 @@ class NodeConnectionManager {
...
+  public findNodeLocal(
+    targetNodeId: NodeId,
+    ctx?: Partial<ContextTimed>,
+  ): PromiseCancellable<Array<NodeAddress>>;
+  @ready(new nodesErrors.ErrorNodeConnectionManagerNotRunning())
+  @timedCancellable(true)
+  public async findNodeLocal(
+    targetNodeId: NodeId,
+    @context ctx: ContextTimed,
+  ): Promise<Array<NodeAddress>> {
```
tegefaulkes: If this is a pretty simple operation then we can specify a default timeout on the `@timedCancellable` decorator. 

`findNode` didn't have one because it's a pretty long running and complex search procedure.

tegefaulkes (changes requested):
Some small things but otherwise seems fine.

Don't forget to lint, I saw an unused import somewhere and linting should catch that.

If all the tests pass then you can merge.

src/PolykeyAgent.ts
amydevs: i forgot to, have added it in

src/PolykeyAgent.ts
amydevs: The `id` paramater used so that MDNS knows what packets were sent from itself, and what were by others on the network.
Anyhow, it has been changed to now instead take the first 2 bytes, so that the `MDNS` id has more chance of being unique.

amydevs:
i think i might've accidentally removed the comment about the node id being used for the `id` param of MDNS. The paramater used so that MDNS knows what packets were sent from itself, and what were by others on the network.
Anyhow, it has been changed to now instead take the first 2 bytes, so that the `MDNS` id has more chance of being unique.

src/nodes/NodeConnectionManager.ts
amydevs: have added in

src/nodes/NodeConnectionManager.ts
amydevs: should this have a default in options?

tegefaulkes: Derive it from the existing options. I think it can just use the normal connection timeout value?

CMCDragonkai: Use the default timeout for finding nodes - shouldn't be different from global vs local.

Also you are using scopes right? Why don't name your suffixes equivalent to the scope keywords?

You have external and local. But not global and local? What about internal? How many more keywords? Did you spec out what an array means for scopes?

amydevs: alright

CMCDragonkai: Re defaults, just use config. It should have relevant defaults for everything.

src/PolykeyAgent.ts
CMCDragonkai: 16 bit prefix of node ID? I'm not sure but I think it's better to use a hashing function. However you won't get more than 256 nodes per local context. Can this be expanded to 32bits?

CMCDragonkai:
I need a better spec for scopes. What is meaning and consequences of it?

What is external? What is local? Why is there no "global"? What can this be expanded to? Copy content to #537.

amydevs:
> I need a better spec for scopes. What is meaning and consequences of it?
> 
> What is external? What is local? Why is there no "global"? What can this be expanded to? Copy content to #537.

I already expanded on it in the spec before:

https://github.com/MatrixAI/Polykey/issues/537

tegefaulkes (approved):

PR #585, opened 2023-10-16
Feature sync node graph fix

tegefaulkes:
### Description

This PR addresses some minor issues with `syncNodeGraph` and the multi connection logic.

### Issues Fixed

* No direct issues fixed, just a collection of small incidental fixes.

### Tasks

* [X] 1. Fix issue with `NodeManager.syncNodeGraph` throwing when a connection fails.
* [X] 2. NCM `establishMultiConnection` should only throw if no connections were established.
* [X] 3. Expand tests relating to multi connection logic.
* [X] 4. `NodeConnection` throws a nodes domain timeout error on connection timeout.
* ~5. Something is holding the process open, related to the `keepAliveIntervalTime`. Investigate.~ https://github.com/MatrixAI/js-timer/issues/15

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/nodes/NodeConnection.ts
+++ b/src/nodes/NodeConnection.ts
@@ -16,9 +16,14 @@ import { CreateDestroy } from '@matrixai/async-init/dist/CreateDestroy';
 import { status } from '@matrixai/async-init';
 import { timedCancellable, context } from '@matrixai/contexts/dist/decorators';
 import { AbstractEvent, EventAll } from '@matrixai/events';
-import { QUICClient, events as quicEvents } from '@matrixai/quic';
+import {
+  QUICClient,
+  events as quicEvents,
+  errors as quicErrors,
+} from '@matrixai/quic';
 import { RPCClient } from '@matrixai/rpc';
 import { middleware as rpcUtilsMiddleware } from '@matrixai/rpc';
+import { errors as contextErrors } from '@matrixai/contexts';
 import * as nodesErrors from './errors';
 import * as nodesEvents from './events';
 import * as networkUtils from '../network/utils';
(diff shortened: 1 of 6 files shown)


tegefaulkes:
Almost done with this, Just investigating task 5.  So far as I can tell, the timer relating to the `KeepAliveIntervalTimer` is not being cleaned up when a connection ends and is left to time out. May need to be fixed is `js-quic`.

tegefaulkes:
There's a weird bug. The process is being held open by a timer reference. The specific timer is the `keepAliveIntervalTimer` in `QUICConnection`. I've run some tests while monkey patching in some logging to the local dependencies and found that... while yes, the `Timer` IS being cancelled. the underlying timer is not being cleared. In fact, the `Timer` is not even rejecting after being cancelled.

This seems to be a problem `@matrixai/timer`. I'll need to investigate it more but I'm confident the problem is out of scope for this PR.

CMCDragonkai:
> There's a weird bug. The process is being held open by a timer reference. The specific timer is the `keepAliveIntervalTimer` in `QUICConnection`. I've run some tests while monkey patching in some logging to the local dependencies and found that... while yes, the `Timer` IS being cancelled. the underlying timer is not being cleared. In fact, the `Timer` is not even rejecting after being cancelled.
> 
> This seems to be a problem `@matrixai/timer`. I'll need to investigate it more but I'm confident the problem is out of scope for this PR.

Can you make an issue, and how to reproduce this in js-timer.

CMCDragonkai:
How do you know the underlying timer is not being cleared. Did you check by logging out the underlying timer object that is inside nodejs?

tegefaulkes:
I added a `console.log` just before each `clearTimeout()` and never saw the output when testing. When I change the timeout delay I see that reflected in the delay for the process closing. I added a `cheeky` `clearTimeout` to the `Timer.cancel()` method and the process started ending immediately after the test ended. When I added a `.finally()` to the Timer` it never got called after the timer was cancelled.

It's pretty clear that the underlying node timer in `Timer` is getting leaked here.

CMCDragonkai:
Ok please create a new issue js-timer, and task 5, shouldn't just be ticked if it wasn't resolved, you can however link to the new issue.

tegefaulkes:
New issue created at https://github.com/MatrixAI/js-timer/issues/15

PR #589, opened 2023-10-17
Integration of js-rpc timeout and middleware changes to fix Auth Provider timeout issue

### LINKED ISSUE #572: Implement `timeoutMiddleware` and Remove `metadata.authorization` for `src/nodes/agent` Domain
Issue is WIP

### Specification

1. `agentService` is a public service, so it should not have `authorisation`, but in the current implementation, it does. Which is to be altered. 
2. `agentService` currently does not use `timeoutMiddleware`, which it should.

### Additional context

js-rpc has a sister PR which addresses the same issues - https://github.com/MatrixAI/js-rpc/pull/42

### Tasks

1. Remove authorisation from `AgentRPCRequestParams` and `AgentRPCResponseResult`
2. Implement timeoutMiddleware for `agentService`

### LINKED ISSUE #588: GitHub Auth Timeout is too low
### Describe the bug

GitHub auth times out before completion due to timeout being too low

### To Reproduce

1. Try claiming github identity.
2. Wait
3. It aborts

### Expected behavior

Should not abort unexpectedly.

 - Version: CLI

addievo:
#### Provider Auth Timeouts

This PR  makes the timeout for the Provider Authentication flow longer. The result of how short it is currently is that the `js-rpc` default timeout set in `src/config`:

```ts
defaultsSystem: {
    rpcCallTimeoutTime: 15_000, // 15 seconds
```

As the Provider Authentication flow utilises the `IdentitiesAuthenticate` handler in `client/handlers`, it will follow this timeout. So there needs to be a way to override the default timeout for the `IdentitiesAuthenticate` handler only, without affecting any other handlers.

Although the aforementioned comment specifies

> Therefore it is expected that specific clients calls and server handlers
     * will override this timeout to cater to their specific circumstances.
    
This is currently not the case, as in js-rpc handler timeout is preferred, only if it is lesser than the default timeout, which is not the case in our desired scenario.

The solution is to make it so that the handler timeout overrides the default timeout set on the `RPCServer`, regardless of whether one is longer than the other.

This has now been updated in the following PR:
https://github.com/MatrixAI/js-rpc/pull/47

#### Moving `timeoutMiddleware` to `js-rpc`

As per https://github.com/MatrixAI/js-rpc/pull/42, `timeoutMiddleware` has been moved out, and is now a default middleware. This will mean that `timeoutMiddleware` will be applied everywhere. 

#### Updated Type Requirements of `js-rpc` Request `params` and Response `results`

The updated types in `js-rpc` of `JSONRPCRequestMessage`, `JSONRPCRequestNotification` and `JSONRPCResponseResult`mean that `JSONRPCRequest*.params` and `JSONRPCResponseResult.result` should now always be objects. This is so that metadata can always be appended to the params. Therefore, we need to make sure in Polykey that non-JSON-objects are never being passed as `result`s or `params`.
#### Long-Running Streaming Calls

Due to Timeout changes, specifically that timeouts no longer after the first server-sent message, long-running streaming calls are now possible without setting timeouts to `Infinity`.
 
### Issues Fixed

* Fixes #588 
* Fixes #572 

### Tasks

- [x] 1. Update `js-rpc` to allow for override of default timeout
- [x] 2. Update timeout for authentication handler to be 2 minutes.
- [x] 3. Remove `timeoutMiddleware` from `Polykey`.
- [x] 4. Check and update (if relevant) the type of message structure.
- [x] 5. Amend the types `ClientRPCRequestParms` and `ClientRPCResponseResult` from client handlers.
- [x] 6. Remove `metadata.authorization` from `AgentRPCRequestParams` and `AgentRPCResponseResult`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
Would this actually solve the problem? Maybe it's better to link to PK CLI to just try the authenticate command?

src/identities/providers/github/GitHubProvider.ts
```diff
@@ -42,7 +42,7 @@ class GitHubProvider extends Provider {
   }
 
   public async *authenticate(
-    timeout: number = 120000,
+    timeout: number = Infinity,
   ): AsyncGenerator<ProviderAuthenticateRequest, IdentityId> {
```
CMCDragonkai: I don't like the idea of an infinite timeout. This could cause the entire application to hang.

What is an appropriate amount of time for the authentication to take place?

It was set to 120000ms yet that was still too low. That's 120 seconds which is 2 minutes.

What is a reasonable amount of time for a person to take to authenticate themselves? Is it 3 minutes, 5 minutes, 10 minutes (in case they had to open up their password manager... etc).

Also you have to balance this against the call timeout time of the RPC calls. What happens if this is much larger than the handler timeout or call timeout. Can you review those as well? And spec how that works in the PR description.

addievo: I agree, I set it to 1200000 (that is 10x of initial timeout earlier, but you said Infinity would make more sense, I'll change it back to 1200k seconds, which would be, 20 minutes. More than enough.

addievo:
This problem is more complex than it first seemed.

The root issue is :- 

1. `RPCServer` has `handlerTimeoutTime` which is initialed in Polykey in `PolykeyAgent`.
2. To alter this timeout, the timeout has to be changed in `CommandStart`, the only place where a PK Agent is started.
3. Changing `streamKeepAliveTimeoutTime` in PolyKeyClient, which occurs in `createPolykeyClient` does not resolve the issue.

So therefore, the only possible solution right now is to alter timeout while `RPCServer` creating in `PolykeyAgent` in `CommandStart`, but that would require changing a global solution to fix a local problem, something which is less than ideal.

Currently exploring other possible solutions but I am fairly certain that this is the only solution here to fix the issue, as the core issue is the fact that RPC times out in exactly 15000ms, which isn't nearly enough time to comfortably authenticate github or other providers.

tegefaulkes:
To fix this properly there needs to be some changes to the handlers, callers and possible RPC.
1. We need to specify the timeouts on a per handler/caller basis. I recall us designing something like this in the past but I don't recall if it's implemented in the RPC.
2. If I recall, there are 3 ways the timeouts are set. first, the global timeout set for the client and server. second, the specific timeout for the handler or caller definition if that is implemented. third, what ever is passed in as the `ctx` timeout for the client call. By design I believe that the minimum of these values was used when creating the timeout timer. So if you want to enable a specific timeout larger than the default that will need to change.


Ultimately we need to be able to
1. Specify a larger timeout for just the specific caller/handler we want to have the longer timeout.
2. allow for timeouts longer than the global timeout value to be used when making a specific call.
3. We probably don't want to allow the client to override the server's timeout to be much longer than it's timeout for that handler.

You'll need to explore the `RPC` code to see how things are implemented currently and likely a new feature needs to be added to the `RPC` to enable this.


If we do need to make changes to the `RPC` for this then be sure to write a spec. It's important you get then intended changes and requirements down. It's an important touch stone for tracking what needs to be done.

addievo:
@tegefaulkes 

1. I did try to increase the RPCClient timeout, however the Server timeout still overrides it. 
 > We need to specify the timeouts on a per handler/caller basis. I recall us designing something like this in the past but I don't recall if it's implemented in the RPC.
2. Hmmm the ctx, yea that might be something worth looking into.

Alright, I'll look into the RPC code. 

Thanks for the reply.

CMCDragonkai:
Please flesh out the spec in the OP @addievo

tegefaulkes:
This requires changes to the `js-rpc` code, is there no link to that work here?

CMCDragonkai:
It's in the github events: https://github.com/MatrixAI/js-rpc/pull/47

addievo:
So this PR grows in complexity by the day.

1. Even after the changes to js-rpc to allow changes to the server timeout through the handler, the timeout shows no difference.

2. Current observations are that while the timeout can be lowered via the caller `ctx` timeout, increasing timeout ends to no avail. 

3. My current thought process is that there is a time bottleneck elsewhere causing the issue. Some theories to suggest this are : 

a. When RPC times out, theres a particular error it throws: 

```ts
> ts-node src/polykey.ts -np ./tmp/nodeA identities authenticate github.com

url             https://github.com/login/device
userCode        5FC2-254D
ErrorRPCTimedOut: RPC has timed out
```

b. Whereas with the normal timeout, it throws a specific error as so: 

```ts
> ts-node src/polykey.ts -np ./tmp/nodeA identities authenticate github.com

url             https://github.com/login/device
userCode        5FC2-254D
ErrorRPCTimedOut: RPC has timed out
```

4. I think the problem domain is as follows:

```ts
// Promise.race does not cancel unfinished promises
    // the finished condition variable is needed to stop the pollAccessToken process
    // the pollTimer is needed to stop the pollTimerP
    let pollTimedOut = false;
    let pollTimer;
    const pollTimerP = new Promise<void>((r) => {
      pollTimer = setTimeout(() => {
        pollTimedOut = true;
        r();
      }, timeout);
    });
    const that = this;
    const pollAccessToken = async () => {
      const payload = new URLSearchParams();
      payload.set('grant_type', 'urn:ietf:params:oauth:grant-type:device_code');
      payload.set('client_id', that.clientId);
      payload.set('device_code', deviceCode);
      const request = new Request(
        'https://github.com/login/oauth/access_token',
        {
          method: 'POST',
          headers: {
            'Content-Type': 'application/x-www-form-urlencoded',
            Accept: 'application/json',
          },
          body: payload.toString(),
        },
      );
      while (true) {
        if (pollTimedOut) {
          this.logger.info('Poll has timed out');
          return;
        }
        const response = await fetch(request);
        if (!response.ok) {
          throw new identitiesErrors.ErrorProviderAuthentication(
            `Provider responded with ${response.status} ${response.statusText}`,
          );
        }
        let data;
        try {
          data = await response.json();
        } catch (e) {
          throw new identitiesErrors.ErrorProviderAuthentication(
            'Provider access token response is not valid JSON',
            { cause: e },
          );
        }
        if (data.error) {
          if (data.error === 'authorization_pending') {
            await utils.sleep(pollInterval);
            continue;
          } else if (data.error === 'slow_down') {
            // Convert seconds to milliseconds
            pollInterval = parseInt(data.get('interval') ?? '1') * 1000;
            await utils.sleep(pollInterval);
            continue;
          }
          throw new identitiesErrors.ErrorProviderAuthentication(
            `Provider access token request responded with: ${data.error}`,
          );
        }
        const providerToken = {
          accessToken: data.access_token,
        };
        return providerToken;
      }
    };
    let providerToken;
    try {
      providerToken = await Promise.race([pollAccessToken(), pollTimerP]);
    } finally {
      clearTimeout(pollTimer);
    }
    if (providerToken == null) {
      throw new identitiesErrors.ErrorProviderAuthentication(
        `Provider authentication flow timed out`,
      );
    }
```

As this is where the specific error being thrown.

I will investigate this further.

The only solution working till now is changing the default timeout (`handlerTimeout`) for the `rpcServer` formed during creation of `PolykeyAgent`.

Also to add on this point further, lets assume the default timeout for the `rpcServer` to be 30000ms.
We call the `authIdentify` handler with a timeout of 120000ms, the server's default timeout will be overridden and set to 120000ms, now this timeout would carry forward till 
a. A new handler sets a different timeout.
b. The server is destroyed and created anew. 

Maybe the rpc requires further changes to set the timeout to default after the rpc call is handled.

CMCDragonkai:
No the solution needs to be simplified. You have to verify that call site timeouts override client timeouts and handler timeouts override server timeouts.

Once that is verified then you have to understand that between client and server, the lowest timeout is chosen.

Therefore you just first give sufficient time on both ends. Using both call site timeout and also handler timeout. This has nothing to do with the client timeout or server timeout.

Afterwards any further fixes happens on the PK CLI command itself.

---

Also the scope isn't increasing - the scope is the requirement. You're supposed to figure out what needs to be done for the requirement.

CMCDragonkai:
This is blocked on https://github.com/MatrixAI/js-rpc/pull/47

tegefaulkes:
I'll be merging the `rpc` PR. We'll have to update this PR after `rpc` has a new release.

tegefaulkes:
https://github.com/MatrixAI/js-rpc/pull/47 has been merged now and doing a release. When that Is done i'll look into finishing this.

tegefaulkes:
I think we're adding some extra fixes to this PR. Things that need to be changed

1. Update the timeout for the authentication handler in the client domain.
2. `Polykey-CLI` needs to update the call for authentication to use the extended timeout as well. (blocked by this)
3. The timeout middleware is built into `rpc` now, so it needs to be removed from `polykey`.
4. Types of the message structure *may* have changed. so that needs to be checked.
5. types like `ClientRPCRequestParams` need to be removed from the client and agent handler types.

addievo:
@tegefaulkes Cleaning up PR and adding tasks

amydevs:
> I think we're adding some extra fixes to this PR. Things that need to be changed
> 
>     1. Update the timeout for the authentication handler in the client domain.
> 
>     2. `Polykey-CLI` needs to update the call for authentication to use the extended timeout as well. (blocked by this)
> 
>     3. The timeout middleware is built into `rpc` now, so it needs to be removed from `polykey`.
> 
>     4. Types of the message structure _may_ have changed. so that needs to be checked.
> 
>     5. types like `ClientRPCRequestParams` need to be removed from the client and agent handler types.

i think types like `ClientRPCRequestParams` still need to stay in order to add properties to the metadata, specifically that for authorization

src/identities/providers/github/GitHubProvider.ts
amydevs: 20 minutes is too much, this caller is setting the timeout anyway with the timeout of the `IdentitiesAuthenticate` handler. So the change should be made there rather than in `GithubProvider`.

addievo:
Looks good

src/client/types.ts
```diff
@@ -18,27 +23,22 @@ import type { Notification } from '../notifications/types';
...
-> = {
+type ClientRPCResponseResult<T extends JSONObject = JSONObject> = {
   metadata?: {
     [Key: string]: JSONValue;
   } & Partial<{
     authorization: string;
-    timeout: number;
   }>;
-} & Omit<T, 'metadata'>;
+} & JSONRPCResult<T>;
```
CMCDragonkai: So question about this type `JSONRPCResult`.

Shouldn't this be called `RPCResponseResult`?

It's already a forgone conclusion to be `JSON`.

Also why do we call it `RPCResponseResult`? It's because this is the result value correct?

Same reason as `RPCRequestParams` to indicate that it is the params key.

HOWEVER if these are not just the `params` key or `result` key, then we should no longer call it `XParams` or `XResult`. It can just be `RPCRequest` and `RPCResponse`.

If however there's "raw RPC" request and response, then, `JSON` prefix can be added to disambiguate.

Can you elaborate on this? @amydevs

src/nodes/agent/handlers/VaultsGitInfoGet.ts
```diff
@@ -28,7 +28,7 @@ class VaultsGitInfoGet extends RawHandler<{
     _cancel,
     meta: Record<string, JSONValue> | undefined,
     _ctx: ContextTimed, // TODO: use
```
CMCDragonkai, line 30: This has a `TODO: use`, what's the situation here? @tegefaulkes @amydevs

src/nodes/agent/handlers/VaultsGitPackGet.ts
```diff
@@ -103,7 +102,7 @@ class VaultsGitPackGet extends RawHandler<{
         sideBand.destroy(e);
       },
     });
-    return [null, outputStream];
+    return [{}, outputStream];
```
CMCDragonkai, line 105: So now if it is an empty object, I'm assuming that this is not the `result` key.

image

Otherwise what was the reason to switch from `null` to `{}`?

src/nodes/agent/types.ts
CMCDragonkai, line 1: Are there tests that test the usage of the timeout in the agent service?

@tegefaulkes @amydevs

src/client/types.ts
CMCDragonkai: Ok I just read:

> Updated Type Requirements of js-rpc Request params and Response results
> The updated types in js-rpc of `JSONRPCRequestMessage`, `JSONRPCRequestNotification` and `JSONRPCResponseResult` mean that JSONRPCRequest*.params and JSONRPCResponseResult.result should now always be objects. This is so that metadata can always be appended to the params. Therefore, we need to make sure in Polykey that non-JSON-objects are never being passed as results or params.

So do you see my confusion here.

Why would there be `JSONRPCResponseResult.result`? I believe originally `JSONRPCResponseResult` IS literally the type of the `result` value inside the message as specified by the JSON RPC 2.0 spec.

If there is a `.result` property... then this type should no longer be called `JSONRPCResponseResult`. Because that doesn't make any sense!!!

Also in what situation do ever use `Message` over `Notification` or vice versa? I don't think we ever use JSON RPC 2.0 notifications right?

CMCDragonkai:
> Looks good

Nope, it doesn't look good. See comments above.

CMCDragonkai:
> Long-Running Streaming Calls

Do we have any calls in the entire client service or agent service that uses this yet?

I don't think so, but it is likely to be needed to provide push-flow abstractions over the network boundary.

CMCDragonkai:
@amydevs earlier @addievo meantioned the `pollAccessToken` process being used in `identities/providers/github/GitHubProvider.ts`.

I'm thinking was originally used to poll the GitHub API. Now with things like timed context, can we refactor that to push-down the timer context? That loop was PRE-invention of the timer contexts. And it seems it can be refactored to improve our system.

Perhaps it is necessary to actually solve the github auth timeout in its entirety?

src/nodes/agent/types.ts
CMCDragonkai: We should have some tests there to sanity check that this in fact works.

The node claiming process itself still hasn't been refactored to use RPC-like conversations, so I don't even know if it works. We need that to work to build up the gestalt graph too.

src/nodes/agent/handlers/VaultsGitPackGet.ts
amydevs: it is the result, because the result MUST be a JSON object now. So that the metadata can fit in there

src/client/types.ts
amydevs: i think `JSONRPCResponseResult` is currently Response with a Result. It was like that from before

src/nodes/agent/handlers/VaultsGitInfoGet.ts
amydevs: i have no idea

src/client/types.ts
amydevs: Imo, `JSONRPCResponseResult` should be renamed to `JSONRPCResponseSuccess`

src/nodes/agent/handlers/VaultsGitInfoGet.ts
tegefaulkes: It should use the CTX for timing out the handler.

amydevs:
@CMCDragonkai 

this would mean that the abstract `Provider` class would have to be updated accordingly to match. I think this would be out of scope with this PR

amydevs:
> > Long-Running Streaming Calls
> 
> Do we have any calls in the entire client service or agent service that uses this yet?
> 
> I don't think so, but it is likely to be needed to provide push-flow abstractions over the network boundary.

i don't think there are any places that need long-running calls at the moment

src/nodes/agent/handlers/VaultsGitInfoGet.ts
amydevs: ah alright, this should probably be added

src/client/types.ts
amydevs: the `JSONRPCResponseResult` and `JSONRPCRequestParams` types now restrict overriding of metadata types, but still allow for user to define metadata properties as long as they don't conflict with the ones already defined (timeout)

CMCDragonkai:
> @CMCDragonkai 
> 
> this would mean that the abstract `Provider` class would have to be updated accordingly to match. I think this would be out of scope with this PR

Ok

src/client/types.ts
CMCDragonkai: Where these type names changed?

src/nodes/agent/handlers/VaultsGitInfoGet.ts
amydevs: this will have to be added later in another PR

src/nodes/agent/types.ts
amydevs: this is going to have to be done later, as it is blocked by implementing usage in `nodes/agent` domain with usage of `ctx` in the handlers

amydevs:
should be ready to merge

src/nodes/agent/types.ts
CMCDragonkai: Ok create a new issue for those.

PR #595, opened 2023-10-19
Fixing identities claim to be in-line with identities discover and identities trust, by introducing a new parser parseGestaltIdentityId

### LINKED ISSUE #28: Crypto - Keypair Creation
It should be possible to create new key pairs. This can be used by other programs for whatever reason, or Polykey may change the master key of the keynode, but this may also involve re-encryption of any symmetric vault keys being used.

This is a low priority feature. Since many other programs usually allow creation of new keypairs to mean new digital identities. But in Polykey we don't really need to do that, since keynodes do not represent a new digital identity representing a person, but a unique agent that participates a in a P2P network.

But it can look like:

```
polykey crypto gen-keypair ./key ./key.pub
polykey crypto gen-keypair ./key
```

addievo:
RELATED TO https://github.com/MatrixAI/Polykey-CLI/pull/29

### Description

For the CLI fix, we need a new function to parse the gestalt identity, this introduces `parseGestaltIdentityId` in Polykey to achieve the same.

The `pk identities claim` currently takes two parameters in order to claim an identity. 

```
pk identities claim github.com cmcdragonkai
```

This is not consistent with other two commands `trust` and `discover`. 

```
pk identities trust github.com:cmcdragonkai
pk identities discover github.com:cmcdragonkai
```

These two commands are not using the `ProviderIdentityId` type. 

```
type ProviderIdentityId = [ProviderId, IdentityId];
```

This is because the encoded form of this type is `["github.com", "cmcdragonkai"]`.

Instead we need to use a parser similar to `binParsers.parseGestaltId`  (which is used by the trust and discover commands), however this parser, parses both node IDs and identities. 

What we need to do is to extract the identity parsing portion of `binParsers.parseGestaltId`.

```typescript
function parseGestaltId(data: any): GestaltId {
  if (typeof data !== 'string') {
    throw new validationErrors.ErrorParse('Gestalt ID must be string');
  }
  const nodeId = decodeNodeId(data);
  if (nodeId != null) {
    return ['node', nodeId];
  }
  const match = (data as string).match(/^(.+):(.+)$/);
  if (match == null) {
    throw new validationErrors.ErrorParse(
      'Gestalt ID must be either a Node ID or `Provider ID:Identity ID`',
    );
  }
  const providerId = parseProviderId(match[1]);
  const identityId = parseIdentityId(match[2]);
  return ['identity', [providerId, identityId]];
}
```

It will be:

```ts
  const match = (data as string).match(/^(.+):(.+)$/);
  if (match == null) {
    throw new validationErrors.ErrorParse(
      'Gestalt ID must be either a Node ID or `Provider ID:Identity ID`',
    );
  }
  const providerId = parseProviderId(match[1]);
  const identityId = parseIdentityId(match[2]);
  return ['identity', [providerId, identityId]];
```

This will be called `parseGestaltIdentityId`.

### Issues Fixed

* Fixes #[28](https://github.com/MatrixAI/Polykey-CLI/issues/28) 

### Tasks

- [x] 1. Introduce `parseGestaltIdentityId`

### Final checklist

* [x] Domain specific tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/ids/index.ts
+++ b/src/ids/index.ts
@@ -338,6 +338,21 @@ function parseGestaltId(data: any): GestaltId {
   return ['identity', [providerId, identityId]];
 }
 
+function parseGestaltIdentityId(data: any): ['identity', ProviderIdentityId] {
+  if (typeof data !== 'string') {
+    throw new validationErrors.ErrorParse('Gestalt identity ID must be string');
+  }
+  const match = (data as string).match(/^(.+):(.+)$/);
+  if (match == null) {
+    throw new validationErrors.ErrorParse(
+      'Gestalt identity ID must be `Provider ID:Identity ID`',
+    );
+  }
+  const providerId = parseProviderId(match[1]);
+  const identityId = parseIdentityId(match[2]);
+  return ['identity', [providerId, identityId]];
+}
+
 function encodeGestaltId(gestaltId: GestaltId): GestaltIdEncoded {
   switch (gestaltId[0]) {
     case 'node':
@@ -476,6 +491,7 @@ export {
   encodeProviderIdentityId,
   decodeProviderIdentityId,
   parseGestaltId,
+  parseGestaltIdentityId,
   encodeGestaltId,
   encodeGestaltNodeId,
   encodeGestaltIdentityId,


src/ids/index.ts
```diff
@@ -315,6 +315,21 @@ function parseGestaltId(data: any): GestaltId {
   return ['identity', [providerId, identityId]];
 }
 
+function parseGestaltIdentityId(data:any) : ProviderIdentityId {
```
tegefaulkes: If it's returning a `ProviderIdentityId` then it should be called `parseProviderIdentityId`.

addievo: @CMCDragonkai

CMCDragonkai: It is actually a `GestaltIdentityId`. That return type is incorrect.

CMCDragonkai: Sorry there's no type defined for `GestaltIdentityId`. It is actually just `GestaltId`.

CMCDragonkai:
```
type GestaltId = ['node', NodeId] | ['identity', ProviderIdentityId];
```

I want to keep it as `parseGestaltIdentityId` since `provider:identity` is part of a gestalt ID, not `ProviderIdentityId`. It's a special format just for gestalt IDs and only used on the CLI.

CMCDragonkai: Actually scratch that, rename this to `parseProviderIdentityId`. I just saw the output.

But please move it to the appropriate location.

CMCDragonkai:
@addievo in the future please name your branches a bit more specifically. We actually leave our feature branches around for a while before GCing them and your name is too generic, leaving plenty of opportunity for conflicts.

src/ids/index.ts
CMCDragonkai: Actually this was just incorrectly produced. The return type is just meant to be `['identity', ProviderIdentityId]`.

CMCDragonkai:
@addievo your commit messages are incorrect, see the new format. Also can you re-read the conventional commits specification.

src/ids/index.ts
CMCDragonkai: This is now done.

PR #600, opened 2023-10-24
Adding privateKeyPath back to PolykeyAgent 

addievo:
### Description

In 

https://github.com/MatrixAI/Polykey/pull/582

Changes were made to `PolykeyAgentOptions `and it was flattened to remove dependency on types, this resulted in `privateKeyPath `being removed from `PolykeyAgentOptions `which is causing issues in `Polykey-CLI`. 

This PR aims to add it back to PolykeyAgentOptions.

```ts
    privateKey: string;
    privateKeyPath: string;
```

### Tasks

- [x] 1. Add  `privateKey` and `privateKeyPath` back to `PolykeyAgentOptions`.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -2,7 +2,7 @@ import type { DeepPartial, FileSystem } from './types';
 import type { PolykeyWorkerManagerInterface } from './workers/types';
 import type { TLSConfig } from './network/types';
 import type { SeedNodes } from './nodes/types';
-import type { Key, PasswordOpsLimit, PasswordMemLimit } from './keys/types';
+import type { Key, PasswordOpsLimit, PasswordMemLimit, PrivateKey } from "./keys/types";
 import path from 'path';
 import process from 'process';
 import Logger from '@matrixai/logger';
@@ -61,6 +61,8 @@ type PolykeyAgentOptions = {
     certDuration: number;
     certRenewLeadTime: number;
     recoveryCode: string;
+    privateKeyPath: string;
+    privateKey: PrivateKey;
   };
   client: {
     keepAliveTimeoutTime: number;


src/PolykeyAgent.ts
```diff
@@ -61,6 +61,7 @@ type PolykeyAgentOptions = {
     certDuration: number;
     certRenewLeadTime: number;
     recoveryCode: string;
+    privateKeyPath: string;
```
tegefaulkes, line 64: Don't just fix what's missing, fix what is broken.

The previous types had 

```
  | {
      recoveryCode: RecoveryCode;
    }
  | {
      privateKey: PrivateKey;
    }
  | {
      privateKeyPath: string;
    }
```

So you want `privateKeyPath` and `privateKey` added back as an option. Double check everything relevant is included.

CMCDragonkai:
@addievo I don't think you lintfixed this?

CMCDragonkai:
* Your PR isn't lintfixed - this is why the lintfixing is meant to be done at the end.
* If the PR doesn't fix anything, delete that section
* Don't tick things off until the end
* Don't we need a default value for the configuration on agent options?

src/PolykeyAgent.ts
```diff
@@ -61,6 +61,8 @@ type PolykeyAgentOptions = {
...
+    privateKey: PrivateKey;
```
CMCDragonkai, line 65: The `privateKey: PrivateKey` should be using a more open type, instead the opaque type because of `PolykeyAgent` taking parameters from outside world.

So if you want to keep this then you should be using `privateKey: Buffer`.

Also `privateKey` should be moved up top.

```
privateKey: Buffer;
privateKeyPath: string;
```

CMCDragonkai: These 2 options aren't even being used, if these 2 options are to be used, they should be passed to the `KeyRing.createKeyRing`.

CMCDragonkai: @addievo link the new PR to fix this.

addievo: https://github.com/MatrixAI/Polykey/pull/601

CMCDragonkai: The `PolykeyAgent.start` should be taking `DeepPartial` of options, and the options need to `keys: { recoveryCode; privateKey; privateKeyPath; }`.

And also in the `KeyRing.createKeyRing` and `KeyRing.start` methods, these need to receive those options if you want to make them useful.

CMCDragonkai: Also the `PolykeyAgentOptions` needs to match:

```
  } & ( // eslint-disable-next-line @typescript-eslint/ban-types
    | {}
    | {
        recoveryCode: RecoveryCode;
      }
    | {
        privateKey: PrivateKey;
      }
    | {
        privateKeyPath: string;
      }
```

Which ensures mutually independent options.

CMCDragonkai: So:

```
keys: ({} | { recoveryCode: string; } | { privateKey: Buffer } | { privateKeyPath: string })
```

And `PolykeyAgent.start` should take this.

PR #602, opened 2023-10-24
PKClient needs to use WebSocketClient and RPCClient in the start/stop procedure

CMCDragonkai:
### Description

The PKClient needs to use WS and RPC in the start/stop so that we can start and stop the PK client to any given client service, while also destroying is for destroying the session state.

### Issues Fixed

* Related https://github.com/MatrixAI/Polykey-CLI/pull/42#issuecomment-1776794054

### Tasks

- [x] 1. Moved `WebSocketClient` and `RPCClient` into start/stop
- [x] 2. Exposed `PolykeyClient.nodeId`
- [x] 3. `PolykeyClient.start` now takes a subset of the creation parameters, and you can stop and start a connection to different client service

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "postversion": "npm install --package-lock-only --ignore-scripts --silent",
     "ts-node": "ts-node",
     "ts-node-inspect": "node --require ts-node/register --inspect",
-    "test": "node --expose-gc ./node_modules/.bin/jest",
+    "test": "jest",
     "lint": "eslint '{src,tests,scripts}/**/*.{js,ts,json}' 'benches/**/*.{js,ts}'",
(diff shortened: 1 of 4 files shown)


CMCDragonkai:
Note that `options` are never saved since they are always re-defaulted.

So when you create, stop, start. On the start call you have pass back whatever options you wanted. 

UX wise it might be unexpected. We could solve this in the future by saving the options on creation, and then having start default to the options set on creation if they exist, otherwise the default config. It would be a good idea to then extend `mergeObjects` to be multi-arity.

PR #611, opened 2023-10-26
Fix nodesGetAll handler

### LINKED ISSUE #44: CLI polish and fixes for deployment
### Specification

While testing hole punching with the testnet, we found a few issues with the CLI.

1. Nodes Get All is throwing an error
2. We are getting a weird output format for nodes connections what is the 0 and what is this NA for?
3. I think nodes connections is better suited for dictionary format - list format or table format (can you check which ones done robustly)
4. The output changes sometimes, to just join the NA after the ip address, so no port is available
5. Summary - the nodes connections command is just very non-deterministic
6. Secret leak is occurring on agent service, there should be no authorisation metadata see --format=json
8. What's usageCount

### Additional context
 
* Related: https://github.com/MatrixAI/Polykey/issues/551

### Tasks

- [x] 1. Fix up `nodes getall` command, it's throwing an error relating to `host` not being defined.
- [x] 2. Review and fix up the `nodes connections` command output formatting. The output is very unclear.
- [x] 3. client call metadata is being included in the `nodes connections --format json` output. This should not be the case.
- [x] 4. `agent status` command outputs too much information. We should strip out the key and certificate information.

### LINKED ISSUE #610: getAll handler is currently broken and needs fixing.
### Describe the bug

 The `NodesGetAll` handler needs to be fixed. I did a monkey patch while testing and it needs to be applied to `Polykey`.

```js
// client/handlers/NodesgetAll.ts

for await (const [index, bucket] of nodeGraph.getBuckets()) {
            for (const [id, info] of bucket) {
                const encodedId = nodesUtils.encodeNodeId(id);
                // For every node in every bucket, add it to our message
                if (ctx.signal.aborted)
                    throw ctx.signal.reason;
                yield {
                    bucketIndex: index,
                    nodeIdEncoded: encodedId,
                    host: info.address.host,
                    port: info.address.port,
                };
            }
        }
```

_Originally posted by @tegefaulkes in https://github.com/MatrixAI/Polykey-CLI/issues/44#issuecomment-1778347617_
            
### To Reproduce

1. Run `nodes getall ` in Polykey-CLI.

### Expected behavior

Shouldn't throw an error and succeed.

addievo:
### Description

`nodesGetAll` handler was obsolete and was breaking operation of `nodes getAll` on Polykey-CLI. 

This simple PR just fixes the handler.
 
Handler previously was: 

```ts
for await (const bucket of nodeGraph.getBuckets()) {
      let index;
      for (const id of Object.keys(bucket)) {
        const encodedId = nodesUtils.encodeNodeId(
          IdInternal.fromString<NodeId>(id),
        );
        // For every node in every bucket, add it to our message
        if (!index) {
          index = nodesUtils.bucketIndex(
            keyRing.getNodeId(),
            IdInternal.fromString<NodeId>(id),
          );
        }
        if (ctx.signal.aborted) throw ctx.signal.reason;
        yield {
          bucketIndex: index,
          nodeIdEncoded: encodedId,
          host: bucket[id].address.host,
          port: bucket[id].address.port,
        };
      }
    }
```

And now is changed to: 

```ts
  for await (const [index, bucket] of nodeGraph.getBuckets()) {
      for (const [id, info] of bucket) {
        const encodedId = nodesUtils.encodeNodeId(keyRing.getNodeId());
        // For every node in every bucket, add it to our message
        if (ctx.signal.aborted) {
          throw ctx.signal.reason;
        }
        yield {
          bucketIndex: index,
          nodeIdEncoded: encodedId,
          host: info.address.host,
          port: info.address.port,
        };
      }
    }
```

### Issues Fixed

* Related https://github.com/MatrixAI/Polykey-CLI/issues/44
* Fixes #610

### Tasks
- [x] 1. Update handler.
- [x] 2. Verify it works on Polykey-CLI.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/NodesGetAll.ts
+++ b/src/client/handlers/NodesGetAll.ts
@@ -26,25 +26,18 @@ class NodesGetAll extends ServerHandler<
   ): AsyncGenerator<ClientRPCResponseResult<NodesGetMessage>> {
     if (ctx.signal.aborted) throw ctx.signal.reason;
     const { nodeGraph, keyRing } = this.container;
-    for await (const bucket of nodeGraph.getBuckets()) {
-      let index;
-      for (const id of Object.keys(bucket)) {
-        const encodedId = nodesUtils.encodeNodeId(
-          IdInternal.fromString<NodeId>(id),
-        );
+    for await (const [index, bucket] of nodeGraph.getBuckets()) {
+      for (const [id, info] of bucket) {
+        const encodedId = nodesUtils.encodeNodeId(id);
         // For every node in every bucket, add it to our message
-        if (!index) {
-          index = nodesUtils.bucketIndex(
-            keyRing.getNodeId(),
-            IdInternal.fromString<NodeId>(id),
-          );
+        if (ctx.signal.aborted) {
+          throw ctx.signal.reason;
         }
-        if (ctx.signal.aborted) throw ctx.signal.reason;
         yield {
           bucketIndex: index,
           nodeIdEncoded: encodedId,
-          host: bucket[id].address.host,
-          port: bucket[id].address.port,
+          host: info.address.host,
+          port: info.address.port,
         };
       }
     }
--- a/tests/client/handlers/nodes.test.ts
+++ b/tests/client/handlers/nodes.test.ts
@@ -2,6 +2,7 @@ import type GestaltGraph from '@/gestalts/GestaltGraph';
 import type { NodeIdEncoded } from '@/ids/types';
 import type { TLSConfig, Host, Port } from '@/network/types';
 import type { Notification } from '@/notifications/types';
+import type { NodeAddress } from '@/nodes/types';
 import fs from 'fs';
 import path from 'path';
 import os from 'os';
@@ -18,12 +19,27 @@ import Sigchain from '@/sigchain/Sigchain';
 import NotificationsManager from '@/notifications/NotificationsManager';
 import NodeConnectionManager from '@/nodes/NodeConnectionManager';
 import ClientService from '@/client/ClientService';
-import { NodesAdd, NodesClaim, NodesFind, NodesPing } from '@/client/handlers';
-import { nodesAdd, nodesClaim, nodesFind, nodesPing } from '@/client/callers';
+import {
+  NodesAdd,
+  NodesClaim,
+  NodesFind,
+  NodesPing,
+  NodesGetAll,
+  NodesListConnections,
+} from '@/client/handlers';
+import {
+  nodesAdd,
+  nodesClaim,
+  nodesFind,
+  nodesPing,
+  nodesGetAll,
+  nodesListConnections,
+} from '@/client/callers';
 import * as keysUtils from '@/keys/utils';
 import * as nodesUtils from '@/nodes/utils';
 import * as networkUtils from '@/network/utils';
 import * as validationErrors from '@/validation/errors';
+import { parseNodeId } from '@/ids';
 import * as testsUtils from '../../utils';
 
 describe('nodesAdd', () => {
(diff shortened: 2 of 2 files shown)


tegefaulkes:
FYI, you don't need to code dump in the issue spec. *Implementation detail* is less relevant than requirements.

You want clear specification of what needs to be done, How the code looks like isn't that use. Focus on what needs to be done and what is required of the code. Not how it will look.

src/client/handlers/NodesGetAll.ts
```diff
@@ -26,25 +26,18 @@ class NodesGetAll extends ServerHandler<
...
     const { nodeGraph, keyRing } = this.container;
-    for await (const bucket of nodeGraph.getBuckets()) {
-      let index;
-      for (const id of Object.keys(bucket)) {
-        const encodedId = nodesUtils.encodeNodeId(
-          IdInternal.fromString<NodeId>(id),
-        );
+    for await (const [index, bucket] of nodeGraph.getBuckets()) {
+      for (const [id, info] of bucket) {
+        const encodedId = nodesUtils.encodeNodeId(keyRing.getNodeId());
```
tegefaulkes: We need to get the `NodeId` from the information we're iterating over. So use the `id` and not the `keyRing.getNodeId()` which would be our own node id..

tegefaulkes (changes requested):
You also need to review the tests for this. no tests caught the problem so either they are insufficient or non-existent. Look into it.

src/client/handlers/NodesGetAll.ts
addievo: Oh, understood.

addievo:
> You also need to review the tests for this. no tests caught the problem so either they are insufficient or non-existent. Look into it.

You are right, currently no tests exist for getAll. I will make one to test the same.

tegefaulkes:
Likely there is no test for the `Nodes Connections` as well.

addievo:
> Likely there is no test for the `Nodes Connections` as well.

Yep, there are currently only tests for `add`, `claim`, `find`, and `ping`.

src/client/handlers/NodesGetAll.ts
addievo: This is now resolved.

tests/client/handlers/nodes.test.ts
```diff
@@ -680,3 +694,266 @@ describe('nodesPing', () => {
     );
   });
 });
+describe('nodesGetAll', () => {
+  const logger = new Logger('nodesPing test', LogLevel.WARN, [
+    new StreamHandler(
```
tegefaulkes: Update the logger name to reflect  what you're actually testing. Same for the other one.

tests/client/handlers/nodes.test.ts
```diff
@@ -680,3 +694,266 @@ describe('nodesPing', () => {
...
+    await fs.promises.rm(dataDir, {
+      force: true,
+      recursive: true,
+    });
+  });
+  test('gets all nodes', async () => {
+    mockedPingNode.mockResolvedValue(false);
+    const response = await rpcClient.methods.nodesGetAll({});
+    expect(response).toBeTruthy();
+  });
```
tegefaulkes: You shouldn't really copy the nodes ping tests wholesale. Make sure you're only using the dependencies you need. We don't need mocking for the ping node here.

Also, test more than there is a response. Make sure the expected data is correct as well.

Same needs to be done for the nodes connections test below.

tegefaulkes (changes requested):
Just a little more polish, Once that is done, so long as all the test pass, you can proceed to merge.

CMCDragonkai:
Good work!

Issue #605, opened 2023-10-25
Fix hole punching to work in both directions


tegefaulkes:
### Specification

While testing the testnet and hole punching. We found that while a connection could be punched between the two test machines. The connection could be established with hole punching from `NodeA` -> `NodeB` just fine. But starting the connection in reverse with hole punching `NodeB` -> `NodeA` could not be done.

With this we can conclude that the hole punching does work, we can see it working with the forward case. But there is some flaw with the hole punching procedure that prevents it working with certain kinds of `NAT`s.

Before we can move forward with testing this, other things need to be cleaned up. 
1. We need more informative logs when t comes to the hole punching procedure.
2. We need to reduce the amount of noise in the logging output. As it stands there is too much logging about connections. We need a discussion about the amount of logging that is appropriate and what kind of logging isn't needed at the `INFO` level.

Then we can move on to Identifying why connections can only be punched one way. I first guess is that we are not providing the correct information when signalling. Right now the signalling protocol provides the src host:port to the target, but I don't think it's providing the target information back to the src node. That is one thing to look into.

### Additional context

* Related #551
* Related #555

### Tasks

1. Identify why a forward connection can be punched, but the reverse can't be punched.
2. Apply a fix

CMCDragonkai:
The local node graph is out of date. The signalling node has the up to date information.

The local node's node graph isn't GCing the old node record. And we are not getting information.

The existing records could be wrong.

CMCDragonkai:
**We are going to be moving to a node only connecting to one of the seed node.** This is necessary for scalability, we can't have a node connecting to every node.

We will need to find the common seed node to do this, if source node is connected to seed node A, and target node is connected to seed node B. Then right now source node has to connect to seed node B to do the hole punching.

Afterwards during a hole punching request to the seed node on the response you could get new information about the target node, which may be more up to date than your current information.

If it does update the current node record, then the current hole punching request has to be cancelled, and restarted with the new node record.

In general we need a better policy for getting up to date node records.

CMCDragonkai:
Up to date node records:

> Effect on Ongoing Operations: If the failed node was part of an ongoing lookup operation, the local node will continue the lookup by querying other closest nodes from its routing table. The failed node is skipped, and its failure impacts only the efficiency but not the correctness of the lookup operation.
>
> Retry Policy: Some implementations incorporate a retry mechanism before marking a node as stale or removing it. This accounts for temporary network issues or delays that might have caused the failed connection attempt.
[...]

PR #624, opened 2023-11-09
feat: connection with ICE now gets target port from signalling node

### LINKED ISSUE #614: NodesListConnections handler is randomly mapping IP addresses as IPv6 mapped IPv4
### Specification

Whilst testing Polykey CLI's `nodes getall` or `nodes connections`, the IP address is getting listed as IPv6 mapped IPv4, this is inaccurate as all the testing has been performed on IPv4 addresses only. 

This behaviour needs to be investigated and consequently resolved.

### Tasks

1. Investigate where this occurs.
2. Investigate why this occurs.
3. Fix mapping.

tegefaulkes:
### Description

This PR tries to address the issue with #605. It's hard to say what the exact problem is but I'm assuming the forward connection isn't getting the most up to date information.

To attempt to fix this, this PR does the following.

1. The `NodesConnectionSignalInitial` returns the host and port of the target node from the perspective of the signalling node.
2. When doing ICE, the `nodesConnectionSignalInitial` is called returns the new address. The new port is then used when attempting the connection. 

### Issues Fixed

* Related #605 - I'm not saying it's fixed because the current solution is based on conjecture. I'll need to do manual testing to know that it addressed the problem. And we'll need some test environments that provide different kinds of NATs to test against.
* Fixes #614 

### Tasks

- 1. [X] Fix the signalling process to return the target's port from the perspective of the signalling node
- 2. [x] Fix `IPv6` mapped `IPv4` addresses being used in the NodeGraph. #614 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/nodes/NodeConnection.ts
+++ b/src/nodes/NodeConnection.ts
@@ -8,6 +8,7 @@ import type {
   QUICSocket,
   ClientCryptoOps,
   QUICConnection,
+  Host as QUICHost,
 } from '@matrixai/quic';
 import type { ContextTimedInput } from '@matrixai/contexts/dist/types';
 import type { X509Certificate } from '@peculiar/x509';
@@ -20,6 +21,7 @@ import {
   QUICClient,
   events as quicEvents,
   errors as quicErrors,
+  utils as quicUtils,
 } from '@matrixai/quic';
 import { RPCClient } from '@matrixai/rpc';
 import { middleware as rpcUtilsMiddleware } from '@matrixai/rpc';
@@ -464,9 +466,11 @@ class NodeConnection<M extends ClientManifest> {
   }) {
     this.validatedNodeId = validatedNodeId;
     this.nodeId = nodeId;
-    this.host = host;
+    this.host = quicUtils.toCanonicalIP(host) as unknown as Host;
     this.port = port;
-    this.localHost = localHost;
+    this.localHost = quicUtils.resolvesZeroIP(
+      localHost as unknown as QUICHost,
+    ) as unknown as Host;
     this.localPort = localPort;
     this.certChain = certChain;
     this.hostname = hostname;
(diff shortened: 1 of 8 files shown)


tegefaulkes:
I made a PR for this to review the changes. We need to coordinate the design between this and the decentralised signalling process that @CMCDragonkai is working on.

CMCDragonkai:
image

If the connection worked from my end, it has to work from your end. The "strict nat" idea does not make sense here, because it's always a simultaneous interaction between both nodes.

The first step is to prove empirically if simultaneous interaction is happening. We know node 1 was able to connect to node 2. But node 2 was not able to connect to node 1 when it was initiating. So dump the logs for sending packets to each other.

CMCDragonkai:
Also we cannot wait until we get a response from the signaller until we start a direct connection. We should start a direct connection, and also ask the signaller, and get a result if it is different, cancel the direct connection and start a new direct connection.

CMCDragonkai:
Also during our initial test, it shouldn't be possible that our local node graph is already out of date, because we got that information about node 2, already from the same signaller.

CMCDragonkai:
So I don't think the current PR fixes the problem. I suspect there's an issue with the signalling logic, we need logs here to show that there's simultaneous holepunching occurring with the correct ports.

CMCDragonkai:
Consider this experiment, what if you initiated signalling and held off from direct connections. And then waited a long enough time for the other side to be attempting a direct connection, then start your own direct connection. Logically one would argue this is the equivalent of when the other side is just connecting to you via ICE.

tegefaulkes:
After some extra testing I've determined that #614 also needs to be addressed in this PR to fix the issue.

tegefaulkes:
I'm going to merge this now and move on to testing it. If the problem is resolved I'll close #605.

PR #618, opened 2023-11-07
Decentralized Signalling

CMCDragonkai:
### Description

This introduces decentralized signalling which allows the PK nodes to:

1. Use SRV records to resolve the seed nodes - this supports #599, but is also necessary to allow dynamic seed nodes
2. Connect to only a subset of the seed nodes, rather than always connecting to all seed nodes - thus allowing us to scale the seed nodes as the network gets larger
3. Find the nearest signalling node to use to try and help connect to a target node, this process is iterative, starting from the final target node, and walking backwards through a chain of nodes that are connected to the final target node
4. It will have to remove the hardcoded seed node list in order to use SRV records, however this PR won't try to solve the #154 problem with having a root certificate of the nodes, so that will be done later to re-introduce trusted seed nodes
5. It should mean that the load on the seed nodes decreases as it can rely on other nodes to participating in the signalling coordination - this should be optional, as some nodes may not want to do so
6. This will also be maintaining random connections in order to maintain connectivity of the network graph assuming all nodes were to be behind the NAT. This randomness may be directed with some policy.

### Issues Fixed

* Fixes #365 
* Related #182 
* Related https://github.com/MatrixAI/Polykey-CLI/issues/40
* Related #537 - especially because the MDNS nodes aren't part of the node graph atm - so they represent a different source of nodes, so ICE connections may need to consider these separately
* Related #372 - this PR was only for simulating code, it is superseded by this PR

### Tasks

- [x] 1. `NodeGraph.getOldestNode` is replaced with `NodeGraph.getBucket` with additional optional`limit` parameter - update all tests
- [x] 2. `NodeGraph.setNode` now respects the `NodeGraph.nodeBucketLimit` and will throw `ErrorNodeGraphBucketLimit` if it goes over the limit when a new node is being set.
- [x] 3. The `NodeGraph.nodeBucketLimit` is no longer hardcoded, it can be set during creation/construction.
- [x] 4. Removed debug logs from `NodeGraph` as these shouldn't be needed after behaviour is verified.
- [x] 5. Added `lastUpdated` parameter to `NodeGraph.setNode`, allowing one to pass down last updated time from a higher context, and the ability to get rid of sleeps by manually setting the time. It defaults to `utils.getUnixtime()`.
- [x] 6. Removed `utils.sleep` usage in `NodeGraph.test.ts` and now all tests execute much faster.
[...]

PR #634, opened 2023-11-16
`audit` Domain

### LINKED ISSUE #628: Setup `audit` domain for tracking user/action events and metrics
### Specification

The `audit` domain of Polykey will be responsible for the auditing user-behaviour of Polykey nodes.

It should expose `JS-RPC` server-streaming handlers that yield audit events, and can also provide summary metrics of those audit events.

The subdomains of `audit` should be based on the domains of Polykey itself, and the audit domain should also contain Observable properties (see #444) derived from the events dispatched by each subdomain.

The `JS-RPC` API should be available via `JS-WS`, so that it is accessible from the services like the mainnet/testnet status page (see #599).

Furthermore, this `JS-RPC` API should replay all accumulated state for each metric upon initial opening of the server-streaming call, so that if the any connected services were to restart, they would be able to get all the existing metric data, much like how the rxjs [shareReplay](https://www.learnrxjs.io/learn-rxjs/operators/multicasting/sharereplay) funciton.

Untitled-2023-10-23-0424 excalidraw(5)

##### Event Flow

Domain class instances will be injected dependencies into the Audit domain. This means that the other domains will be able to expose any data they want to record via Events without any semantics regarding the Audit domain. The Audit domain can listen to these events and record them in the database.

##### Database Schema

Using `js-db`, there will be several level for the `Audit` domain:

- `audit/` - The base Level
- `audit/topic/{topicId}` - Topics Level
- `audit/events/{eventId}` - Events level

`eventId`s will be made using `IdSortable`, so that they are completely monotonic. Furthermore, events can be accessed by iterating over a topic level (`audit/topic/{topicId}`), which yields multiple `eventId`s. This will be used to reference the events stored in the `audit/events/{eventId}`. By doing this, events are able to be apart of multiple topics as well.

Topics can be nested meaning that querying topic path of `['node', 'connection']` will return all audit events from their children (`['node', 'connection', 'reverse']` and `['node', 'connection', 'forward']`)

##### API

The basic API will use an AsyncGenerator that yields events from a specified topic:

```ts
function* getAuditEvents(topicPath: Array<string>, options: { seek?: EventId, seekEnd?: EventId, order?: 'asc' | 'desc', limit?: number }, tran?: DBTransaction): AsyncGenerator<AuditEvent>
```

The options offer pagination, where the user can limit the number of audit events that the generator will yield and call the generator again with `seek` set to the EventId of the last element that was yielded.

The second API method yields events live as they are being dispatched:

```ts
function* getAuditEventsLongRunning(topicPath: Array<string>, options: { seek?: EventId, seekEnd?: EventId, limit?: number }): AsyncGenerator<AuditEvent>
```

The reason why the paramaters are different is because that the iteration of new events beyond what is currently stored within the DB cannot be in any order other than chronologically ascending. Furthermore, as this method requires for multiple db transaction snapshots, there is no point for the caller to pass in a transaction to perform on. Note that `generator.return()` or `generator.throw()` must be called on the returned `AsyncGenerator` when either `seekEnd` or `limit` is not specified, as this call will run indefinitely until either `throw` or `return` is called, `seekEnd` or `limit` is reached, or `audit.stop({ force: true })` is called.

#### Metrics

Metrics will need to be specc'd out further. However, currently, metrics are indexed by a `MetricPath` similar to `AuditEvents`. However, they are not stored in the DB, but rather derived from data within the DB.

#### API

The basic API returns a metric   based on a `topicPath` and allows for the input of `seek` and `seekEnd` to specify a specific timeframe for the metric results. Metrics will have to be implemented on a case by case basis.

```ts
async function getAuditMetric(topicPath: Array<string>, options: { seek?: EventId, seekEnd?: EventId }): Promise<AuditMetric>
```

#### Possibly Relevant Metrics

Some Specific Metrics Include:
- Active Connections
- Node Graph Size
- Gestalt Graph Size
- Sigchain Claims
- Geographical Data
- etc.

### Additional context

- https://github.com/MatrixAI/Polykey/issues/386
- https://github.com/MatrixAI/Polykey/issues/444
- https://github.com/MatrixAI/Polykey/issues/599
- https://github.com/MatrixAI/Polykey-CLI/issues/36
- #237 
- #179 - an old issue about audit logging

### Tasks

1. Expose metrics from Polykey domains as events
2. Use Observables to convert EventTargets to Streams
3. Convert Observable Streams to WebStreams/AsyncIterables for Usage with `JS-RPC`.

amydevs:
### Description
This PR implements an `audit` domain that collects events from various domains and stores them in `js-db`, serving them via a `js-rpc` API to external applications. 

This PR currently only implements AuditEvents and AuditMetrics for `['node', 'connection']` topics/metrics.

More will be implemented incrementally as needed.

### Issues Fixed

* Fixes #628 

### Tasks

- [x] 1. Design Topic and Event types
- [x] 2. Audit Domain
- [x] 3. Implement Audit domain js-rpc handlers

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -749,6 +749,7 @@ class PolykeyAgent {
         manifest: clientServerManifest({
           polykeyAgent: this,
           acl: this.acl,
+          audit: this.audit,
           certManager: this.certManager,
           db: this.db,
           discovery: this.discovery,
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -13,6 +13,7 @@ import {
   ready,
 } from '@matrixai/async-init/dist/CreateDestroyStartStop';
 import { WorkerManager } from './workers';
+import Audit from './audit/Audit';
 import KeyRing from './keys/KeyRing';
 import CertManager from './keys/CertManager';
 import Status from './status/Status';
@@ -220,6 +221,7 @@ class PolykeyAgent {
     let vaultManager: VaultManager | undefined;
     let sessionManager: SessionManager | undefined;
     let clientService: ClientService | undefined;
+    let audit: Audit | undefined;
     try {
       status = new Status({
         statusPath,
@@ -442,6 +444,12 @@ class PolykeyAgent {
         rpcParserBufferSize: optionsDefaulted.client.rpcParserBufferSize,
         logger: logger.getChild(ClientService.name),
       });
+      audit = await Audit.createAudit({
+        db,
+        nodeConnectionManager,
+        fresh,
+        logger: logger.getChild(Audit.name),
+      });
     } catch (e) {
       logger.warn(`Failed Creating ${this.name}`);
       await sessionManager?.stop();
@@ -455,6 +463,7 @@ class PolykeyAgent {
       await sigchain?.stop();
       await certManager?.stop();
       await taskManager?.stop();
+      await audit?.stop();
       await db?.stop();
       await keyRing?.stop();
       await schema?.stop();
@@ -463,6 +472,7 @@ class PolykeyAgent {
     }
     const pkAgent = new this({
       nodePath: optionsDefaulted.nodePath,
+      audit,
       status,
       schema,
       keyRing,
@@ -502,6 +512,7 @@ class PolykeyAgent {
   }
 
   public readonly nodePath: string;
+  public readonly audit: Audit;
   public readonly status: Status;
   public readonly schema: Schema;
   public readonly keyRing: KeyRing;
@@ -548,6 +559,7 @@ class PolykeyAgent {
 
   constructor({
     nodePath,
+    audit,
     status,
     schema,
     keyRing,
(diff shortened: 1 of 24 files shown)


CMCDragonkai:
@amydevs can you review https://github.com/MatrixAI/js-quic/issues/64 in relation to this work? It may conflict, we had some ideas about optimising the event dispatching.

amydevs:
@CMCDragonkai i'm more concerned with this conflicting with #618 because of the changes made to the nodes domains, i'm going to have to rewrite the handlers and reinsert the event dispatches in the correct places once the nodes domain api is changed. I would rather that PR be merged so i can rebase and write my event changes on top of it so that i don't have to do the same work twice.

CMCDragonkai:
Any expected changes, make sure you're putting a comment on #618, also rebase on top of staging cause @tegefaulkes made some reversions to the diagnostics data that should become a `diagnostics` domain.

PR #654, opened 2023-12-13
Fixes the vault cloning issue when trying to clone a vault with the same name as one already present 

### LINKED ISSUE #652: Vaults clone with name of a vault already existing fails
### Specification

If a vault already exists with the same name as the one being cloned, polykey throws "undefined". 

This is a bug and the clone option needs to be expanded with the proper errors to demonstrate why it is failing and expand the clone option to allow force cloning.

### Tasks

1. Fix vaultManager
2. Fix handler
3. Fix CLI

Related https://github.com/MatrixAI/Polykey-CLI/issues/73

addievo:
### Description

The vaults clone handler now has a force flag to override the default cloning behaviour of throwing undefined when the name was  already in the Vault Map.

The override flag has also been amended in the types.

```ts

        // Get the list of existing vaults
        const existingVaults = await this.listVaults(tran);

        if (existingVaults.has(newVaultName)) {
          if (force) {
            newVaultName = `${newVaultName}_1`;
          } else {
            throw new vaultsErrors.ErrorVaultsNameConflict(
              `A vault with the name '${newVaultName}' already exists. Use --override to force cloning.`,
            );
          }
        }
```

Now the vaultsClone function checks with listVaults if a vault with the same name exists and then throws a error (default behaviour) 

The user can use the --override call to force clone the vault, the new vault name is appended with a "_1".

The default error message includes the information about the override call.

### Issues Fixed

* Fixes #652

### Tasks

- [x] 1. Update VaultClone method
- [x] 2. Update the types for the handler
- [x] 3. Update the handler

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsClone.ts
+++ b/src/client/handlers/VaultsClone.ts
@@ -39,9 +39,10 @@ class VaultsClone extends UnaryHandler<
         nodeId: input.nodeIdEncoded,
       },
     );
+    const force = input.force ?? false;
     // Vault id
     await db.withTransactionF(async (tran) => {
-      await vaultManager.cloneVault(nodeId, input.nameOrId, tran);
+      await vaultManager.cloneVault(nodeId, input.nameOrId, tran, force);
     });
     return {
       success: true,
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -228,7 +228,10 @@ type VaultIdentifierMessage = {
   nameOrId: VaultIdEncoded | VaultName;
 };
 
-type CloneMessage = NodeIdMessage & VaultIdentifierMessage;
+type CloneMessage = NodeIdMessage &
+  VaultIdentifierMessage & {
+    force?: boolean;
+  };
 
 type VaultListMessage = VaultNameMessage & VaultIdMessage;
 
--- a/src/vaults/VaultManager.ts
+++ b/src/vaults/VaultManager.ts
@@ -686,6 +686,7 @@ class VaultManager {
     nodeId: NodeId,
     vaultNameOrId: VaultId | VaultName,
     tran?: DBTransaction,
+    force: boolean = false,
   ): Promise<VaultId> {
     if (tran == null) {
       return this.db.withTransactionF((tran) =>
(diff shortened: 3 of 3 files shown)


addievo:
Related

https://github.com/MatrixAI/Polykey-CLI/issues/73
https://github.com/MatrixAI/Polykey-CLI/pull/79

tegefaulkes:
This seems like a bad fix. AFAIK the vaults domain is already meant to increment the name if it already exists. That must be broken right now.

Have a look at `src/vaults/VaultManager.ts:719`.

tegefaulkes:
Oh, that's where you made the change. The code right below it already deals with incrementing the name.

```ts
     while (true) {
          const existingVaultId = await tran.get([
            ...this.vaultsNamesDbPath,
            newVaultName,
          ]);
          if (existingVaultId == null) break;
          newVaultName = `${baseVaultName}-${attempts}`;
          if (attempts >= 50) {
            throw new vaultsErrors.ErrorVaultsNameConflict(
              `Too many copies of ${baseVaultName}`,
            );
          }
          attempts++;
        }
```

Is this not working? You need to investigate.

CMCDragonkai:
Yea this is not intended. The vault name is supposed to be incremented.

CMCDragonkai:
Not accepting this change. The vaults manager is supposed to increment the name by adding a suffix.

CMCDragonkai:
The original design was to behave similar to Dropbox.

CMCDragonkai:
Along with a warning that the end user should see.

CMCDragonkai:
Actually the CLI could have a prompt for the end user to give a different name.

Vault name doesn't actually matter. 2 agents can have the same vault with different names. The user could supply an alternative name.

addievo:
> Oh, that's where you made the change. The code right below it already deals with incrementing the name.
> 
> ```ts
>      while (true) {
>           const existingVaultId = await tran.get([
>             ...this.vaultsNamesDbPath,
>             newVaultName,
>           ]);
>           if (existingVaultId == null) break;
>           newVaultName = `${baseVaultName}-${attempts}`;
>           if (attempts >= 50) {
>             throw new vaultsErrors.ErrorVaultsNameConflict(
>               `Too many copies of ${baseVaultName}`,
>             );
>           }
>           attempts++;
>         }
> ```
> 
> Is this not working? You need to investigate.

Yea this was broken @tegefaulkes 

I'll investigate and fix

tegefaulkes:
Since this is already started, I'm taking it over.

tegefaulkes:
I'm going to close this PR and start from scratch. It's using a cloned repo rather than a feature branch on this repo.

PR #670, opened 2024-02-06
fix: fixed vaults cloning handling duplicate vault name

### LINKED ISSUE #652: Vaults clone with name of a vault already existing fails

tegefaulkes:
### Description

This PR addresses the issue where cloning a vault with a name that already exists fails.

### Issues Fixed

Fixes #652 

### Tasks
- [x] 1. Find and fix problem
- [x] 2. Add a test to check for regressions

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/vaults/VaultManager.ts
+++ b/src/vaults/VaultManager.ts
@@ -720,10 +720,10 @@ class VaultManager {
         let newVaultName = baseVaultName;
         let attempts = 1;
         while (true) {
-          const existingVaultId = await tran.get([
-            ...this.vaultsNamesDbPath,
-            newVaultName,
-          ]);
+          const existingVaultId = await tran.get(
+            [...this.vaultsNamesDbPath, newVaultName],
+            true,
+          );
           if (existingVaultId == null) break;
           newVaultName = `${baseVaultName}-${attempts}`;
           if (attempts >= 50) {
--- a/tests/vaults/VaultManager.test.ts
+++ b/tests/vaults/VaultManager.test.ts
@@ -934,7 +934,7 @@ describe('VaultManager', () => {
     //           });
     //         },
     //       );
-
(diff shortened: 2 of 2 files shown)


CMCDragonkai:
What is the change in behaviour?

tegefaulkes:
No change is behaviour. There was existing logic that was broken. We lacked a test that caught any regressions so it's been broken for some time.

So after the fix it has the same behaviour where the name ins incremented with `${name}-${attempt}` until it finds a free name up to attempt 50.

CMCDragonkai:
So it's just a matter of incrementing the name in this case. If so, no need to hardcode a limit of 50. Also why don't we read the last name and add 1 to it? It's all transactional now with the DB. It does require re-attempting each time?

tegefaulkes:
It's pretty old code. I was just getting it working and not really overhauling it since that's a bit out of scope right now.

There's not really a way to know the latest name for any given vaultname?

If we want to change how this works we need to leave it to a new issue for now. Maybe spec out something more robust.

CMCDragonkai:
Aren't vault names in the DB, they aren't indexed by sort order?

Either way, an arbitrary limit should not be a fix, that's way to easy to create entropy.

CMCDragonkai:
And yes it should be specced out what this flow should be.

Issue #76, opened 2020-08-12
Vault Schema


CMCDragonkai:
Vault schemas describe the content of a vault and limits the media types that is allowed to go into the vault files.

A vault can have an attached schema. When a schema is attached, it autogenerates the marshaling code and acts like a validation layer.

This allows us to represent things like credit card data, or username password combos, or tls certificates, public and private keys.

Can reuse ideas from openapi, json schema and semantic web.

For research now, not priority.

robert-cronin:
Closing on account of migration to gitlab

PR #675, opened 2024-02-14
Adding `secrets env command` RPC logic

tegefaulkes:
### Description

This PR adds the RPC method required by the `Polykey-CLI` `pk secrets env` command.

### Issues Fixed

* Related https://github.com/MatrixAI/Polykey-CLI/pull/129

### Tasks
- [x] 1. Add `VaultsSecretsEnv` RPC method.
- [x] 2. Needs to stream back requested ENV variables and their contents.
- ~3. Needs to support globbing and possibly regex selection~ To complex for now, focusing on basic functionality
- [x] 4. Specifying a directory path will get all contents of that directory

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/callers/index.ts
+++ b/src/client/callers/index.ts
@@ -60,6 +60,7 @@ import vaultsRename from './vaultsRename';
 import vaultsScan from './vaultsScan';
 import vaultsSecretsDelete from './vaultsSecretsDelete';
 import vaultsSecretsEdit from './vaultsSecretsEdit';
+import vaultsSecretsEnv from './vaultsSecretsEnv';
 import vaultsSecretsGet from './vaultsSecretsGet';
 import vaultsSecretsList from './vaultsSecretsList';
 import vaultsSecretsMkdir from './vaultsSecretsMkdir';
@@ -135,6 +136,7 @@ const clientManifest = {
   vaultsScan,
   vaultsSecretsDelete,
   vaultsSecretsEdit,
+  vaultsSecretsEnv,
   vaultsSecretsGet,
   vaultsSecretsList,
   vaultsSecretsMkdir,
@@ -209,6 +211,7 @@ export {
   vaultsScan,
   vaultsSecretsDelete,
   vaultsSecretsEdit,
+  vaultsSecretsEnv,
   vaultsSecretsGet,
   vaultsSecretsList,
   vaultsSecretsMkdir,
--- a/src/client/callers/vaultsSecretsEnv.ts
+++ b/src/client/callers/vaultsSecretsEnv.ts
@@ -1,0 +1,12 @@
+import type { HandlerTypes } from '@matrixai/rpc';
+import type VaultsSecretsEnv from '../handlers/VaultsSecretsEnv';
+import { DuplexCaller } from '@matrixai/rpc';
+
+type CallerTypes = HandlerTypes<VaultsSecretsEnv>;
+
+const vaultsSecretsEnv = new DuplexCaller<
+  CallerTypes['input'],
+  CallerTypes['output']
+>();
+
+export default vaultsSecretsEnv;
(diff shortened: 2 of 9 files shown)


tegefaulkes:
I've added the ability to specify a directory when selecting secrets. This will walk the directory and return all of the secrets in that tree. This will be useful for selecting groups of envs you want to use. The grouping could get pretty complex if we use linking but I don't think we really support that in our vault ops right now.

While making these changes I've come across some questions.
1. How do we want to handle duplicate secrets? It would be possible to have secrets with the same name between directories. Or even specify a secret twice. How do we want to handle this? We could either throw an error due to duplicates or we take the first/last secret with that name if it's duplicate.
2. Do we want to handle file system cycles with linking? Do we even bother right now? As I said I don't think we support linking in any of our secret operations.
3. Do we want some way to rename an env when selecting it. We can already parse this in the CLI command where `secrets env -e vault1:secret1=NEW_NAME`. Question is, do we bother with this? How would we handle re-naming when selecting a group like a secret directory? Likely just ignore re-naming for groups. Or group renaming would be a prefix to the secret name?
4. I already have the ability to select a directory of secrets. Do we want to support globbing or regex selection at all? For globbing I'll need to convert the globbing pattern to actually filtering the secrets. The regex is mostly a problem of specifying and parsing the regex in the CLI.

tegefaulkes:
I'm going to skip the regex and globbing for now. I want to get the rest of the functionality going before I dive into those weeds.

CMCDragonkai:
> I've added the ability to specify a directory when selecting secrets. This will walk the directory and return all of the secrets in that tree. This will be useful for selecting groups of envs you want to use. The grouping could get pretty complex if we use linking but I don't think we really support that in our vault ops right now.
> 
> While making these changes I've come across some questions.
> 
> 1. How do we want to handle duplicate secrets? It would be possible to have secrets with the same name between directories. Or even specify a secret twice. How do we want to handle this? We could either throw an error due to duplicates or we take the first/last secret with that name if it's duplicate.
> 2. Do we want to handle file system cycles with linking? Do we even bother right now? As I said I don't think we support linking in any of our secret operations.
> 3. Do we want some way to rename an env when selecting it. We can already parse this in the CLI command where `secrets env -e vault1:secret1=NEW_NAME`. Question is, do we bother with this? How would we handle re-naming when selecting a group like a secret directory? Likely just ignore re-naming for groups. Or group renaming would be a prefix to the secret name?
> 4. I already have the ability to select a directory of secrets. Do we want to support globbing or regex selection at all? For globbing I'll need to convert the globbing pattern to actually filtering the secrets. The regex is mostly a problem of specifying and parsing the regex in the CLI.

We should keep this command simple first, without increasing its automation. Directories aren't supposed to work like that, remember we still have vault schema concepts #76 that hasn't been fully worked out. So I'd rather keep it simple, and you should just specify a path to the secret path within a vault like `vaultName:path/to/secret`.

CMCDragonkai:
Also can you create a `js-exec` repo for our own native execution of it. You can base it off `js-quic` or `js-mdns` for the stubs on windows.

tegefaulkes:
I am keeping it simple for now. Right now we support selecting a specific secret or a directory. When specifying a directory all the secrets within it are returned. Right now it's just using the normal file system rather than building a schema on top of it. So if you think of the env stored as a key:value pair. the the key is the file name and the contents is the value. I think this is a good compromise for now without building an actual schema structure while getting the most functionality.

so for a file tree of 

```
.
├── aws
│   ├── AWS_ACCESS_KEY_ID
│   ├── AWS_DEFAULT_REGION
│   └── AWS_SECRET_ACCESS_KEY
└── REGISTRY_AUTH_FILE

```

we end up with the following envs

```
AWS_DEFAULT_REGION
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
REGISTRY_AUTH_FILE
```

if we select just the `aws` directory we end up with just

```
AWS_DEFAULT_REGION
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
```

Supporting the directories like this allows for basic grouping.

PR #681, opened 2024-03-12
fix: changed umask to 0007

brynblack:
### Description

This PR changes the default `umask` on all files Polykey Agent creates to a value of `0007` (translates to 0770 permissions), allowing users in a given group, like `polykey` for example, to access/interact with the files and CLI commands without error.

The idea here is that the files were originally only able to be accessed by the user who created them. This has implementations when trying to "daemonise" Polykey Agent using a Docker daemon-like approach. The only way to make this idea work is to set a default of `0770` permissions on the state files, allowing users outside of the user that created the file to interact with the state directory, as long as they are in the appropriate group to interact with the files it creates.

### Tasks

- [x] 1. Change the `umask` value

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/src/bootstrap/utils.ts
+++ b/src/bootstrap/utils.ts
@@ -61,7 +61,7 @@ async function bootstrapState({
   fs?: FileSystem;
   logger?: Logger;
 }): Promise<RecoveryCode | undefined> {
-  const umask = 0o077;
+  const umask = 0o007;
   logger.info(`Setting umask to ${umask.toString(8).padStart(3, '0')}`);
   process.umask(umask);
   logger.info(`Setting node path to ${nodePath}`);


tegefaulkes:
Having a quick overview of how groups work. It seems that when files/directories are created they are assigned a default depending of the directory it's in and it's user. For example If I as user `someUser` create a file it's owned by `someUser` and `operators` group. Rather than giving all operators access I think it's better to default to only allowing the user to access the files.

That said, It's not like any admin can't just have access anyway using admin magic, but reducing the scope of access is probably best here.

But in our case, when running as a daemon, our main method of access will be via the `polykey` group. So in that use case we do want the group to have access.

But let's take a step back, what files does the `Polykey-CLI` need to access? As far as I know it's *just* the `status.json` file. All other state will only be accessed by the `polykey` agent running as the service. so technically we don't need to give group access to the state files besides the status file. And status/PID files should be read by anyone anyway?

So it seems to me we don't actually need group permissions here? But the service should still run with a group. Only the `status.json` file needs to be accessible by anyone.

Got any insights @CMCDragonkai?

tegefaulkes:
Do we still need this? I think we determined that we're going to keep state files as user permissions only. But I think we need the `status.json` file to have polykey group permissions? Or possibly read only permissions for everyone.

What are we doing with that @brynblack?

tegefaulkes:
Closing this since we determined it's not needed.

PR #687, opened 2024-03-20
Adding `NodeConnectionAddress` limit for each node

### LINKED ISSUE #537: `MDNS` integration and `NodeGraph` structure expansion

tegefaulkes:
### Description

This PR aims to solve the last task for #537 by adding a limit for the number of addresses stored for each NodeId and removing the oldest addresses when full.

### Issues Fixed

* Fixes #537 

### Tasks

- [x] 1. Add a `NodeConnectionAddressLimit` config to the `NodeGraph`
- [x] 2. Add logic for enforcing this limit on a per `NodeId` basis.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -287,6 +287,13 @@ const config = {
      * a node and resource usage per node and across the network.
      */
     nodesGraphBucketLimit: 20,
+    /**
+     * Node graph node specific NodeContactAddress limit. The node graph keeps
+     * a list of addresses for each node. This sets the limit on the number of
+     * different addresses we keep track of. If an added address exceeds this
+     * limit then the oldest addresses are removed first.
+     */
+    nodesGraphNodeContactAddressLimit: 5,
 
     /**
      * Multicast group addresses that the MDNS stack will operate on.
--- a/src/nodes/NodeGraph.ts
+++ b/src/nodes/NodeGraph.ts
@@ -428,7 +428,7 @@ class NodeGraph {
       nodeContactAddressData.connectedTime,
       tran,
     );
-    await this.purgeAddresses(nodeId, this.nodeContactAddressLimit);
+    await this.purgeAddresses(nodeId, this.nodeContactAddressLimit, tran);
   }
 
   /**
@@ -1022,15 +1022,15 @@ class NodeGraph {
     }
 
     // Start by getting all the existing addresses
-    const asd = await this.getNodeContact(nodeId, tran);
-    if (asd == null) return;
-    if (Object.keys(asd).length <= num) return;
+    const nodeContact = await this.getNodeContact(nodeId, tran);
+    if (nodeContact == null) return;
+    if (Object.keys(nodeContact).length <= num) return;
 
     const list: Array<[NodeContactAddress, number]> = [];
-    for (const nodeAddress of Object.keys(asd)) {
+    for (const nodeAddress of Object.keys(nodeContact)) {
       list.push([
         nodeAddress as NodeContactAddress,
-        asd[nodeAddress].connectedTime,
+        nodeContact[nodeAddress].connectedTime,
       ]);
     }
     list.sort(([, connectedTimeA], [, connectedTimeB]) => {
--- a/src/nodes/NodeGraph.ts
+++ b/src/nodes/NodeGraph.ts
@@ -57,6 +57,8 @@ class NodeGraph {
     keyRing,
     nodeIdBits = 256,
     nodeBucketLimit = config.defaultsSystem.nodesGraphBucketLimit,
+    nodeContactAddressLimit = config.defaultsSystem
+      .nodesGraphNodeContactAddressLimit,
     logger = new Logger(this.name),
     fresh = false,
   }: {
@@ -64,6 +66,7 @@ class NodeGraph {
     keyRing: KeyRing;
     nodeIdBits?: number;
     nodeBucketLimit?: number;
+    nodeContactAddressLimit?: number;
     logger?: Logger;
     fresh?: boolean;
   }): Promise<NodeGraph> {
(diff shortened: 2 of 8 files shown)


tegefaulkes:
I went with a simple logic where each time an addresses is added for a node, any excess is checked and removed. I opted for a shared limit here just to simplify it. But that leaves the possibility of a `hostname` being pushed out of the list. If that's a problem I'll have to explore a more complex alternative of either sticky addresses or multiple limits based on address type.

There is a *possibility* of transaction conflicts here. Where an old addresses is seen again and a new address would push the old one out. I might have to add locking to address this.

CMCDragonkai:
Can you verify your logic - transactional consistency is important here.

tegefaulkes:
Looking into it, while the `NodeGraph` doesn't do any locking directly. The `NodeManager` does lock the transaction on the bucket when adding or updating nodes. This should cover our use and protect from conflicts.

I'll add tests covering the new logic on the `NodeGraph` and `NodeManager` level.

tegefaulkes:
Added some tests and checked. There are no transaction conflicts since the locking on the bucket is serialising the whole process anyway. So that's all good.

I also added some logic to filter out any `local` only addresses when getting closest local addresses.

This should be good to merge after some review.

PR #689, opened 2024-03-22
Feature general fixes

tegefaulkes:
### Description

This PR is just a collection of general small fixes in `Polykey`.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/nodes/NodeConnectionManager.ts
+++ b/src/nodes/NodeConnectionManager.ts
@@ -835,7 +835,7 @@ class NodeConnectionManager {
       async (conn) => {
         const client = conn.getClient();
         const nodeIdSource = this.keyRing.getNodeId();
-        // Data is just `<sourceNodeId><targetNodeId>` concatenated
+        // Creating signature verifying request, data is just `<sourceNodeId><targetNodeId>` concatenated.
         const data = Buffer.concat([nodeIdSource, nodeIdTarget]);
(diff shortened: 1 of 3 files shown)


CMCDragonkai:
Usually random fixes like this I'd just push to staging. But shouldn't this all be part of issues?

tegefaulkes:
I'd agree with that but I wanted a paper trail for what I was doing. I don't think there are existing issues for any of theses. 

The fix for the intermittent test failure has an existing issue related to it. But this far from fully addresses that.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1278,9 +1287,15 @@ class NodeConnectionManager {
...
       });
-    }).finally(() => {
-      this.activeSignalFinalPs.delete(connProm);
-    });
+    })
+      // Ignore results and failures, then are expected to happen and are allowed
+      .then(
+        () => {},
+        () => {},
+      )
```
tegefaulkes: I should really only ignore connection specific failures here. I'm going to fix that quickly.

tegefaulkes: Addressed, I'm having it check if the error is connection related and throwing if not.

PR #695, opened 2024-04-08
Backgrounding of Notifications Domain

### LINKED ISSUE #703: Backgrounding of `NotificationsManager`for Fault Tolerance
### Specification

Currently the sending of notifications is entirely synchronous. The `NotificationsManager.sendNotifcation`currently blocks until the Notification has either been sent or failed to be sent. 

In order to make the `NotificationManager` more fault tolerant, the sending of notifications needs to be assigned to a `TaskManager` task instead. This task needs to reschedule itself with a delay that doubles each retry. This would require a refactor of the notifications domain to abstract over the TaskManager in queuing notifications, but it would give us an outbox/inbox system similar to emails.

#### Expanded RPC Handlers

The RPC handlers need to be expanded to add ability to both clear the notifications of the outbox and remove a pending notification from the outbox.

#### NotificationIds

NotifcationIds need to be exposed on the notifications themselves, much like how AuditEventIds are exposed on the AuditEvents. The reason for this is because the user should be able to delete Notifications / cancel a pending Notification in the outbox. This was already discussed as an improvement during https://github.com/MatrixAI/Polykey-CLI/pull/50, but there is an actual need for it now. The Encoded version of the NotificationId will be available both in the Notifications returned by `read*Notificaitons`, but not in the database itself. The reason for this is because the database already stores the ID as the key, which is then filled in as a field in the Notificaiton returned by `read*Notifications` methods.

### Tasks

1. Implement non-blocking `NotificaitonsManager.sendNotification`
2. Implement exposing NotificationIds on Notifications
3. Implement `readOutboxNotifications`, `removeOutboxNotification`, and `clearOutboxNotifications`
4. Rename `readNotifications`, `removeNotification`, and `clearNotifications` to `readInboxxNotifications`, `removeInboxNotification`, and `clearInboxNotifications` respectively

amydevs:
### Description

This PR applies backgrounding to the Notifications domain by using `TaskManager`.

The sending of notifications can be either specified as blocking, or non-blocking, and with any amount of retries.

By doing this, errors on attempts to send notifications are now caught and turned into warning logs instead for non-blocking notifications.

Methods and RPC handlers to list and cancel the pending notifications are also to be added.

### Issues Fixed

* Related https://github.com/MatrixAI/Polykey-CLI/issues/168
* Fixes #703 

### Tasks

- [x] 1. Implement non-blocking sending of notifications.
- [x] 2. Implement outbox notification listing 
- [x] 3. Implement outbox notification cancelling
- [x] 4.  Tests for outbox notifications in NotificationsManager 
- [x] 5. Tests for outbox notification RPC handlers

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
I'm thinking that if you're going to structure the DB to have outbox, will you also be creating sublevels for inbox? I think we would want to keep our "async messaging" layer to a minimal functionality, let's not recreate email inside Polykey. So from an async messaging layer perspective, the inbox is just all the messages we have received. But an outbox is all the messages we have sent out. Both have their own `IdSortable` ids, and is sorted by time.

What's interesting is that outbound messages would only have our own local `IdSortable` - sender notification ID.

Whereas inbound messages would have both the sender notification ID and our own received notification ID.

Furthermore I want us all to realise (@MatrixAI/matrixai-engineers) that we have a sync layer, and now by wrapping certain operations into the task manager, we introduce an async layer, for delay tolerance.

---

This means notifications which used to be synchronous is entirely changed to be asynchronous, which is what it is intended to be.

So therefore - this level of architectural change is a major thing. @amydevs you should be writing up in the title and spec that this is changing how the notifications domain works entirely so that notification messaging is an asynchronously operated domain.

Other domains that are also natively asynchronous are the like the discovery domain - and the way we discover our gestalt graph.

amydevs:
@CMCDragonkai 

Currently the outbox is driven by the state of the parameters of the Task that stores the Notification and the corresponding Notification ID. I did this to simplify the state, and not have to manage 2 sets of state that are not relational be about the same notification. Also, since we're not rlly implementing a full messaging system, I thought that history for sent notifications wouldn't rlly be needed, so using the TaskManager for the pending notification state would be fine. The only concern would be that starting a fresh TaskManager would explicitly empty the Tasks for outbox notifications, but that would implicitly delete all outbox notifications.

src/client/handlers/NotificationsOutboxRead.ts
```diff
@@ -0,0 +1,43 @@
...
+        tran,
+      }),
+    );
+    for (const notification of notifications) {
+      if (ctx.signal.aborted) throw ctx.signal.reason;
+      yield {
+        notification: notification,
+      };
+    }
+  }
```
tegefaulkes: Generally we want to avoid large collections like this in array form. Ideally the `readOutboxNotifications` should be a generator.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+    NOTIFICATIONS_COUNT_KEY,
+  ];
+  protected notificationsManagerOutboxNotificationsDbPath: LevelPath = [
+    ...this.notificationsManagerOutboxDbPath,
+    'notifications',
+  ];
+  protected notificationsManagerOutboxNotificationsCounterDbPath: KeyPath = [
+    ...this.notificationsManagerOutboxDbPath,
+    NOTIFICATIONS_COUNT_KEY,
+  ];
```
tegefaulkes: Normally we include a description of what the `LevelPath` looks like. You can look at the `GestaltGraph` to see how. E.G.

```ts
 /**
   * Gestalt links.
   * `GestaltGraph/links/{GestaltLinkId} -> {json(GestaltLink)}`
   */
  public readonly dbLinksPath: Readonly<LevelPath> = [
    this.constructor.name,
    'links',
  ];

  /**
   * Node information
   * `GestaltGraph/nodes/{GestaltKey} -> {json(GestaltNodeInfo)}`
   */
  public readonly dbNodesPath: Readonly<LevelPath> = [
    this.constructor.name,
    'nodes',
  ];
```

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
   /**
-   * Top level stores MESSAGE_COUNT_KEY -> number (of messages)
+   * Map for hashing NotificationIds to find TaskIds.
+   * The key must be NotificationIdEncoded, so that the string representation is hashed rather than the reference to the NotificationId object.
    */
-  protected notificationsDbPath: LevelPath = [this.constructor.name];
+  protected outboxNotificationIdEncodedToTaskIdMap: Map<
+    NotificationIdEncoded,
+    TaskId
+  >;
```
tegefaulkes: I'm not sure I like the in memory map for this. I think it's possible to use the task path to handle this. Let me look into it more.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+    ...this.notificationsManagerOutboxDbPath,
+    NOTIFICATIONS_COUNT_KEY,
+  ];
+
+  protected inboxNotificationIdGenerator: () => NotificationId;
+  protected outboxNotificationIdGenerator: () => NotificationId;
 
-  protected notificationIdGenerator: () => NotificationId;
+  public readonly sendNotificationHandlerId =
+    `${this.constructor.name}.sendNotificationHandler` as TaskHandlerId;
```
tegefaulkes: I've always defined this after the handler arrow function. Let's keep it consistent.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+      attempt: number;
+    },
+  ) => {
+    const nodeId = nodesUtils.decodeNodeId(nodeIdEncoded)!;
+    const notificationId = notificationsUtils.decodeNotificationId(
+      notificationIdEncoded,
+    )!;
+    const notificationKeyPath = [
+      ...this.notificationsManagerOutboxNotificationsDbPath,
+      idUtils.toBuffer(notificationId),
```
tegefaulkes: You don't need to use `idUtils.toBuffer()`. Any `Id` type has `toBuffer()` built in. So to keep consistent `notificationId.toBuffer()` is preferred here.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+    ];
+
+    const notificationDb = await this.db.withTransactionF(
+      async (tran) => (await tran.get<NotificationDB>(notificationKeyPath))!,
+    );
```
tegefaulkes: This is odd. If it's just a single thing in the transaction then just use `this.db.get()` instead. But I can see another transaction within a `finally` block further down. Doesn't that need to be consistent with this `get`? If so then pretty much the whole handler needs to be wrapped within a transaction.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+          )} responded with error: ${e.cause.description}`,
+        );
+        throw e;
+      } else {
+        this.logger.warn(
+          `Could not send to ${
+            notificationDb.data.type
+          } notification to ${nodesUtils.encodeNodeId(nodeId)}`,
+        );
+      }
```
tegefaulkes: `ErrorPolykeyRemote` isn't a connection related error, wraps errors thrown by the other side of the connection. A connection needs to be established first for that to happen. To check for a connection error we have a utility at `isConnectionError` `src/nodes/utils.ts:317` to check if an error is a failure to connect in some fashion.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+            {
+              nodeIdEncoded,
+              notificationIdEncoded,
+              retries,
+              attempt: attempt + 1,
+            },
+          ],
+          delay: 1000 * 2 ** attempt,
+          lazy: false,
+        });
```
tegefaulkes: I don't like the unbounded delay here. Also a 1 second retry seems to short. I don't see connection issues resolving within that time. Maybe start at 1 hour and limit it to 1 day delay? That'll cap out after 5 retries. How many attempts would we normally expect?

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+            counter != null ? counter - 1 : 0,
+          );
+        } else {
+          this.outboxNotificationIdEncodedToTaskIdMap.set(
+            notificationIdEncoded,
+            newTaskId,
+          );
+        }
+      });
+    }
```
tegefaulkes: The logic flow of this whole try-catch-finally block seems a little hard to follow. So the logic is

1. Send succeeds - Preform any needed cleanup in database and state indicating success
2. Send fails and out of attempts - Clean up state indicating no more retries.
3. Send fails and more attempts - schedule new attempt and modify any needed state.

Right now this is spread across the try-catch-finally block. which is why it's confusing. I think we can flatten this out a little in just the try-catch sections. Doing a state update before returning or throwing.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+    {
+      nodeIdEncoded,
+      notificationIdEncoded,
+      retries,
+      attempt,
+    }: {
+      nodeIdEncoded: NodeIdEncoded;
+      notificationIdEncoded: NotificationIdEncoded;
+      retries: number;
+      attempt: number;
```
tegefaulkes: Unless we really need to know the total attempts that were allowed. We only really need to count down the attempts left and track 1 number for the logic.

So rather than count up till the retry limit, start at the retry limit and count to 0.

src/notifications/NotificationsManager.ts
tegefaulkes: Yeah, you can make the path of the scheduled task be `[this.sendNotificationHandlerId, notificationIdEncoded]` and that will maintain the `task -> notification` relationship`. To find the `notificationId` of a task you just look at the path. To get the reverse you just need to iterate over the path with that `notificationId`. The iteration of the task path is annoying, yes, but also fine for this application. The we don't maintain a separate memory map for this information reducing the complexity.

You can make a protected utility function to get the `taskId` -> `notificationId` if you need it in a bunch of places.

src/notifications/NotificationsManager.ts
```diff
@@ -161,26 +353,210 @@ class NotificationsManager {
...
+    }: {
+      nodeId: NodeId;
+      data: NotificationData;
+      retries?: number;
+    },
+    tran?: DBTransaction,
+  ): Promise<{
+    notificationId: NotificationId;
+    sendProm: Promise<void>;
+  }> {
```
tegefaulkes: Probably call `sendProm` something like `sendP` or just `p`. We're avoiding using the `prom` name now.

src/notifications/NotificationsManager.ts
```diff
@@ -161,26 +353,210 @@ class NotificationsManager {
...
   public async sendNotification(
-    nodeId: NodeId,
-    data: NotificationData,
-  ): Promise<void> {
-    const notification: Notification = {
+    {
+      nodeId,
+      data,
+      retries = 64,
+    }: {
```
tegefaulkes: Avoid magic numbers, If we need a default here then take it from `this.defaultRetries` and set that in the constructor and `createX`. I'm not sure if we need to go as far as adding the default value for this to the config. so it should be fine to have the magic number specified in the `createX` configuration.

src/notifications/NotificationsManager.ts
```diff
@@ -161,26 +353,210 @@ class NotificationsManager {
...
+      this.outboxNotificationIdEncodedToTaskIdMap.set(
+        notificationIdEncoded,
+        pendingTask.id,
+      );
+    };
+    if (tran == null) {
+      await this.db.withTransactionF(transactionCallback);
+    } else {
+      await transactionCallback(tran);
+    }
```
tegefaulkes: We use recursion to ensure a transaction is created when it's not provided. Do something like this to the top of a method.

```ts
  public async setNode(
    nodeInfo: GestaltNodeInfo,
    tran?: DBTransaction,
  ): Promise<['node', NodeId]> {
    if (tran == null) {
      return this.db.withTransactionF((tran) => this.setNode(nodeInfo, tran));
    }
    //...
  }
```

src/notifications/NotificationsManager.ts
```diff
@@ -161,26 +353,210 @@ class NotificationsManager {
...
+      this.notificationsManagerOutboxNotificationsDbPath,
+      { valueAsBuffer: false },
+    );
+    for await (const [keyPath] of messageIterator) {
+      const key = keyPath[0] as Buffer;
+      const notificationId = IdInternal.fromBuffer<NotificationId>(key);
+      notificationIds.push(notificationId);
+    }
+    return notificationIds;
+  }
```
tegefaulkes: This should be a generator. Generally we want to avoid collecting generator outputs into an array unless we have to.

src/notifications/NotificationsManager.ts
```diff
@@ -161,26 +353,210 @@ class NotificationsManager {
...
+  public async readOutboxNotifications({
+    number = 'all',
+    order = 'newest',
+    tran,
+  }: {
+    unread?: boolean;
+    number?: number | 'all';
+    order?: 'newest' | 'oldest';
+    tran?: DBTransaction;
+  } = {}): Promise<Array<Notification>> {
```
tegefaulkes: Likewise, this should be a generator.

src/notifications/NotificationsManager.ts
```diff
@@ -73,37 +94,174 @@ class NotificationsManager {
...
+   * Notifications stores NotificationId -> Notification
+   */
+  protected notificationsManagerInboxNotificationsDbPath: LevelPath = [
+    ...this.notificationsManagerInboxDbPath,
+    'notifications',
   ];
+  protected notificationsManagerInboxNotificationsCounterDbPath: KeyPath = [
+    ...this.notificationsManagerInboxDbPath,
+    NOTIFICATIONS_COUNT_KEY,
+  ];
```
tegefaulkes: Is this a counter for the number of inbox messages? Don't do this because it WILL conflict without locking. Even with locking it's not a good idea. We already have a way to track the number of entries in the DB. there's a `tran.count` method that should handle this. See if you can try using that instead.

tegefaulkes (commented):
@CMCDragonkai Asked me to have a look over. Seems some things can be simplified and other things to be avoided. But otherwise looking fine.

I'll look deeper into the tests later.

tegefaulkes:
Oh, and be sure replace any implicit return types with explicit ones. The implicitness is generally something we want to avoid.

src/notifications/NotificationsManager.ts
amydevs: the inbox count already existed before i worked on the notifications, i just changed the db path

src/notifications/NotificationsManager.ts
amydevs: should i change the readNotifications fn to also be an asyncgenerator?

src/notifications/NotificationsManager.ts
amydevs: i've updated this to follow the suggestion

src/notifications/NotificationsManager.ts
amydevs: this has been done

src/notifications/NotificationsManager.ts
amydevs: this has been done

src/client/handlers/NotificationsOutboxRead.ts
amydevs: should readNotifications also be made an asyncgenerator?

src/notifications/NotificationsManager.ts
amydevs: this has been done

src/notifications/NotificationsManager.ts
amydevs: i tried to do that originally, but the attempts are needed to keep track of the exponential timeout

src/notifications/NotificationsManager.ts
amydevs: i've removed this for now

src/notifications/NotificationsManager.ts
amydevs: this has been done, i dont ever rlly need to reference the taskid the other way around so it's alright

src/notifications/NotificationsManager.ts
amydevs: i'm now using isConnectionError to check instead now

src/notifications/NotificationsManager.ts
amydevs: i've encapsulated the entirety of sendNotificationHandler with the db transaction

src/notifications/NotificationsManager.ts
amydevs: it's been made much simpler now with the removal of the NotificationID -> TaskID map

src/client/handlers/NotificationsOutboxRead.ts
tegefaulkes: Yeah, anything that can return a potentially unbounded array should be a generator. Doubly so if it gets streamed over RPC.

src/notifications/NotificationsManager.ts
tegefaulkes: You can just double the previous delay. Unless you can't access the task info inside of the handler?

src/notifications/NotificationsManager.ts
tegefaulkes: Yes.

src/notifications/NotificationsManager.ts
amydevs: this has been done

src/notifications/NotificationsManager.ts
amydevs: this has been done.

src/notifications/NotificationsManager.ts
amydevs: this is partially done, I'm still fixing the tests for this.

CMCDragonkai:
I think you should keep the retry delay simple as a 2x factor. It's the same elsewhere. Any change to this would need to be a separate issue.

Also there will need to be a corresponding PR in the CLI too. So that issue needs to be specced out.

src/client/handlers/NotificationsOutboxRead.ts
amydevs: this has been done

src/notifications/NotificationsManager.ts
amydevs: this has been done

CMCDragonkai:
> @CMCDragonkai
> 
> Currently the outbox is driven by the state of the parameters of the Task that stores the Notification and the corresponding Notification ID. I did this to simplify the state, and not have to manage 2 sets of state that are not relational be about the same notification. Also, since we're not rlly implementing a full messaging system, I thought that history for sent notifications wouldn't rlly be needed, so using the TaskManager for the pending notification state would be fine. The only concern would be that starting a fresh TaskManager would explicitly empty the Tasks for outbox notifications, but that would implicitly delete all outbox notifications.

Is this still the case? Task state shouldn't really be used for keeping track of notifications. It should keep track of tasks. Separation of concerns. The task itself can reference the notification though.

src/notifications/NotificationsManager.ts
amydevs: i have made it so that the delay starts at 1 hr and is capped at 1 day, doubling each retry. The default retries is now 8.

src/notifications/NotificationsManager.ts
amydevs: i've made it so that the delay is doubled instead

src/client/handlers/NotificationsOutboxRead.ts
```diff
@@ -0,0 +1,58 @@
...
+    ctx,
+  ): AsyncGenerator<ClientRPCResponseResult<NotificationOutboxMessage>> {
+    if (ctx.signal.aborted) throw ctx.signal.reason;
+    const { db, notificationsManager } = this.container;
+    const notifications = await db.withTransactionF(async (tran) =>
+      notificationsManager.readOutboxNotifications({
+        number: input.number,
+        order: input.order,
+        tran,
+      }),
```
tegefaulkes: Technically we'd support seeking, just as a standard way of handling streaming like this. We always have seek, order, limit.

 But would seek even be useful here? Would we want to list all notifications before or after a certain time?

tegefaulkes (commented):
Seems good.

src/client/handlers/NotificationsOutboxRead.ts
CMCDragonkai: If it can support it, it's better to be consistently supported.

amydevs: this has been done

PR #701, opened 2024-04-19
fix: fixed bugs with `getAuditEventsLongRunning` not handling cornercases properly

tegefaulkes:
### Description

There were some assumptions made about the data that led to bugs in certain circumstances. Mainly if you attempted a long stream with no existing data it would just return without waiting for future events.

### Issues Fixed

* Fixes ENG-28

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/audit/Audit.ts
+++ b/src/audit/Audit.ts
@@ -341,14 +341,15 @@ class Audit {
           order: 'asc',
         });
         for await (const auditEvent of iterator) {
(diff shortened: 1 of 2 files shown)


tegefaulkes:
There's a bug with `auditEventGet > gets connetion events` test that I need to fix still. But otherwise @amydevs can you have a look over this?

PR #709, opened 2024-05-03
Refactoring `git` domain to use async generators, remove readable-stream, simplify utilities

### LINKED ISSUE #298: Review and Refactor of `git/utils.ts` for Serving Vault Git Repositories
### Specification

The underlying workings of `git/utils.ts` is largely a mystery to our team. We want to perform a deep dive into the workings of this code, and remove/refactor any of our own implementation, potentially also replacing their usage in favour of the `isomorphic-git/internal-apis`. There seems to be some overlap with what our git utils perform, and what `iso-git` provides.

To clarify... we use git in 2 ways: as a client and as a server.

When using isogit, it only provides the primitives to use git as a client. It does not provide the primitives to use git as a server.

Why do we use git as a client and server? Well our entire secrets management system is built as a stack, from bottom to top:

1. Encrypted Key Value Database (soon to have transparent block based encryption) - js-db
2. Virtual Filesystem - js-encryptedfs
3. Virtual Git plus Filesystem - isogit (this is what constitutes a "vault")
4. Virtual Unix Operations - MatrixAI/Polykey-CLI#32 (this is what creates "vault operations")
5. Frontends - Polykey-CLI, Polykey-Desktop, Polykey-Mobile (and then presented over an RPC interface to different front-ends)

As you can see, the "Virtual Git plus Filesystem" layer makes use of isogit. A virtual git repository is created on the virtual filesystem. This is how we are able to then stack unix operations with roughly similar semantics later, in order to maintain API consistency with what developers are used to.

Now the issue is that isogit by itself can act like a git client. That means it can clone/pull repositories, it can create repositories, it can create commits, it can manage the git objects... etc.

However isogit cannot "serve" the repositories. And this is essential to be able to share our vaults.

To serve our repositories we need to know how to act like a "git server" when we use isogit or something were to use git to try and clone/pull our vaults.

So a long time ago, we developed the git server primitives and put them all into `git/utils.ts`. However this code has not been heavily tested, nor has it gone through any refactoring since then. And since then we have forgotten how it all works, and it's not in the programming style that we have developed over the last few years working on Polykey.

So the `git/utils.ts` needs to go under review, and most likely some heavy refactoring with the addition of tests, to ensure that acting like a git server makes sense here.

Now there is definitely cross over between isogit and the `git/utils.ts`. However we aren't aware of what these are. And there probably should be a bunch of "shared types" and similar constructs. We just need to get into it, and get it done.

### How this was addressed

After reviewing the code as it was, we ended up just cleaning it up by refactoring the `http` response to be streamed using an async generator. Most of the utility functions were removed in favour of using the `isomorphic-git` plumbing functions for interacting with the git data structure.

Ultimately this didn't speed things up very much, however, it did significantly reduce the complexity of the git domain. So it is much easier to follow along with what is happening. There is also a slight improvement with pulling a repo. We use the `haves` provided by the client to only grab the minimum objects required for the pull.

### Additional context

* initial discussion of git utils in sprint 11 meeting (roughly around the 1:02:00 mark): https://vimeo.com/manage/videos/650965464

### Tasks

1. Review the whole git domain
2. Refactor the git domain with the intention of simplifying the code and improving performance when cloning/pulling a vault.

### LINKED ISSUE #523: Replace `readable-stream` usage with Web Streams in the Vaults and Git Domains
### Specification

Some parts of the code are still using `readable-stream`. these need to be replaced with webstreams and the `readable-stream` dependency removed from the `package.json`.

These usages can be found at 
```

[nix-shell:~/Projects/Polykey]$ ag "from 'readable"
src/vaults/VaultManager.ts
24:import { PassThrough } from 'readable-stream';

src/git/types.ts
1:import type { PassThrough } from 'readable-stream';

src/git/utils.ts
22:import { PassThrough } from 'readable-stream';
```

### Additional context

* https://github.com/MatrixAI/js-encryptedfs/issues/75 - Git related streams may have some relationship with EFS readable stream changes, can also fix that up too (however EFS will cannot use web streams, since it has to simulate node FS streams)

### Tasks

1. ~[ ] 1. Replace any `readable-stream` usage with `stream/web`.~ - replacing with async generator
2. [x] 2. Remove `readable-stream` as a dependency from the `package.json`.

tegefaulkes:
### Description
This PR addresses refactoring the `git` domain to be more clear and more in line with our latest practices.

### Issues Fixed

* Fixes #298 
* Ref ENG-68
* Fixes #523 

### Tasks

- [x] 1. Factor out http stream construction to a new `http.ts` file.
- [x] 2. Keep git and FS interaction utility in `utils.ts`.
- [x] 3. Refactor existing code that we are keeping to be more in line with current practices.
- [x] 4. Refactor the HTTP response to be streaming focused. Using generators for most of the back end logic and convert to a webstream for the rpc stage.
- [x] 5. All types should be formalised where appropriate.
- [x] 6. Inline documentation should be updated and all relevant spec details should be included along side the code.
- [x] 7. Tests need to updated and refactored to reflect the new refactoring changes.
- ~8. Performance is an issue here, so benchmarks should be added to test for future regressions.~ - Differed to a new PR. It will test the current changes but nothing here will depend on benchmarks, so I'm just going to make it separate.
- [x] Check if #523 has been addressed fully

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/git/http.ts
+++ b/src/git/http.ts
@@ -1,0 +1,384 @@
+/**
+ * Reference discovery
+ * Notes:
+ *
+ * Server SHOULD terminate each non-flush line using LF ("\n") terminator;
+ * client MUST NOT complain if there is no terminator.
+ *
+ * The returned response is a pkt-line stream describing each ref and its current value.
+ * The stream MUST be sorted by name according to the C locale ordering.
+ *
+ * If HEAD is a valid ref, HEAD MUST appear as the first advertised ref.
+ * If HEAD is not a valid ref, HEAD MUST NOT appear in the advertisement list at all, but other refs may still appear.
+ *
+ * The stream MUST include capability declarations behind a NUL on the first ref.
+ * The peeled value of a ref (that is "ref^{}") MUST be immediately after the ref itself, if presented.
+ * A conforming server MUST peel the ref if it’s an annotated tag.
+ *
+ *   advertised-refs  =  (no-refs / list-of-refs)
+ *                       *shallow
+ *                       flush-pkt
+ *
+ *   no-refs          =  PKT-LINE(zero-id SP "capabilities^{}"
+ *                       NUL capability-list LF)
+ *
+ *   list-of-refs     =  first-ref *other-ref
+ *   first-ref        =  PKT-LINE(obj-id SP refname
+ *                       NUL capability-list LF)
+ *
+ *   other-ref        =  PKT-LINE(other-tip / other-peeled)
+ *   other-tip        =  obj-id SP refname LF
+ *   other-peeled     =  obj-id SP refname "^{}" LF
+ *
+ *   shallow          =  PKT-LINE("shallow" SP obj-id)
+ *
+ *   capability-list  =  capability *(SP capability)
+ *   capability       =  1*(LC_ALPHA / DIGIT / "-" / "_")
+ *   LC_ALPHA         =  %x61-7A
+ */
+
+import type { CapabilityList, ObjectGenerator, ObjectId } from './types';
+import type { EncryptedFS } from 'encryptedfs';
+import { Buffer } from 'buffer';
+import git from 'isomorphic-git';
+import * as gitUtils from './utils';
+import * as utils from '../utils';
+
+/**
+ * Smart ref discovery response looks like
+ *
+ * ```
+ * S: 200 OK
+ * S: Content-Type: application/x-git-upload-pack-advertisement
+ * S: Cache-Control: no-cache
+ * S:
+ * S: 001e# service=git-upload-pack\n
+ * S: 0000
+ * S: 004895dcfa3633004da0049d3d0fa03f80589cbcaf31 refs/heads/maint\0multi_ack\n
+ * S: 003fd049f6c27a2244e12041955e262a404c7faba355 refs/heads/master\n
+ * S: 003c2cb58b79488a98d2721cea644875a8dd0026b115 refs/tags/v1.0\n
+ * S: 003fa3c2e2402b99163d1d59756e5f207ae21cccba4c refs/tags/v1.0^{}\n
+ * S: 0000
+ * ```
+ *
+ * smart_reply     =  PKT-LINE("# service=$servicename" LF)
+ *                    "0000"
+ *                    *1("version 1")
+ *                    ref_list
+ *                    "0000"
+ * ref_list        =  empty_list / non_empty_list
+ * empty_list      =  PKT-LINE(zero-id SP "capabilities^{}" NUL cap-list LF)
+ * non_empty_list  =  PKT-LINE(obj-id SP name NUL cap_list LF)
+ *                    *ref_record
+ * cap-list        =  capability *(SP capability)
+ * capability      =  1*(LC_ALPHA / DIGIT / "-" / "_")
+ * LC_ALPHA        =  %x61-7A
+ * ref_record      =  any_ref / peeled_ref
+ * any_ref         =  PKT-LINE(obj-id SP name LF)
+ * peeled_ref      =  PKT-LINE(obj-id SP name LF)
+ *                    PKT-LINE(obj-id SP name "^{}" LF
+ *
+ *   NUL       =  %x00
+ *   zero-id   =  40*"0"
+ *   obj-id    =  40*(HEXDIGIT)
+ */
+
+// Total number of bytes per pack line minus the 4 size bytes and 1 channel byte
+const chunkSize = 65520 - 4 - 1;
+const headRef = 'HEAD';
+const sideBand64Capability = 'side-band-64k';
+const agentCapability = 'agent=git/isomorphic-git@1.8.1';
+const spaceString = ' ';
+const channelData = 1;
+const channelProgress = 2;
+const channelError = 3;
+
+// Initial string sent when doing a smart http discovery request
+const referenceDiscoveryHeader = Buffer.from('# service=git-upload-pack');
+// NUL       =  %x00
+const nullBuffer = Buffer.from('\0');
+// LF
+const lineFeedBuffer = Buffer.from('\n');
+// Zero-id   =  40*"0"
+const zeroIdBuffer = Buffer.from('0'.repeat(40));
+// Magic string used when no refs are provided
+const emptyListCapabilitiesBuffer = Buffer.from('capabilities^{}');
+// SP
+const spaceBuffer = Buffer.from(spaceString);
+// Flush-pkt    = "0000", used to indicate a special step or end of the stream.
+// This will not be padded with the `PKT-LINE` delimiter. In essence, it's a special delimiter
+// since a 0-len line would include the 4 bytes `0004` length delimiter which is explicitly not
+// allowed.
+const flushPacketBuffer = Buffer.from('0000');
+// Used to indicate no common commits during ref negotiation phase.
+const nakBuffer = Buffer.from('NAK\n');
+const dummyProgressBuffer = Buffer.from('progress is at 50%');
+
+/**
+ * Smart_reply     =  PKT-LINE("# service=$servicename" LF)
+ *                    "0000"
+ *                    *1("version 1")
+ *                    ref_list
+ *                    "0000"
+ */
+async function* advertiseRefGenerator(
+  fs: EncryptedFS,
+  dir: string = '.',
+  gitDir: string,
+): AsyncGenerator<Buffer, void, void> {
+  const capabilityList = [
+    sideBand64Capability,
+    await gitUtils.refCapability(fs, dir, gitDir, headRef),
+    agentCapability,
+  ];
+  const objectGenerator = gitUtils.listReferencesGenerator(fs, dir, gitDir);
+
+  yield packetLineBuffer(referenceDiscoveryHeader);
+  yield flushPacketBuffer;
+  yield* referenceList(objectGenerator, capabilityList);
+  yield flushPacketBuffer;
+}
+
+/**
+ *
+ * Ref_list        =  empty_list / non_empty_list
+ * empty_list      =  PKT-LINE(zero-id SP "capabilities^{}" NUL cap-list LF)
+ * non_empty_list  =  PKT-LINE(obj-id SP name NUL cap_list LF)
+ *                    *ref_record
+ * ref_record      =  any_ref / peeled_ref
+ * any_ref         =  PKT-LINE(obj-id SP name LF)
+ * peeled_ref      =  PKT-LINE(obj-id SP name LF)
+ *                    PKT-LINE(obj-id SP name "^{}" LF
+ * cap-list        =  capability *(SP capability)
+ */
+async function* referenceList(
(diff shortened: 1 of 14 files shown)


tegefaulkes:
The reference discovery stage has been completed. I implemented it as `AsyncGenerator`s which made it stream-able but also very clean logic wise. I opted for generators over `webstreams` for this stage since composing streams together is a lot cleaner with generators than streams.

CMCDragonkai:
Generators are a higher level construct indeed. I think streams represent a lower level construct, and also streams are actually synchronous. Whereas generators represent coroutines in the javascript runtime. So if you're dealing at the object-level then async generator abstraction is better abstraction.

CMCDragonkai:
In the future, you should write up your own branch name.

CMCDragonkai:
New types and new classes need to be documented into the spec. Like this:

#### Module

```
type X = ...
```

```
type Y = ...
```

Give a reason as why they exist now.

src/git/http.ts
```diff
@@ -0,0 +1,321 @@
...
+ * S: Content-Type: application/x-git-upload-pack-advertisement
+ * S: Cache-Control: no-cache
+ * S:
+ * S: 001e# service=git-upload-pack\n
+ * S: 0000
+ * S: 004895dcfa3633004da0049d3d0fa03f80589cbcaf31 refs/heads/maint\0multi_ack\n
+ * S: 003fd049f6c27a2244e12041955e262a404c7faba355 refs/heads/master\n
+ * S: 003c2cb58b79488a98d2721cea644875a8dd0026b115 refs/tags/v1.0\n
+ * S: 003fa3c2e2402b99163d1d59756e5f207ae21cccba4c refs/tags/v1.0^{}\n
+ * S: 0000
```
CMCDragonkai, line 66: Beware of the parser generator standard.

CMCDragonkai: There's examples of that in the `ids`. And there's an old issue describing the whole concept.

CMCDragonkai: Still to align with that standard.

That information needs to be put into the development docs.

Polykey-Docs - MatrixAI-Graph.

CMCDragonkai: The `claims` domain also uses it.

CMCDragonkai:
Expected to take the whole cycle.

CMCDragonkai:
image

This is a bug with the vault log - pull system. Needs to be fast forward merge.

tegefaulkes:
Progress update: I've converted to using generators and swapped over to using them in the code. The test I've been using as a sanity check is passing as well. I'm going to go back through the code and clean it up a little now.

tegefaulkes:
`isometric-git` actually provides a lot of the functionality we needed in the git domain. I was actually able to replace most of the `utils.ts` functions with it. I did some cursory checks and the things run just as fast, the whole git domain has been simplified immensely.

It seems odd that `isometic-git` wasn't used to handle interacting with the git structure in the first place. Was the git domain created before we started using `isometric-git`?

CMCDragonkai:
Originally isogit did not have server-side functionality. Only client side. You were able to replace it all? I thought there was git server functionality that was missing. Remember in the future we may want to expose an http API.

tegefaulkes:
It is missing the server side functionality, but it does provide a lot of plumbing functionality for interacting with the git structure that the `utils.ts` was replicating.

For example, Listing all the existing refs, resolving them to objectIds, then finding all of the dependent objectIds, and finally creating the packfile format that gets sent over can all be done with isogit. The only thing I had to do was generate the HTTP responses in the proper format.

CMCDragonkai:
Then yes it was a quick and dirty job.

CMCDragonkai:
Originally we even wanted to use libgit2 but it was easily adapted to work on a in-memory virtual filesystem. Too tied into posix fs semantics.

tegefaulkes:
Mostly done now. Just need to fix some tests in the vaults domain now. Along with doing the benchmarks.

I think the benchmarks can be moved to a 2nd stage PR. I also think the whole vaults domain needs a general review and touch up, also in a separate PR.

tegefaulkes:
Tests are fixed, I added a end-to-end integration tests of of cloning and pulling in the `http.test.ts` tests. This does a git clone/pull using just the new `http` code directly and the normal file system. So no `efs` or networking. It should be a good basis of comparison between adding `efs` and networking separately in the benchmarks. 

I'm going to look into the clone/pull bug mentioned from before and then move on the merge prep. I'll start a new PR for adding the benchmarks.

tegefaulkes:
1 test relating to discovery and 2 tests relating to MDNS are consistently failing. I'm not sure why and they're wait out of scope of this work so it shouldn't be affected.

tegefaulkes:
This is ready for review. I'm going to start the benchmarks in a new PR for now.

src/git/http.ts
CMCDragonkai: What's the status with this?

tests/git/utils.ts
```diff
@@ -1,72 +1,270 @@
...
+import type fs from 'fs';
 import path from 'path';
 import git from 'isomorphic-git';
+import fc from 'fast-check';
+import * as gitUtils from '@/git/utils';
+import * as gitHttp from '@/git/http';
+import * as utils from '@/utils';
 
+// Just to avoid confusing the type with the name
+type FsType = typeof fs;
```
CMCDragonkai: Can we use the filesystem type?

There's already a `import type fs from 'fs'` in `src/types.ts` in Polykey. Probably one in PK CLI too?

tests/git/utils.ts
```diff
@@ -1,72 +1,270 @@
...
+    }
+  | {
+      type: 'have';
+      objectId: string;
+    }
+  | {
+      type: 'SEPARATOR' | 'done' | 'none';
+    };
+
+function generateTestNegotiationLine(data: NegotiationTestData, rest: Buffer) {
```
CMCDragonkai: Can we get some docblocks for this. Also `generateTest...` sounds a bit weird?

tests/git/utils.ts
```diff
@@ -1,72 +1,270 @@
...
+      return Buffer.concat([Buffer.from('0009done\n'), rest]);
+    case 'none':
+      return Buffer.alloc(0);
+    default:
+      utils.never();
+  }
+}
+
+// Used to print out the contents of an `Buffer` iterable for testing
+async function* tapGen(
```
CMCDragonkai: Docblock format please.

CMCDragonkai: This function's name is a bit weird? Should it be expanded.

tests/git/utils.ts
```diff
@@ -1,72 +1,270 @@
...
+  gen: AsyncIterable<Buffer>,
+): AsyncGenerator<Buffer, void, void> {
+  let acc = '';
+  for await (const line of gen) {
+    acc += line.toString();
+    yield line;
+  }
+  // eslint-disable-next-line no-console
+  console.log(acc);
```
CMCDragonkai: No `console.log` allowed.

tests/git/utils.ts
```diff
@@ -1,72 +1,270 @@
...
+  listGitObjects,
+  generateTestNegotiationLine,
+  tapGen,
+  request,
+  objectIdArb,
+  capabilityArb,
+  capabilityListArb,
+  restArb,
+  wantArb,
+  haveArb,
```
CMCDragonkai: I think all these names should be expanded a bit. Like `wantArb`? That's way too simple. It should be specific to particular data structures within git.

tests/git/utils.ts
```diff
@@ -1,72 +1,270 @@
...
+  lineDataArb,
```
CMCDragonkai: This too!

CMCDragonkai: I would suggest prefixing it like `gitXYZArb`.

src/git/http.ts
```diff
@@ -0,0 +1,408 @@
...
+    yield packetLineBuffer(subBuffer, gitUtils.CHANNEL_DATA);
+  } while (packFileBuffer.byteLength > chunkSize);
+}
+
+export {
+  advertiseRefGenerator,
+  packetLineBuffer,
+  parsePackRequest,
+  generatePackRequest,
+  generatePackData,
```
CMCDragonkai, line 424: I would like docblocks here to make it clear exactly the dependency tree, or call graph of the http protocol itself.

As in what is considered the main function in relation to the cloning and pulling operations? And how does it decompose subsequently. Docblocks should use ASCII diagrams to make this clear.

CMCDragonkai:
I added some additional review comments with a first glance. Also this review comment previously should be addressed: https://github.com/MatrixAI/Polykey/pull/709#pullrequestreview-2039845891.

Otherwise is there specific things you want reviewed?

CMCDragonkai:
The final checklist should be ticked off.

tegefaulkes:
Nothing specific needs to be reviewed, just need the extra pair of eyes for the due diligence.

I'm leaving the checklist along for when the review changes have been addressed.

src/git/http.ts
tegefaulkes: I modified the parsers to be consistent with how the claims domain does things. I ended up creating an `isX()` and `assertX()` to use.

tests/git/utils.ts
tegefaulkes: It makes more sense to use `FileSystemWritable` or `EncryptedFs` as the type here. `isomorphic-git` doesn't like the `FileSystem` type for reasons I haven't dug into.

I think I'll just make sure everything that uses the `fs` works with the `EFS` and for testing we use `fs` but override the type where needed.

tegefaulkes: Switched everything relevant to `efs: EncryptedFs`. This should make the domain more consistent.

tests/git/utils.ts
tegefaulkes: Added some clarification for it.

tests/git/utils.ts
tegefaulkes: If we're that strict about the `console.log` then I'll just remove the `tapGen` function. It's only there for testing and not even used.

tests/git/utils.ts
tegefaulkes: `tapGen` has been removed.

tests/git/utils.ts
tegefaulkes: `tapGen` has been removed.

tests/git/utils.ts
tegefaulkes: I have cleaned these up. Added some comments and made the names more specific.

tests/git/utils.ts
tegefaulkes: Fixed along with above.

src/git/http.ts
tegefaulkes: Updated the relevant docblocks.

tegefaulkes:
Done and ready to merge. I'm going to rebase the related PRs before merging this.

Issue #651, opened 2023-12-13
Intermittent test failures due to Vaults and Tasks in CI


tegefaulkes:
### Specification

Currently there are intermittent failures in the CI for the `vaults` domain and `tasks` domain. These need to be investigated and fixed up.

Test failures can be found at
1. https://gitlab.com/MatrixAI/open-source/Polykey/-/jobs/5691640837
2. https://gitlab.com/MatrixAI/open-source/Polykey/-/jobs/5689124352

The failures are as follows.
<details>

Vaults domain failures

```
  ● VaultManager › with remote agents › can pull a cloned vault
    ErrorRPCHandlerFailed:
      373 |     // Rpc errors
      374 |     case 1:
    > 375 |       return new rpcErrors.ErrorRPCHandlerFailed();
          |              ^
      376 |     case 2:
      377 |       return new rpcErrors.ErrorRPCMessageLength();
      378 |     case 3:
      at constructor_.codeToReason (src/nodes/utils.ts:375:14)
      at constructor_.read (node_modules/@matrixai/quic/src/QUICStream.ts:423:31)
      at constructor_.read (node_modules/@matrixai/async-init/src/CreateDestroy.ts:203:18)
      at constructor_.processStreams (node_modules/@matrixai/quic/src/QUICConnection.ts:982:18)
      at node_modules/@matrixai/quic/src/QUICConnection.ts:812:14
      at node_modules/@matrixai/async-locks/src/Lock.ts:57:63
      at withF (node_modules/@matrixai/resources/src/utils.ts:24:18)
      at constructor_.recv (node_modules/@matrixai/quic/src/QUICConnection.ts:731:5)
      at Socket.handleSocketMessage (node_modules/@matrixai/quic/src/QUICSocket.ts:117:7)
  ● VaultManager › with remote agents › pullVault respects locking
    ErrorRPCHandlerFailed:
      373 |     // Rpc errors
      374 |     case 1:
    > 375 |       return new rpcErrors.ErrorRPCHandlerFailed();
          |              ^
      376 |     case 2:
      377 |       return new rpcErrors.ErrorRPCMessageLength();
      378 |     case 3:
      at constructor_.codeToReason (src/nodes/utils.ts:375:14)
      at constructor_.read (node_modules/@matrixai/quic/src/QUICStream.ts:423:31)
      at constructor_.read (node_modules/@matrixai/async-init/src/CreateDestroy.ts:203:18)
[...]

PR #716, opened 2024-05-09
Claim Discovery Task Atomization to Apply Task Deadline Individually Across Each Claim Request

### LINKED ISSUE #181: Discovering identities fails if the target user has multiple claims
### Describe the bug

If the user has claimed their github identity multiple times, then discovering the user fails with the following error:
`WARN:polykey.PolykeyAgent.task v0phkmdj0tlo0146ppc9ce79g0k:Failed - Reason: ErrorTaskTimeOut`

However, when I delete the second claimed gist, then the discovery command works fine once again.

### To Reproduce

1. Authenticate with Github
2. Claim identity
3. Discovery works fine
4. Claim identity again
5. Discovery fails
6. Delete gist resultant from second claiming
7. Discovery works

### Expected behavior

The discovery command should work regardless of there being multiple claims.

Should it even be possible to perform multiple claims?

* Device: Acer Spin 3 (SP314-54N)
* OS: Arch Linux (Kernel 6.8.5-arch1-1 64-bit)
* Version: 1.2.3-alpha.4

### Additional context

> I think this can be a solution
>
> image

### Notify maintainers

@tegefaulkes @amydevs

- [ ] Also notify @CryptoTotalWar once this bug is fixed so a mention can be made in the pk docs

### LINKED ISSUE #723: `Discovery` domain does not reschedule lazily executed `Task`s
### Describe the bug

The `Discovery` domain does not reschedule lazily executed `Task`s when stopping. The trace for this is as follows:

1. `Discovery.start` is called to register handlers onto `TaskManager` before it starts processing.
2. `TaskManager.startProcessing` is called to start task processing.
3. The `Discovery.discoverVertexHandler` is executed lazily.
4. `TaskManager.stopProcessing` is called in order to stop task processing.
5. `Discovery.discoverVertexHandler` catches an `ErrorTaskStop` on `ctx.signal.reason`.
6. It calls `Discovery.scheduleDiscoveryForVertex` to reschedule the task.
7. `Discovery.scheduleDiscoveryForVertex` calls `TaskManager.getTasks`, seeing that the task that is currently executing cleanup already exists.
8. The new task is not rescheduled.
9. `TaskManager.startProcessing` is called, the task to resume discovery is not executed.

### To Reproduce

As the code to reproduce this only exists within the context of my `Discovery` domain pagination changes, please refer to https://github.com/MatrixAI/Polykey/blob/167b041cabb6f0e0bb1c2bff2cdc51c2f989a7b5/tests/discovery/Discovery.test.ts#L494

### Expected behavior

The discovery task should be rescheduled so that discovery of vertices can be resumed after restarts.

### Additional context

Found whilst working on https://github.com/MatrixAI/Polykey/pull/716

### Notify maintainers
@tegefaulkes

amydevs:
### Description

The goal of this PR is to refactor the `Discovery` domain so that `discoverVertexHandler` does not timeout upon making multiple GET requests for claims on the identity provider.

This will fix the issue listed as currently, what is failing the discovery of multiple claims is purely the fact that `Discovery.verifyIdentityClaims` is not an atomic operation. The method will only return once all ClaimIds and their associated Claims have been retrieved from the identity provider.

In the case of GitHub, where `n = number of claims`, the requests made for the discovery of a given identity will be `ceiling(n / 10) + n`. The reason for the `ceiling(n / 10)` requests is that GitHub will only show 10 Gists per page, meaning that we have to paginate to get all ClaimIds.

Given that the amount of requests, and therefore amount of time taken scales linearly, that is why `discoverVertexHandler` times out when processing more than 1 Claim.

#### Task Aggregation

Upon realizing debugging many, many, many tasks might be a pain, I've realized the most painless solution.

Instead of scheduling a task for each request made, we can simply refresh the ctx to emulate this behaviour. Refreshing a ctx will not reset the timer if the timer has already hit the deadline, so it makes it very easy to emulate this behaviour. Furthermore, since the task is rescheduled when lazily cancelled by the task manager stopping, we can just paramaritise our progress before the task manager shuts down.

From here on out, please keep this in mind when I talk about creating new tasks, as I am now doing this instead.

#### Claim Pagination

Given that the time taken to paginate across a gist search to find all ClaimIds is `ceiling(n / 10)`, we could run into a timeout on this task. This would happen with the current implementation:

In order to apply our timeout deadline to each request individually, we can instead have the task that processes the first page schedule a task to process the second page once it is done:

In order to do this we need a way to paginate through each page of Claims that is:

1. the least common denominator of all APIs
2. is serializable

The way we can do this is by returning a generic `Opaque` `String` that represents the `PaginationToken` to to be handed to the next task to execute their query. This type will have no meaning other than being what should be provided to the next function to get the next page.

#### Claim Requests

Each task to process a page can represent a pipeline to get all of the contents of the claims on that particular page. The directed graph in the order of which tasks are scheduled can be represented as so:

This way, we can streamline the meaning of `Discovery.discoverVertexTimeoutTime` to be the timeout time of each request in the Discovery task pipelines. By binding the timeout time with a specific expectation, we can expect less unexpected errors such as that found in the issue. 

#### Identities `Provider` Abstract Class Interface

Because we now have to account for __pagination__  + __getting claim ids to schedule tasks for requesting each claim__, we need to have a `getClaimIds` async generator method on `Provider` interface.

We will also need to add paginated versions of the `getClaims` and `getClaimIds` methods that will also return a paginationToken. The reason for this, is so that the task to process each request triggered by the method call to a paginated `get...` method, as well as resume if the task were stopped. This will be a bit messy as we cannot follow our existing pagination conventions, considering that there different providers will use different pagination methods. GitHub will use page index, but Twitter will use tokens and timestamps.

### Issues Fixed

* Fixes https://github.com/MatrixAI/Polykey-CLI/issues/181
* Fixes https://github.com/MatrixAI/Polykey/issues/723

### Tasks

- [x] 1. Update `Identities.Provider` abstract class to include paginated API options
- [x] 2. Change `Discovery.verifyIdentityClaims` so that we can schedule a task for requesting and processing each claim separately.
- [x] 3. Implementing progress saving to the `Discovery` handler
- [x] 4. Implement flag that will prefer usage of `getClaimPages` directly in the case where data can be aggregated more efficiently

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -624,16 +624,28 @@ class Discovery {
       identityId,
       ctx,
     );
+    let lastProviderPaginationToken = await this.gestaltGraph
+      .getIdentity(providerIdentityId)
+      .then((identity) => identity?.lastProviderPaginationToken);
     // If we don't have identity info, simply skip this vertex
     if (vertexIdentityInfo == null) {
       return;
     }
     // Getting and verifying claims
-    const claims = await this.verifyIdentityClaims(providerId, identityId, ctx);
+    const {
+      identityClaims,
+      lastProviderPaginationToken: lastProviderPaginationToken_,
+    } = await this.verifyIdentityClaims(
+      providerId,
+      identityId,
+      lastProviderPaginationToken,
+      ctx,
+    );
+    lastProviderPaginationToken = lastProviderPaginationToken_;
+    const isAborted = ctx.signal.aborted;
     // Link the identity with each node from its claims on the provider
-    // Iterate over each of the claims
-    for (const [claimId, claim] of Object.entries(claims)) {
-      if (ctx.signal.aborted) throw ctx.signal.reason;
+    // Iterate over each of the claims, even if ctx has aborted
+    for (const [claimId, claim] of Object.entries(identityClaims)) {
       // Claims on an identity provider will always be node -> identity
       // So just cast payload data as such
       const linkedVertexNodeId = nodesUtils.decodeNodeId(claim.payload.iss);
@@ -644,7 +656,10 @@ class Discovery {
       };
       await this.gestaltGraph.linkNodeAndIdentity(
         linkedVertexNodeInfo,
-        vertexIdentityInfo,
+        {
+          ...vertexIdentityInfo,
+          lastProviderPaginationToken,
+        },
         {
           claim: claim,
           meta: {
@@ -669,6 +684,11 @@ class Discovery {
         );
       }
     }
+    // Throw after we have processed the node claims if the signal aborted whilst running verifyIdentityClaims
+    if (isAborted) {
+      throw ctx.signal.reason;
+    }
+    // Only setVertexProcessedTime if we have succeeded in processing all identities
     await this.gestaltGraph.setVertexProcessedTime(
       ['identity', providerIdentityId],
       Date.now(),
@@ -836,22 +856,33 @@ class Discovery {
    * Helper function to retrieve and verify the claims of an identity on a given
    * provider. Connects with each node the identity claims to be linked with,
    * and verifies the claim with the public key of the node.
+   *
+   * This method never throws if ctx has aborted, opting instead to return early
+   * with a lastProviderPaginationToken so that the caller can process the partially
+   * requested Claims as well as resume progress when calling again.
    */
   protected async verifyIdentityClaims(
     providerId: ProviderId,
     identityId: IdentityId,
+    providerPaginationToken: ProviderPaginationToken | undefined,
     ctx: ContextTimed,
-  ): Promise<Record<ProviderIdentityClaimId, SignedClaim<ClaimLinkIdentity>>> {
+  ): Promise<{
+    identityClaims: Record<
+      ProviderIdentityClaimId,
+      SignedClaim<ClaimLinkIdentity>
+    >;
+    lastProviderPaginationToken?: ProviderPaginationToken;
+  }> {
     const provider = this.identitiesManager.getProvider(providerId);
     // If we don't have this provider, no identity info to find
     if (provider == null) {
-      return {};
+      return { identityClaims: {} };
     }
     // Get our own auth identity id
     const authIdentityIds = await provider.getAuthIdentityIds();
     // If we don't have one then we can't request data so just skip
     if (authIdentityIds.length === 0 || authIdentityIds[0] == null) {
-      return {};
+      return { identityClaims: {} };
     }
     const authIdentityId = authIdentityIds[0];
     const identityClaims: Record<
@@ -859,7 +890,6 @@ class Discovery {
       SignedClaim<ClaimLinkIdentity>
     > = {};
 
-    let nextPaginationToken: ProviderPaginationToken | undefined;
     const identitySignedClaimDb = (
       identitySignedClaim: IdentitySignedClaim,
     ) => {
@@ -875,6 +905,8 @@ class Discovery {
         identityClaims[identitySignedClaim.id] = claim;
       }
     };
+    let nextPaginationToken: ProviderPaginationToken | undefined =
+      providerPaginationToken;
     while (true) {
       // Refresh before each request made with identitySignedClaimGenerator
       ctx.timer.refresh();
@@ -890,7 +922,10 @@ class Discovery {
           // 1. throw if the getClaimIdsPage takes too much time
           // 2. the rest of this loop iteration takes too much time
           if (ctx.signal.aborted) {
-            throw ctx.signal.reason;
+            return {
+              identityClaims: identityClaims,
+              lastProviderPaginationToken: nextPaginationToken,
+            };
           }
           const claimId = wrapper.claimId;
           nextPaginationToken = wrapper.nextPaginationToken;
@@ -918,7 +953,10 @@ class Discovery {
           // 1. throw if the getClaimIdsPage takes too much time
           // 2. the rest of this loop iteration takes too much time
           if (ctx.signal.aborted) {
-            throw ctx.signal.reason;
+            return {
+              identityClaims: identityClaims,
+              lastProviderPaginationToken: nextPaginationToken,
+            };
           }
           nextPaginationToken = wrapper.nextPaginationToken;
           // Claims on an identity provider will always be node -> identity
@@ -929,7 +967,9 @@ class Discovery {
         break;
       }
     }
-    return identityClaims;
+    return {
+      identityClaims,
+    };
   }
 
   /**
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -177,6 +177,7 @@ class Discovery {
           undefined,
           undefined,
           gestaltsUtils.decodeGestaltId(parent ?? undefined),
+          true,
         );
         return;
       }
@@ -503,9 +504,9 @@ class Discovery {
       );
       return;
     }
-    const linkedVertexNodeId = node1Id.equals(nodeId) ? node2Id : node1Id;
+    const linkedNodeId = node1Id.equals(nodeId) ? node2Id : node1Id;
     const linkedVertexNodeInfo: GestaltNodeInfo = {
-      nodeId: linkedVertexNodeId,
+      nodeId: linkedNodeId,
     };
     await this.gestaltGraph.linkNodeAndNode(
       {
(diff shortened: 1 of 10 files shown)


tegefaulkes:
This is really well explained, good work!

CMCDragonkai:
Nice! I want to point out that we do have some notion of "time destructuring". https://github.com/MatrixAI/MatrixAI-Graph/issues/42

https://github.com/MatrixAI/MatrixAI-Graph/issues/43 - this is the right one.

CMCDragonkai:
Regarding your ideas on parallelising the request process. Just beware of a couple of things:

1. Third party IdPs may change their UI - so it's best to rely on things that are unlikely to change alot (as in prefer things like URL patterns over like specific buttons/text located in specific HTML elements)
2. So if you want the count of gists... that may not always be the most reliable way of doing this.
3. Beware of "scraping" detectors, you really need to balance the throughput with reliability.
4. At the end of the day, I do believe in step-through timing, things need to run in the background with minimal CPU usage... etc - inch by inch is a better idea.

amydevs:
hmm alright, i think i'll just stick with the step-through timing then

amydevs:
#### Parallelism Draft

However, this is still kind of slow. We would ideally want the pages to be processed in parallel. GH search will provide the amount of total Gists returned from a specific query:

image

We can simply take this number and apply it to `ceiling(n / 10)` to figure out how many pages we need to request.

This will dramatically cut down on the amount of time required to get the ClaimIds.

Note, when there are no gist results, a different message is shown instead, so we should expect that the element to find amount of matching gists may not exist.

image

amydevs:
according to the new implementation notes in the PR description, I have solved the issue of multiple claims. There are still some optimizations that still need to be done tho (listed in tasks) :333

image

amydevs:
I can abunch of warnings about dangling tasks being put in to the queue each time i

1. run `polykey identities discover`
2. run  `polykey identities list` to see that the discovery background task has completed
3. ctrl-c in the agent window
4. the agent shuts down gracefully
5. run `polykey agent start` again

image

I'm not exactly sure what these tasks are, given that the task manager is shutting down gracefully and that the discovery background task has already long-completed before i shut down the agent. This problem only happens sometimes, so not what sure it could be... I was consistently getting it, but after my latest commits, i am not

CMCDragonkai:
I think in general tasks need to be checked for any weird errors. The latest staging version still comes up with things like this:

```
WARN:polykey.PolykeyAgent.task v0pi41vb17do01e9e5btfbaadtc:Failed - Reason: ErrorUtilsUndefinedBehaviour
```

amydevs:
# THIS IS NOT A PROBLEM, DISREGARD THIS! THE ROOT CAUSE HAS BEEN IDENTIFIED IN THE PROCEEDING COMMENT
for discovery resumption there needs to be changes in the code path. I have implemented the ProviderPaginationToken in order to support resumption of identity vertex discovery if the TaskManager is shutdown. However, I have come to realize that the discovery of a Node Vetex and all connected Vertices are all performed in a single Task. Therefore, passing a single `ProviderPaginationToken` around is not enough, as the discovery of a Node Vertex will potentially cause the discovery of multiple other Identity Vertices in one Task.

The solutions are either:
- Split the discovery of each Vertex to be a Task each. This might be also useful for streamlining the `queueDiscoveryByIdentity` and `queueDiscoveryByNode` methods, as it will allow us to avoid rescheduling discovery for an identity if it is already scheduled to be discovered by the discovery task of a neighbor Vertex.
- Pass around a map of ProviderPaginationTokens instead for the claims discovery of each Identity Vertex
- Store ProviderPaginationTokens in the Database so that the same Task can be run multiple times, but will resume from the ProviderPaginationToken in the DB.

amydevs:
The main problem right now, is that I've found discovering a Node vertex will make a Polykey agent request and iterate through the pages of claims on an identity in order to just obtain the `providerIdentityClaimId` used to index the claim on the `Provider` service. This is very inefficient, and has been a blocker as it would require us to keep the PaginationToken for Discovery resumption logic per Identity Claim on a given discovered node. Ideally we would want to be able to obtain the `providerIdentityClaimId` from the peer node directly.

There are 3 options for this I have discussed with @tegefaulkes about:

1.  Create the Claim locally, upload the Claim to the provider, get the `ProviderIdentityClaimId`, append it to the Claim locally, edit the corresponding remote claim on the provider, publish the claim. This will mean that the `ProviderIdentityClaimId` is stored identitcally both on the local claim in the SIgChain, as well on the Provider. However, this would rely on the Provider having the ability to a Claim, which may not exist on a service like Twitter. Furthermore, the `ProviderIdentityClaimId` is redundant when being stored on the Provider, as there is always some way to retrieve it from the Provider already.
2. Create the Claim locally, publish the Claim to the provider, get the `ProviderIdentityClaimId`, append it to the Claim locally. This would solve the problem on depending the Provider to have functionality to edit a Claim. However, this would mean that the contents of the Claim deviate from the local copy in the SigChain
3. Create the Claim locally, publish the Claim to the provider, get the `ProviderIdentityClaimId`, store it locally in the database. When a peer node calls our `nodesClaimsGet` RPC handler, we respond not only with the `ClaimIdEncoded` and `SignedTokenEncoded`, but also the `ProviderIdentityClaimId`. However, the issue with this, is that this information is not apart of the provenance history in the sigchain.

Edit:

It has been deemed that solution 2 is the probably the way to go.

amydevs:
@tegefaulkes roger wanted to let you know that the PR can be handed off to you. The last part is just fixing #723 and then it should be ready to merge

tegefaulkes:
Sure thing, I'll get to it after addressing #720 and #651

tegefaulkes:
Alright, I've fixed the main problem that was left. The main solution was to add a flag to `scheduleDiscoveryForVertex` called `ignoreActive` to ignore any active tasks when scheduling a new one. We only use this when rescheduling the task inside the task handler.

Besides that I had to apply some small fixes here and there.

This should be ready for review now.

src/discovery/Discovery.ts
```diff
@@ -165,10 +167,10 @@ class Discovery {
...
       );
     } catch (e) {
-      if (
-        e instanceof tasksErrors.ErrorTaskStop ||
-        e === discoveryStoppingTaskReason
-      ) {
+      // We do not check for ErrorTaskStop because the
+      // below code relies on TaskManager being running
+      // to reschedule the vertex discovery task.
+      if (e === discoveryStoppingTaskReason) {
```
tegefaulkes: Better to explain what and why we do this here rather than what we're not doing.

src/identities/Provider.ts
tegefaulkes, line 1: I understand why we have two ways of getting the claims for `GitHub`, but is this generic enough to apply to all providers? Couldn't this be purely internal to the specific provider implementation?

What's your reason for expanding the provider abstract definitions to include this functionality?

tests/discovery/Discovery.test.ts
```diff
@@ -33,8 +33,8 @@ import * as keysUtils from '@/keys/utils';
 import * as gestaltsUtils from '@/gestalts/utils';
 import * as testNodesUtils from '../nodes/utils';
 import TestProvider from '../identities/TestProvider';
-import { encodeProviderIdentityId } from '../../src/ids';
 import 'ix/add/asynciterable-operators/toarray';
+import { encodeProviderIdentityId } from '../../src/ids';
```
tegefaulkes: Odd import path

src/identities/types.ts
```diff
@@ -6,6 +6,7 @@ import type {
 } from '../ids/types';
 import type { SignedClaim } from '../claims/types';
 import type { ClaimLinkIdentity } from '../claims/payloads';
+import type { Opaque } from 'encryptedfs';
```
tegefaulkes: Weird place to import the Opaque type.

src/discovery/Discovery.ts
tegefaulkes: Now that I think about it, Are we actually rescheduling the task if it timed out? What if it failed in some way, say we failed to fetch the data. Do we reschedule then? or leave it to rediscovery?

We may need to think about this a little bit.

amydevs: The original implementation schedules rediscovery when Discovery is stopping. It is possible to have this include rescheduling for other errors but we'll have to put more thought into it later (need retry intervals, max retries, etc.). This is out of scope with this PR I would say.

src/identities/Provider.ts
amydevs: This should be generic enough to apply to all providers. There are default implementations for `getClaimsPage` and `getClaims` that depend on `getClaimIdsPage` and `getClaim`. The reason for preferring `getClaimsPage`/`getClaims` is so that providers, especially those that have a graphql endpoint, can have significant speedups with iterating over claims on a given identity by being able to gather both pages of claimIds and related claims within a single request.

amydevs:
> Alright, I've fixed the main problem that was left. The main solution was to add a flag to `scheduleDiscoveryForVertex` called `ignoreActive` to ignore any active tasks when scheduling a new one. We only use this when rescheduling the task inside the task handler.
> 
> Besides that I had to apply some small fixes here and there.
> 
> This should be ready for review now.

everything seems good, merging now

PR #719, opened 2024-05-14
Adding benchmarks for git domain

### LINKED ISSUE #717: Adding benchmarks for `git` domain to `polykey`
### Specification

After refactoring the git domain in #709 and simpifying the git code significantly. I noticed that the performance when cloning wasn't that much improved. This was purely a vibes based observation so we need to add benchmarks to for the git domain to determine performance changes later down the line. This is very important so we can compare improvements when updating the `efs` and `js-db` repos to be more performant.

Benchmarks will be added for each `git/http.ts` and `git/utils.ts` function to get an idea of which parts do the most processing. We will also add a full integration benchmark for doing an clone and pull example using `isomorphic-git`.

Each benchmark will compare the following where apropriate.
1. Running using `fs`.
2. Running using `efs`.
3. Running over the `RPC`.
4. Running over `RPC` AND `efs`.

We may not need to cover every function or example but being able to compare baseline `fs` with `efs` and `rpc` examples would be good for determining where the slowdowns are. I'm already pretty sure it's the `db` and `efs` being the core source of the slowdown.

### Additional context

* Related: #298 
* Related: #709 

### Tasks

1. Create benchmarks for each relevant `git` domain function and generator.
2. Ensure we test each case of using the `fs`, `efs` `rpc` and the combo of `efs`+`rpc`.

tegefaulkes:
### Description

This PR adds benchmarks for the git domain. Mostly to have a good point of reference when making optimisation upgrades to the `DB` and `EFS`.

NOTE: this is targeting another feature branch while it is worked on. This is blocked by the git refactoring and requires that to be merged before switching the target branch to staging and merging.

### Issues Fixed

* Fixes #717 
* Ref ENG-314

### Tasks

- [ ] 1. Add appropriate benchmarks for git domain.

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/benches/index.ts
+++ b/benches/index.ts
@@ -12,7 +12,10 @@ async function main(): Promise<void> {
   for await (const suitePath of fsWalk(suitesPath)) {
     // Skip over non-ts and non-js files
     const ext = path.extname(suitePath);
-    if (ext !== '.ts' && ext !== '.js') {
+    if (
+      (ext !== '.ts' && ext !== '.js') ||
+      path.basename(suitePath) !== 'git_utils_fs.ts'
+    ) {
       continue;
     }
     const suite: () => Promise<Summary> = (await import(suitePath)).default;
--- a/benches/index.ts
+++ b/benches/index.ts
@@ -14,7 +14,7 @@ async function main(): Promise<void> {
     const ext = path.extname(suitePath);
     if (
       (ext !== '.ts' && ext !== '.js') ||
-      path.basename(suitePath) !== 'git_utils_fs.ts'
+      path.basename(suitePath) !== 'gitClone.ts'
     ) {
       continue;
     }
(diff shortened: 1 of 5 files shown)


tegefaulkes:
Switched base over staging now that git refactoring has been merged.

CMCDragonkai:
I feel like the suites need to be named better.

tegefaulkes:
I think we can merge what we have so far and revisit the issue again later. This will need a quick review and I can work on merging it soon.

CMCDragonkai:
The benchmarks right now look quite messy.

CMCDragonkai:
But before further benchmarking, the infrastructure should be upgraded with a bot so we can compare statistics and monitor for performance regressions @brynblack.

tegefaulkes:
This has been orphaned for quite a while now. I'm going to close it.

Issue #461, opened 2022-09-29
Allow `NodeManager` to start lazily without network entry procedure


tegefaulkes:
### Specification

When starting the `PolykeyAgent` the `NodeManager` takes two steps to enter the network. When `NodeManager` is started it calls `this.setupRefreshBucketTasks()` and shortly after starting the `PolykeyAgent calls `this.nodeManager.syncNodeGraph`. The fact that these two steps happen by default when starting a `PolykeyAgent` can interfere with testing and adds a bunch of processing that doesn't need to be done.

We need to update this behaviour as specified in this comment https://github.com/MatrixAI/Polykey/pull/451#issuecomment-1261697179.

### Additional context

- Related #451 
- https://github.com/MatrixAI/Polykey/pull/451#issuecomment-1261697179

### Tasks

1. Add a `lazy` parameter to `NodeManager.start` that when true, skips the creation of `refreshBucket` tasks with `setupRefreshBucketTasks` and skips running `syncNodeGraph`.
2. Add a method to `NodeManager` called `startSync` that combines the `setupRefreshBucketTasks` and `syncNodeGraph` operations
3. rename `syncNodeGraph` to `syncSeedNodes`
4. Have the `PolykeyAgent` creation and start optionally skip the `startSync` process based on a parameter.
5. Update any `PolykeyAgent` usage in testing to skip the sync process where needed.
6. Remove any mocking of `refreshBucket` to stub it out and replace it with usage of 5.

CMCDragonkai:
Would this be something we should do as part of 6th testnet deployment?

tegefaulkes:
Not strictly needed but is nice to have for testing and should be reasonably quick to implement.
I'll add it in.

tegefaulkes:
I think this is the case now. You can start the `NodeManager` and then need to call `syncNodeGraph`. I'd like to extend this to all background functionality in `NodeGraph`. So we can disable all background operations for tests. They tend to add a bunch of noise when testing and can interfere in some cases.

I'd also like to add a option `PolykeyAgent` to disable background tasks such as these.

CMCDragonkai:
I suggest we use the stealth mode #199 feature to side-step this problem. Essentially it would temporarily (togglable) disable all background network tasks (that includes social discovery, node discovery, mdns... etc) across the board, then we don't need to add special configuration just to disable certain things for testing.

CMCDragonkai:
[...]

PR #720, opened 2024-05-14
General vaults review and fixes

### LINKED ISSUE #260: A single digit modification is not detected by isomorphic-git's `status` call
### Describe the bug

When making a change in a file using `await vault.commit(...)` and changing only a single number on that file will sometimes cause isomorphic-git to not recognise that the file has been modified and therefore if no other changes were made a commit will not be made even though a file was changed.

### To Reproduce

1. Add a file using `await vault.commit`
```
      const secretVer1 = {
        name: secretName,
        content: 'Secret-1-content-ver1',
      };
```
2. Modify the file using `await vault.commit`
```
      const secretVer2 = {
        name: secretName,
        content: 'Secret-1-content-ver2',
      };
```
3. Check the log using `await vault.log` to see if the most recent modification shows up.

NOTE: This won't happen ALL the time. It seems to be most common when running large groups of tests together?

### Expected behavior

We expect any file change, even a single digit change to be detected by isomorphic-git and therefore create a new commit when `vault.commit` is used.

### Additional context

- Most likely related to this [issue](https://github.com/isomorphic-git/isomorphic-git/issues/1245) on `isomorphic-git` 
- This issue has given a possible workaround by using `resetIndex` on each file in the vault
- Discussion about the problem is [here](https://gitlab.com/MatrixAI/Engineering/Polykey/js-polykey/-/merge_requests/205#note_711060987)
- This [issue](https://github.com/MatrixAI/js-encryptedfs/issues/55) created for EFS has the potential to resolve this issue

### Resolved

This was addressed by doing the following.

1. All contents of the vault are added using `git.add` before processing the `statusMatrix`. Making sure everything is staged beforehand essentially solves the issue. 
2. Simplified the logic for handling the `statusMatrix` to a much simpler `switch` statement.

### LINKED ISSUE #714: Inconsistent vault log when pulling secrets from a modified vault (received vault log is different) - should use fast forward merge
### Describe the bug

When I pull the changes from a modified vault on my local system, the log on both machines were different, while they should be identical.

```
cmcdragonkai ➜ matrix-framework-13-ryzen-7040  ➜ ~/Projects/Polykey-CLI
 $ ./dist/polykey.js vaults log myvault
commitId     6641db528e0e79412898425a3ea84e349aa5c82b
committer    vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp    Fri May 03 2024 19:53:49 GMT+1000 (Australian Eastern Standard Time)
message      "abc modified\n"
commitId     4e648a16325ebedd0847473ec04f02a0891b1be1
committer    vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp    Fri May 03 2024 19:45:41 GMT+1000 (Australian Eastern Standard Time)
message      "abc modified\n"
commitId     c987ff12719344b962e9586aa25b7a28bb328bc8
committer    vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp    Fri May 03 2024 19:45:13 GMT+1000 (Australian Eastern Standard Time)
message      "abc added\n"
commitId     88e97c3937ac2ee86595918d5ef136b9b92d0da0
committer    vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp    Fri May 03 2024 19:44:02 GMT+1000 (Australian Eastern Standard Time)
message      "Initial Commit\n"
```

```
aryanj@supercomputer ❯ ./pk vaults log myvault                                                       
commitId        aa08262fd12acc954561093cd1d4784cf772d588
committer       vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp       Fri May 03 2024 19:54:14 GMT+1000 (Australian Eastern Standard Time)
message         "Merge branch \'HEAD\' of http://\n"
commitId        6641db528e0e79412898425a3ea84e349aa5c82b
committer       vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp       Fri May 03 2024 19:53:49 GMT+1000 (Australian Eastern Standard Time)
message         "abc modified\n"
commitId        dd0d131b9997c74cabfe7202e68d3044fe59a5de
committer       vet1dhoduhkmn4749r8usiopivklr8i4qoh9kjnhtgeg5rie4nvh0
timestamp       Fri May 03 2024 19:48:42 GMT+1000 (Australian Eastern Standard Time)
message         "Initial Commit\n"
commitId        4e648a16325ebedd0847473ec04f02a0891b1be1
committer       vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp       Fri May 03 2024 19:45:41 GMT+1000 (Australian Eastern Standard Time)
message         "abc modified\n"
commitId        c987ff12719344b962e9586aa25b7a28bb328bc8
committer       vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp       Fri May 03 2024 19:45:13 GMT+1000 (Australian Eastern Standard Time)
message         "abc added\n"
commitId        88e97c3937ac2ee86595918d5ef136b9b92d0da0
committer       vcj3ufun2506jn44q7rr9bc1hv4c6eebicusfj1bl9sr27ookknlg
timestamp       Fri May 03 2024 19:44:02 GMT+1000 (Australian Eastern Standard Time)
message         "Initial Commit\n"
```

### To Reproduce

1. Add a secret to a vault
2. Share the vault
3. Modify the secret in the vault
4. Pull changes for the log
5. Observe different logs

### Expected behavior

The logs should be identical for any and all instances of the vault, irrelevant of the vault pulling new changes or not.

* Device: Acer Spin 3 (SP314-54N)
* OS: Arch Linux (Kernel 6.8.5-arch1-1 64-bit)
* Version: 1.2.3-alpha.4

### Additional context

> image
>
> This is a bug with the vault log - pull system. Needs to be fast forward merge.
> \_Originally posted by @CMCDragonkai in [https://github.com/MatrixAI/Polykey/issues/709#issuecomment-2095014382\_](https://github.com/MatrixAI/Polykey/issues/709#issuecomment-2095014382\_)

### Resolved

This was adressed by doing the following.

1. When calling `git.clone` and `git.pull` I made sure to target the `canonicalBranch` ref name. This really was all that was needed to make sure that the history was correct.
2. I also enabled the `fastForward` and `fastForwardOnly` flags for `git.pull` to ensure only linear history was pulled from the remote.

### LINKED ISSUE #718: Minor review and refactor of vaults domain for 'polykey'
### Specification

As part of refactoring the `git` domain in #298 we need to do a small review of the vaults domain for any fixes and improvements. This is a general refactor and review. We're looking for any places that can be tidied up if needed. Some tests need to be re-enabled as well.

### Additional context

* Related: #298
* Related #709
* Related #717
* Related #714
* Related #523

### Tasks

1. General review of the vaults domain.
2. Have a closer look at the vaults committing secrets and git garbage collection code.
3. Determine and apply any optimisations that can be made.
4. Review vaults tests and clean them up if needed.
5. Re-enable any tests that have been disabled.
6. Possibly split up some test files so the vaults domain runs quicker in CI. vault ops is currently very slow.

tegefaulkes:
### Description

This PR handles a general review of the vaults domain. General optimisations if they can be applied. Minor fixes for vault related issues and some touch ups to the vaults tests.

NOTE: this branches off from the git refactoring PR and is targeting that for now. But once the git refactoring is merged then this needs to rebase and target staging.

### Issues Fixed

* Ref ENG-316
* Fixes #718 
* Fixes #260 
* Fixes #714 

### Tasks

- [x] 1. Review vaults domain, clean up and refactor anything as needed
- [x] 2. Address #260 
- [x] 3. Address #714 
- [x] 4. Have a closer look at the vaults committing secrets and git garbage collection code.
- [x] 5. Determine and apply any optimisations that can be made.
- [x] 6. Review vaults tests and clean them up if needed.
- [x] 7. Re-enable any vaults tests that have been disabled.
- [x] 8. Possibly split up some test files so the vaults domain runs quicker in CI. vault ops is currently very slow.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/git/http.ts
+++ b/src/git/http.ts
@@ -400,6 +400,8 @@ async function* generatePackData({
   chunkSize?: number;
 }): AsyncGenerator<Buffer, void, void> {
   let packFile: PackObjectsResult;
+  // In case of errors we don't want to throw them. This will result in the error being thrown into `isometric-git`
+  // when it consumes the response. It handles this by logging out the error which we don't want to happen.
   try {
     packFile = await git.packObjects({
       fs: efs,
@@ -407,12 +409,18 @@ async function* generatePackData({
       gitdir: gitDir,
       oids: objectIds,
     });
-  } catch (e) {
+  } catch {
     // Return without sending any data
     return;
   }
-  if (packFile.packfile == null) utils.never('failed to create packFile data');
-  let packFileBuffer = Buffer.from(packFile.packfile.buffer);
+  // Pack file will only be undefined if it was written to disk instead
+  if (packFile.packfile == null) return;
+  // Convert to a buffer without copy, so we can process it
+  let packFileBuffer = Buffer.from(
+    packFile.packfile.buffer,
+    0,
+    packFile.packfile.byteLength,
+  );
 
   // Streaming the packFile as chunks of the length specified by the `chunkSize`.
   // Each line is formatted as a `PKT-LINE`
--- a/src/git/http.ts
+++ b/src/git/http.ts
@@ -5,6 +5,7 @@ import type {
   ObjectIdList,
 } from './types';
 import type { EncryptedFS } from 'encryptedfs';
+import type { PackObjectsResult } from 'isomorphic-git';
 import { Buffer } from 'buffer';
 import git from 'isomorphic-git';
 import * as gitUtils from './utils';
@@ -398,12 +399,18 @@ async function* generatePackData({
   objectIds: Array<ObjectId>;
   chunkSize?: number;
 }): AsyncGenerator<Buffer, void, void> {
-  const packFile = await git.packObjects({
-    fs: efs,
-    dir,
-    gitdir: gitDir,
-    oids: objectIds,
-  });
+  let packFile: PackObjectsResult;
+  try {
+    packFile = await git.packObjects({
+      fs: efs,
+      dir,
+      gitdir: gitDir,
+      oids: objectIds,
+    });
+  } catch (e) {
+    // Return without sending any data
+    return;
+  }
   if (packFile.packfile == null) utils.never('failed to create packFile data');
   let packFileBuffer = Buffer.from(packFile.packfile.buffer);
 
--- a/src/vaults/VaultInternal.ts
+++ b/src/vaults/VaultInternal.ts
@@ -925,78 +925,71 @@ class VaultInternal {
     const message: string[] = [];
     // Get the status of each file in the working directory
     // https://isomorphic-git.org/docs/en/statusMatrix
+    await git.add({
+      fs: this.efs,
+      dir: this.vaultDataDir,
+      gitdir: this.vaultGitDir,
+      filepath: '.',
+    });
     const statusMatrix = await git.statusMatrix({
       fs: this.efs,
       dir: this.vaultDataDir,
       gitdir: this.vaultGitDir,
     });
-    for (let [
+    for (const [
       filePath,
       HEADStatus,
       workingDirStatus,
       stageStatus,
     ] of statusMatrix) {
-      // Reset the index of files that are marked as 'unmodified'
-      // The working directory, HEAD and staging area are all the same
-      // https://github.com/MatrixAI/js-polykey/issues/260
-      if (HEADStatus === workingDirStatus && workingDirStatus === stageStatus) {
-        await git.resetIndex({
-          fs: this.efs,
-          dir: this.vaultDataDir,
-          gitdir: this.vaultGitDir,
-          filepath: filePath,
-        });
-        // Check if the file is still 'unmodified' and leave
-        // it out of the commit if it is
-        [filePath, HEADStatus, workingDirStatus, stageStatus] = (
-          await git.statusMatrix({
-            fs: this.efs,
-            dir: this.vaultDataDir,
-            gitdir: this.vaultGitDir,
-            filepaths: [filePath],
-          })
-        ).pop()!;
-        if (
-          HEADStatus === workingDirStatus &&
-          workingDirStatus === stageStatus
-        ) {
-          continue;
-        }
-      }
-      // We want files in the working directory that are both different
-      // from the head commit and the staged changes
-      // If working directory and stage status are not equal then filepath has un-staged
-      // changes in the working directory relative to both the HEAD and staging
-      // area that need to be added
-      // https://isomorphic-git.org/docs/en/statusMatrix
-      if (workingDirStatus !== stageStatus) {
-        let status: 'added' | 'modified' | 'deleted';
-        // If the working directory status is 0 then the file has
-        // been deleted
-        if (workingDirStatus === 0) {
-          status = 'deleted';
+      /*
+        Type StatusRow     = [Filename, HeadStatus, WorkdirStatus, StageStatus]
+        The HeadStatus status is either absent (0) or present (1).
+        The WorkdirStatus status is either absent (0), identical to HEAD (1), or different from HEAD (2).
+        The StageStatus status is either absent (0), identical to HEAD (1), identical to WORKDIR (2), or different from WORKDIR (3).
+
+        ```js
+        // example StatusMatrix
+        [
+          ["a.txt", 0, 2, 0], // new, untracked
+          ["b.txt", 0, 2, 2], // added, staged
+          ["c.txt", 0, 2, 3], // added, staged, with unstaged changes
+          ["d.txt", 1, 1, 1], // unmodified
+          ["e.txt", 1, 2, 1], // modified, unstaged
+          ["f.txt", 1, 2, 2], // modified, staged
+          ["g.txt", 1, 2, 3], // modified, staged, with unstaged changes
+          ["h.txt", 1, 0, 1], // deleted, unstaged
+          ["i.txt", 1, 0, 0], // deleted, staged
+        ]
+        ```
+       */
+      const status = `${HEADStatus}${workingDirStatus}${stageStatus}`;
+      switch (status) {
+        case '022': // Added, staged
+          message.push(`${filePath} added`);
+          break;
+        case '111': // Unmodified
+          break;
+        case '122': // Modified, staged
+          message.push(`${filePath} modified`);
+          break;
+        case '101': // Deleted, unStaged
+          // need to stage the deletion with remove
           await git.remove({
             fs: this.efs,
             dir: this.vaultDataDir,
             gitdir: this.vaultGitDir,
             filepath: filePath,
           });
-        } else {
-          await git.add({
-            fs: this.efs,
-            dir: this.vaultDataDir,
-            gitdir: this.vaultGitDir,
-            filepath: filePath,
-          });
-          // Check whether the file already exists inside the HEAD
-          // commit and if it does then it is unmodified
-          if (HEADStatus === 1) {
-            status = 'modified';
-          } else {
-            status = 'added';
-          }
-        }
-        message.push(`${filePath} ${status}`);
+        // Fall through
+        case '100': // Deleted, staged
+          message.push(`${filePath} deleted`);
+          break;
+        default:
+          // We don't handle untracked and partially staged files since we add all files to staging before processing
+          utils.never(
+            `Status ${status} is unhandled because it was unexpected state`,
+          );
       }
     }
     // Skip commit if no changes were made
--- a/src/vaults/VaultInternal.ts
+++ b/src/vaults/VaultInternal.ts
@@ -33,6 +33,7 @@ import * as vaultsEvents from './events';
 import { tagLast } from './types';
 import * as ids from '../ids';
 import * as nodesUtils from '../nodes/utils';
+import * as gitUtils from '../git/utils';
 import * as utils from '../utils';
 
 type RemoteInfo = {
@@ -754,7 +755,7 @@ class VaultInternal {
         // This ensures that any uncommitted state is dropped
         await this.cleanWorkingDirectory();
         // Do global GC operation
-        await this.garbageCollectGitObjects();
+        await this.garbageCollectGitObjectsGlobal();
 
         // Setting dirty back to false
         await tran.put(
@@ -998,27 +999,7 @@ class VaultInternal {
       });
       // We clean old history if a commit was made on previous version
       if (headRef !== masterRef) {
-        // Delete old commits following chain from masterRef -> headRef
-        let currentRef = masterRef;
-        while (currentRef !== headRef) {
-          // Read commit info
-          const commit = await git.readCommit({
-            fs: this.efs,
-            dir: this.vaultDataDir,
-            gitdir: this.vaultGitDir,
-            oid: currentRef,
-          });
-          // Delete commit
-          await vaultsUtils.deleteObject(
-            this.efs,
-            this.vaultGitDir,
-            commit.oid,
-          );
-          // Getting new ref
-          const nextRef = commit.commit.parent.pop();
-          if (nextRef == null) break;
-          currentRef = nextRef;
-        }
+        await this.garbageCollectGitObjectsLocal(masterRef, headRef);
       }
     }
   }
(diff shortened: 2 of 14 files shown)


CMCDragonkai:
What's the order of merging in relation to #709?

tegefaulkes:
#709 merges first and then this one.

tegefaulkes:
Switched base over staging now that git refactoring has been merged.

tegefaulkes:
In the `pullVault` function in `VaultInternal`. There is an `let error` variable that is never set or modified but it affects some logic. I need to dig into this since it can't be intended as it is.

tegefaulkes:
One of the vaults test for pulling was failing randomly. With some testing I found that it's random network traffic with the mainnet.

image


I'll need to make sure any tests using a node connection don't actually try to join the mainnet when testing. In this case we have a `NodeConnectionManager` for testing but give it an empty manifest. On the same note, it's very odd that the test is failing for failing to handle incoming RPC requests. This shouldn't crash us so maybe I need to look deeper into that.

So yeah, there are two problems here. 
1. I have a minimal `NodeConnectionManager` needed to start forward connections but it's receiving reverse connections from the mainnet. Likely from trying to join it itself. We shouldn't be starting or receiving external connections unless the test explicitly wants to do this. 
2. There is an `RPC` error for a missing handler propagating outwards causing the test to fail. By design this should only bubble through the stream back through the `RPC`. This isn't a critical error that should ever leave the `RPC`. Something is wrong here.

tegefaulkes:
I may need to add #461 to this to address the intermittent failure mentioned above. Though at the same time, I feel like this PR is getting too large in scope? I'll think about it.

tegefaulkes:
> One of the vaults test for pulling was failing randomly. With some testing I found that it's random network traffic with the mainnet.
> 
> image
> 
> I'll need to make sure any tests using a node connection don't actually try to join the mainnet when testing. In this case we have a `NodeConnectionManager` for testing but give it an empty manifest. On the same note, it's very odd that the test is failing for failing to handle incoming RPC requests. This shouldn't crash us so maybe I need to look deeper into that.
> 
> So yeah, there are two problems here.
> 
> 1. I have a minimal `NodeConnectionManager` needed to start forward connections but it's receiving reverse connections from the mainnet. Likely from trying to join it itself. We shouldn't be starting or receiving external connections unless the test explicitly wants to do this.
> 2. There is an `RPC` error for a missing handler propagating outwards causing the test to fail. By design this should only bubble through the stream back through the `RPC`. This isn't a critical error that should ever leave the `RPC`. Something is wrong here.

I made a new issue #722 And it will be addressed later, outside this PR.

tegefaulkes:
`vaultOps.addSecretDirectory` is a weird case. All other vaultOps operate purely within the EFS with an secret contents passed directly into it.  but `addSecretDirectory` seems to access the `fs` directly reading from it to copy a directory into a vault. 

I don't like this. It makes it the only secret command that functions against the node's filesystem and not the `Polykey-CLI`'s file system. We need to review this and work out what we want to do with it.

tegefaulkes:
So `addSecretDirectory` needs to be refactored in how it works. currently it accesses the fs to read the directory when copying it to the vault. However, as mentioned in all other vaultOps that write data, the data is passed into the function. From the `CLI` perspective, the files are sent over the RPC when adding to the vault.

The problem here is `addSecretDirectory` is the odd duck out. The data isn't sent over the RPC, it's read directly from the `fs`. So we need to refactor it. There are two options

1. We refactor it so the directory and all it's contents are sent over the RPC to the `addSecretDirectory` command. A bit of a hassle but do-able. We'd have to encode directory and file contents in a binary stream. I'd rather handle this in a separate PR.
2. We remove the `addSecretDirectory` and it's `CLI` command. Seems like a QOL function.

tegefaulkes:
I'm working on #714 right now. It's a pretty simple fix. However I found that we don't test the case where you pull a vault that deviated from the original history.

I modified a test to check this and `isomorphic-git` doesn't actually support handling that.

```
Merges with conflicts are not supported yet.
ErrorVaultsMergeConflict: Merges with conflicts are not supported yet.
    at constructor_.pullVault (/home/brian/workspace/Polykey/src/vaults/VaultInternal.ts:651:15)
    at /home/brian/workspace/Polykey/src/vaults/VaultManager.ts:789:9
    at withF (/home/brian/workspace/Polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)
    at constructor_.pullVault (/home/brian/workspace/Polykey/src/vaults/VaultManager.ts:784:5)
    at withF (/home/brian/workspace/Polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)
    at Object.<anonymous> (/home/brian/workspace/Polykey/tests/vaults/VaultManager.test.ts:1138:9)
```

So I'm not actually sure it's possible to handle this case. We'd have to clone the vault from scratch to handle it really. Do we just throw an error urging the user to clone the vault fresh due to the deviating history?

CMCDragonkai:
When you a pull a vault into a deviated history. This is what supposed to happen:

1. You are not allowed to do it.

At a higher level, the user upon editing a vault that is "pulled", should be doing a COW operation, the vault must be duplicated and separated from the old vault. The old vault maintains its ability to be pulled.

This means you cannot have a deviated history at all. It's not allowed by the constraints of the system.

tegefaulkes:
I'm aware that we restrict editing of a vault that has been cloned. But the remote vault we clone from can do a branching commit and completely change it's history. Do we apply the same restraint in this case? Where we disallow pulling a vault who's history no longer matches our cloned vault's history?

Different sides of the same coin really. Just want to be clear since the example you gave was the opposite side to the example I described.

CMCDragonkai:
> I'm aware that we restrict editing of a vault that has been cloned. But the remote vault we clone from can do a branching commit and completely change it's history. Do we apply the same restraint in this case? Where we disallow pulling a vault who's history no longer matches our cloned vault's history?
> 
> Different sides of the same coin really. Just want to be clear since the example you gave was the opposite side to the example I described.

It must be disallowed on both sides.

Linear history must be maintained. If the other side corrupted the vault, you just decline the ability to pull. You indicate why.

tegefaulkes:
I still need to do one last once-over, but it's ready for review. Doesn't need a deep review. Just so you are aware of changes and the more eyes on it the better. @CMCDragonkai @amydevs 

I'll run through the checklist after review. No point doing most of the checklist if there are changes that need to be made still.

src/vaults/VaultInternal.ts
```diff
@@ -953,12 +953,12 @@ class VaultInternal {
...
+          ["c.txt", 0, 2, 3], // added, staged, with unStaged changes
           ["d.txt", 1, 1, 1], // unmodified
-          ["e.txt", 1, 2, 1], // modified, unstaged
+          ["e.txt", 1, 2, 1], // modified, unStaged
           ["f.txt", 1, 2, 2], // modified, staged
-          ["g.txt", 1, 2, 3], // modified, staged, with unstaged changes
-          ["h.txt", 1, 0, 1], // deleted, unstaged
+          ["g.txt", 1, 2, 3], // modified, staged, with unStaged changes
+          ["h.txt", 1, 0, 1], // deleted, unStaged
           ["i.txt", 1, 0, 0], // deleted, staged
```
CMCDragonkai: You might want to align your case sensitivity here.

Also should `Type StatusRow` be an actual type somewhere in `types.ts`?

src/vaults/VaultManager.ts
```diff
@@ -806,7 +806,8 @@ class VaultManager {
     tran?: DBTransaction,
   ): AsyncGenerator<Buffer, void, void> {
     if (tran == null) {
-      const handleInfoRequest = (tran) => this.handleInfoRequest(vaultId, tran);
+      const handleInfoRequest = (tran: DBTransaction) =>
+        this.handleInfoRequest(vaultId, tran);
       return yield* this.db.withTransactionG(async function* (tran) {
         return yield* handleInfoRequest(tran);
       });
```
CMCDragonkai, line 813: This seems rather pointless, it's used once. Maybe just inline it?

CMCDragonkai: Also later it says `const efs = this.efs;` also seems pointless, make the code simpler... no need for internal assignment.

src/vaults/utils.ts
```diff
@@ -55,8 +55,8 @@ async function* readDirRecursively(
   fs,
   dir = '.',
 ): AsyncGenerator<string, void, void> {
-  const dirents = await fs.promises.readdir(dir);
-  for (const dirent of dirents) {
+  const dirEntities = await fs.promises.readdir(dir);
+  for (const dirent of dirEntities) {
```
CMCDragonkai: I believe this is `dirEntries` and `dirEntry`.

tests/vaults/VaultOps.test.ts
```diff
@@ -476,7 +476,7 @@ describe('VaultOps', () => {
       expect(secretList).toHaveLength(0);
     });
   });
-  // Not sure if hidden file names are a special case but I'm keeping the tests just in case
+  // Not sure if hidden file names are a special case, but I'm keeping the tests just in case
   test('adding hidden files and directories', async () => {
```
CMCDragonkai: Let's actually test this, and remove the comment.

CMCDragonkai:
How was #260 and #714 fixed? Can you describe the solution in the spec?

CMCDragonkai:
> > I'm aware that we restrict editing of a vault that has been cloned. But the remote vault we clone from can do a branching commit and completely change it's history. Do we apply the same restraint in this case? Where we disallow pulling a vault who's history no longer matches our cloned vault's history?
> > Different sides of the same coin really. Just want to be clear since the example you gave was the opposite side to the example I described.
> 
> It must be disallowed on both sides.
> 
> Linear history must be maintained. If the other side corrupted the vault, you just decline the ability to pull. You indicate why.

And do you have a corresponding exception for these cases now?

src/vaults/VaultManager.ts
tegefaulkes: Generic generators don't handle the `this` context very well compared to a normal arrow function. Anything that makes use of `this` in a generator needs to be put into a variable or an arrow function outside the generator for it to work properly. We do it in a bunch of places.

src/vaults/VaultInternal.ts
tegefaulkes: The type is defined by `isomorphic-git` and we don't use it in any form besides consuming it in this one part of the code. It is only named in this comment block to make it clear what we're getting and how to interpret it.

src/vaults/utils.ts
tegefaulkes: Fixed.

tests/vaults/VaultOps.test.ts
tegefaulkes: Comment has been removed.

tegefaulkes:
> > > I'm aware that we restrict editing of a vault that has been cloned. But the remote vault we clone from can do a branching commit and completely change it's history. Do we apply the same restraint in this case? Where we disallow pulling a vault who's history no longer matches our cloned vault's history?
> > > Different sides of the same coin really. Just want to be clear since the example you gave was the opposite side to the example I described.
> > 
> > 
> > It must be disallowed on both sides.
> > Linear history must be maintained. If the other side corrupted the vault, you just decline the ability to pull. You indicate why.
> 
> And do you have a corresponding exception for these cases now?

Yes, there was an existing error for this called `ErrorVaultsMergeConflict`

src/vaults/VaultManager.ts
CMCDragonkai: Don't do the binding pattern instead?

src/vaults/VaultInternal.ts
CMCDragonkai: No I mean `unStaged` vs `unstaged`.

CMCDragonkai:
Does `ErrorVaultsMergeConflict` work both ways? It generically means both conflict during a pull and any attempt to conflict it locally? Or are you just ensuring that one cannot corrupt one's own vault automatically?

Are you preventing the node from mutating its own (pulled/cloned) vault, and breaking history for a vault that has been pulled or cloned...? Or maybe the error only shows up on the pulling/cloning side?

CMCDragonkai:
You should update the spec of this PR describing any protocol changes - and how was the solution implemented in a human readable manner. Including in reference to the above comments.

src/vaults/VaultManager.ts
tegefaulkes: As a personal rule I'll only use `.bind` if I absolutely have to. In this case I think it's cleaner to still treat it as a closure but to assign what we need to a variable just before it. it *looks* unnecessary, but its still very clear what's happening.

tegefaulkes:
> Does `ErrorVaultsMergeConflict` work both ways? It generically means both conflict during a pull and any attempt to conflict it locally? Or are you just ensuring that one cannot corrupt one's own vault automatically?
> 
> Are you preventing the node from mutating its own (pulled/cloned) vault, and breaking history for a vault that has been pulled or cloned...? Or maybe the error only shows up on the pulling/cloning side?

Yes, weather the local history diverged or the remove diverged, it looks like a disconnect in the history that requires a merge. So the same error in both cases.

We can't modify local history anyway since any attempt to do so will result in a `ErrorVaultRemoteDefined: Vault is a clone of a remote vault and can not be mutated` error. So cloned vaults are immutable.

src/vaults/VaultManager.ts
CMCDragonkai: I think `const efs = this.efs;` is however useless though right?

tegefaulkes:
Ready to merge. I'll write up the changes in a moment.

src/git/http.ts
```diff
@@ -398,12 +399,18 @@ async function* generatePackData({
...
+      dir,
+      gitdir: gitDir,
+      oids: objectIds,
+    });
+  } catch (e) {
+    // Return without sending any data
+    return;
+  }
   if (packFile.packfile == null) utils.never('failed to create packFile data');
   let packFileBuffer = Buffer.from(packFile.packfile.buffer);
```
CMCDragonkai: The `never` should be reserved for programming errors. If it can happen, it should be an exception.

tegefaulkes: I mis-remembered, the `never` here is just to enforce the fact that `packfile` is defined. By our usage it should always be defined. So the `never` is proper here, just the message is off point.

```ts
    /**
     * - The packfile contents. Not present if `write` parameter was true, in which case the packfile was written straight to disk.
     */
    packfile?: Uint8Array;
};
```

CMCDragonkai:
Ok message should be updated.

20 May 2024 14:36:00 Brian Botha ***@***.***>:

src/git/http.ts
tegefaulkes: I made some comments in the code to clarify things, Also removed the never for a `return` to align with the above catch block.

PR #725, opened 2024-05-23
Fixing intermittent failures due to `js-quic` stream errors

### LINKED ISSUE #722: Fix random test failures caused by `mainnet` connections
### Specification

While working on #720 I was re-enabling disabled tests. One of them `can pull a cloned vault` at `tests/vaults/VaultManager.test.ts:932` was failing intermittently. After some digging I found that it was caused by external connections and RPC requests being made to the test `NodeConnectionManager` being used within the test. 

image

There are two problems with this. 
1. `NodeConnectionManagers` within tests should not be starting or receiving any connections during the test. This is just to maintain test isolation since test failures should ideally be directly related to what the test is checking. This would depend on #461 being completed and then updating all relevant tests.
2. The error causing the test to fail is an RPC error for missing a handler. This shouldn't be causing the test to fail since this error should only be thrown back through the RPC call being made. I'll have to find out why it's causing the test to fail since it's very likely a bug. Unless there's some weird interaction where our test `NodeConnectionManager` is attempting the `RPC` call to a `RPCServer` that doesn't have the handler? Needs more digging.

### Additional context

- Related: #720 
- Related by #461 

### Tasks

1. Ensure all tests `PolykeyAgent`s and `NodeConnectionManager`s don't attempt to interact with the wider mainnet while testing by starting with network entry disabled.
3. Dig into why the `RPC` error is causing the test to fail and fix any bug that may be revealed.

tegefaulkes:
### Description

This PR addressed #722 

The main problem was during certain circumstances an error that should've been thrown through a stream was escaping the call stack and crashing the program. This was due to improper usage of throw within the stream handlers. This has been resolved within `js-quic`.

Note that the branch name doesn't describe the problem scope very well. It was created while I was still exploring what the actual issue was. While mainnet connections in tests could still happen, they shouldn't be causing this issue regardless. No amount of interacting with random network traffic should crash the program.

#461  was removed from this PR, it turns out to not actually be a requirement.

### Issues Fixed

* Fixes #722 
* Ref ENG-317

### Tasks

- [X] 1. find the problem causing intermittent errors.
- [X] 2. Apply the fix to `js-quic` and update the dependency version pulling in the fix
- [X] 3. Applied any other minor fixes

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@matrixai/id": "^3.3.6",
         "@matrixai/logger": "^3.1.2",
         "@matrixai/mdns": "^1.2.6",
-        "@matrixai/quic": "^1.2.6",
+        "@matrixai/quic": "^1.2.7",
         "@matrixai/resources": "^1.1.5",
         "@matrixai/rpc": "^0.5.1",
         "@matrixai/timer": "^1.1.3",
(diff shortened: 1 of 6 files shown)


CMCDragonkai:
What are the implications of doing this:

* https://github.com/MatrixAI/js-quic/commit/0c9bd4fd393108720029ca43194ec84631d3e9f8
* https://github.com/MatrixAI/js-quic/commit/1f5de39a5f40fcb0e721a769ede17be9e07f699f

Given that `throw reason;` and related were supposed to be "proper".

tegefaulkes:
Everything functions the same way, we just remove the chance of that `throw` inside the stream handler bubbling up and crashing. It's not a good idea to throw the errors inside the stream like that, I think I pointed it out at the time as well. So I've removed the `throw` in the cases where the errors are only expected to be thrown through the stream.

CMCDragonkai:
If it still happens on CI, are you sure the fixes actually fix this? If not, perhaps it's better to leave it as is, and find the actual deterministic root cause.

PR #735, opened 2024-06-06
Handle failure to resolve seednodes

### LINKED ISSUE #729: DNS failure reporting and fallback
### Specification

As mentioned in this issue [https://github.com/MatrixAI/Polykey-CLI/issues/104](https://github.com/MatrixAI/Polykey-CLI/issues/104) It's possible for the OS DNS configuration to cause DNS resolution for our seed nodes to fail. In this case we need some amount of error reporting indicating that we failed to resolve the seed nodes.

It is also mentioned that we should resort to in app DNS query as a fallback for theses cases.

### Additional context

* Related [https://github.com/MatrixAI/Polykey-CLI/issues/104](https://github.com/MatrixAI/Polykey-CLI/issues/104)

### Tasks

1. Add error reporting for failing to resolve seed nodes using DNS.
2. ~We should use a fall-back in app DNS resolution if the normal method fails.~ - Extracted out to https://github.com/MatrixAI/Polykey-CLI/issues/202

tegefaulkes:
### Description

This PR adds handling for failing to resolve seed nodes with DNS resolution.

### Issues Fixed

* Fixes #729 

### Tasks

- [x] 1. Add error reporting for failing to resolve seed nodes using DNS.
- ~2. We should use a fall-back in app DNS resolution if the normal method fails.~ - Pulled into a new issue at https://github.com/MatrixAI/Polykey-CLI/issues/202

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/nodes/NodeManager.ts
+++ b/src/nodes/NodeManager.ts
@@ -227,6 +227,11 @@ class NodeManager {
         const resolvedHosts = await networkUtils.resolveHostnames([
           nodeAddress,
         ]);
+        if (resolvedHosts.length === 0) {
+          throw new nodesErrors.ErrorNodeManagerResolveNodeFailed(
+            `Failed to resolve '${nodeAddress[0]}'`,
+          );
+        }
         const nodeId = nodesUtils.decodeNodeId(nodeIdEncoded);
         if (nodeId == null) utils.never();
         return this.nodeConnectionManager.createConnectionMultiple(
@@ -240,7 +245,17 @@ class NodeManager {
       (r) => r.status === 'fulfilled',
     ).length;
     if (successfulConnections === 0) {
-      throw new nodesErrors.ErrorNodeManagerSyncNodeGraphFailed();
+      const failedConnectionErrors = connectionResults
+        .filter((r) => r.status === 'rejected')
+        .map((v) => {
+          if (v.status === 'rejected') return v.reason;
+        });
+      throw new nodesErrors.ErrorNodeManagerSyncNodeGraphFailed(
+        `Failed to establish any connections with the following errors '[${failedConnectionErrors}]'`,
+        {
+          cause: new AggregateError(failedConnectionErrors),
+        },
+      );
     }
     if (ctx.signal.aborted) return;
 
--- a/src/nodes/errors.ts
+++ b/src/nodes/errors.ts
@@ -25,6 +25,11 @@ class ErrorNodeManagerFindNodeFailed<T> extends ErrorNodeManager<T> {
   exitCode = sysexits.TEMPFAIL;
 }
 
+class ErrorNodeManagerResolveNodeFailed<T> extends ErrorNodeManager<T> {
+  static description = 'Failed to resolve node address using DNS';
+  exitCode = sysexits.TEMPFAIL;
+}
+
 class ErrorNodeManagerSyncNodeGraphFailed<T> extends ErrorNodeManager<T> {
   static description = 'Failed to enter the network with syncNodeGraph';
   exitCode = sysexits.TEMPFAIL;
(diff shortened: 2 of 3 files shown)


tegefaulkes:
Digging deeper into the current implementation.

It looks like that if a DNS lookup fails it will just be treated as a connection failure for that seed node. so if `syncNodeGraph` fails to resolve any seed nodes then it just throws the `ErrorNodeManagerSyncNodeGraphFailed`.

So far I've updated this to include more details about why `syncNodeGraph` failed by including details for each connection attempt. This looks like the following.

```
    ErrorNodeManagerSyncNodeGraphFailed: 
        at constructor_.syncNodeGraphHandler (/home/brian/workspace/Polykey/src/nodes/NodeManager.ts:249:13)
        at constructor_.syncNodeGraph (/home/brian/workspace/Polykey/src/nodes/NodeManager.ts:2050:7) {
      data: {},
      cause: AggregateError: 
          at constructor_.syncNodeGraphHandler (/home/brian/workspace/Polykey/src/nodes/NodeManager.ts:252:18)
          at constructor_.syncNodeGraph (/home/brian/workspace/Polykey/src/nodes/NodeManager.ts:2050:7) {
        [errors]: [
          [ErrorNodeManagerResolveNodeFailed],
          [ErrorNodeConnectionTimeout]
        ]
      },
      timestamp: 2024-06-06T05:13:02.342Z,
      exitCode: 75
    }
```

So this should make it much more clear what is happening. That said, when running `syncNodeGraph` with `blocking: false` it runs in the background as a task in the `TaskManager` and any error will only be reported as a logger warning as the following.

```
WARN:NodeManager test.TaskManager.task v0pj18su3rdo018f2m5d4nkf5dc:Failed - Reason: ErrorNodeManagerSyncNodeGraphFailed
```

So even with the new details within the error, the details won't be included in the logger line. But I could include a short description within the the error's message.

```
WARN:NodeManager test.TaskManager.task v0pj191kmmho0149ek5ibna8b0s:Failed - Reason: ErrorNodeManagerSyncNodeGraphFailed: some details
```

tegefaulkes:
The logger message is now the following which provides much more information.

```
WARN:NodeManager test.TaskManager.task v0pj199sa7lo016obtc1jotm8rc:Failed - Reason: ErrorNodeManagerSyncNodeGraphFailed: Failed to establish any connections with the following errors '[ErrorNodeManagerResolveNodeFailed: Failed to resolve the following address 'some.random.host,55555',ErrorNodeConnectionTimeout]'
```

tegefaulkes:
Digging into how the DNS resolver is implemented. It seems that it's done in a way that will always do a network DNS query. So it's not using the OS DNS table for the query at all. If it was implemented with `resolver.lookup()` then it would be depending on the OS DNS. 

So technically its already using an in app dns resolution in this case? I'm not sure what we could use for fallback here besides falling back to the OS level DNS? I did check and the DNS servers list is defaulted to the OS settings. In my testing it was `100.100.100.100` which I think is the `Tailscale` magic DNS? So we can also try adding fallback DNS servers. Possibly our own DNS server for the seed nodes?

So I think task 2 is not really needed? Any input @CMCDragonkai

CMCDragonkai:
There's a reason to use the app DNS. Also it uses the os provided DNS servers. Users can override the DNS servers used - look up the options. And also finally if a DNS failure has occurred, it needs to be reported separately from just connection failure.

tegefaulkes:
I'm not sure that answers my question. I've already fixed it so we have better error reporting for *why* the seed node connections fail. Including the logger warning including these details.

I'm just unsure what to do about the fallback aspect of the task. Initially I took that to mean we'd use a app implementation of DNS that skips over the OS implementaiton. But that's *already* how it works. So how do we do a fallback then? 

1. Do I just leave it as is? 
2. Do I add a fallback default DNS server to the server list? 
3. Do I add the OS level lookup to the resolver logic?
4. All of the above?

PR #736, opened 2024-06-07
Adding social and permission level discovery

### LINKED ISSUE #702: Follow permission and social links during discovery
### Specification

(copied from https://github.com/MatrixAI/Polykey/issues/692#issuecomment-2050944955)

Here is a general diagram of what a social network would look like.

Untitled-2024-01-23-1145 excalidraw

This is the kind of network we need to preform discovery on. The network is esentially made up of a graph, containing verticies made up of identities and polykey nodes, and edges forming links between them.

There are 3 tiers of edges. 

1. Cryptolinks, These are the most concrete form of a link. You can think of a gestalt as a fully formed distinct sub-graph made up of JUST cryptolinks. Cryptolinks are depicted as black arrow edges above. The circles grouping them are the gestalts.
2. Trust and permission links, depicted as the blue arrows above. These are the main relationships between nodes. There are gestalt level permissions such as trusting that gestalt. And node-node level permissions such as sharing a vault. These edges form a relationship between gestalts such that we want to know more about them since we're directly interacting with them.
3. There are weak relationships between identities. Depending on the kind of identity they could be friends, followers, part of the same group, whatever. It just implies a social relation between two identities. These exist outside of the Polykey ecosystem and don't really affect the interaction within Polykey. But it's useful to know about for inviting friends into the polykey ecosystem, for finding friends already using polykey. 

Currently Polykey discovery only operates on the first tier of edges. So only whole gestalts are discovered and the user needs to manually trigger discovery on each gestalt to discover them.

To address task 4 in the above issue description, `Quality of life and streamline features such as automatic friend discovery for identities` we need to make some upgrades to the discovery system. We need the ability to do the following

1. Follow permission links between gestalts to discover them in the background. We only really need to follow our own permissions.
2. Allow the ability to trust or set permissions between gestalts or nodes without having to discover them first.
3. Trusting or sharing should trigger background discovery.
4. Starting up Polykey should trigger initial discovery on our own node moving outwards.
5. We should check social level (tier 3) edges to enable the following.
   1. Compile a list of friends/followers to invite to use Polykey.
   2. Find friend/followers that already use Polykey.
6. We'd need to have a priority system for processing tier 2-3 edges. Social edges alone could crowd out all other forms of discovery and grind useful discovery to a halt.

The first 2 tiers of links can be fully explored. But to limit the exploration space during discovery we only want to explore first order social links. That is, only the people we follow and their tier 1 links. That makes the full exploration space to be our gestalt, gestalts of anyone with permissions and the full gestalts of people we follow.

### Additional context

Related: https://github.com/MatrixAI/Polykey/issues/692#issuecomment-2050944955

### Tasks

1. Identities have an optional implementation for listing all friends or social links. THis would be an generator that yields each link with associated metadata such as PK claims, usernames, possible emails etc etc.
2. Need an generator for listing out all trust links and an generator for all permission links
3. Implement the ability to trust and share without discovering a node or identity first.
4. Discovery should be seeded by known claims, social links, trust and permissions. It should be seeded when an agent starts.
5. We need to categorise each type of link into a priority level and preferentially process them. The tasks domain allows priority but this is a global level.

tegefaulkes:
### Description

This PR contains work for adding social and permission level discovery to the discovery domain.

### Issues Fixed

* Related #702 

### Tasks

- [ ] 1. Implement the ability to trust and give permissions to nodes without discovering them first.
- [ ] 2. Implement some way to list all social links for a provider. these need to include metadata such as existing claims and some useful information such as name, email and the like.
- [ ] 3. Implement an iterator for listing trust links
- [ ] 4. Implement an iterator for listing permission links. If permissions require trust then it might not be needed.
- [ ] 5. Discovery should trigger on start up and use existing know vertices as the starting point.
- [ ] 6. Implement priority queue when processing these social and permission links.

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -42,8 +42,6 @@ import { never } from '../utils';
 import Token from '../tokens/Token';
 import { decodeClaimId } from '../ids';
 
-// TODO: add social and permission level discovery
-
 /**
  * This is the reason used to cancel duplicate tasks for vertices
  */
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -42,6 +42,8 @@ import { never } from '../utils';
 import Token from '../tokens/Token';
 import { decodeClaimId } from '../ids';
 
+// TODO: add social and permission level discovery
+
 /**
  * This is the reason used to cancel duplicate tasks for vertices
  */
--- a/src/gestalts/GestaltGraph.ts
+++ b/src/gestalts/GestaltGraph.ts
@@ -1104,7 +1104,12 @@ class GestaltGraph {
     switch (type) {
       case 'node': {
         if ((await tran.get([...this.dbNodesPath, gestaltKey], true)) == null) {
-          throw new gestaltsErrors.ErrorGestaltsGraphNodeIdMissing();
+          await this.setNode(
+            {
+              nodeId: id,
+            },
+            tran,
+          );
         }
         await this.acl.setNodeAction(id, action, tran);
         return;
@@ -1498,6 +1503,7 @@ class GestaltGraph {
     return gestalt;
   }
 
+  // FIXME: this returns only 1 node that is linked to an identity, It should really return all of them. Maybe convert to a generator?
   protected async getIdentityLinkedNodeId(
     providerIdentityId: ProviderIdentityId,
     tran: DBTransaction,
--- a/tests/client/handlers/gestalts.test.ts
+++ b/tests/client/handlers/gestalts.test.ts
@@ -1948,6 +1948,17 @@ describe('gestaltsGestaltTrustByNode', () => {
       notify: null,
     });
   });
+  test('trusts a node (not in gestalt graph)', async () => {
+    const request = {
+      nodeIdEncoded: nodeIdEncodedRemote,
+    };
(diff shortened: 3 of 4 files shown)


tegefaulkes:
I can easily change the check in `setGestaltAction` so that we don't need a pre-discovered node to set a permission for it. The act of setting the action will add the node to the gestalt graph.

However, an identity can't have an action associated with it in the `ACL`. The permission is applied to a node it's linked to. So to be able to trust a gestalt from an identity, the identity will need to be discovered first. OR we can expand the `ACL` to support identities as well.

It shouldn't be too hard to add identities to the `ACL`, they would only exist to placeholder permissions to be linked with the gestalt when links are found.

CMCDragonkai:
Permissions only apply to nodes for a reason. Identities only exist in the context of gestalt graph. You're supposed to go from identity to node, only then would it even matter. Everytime you discover an identity if you don't have a corresponding node, there's no point in adding a permission to it... unless we expect to add permissions to identities even when they have no node associated. That could be an interesting away of doing "optimistic sharing".

tegefaulkes:
I'm mothballing this PR for now since our focus has shifted. It isn't currently worked on.

PR #741, opened 2024-06-13
Backwards Compatibility for `Discovery.processClaimLinkIdentity`

### LINKED ISSUE #740: Backwards Compatibility of Discovery Mechanism
### Specification

In order to adhere to a federated system, we need to adhere to the following principles when dealing with non-breaking changes:

1. version schema is a single number that changes if it is truly a compat break
2. the whole discovery system must be forwards compatible
3. the whole discovery system must be backwards compatible

What happened here, is that it was a field addition, but the logic that would process a claim without the field addition was removed.

For forwards compatibility, this would still allow an old node would be able to discover a new node.

But for backwards compatibility, this prevented the ability, the removed logic would no longer be able to discover an old node.

### Additional context

https://github.com/MatrixAI/Polykey/issues/737

### Tasks

1. Return old discovery logic

amydevs:
### Description

This PR re-adds backwards compatibility to the `Discovery` system to deal with claims that do not include the `providerIdentityClaimId`.

### Issues Fixed

* Fixes https://github.com/MatrixAI/Polykey/issues/740

### Tasks

- [x] 1. Add backwards compatibility logic for claims w/o `providerIdentityClaimId``

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -570,9 +570,32 @@ class Discovery {
       );
       return;
     }
-    // Need to get the corresponding claim for this
-    const providerIdentityClaimId = signedClaim.payload
+    // Need to get the corresponding claim on provider for this
+    let providerIdentityClaimId = signedClaim.payload
       .providerIdentityClaimId as ProviderIdentityClaimId | null;
+    // If we cannot get the corresponding claim directly, we will attempt to get it via iterating over the provider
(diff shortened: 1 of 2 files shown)


amydevs:
image

I managed to discover the highlighted node using a claim that was created using polykey-cli `v0.4.0`

PR #772, opened 2024-07-12
Add lazy WebSocketConnection creation to PolykeyClient

### LINKED ISSUE #771: Lazy Creation of WebSocketClient within PolykeyClient
### Specification

After experimenting with PolykeyClient in Polykey Network Status. I've realized that the error handling is really tedious.

The main issues happen when a connection fails, the error you will receive from the PolykeyClient will be inconsistent depending on the point at which the code throws during the the stopping process.

This is because the Stopping signal propagation of WebSocketClient to the PolykeyClient is entirely asynchronous through EventListeners. This makes for a huge pain when attempting connection retries like so:

https://github.com/MatrixAI/Polykey-Network-Status/blob/staging/src/plugins/polykeyClientPlugin.ts

This initially gave me the impression that error handling with RPCClient + WebSocketClient was really messy.

But after playing around with it in React, I managed to come up with a small React context provider component that handles it all really elegantly: https://github.com/MatrixAI/Polykey-Enterprise/blob/feature-ws/src/app/contexts/RPCProvider.tsx

The gist is that instead of the lifecycle of PolykeyClient being flakily tied to WebSocketClient, the PolykeyClient will instead always be started, creating WebSocketConnections lazily when an RPCClient call is being attempted.

This has several benefits:

- Errors from PolykeyClient RPC calls are more predictable, as making an RPC call can no longer error from either
  1. The connection error from the RPC call
  2. The connection error from the WebSocketClient
  3. The WebSocketClientStopped error
- Reconnection + reinitate RPC call logic can be entirely encapsulated within a single loop of the RPC call, rather than 2 independent loops. This is because WebSocketConnection errors will now always occur on the RPC call, rather than possibly receiving three different errors in this circumstance.

#### Implementation

A `lazy` param will be simply added to the PolykeyClient constructor. There will be no API changes.

There WILL however be behaviour changes that need to be documented:

- PolykeyClient will never stop unless `.stop` is called.
- RPC calls will throw WebSocketConnection errors if the lazily created WebSocketConnection is unsuccessful in connecting

### Tasks

1. Add `lazy` option to `PolykeyClient`

amydevs:
### Description
Adds lazy WebSocketConnection creation to PolykeyClient

### Issues Fixed

* Fixes #771 

### Tasks

- [x] 1. Add lazy parameter to PolykeyClient

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/src/PolykeyClient.ts
+++ b/src/PolykeyClient.ts
@@ -80,6 +80,7 @@ class PolykeyClient {
       port: number;
       options?: DeepPartial<PolykeyClientOptions>;
       fresh?: boolean;
+      lazy?: boolean;
       fs?: FileSystem;
       logger?: Logger;
     },
@@ -99,6 +100,7 @@ class PolykeyClient {
       // Options
       options = {},
       fresh = false,
+      lazy = false,
       // Optional dependencies
       fs = require('fs'),
       logger = new Logger(this.name),
@@ -108,6 +110,7 @@ class PolykeyClient {
       port: number;
       options?: DeepPartial<PolykeyClientOptions>;
       fresh?: boolean;
+      lazy?: boolean;
       fs?: FileSystem;
       logger?: Logger;
     },
@@ -153,6 +156,7 @@ class PolykeyClient {
           rpcParserBufferSize: optionsDefaulted.rpcParserBufferSize,
         },
         fresh,
+        lazy,
       },
       ctx,
     );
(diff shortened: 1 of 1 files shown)


tegefaulkes:
Looks fine, two things stand out though.

1. Before it was expected that parameters like ports and addresses would be define while the client is running. This isn't true anymore so it may brake some things making that assumptions. Usually testing if that?
2. You don't have any tests for the new behaviour. You'll need a few to demostrate the lazy feature along with making sure it works as expected and for catching future regressions.

amydevs:
Need to be redone in the future, more considerations must be made as the API will actually change if the PolykeyClient is lazy. We should probably just make 2 separate classes instead for lazy + non-lazy respectively.

PR #775, opened 2024-07-18
Network Segregation

### LINKED ISSUE #770: EPIC: Network Segregation
### Specification

In order to keep nodes from networks connecting to other networks that they are not supposed to, we need to implement some mechanism during the connection establishment process that will reject connections if say, a mainnet node tries to connect to the testnet network.

After talking with @tegefaulkes, we have come up with 2 solutions:

#### Using JWTs/Tokens with Verification Metadata

In order to show proof of access into a network, nodes should need to show some token to the seednodes, and peer nodes of that network in order to verify they are apart of that network.

These JWT tokens should contain information about:

From the point of view of proving that a node has access to a network we need the following.

1. Some way to show proof of access. Ideally this will be a token, dumb (JWT) or otherwise (biscuits and capabilities) proving the following facts. This will provide verifiable proof of access without needing to ask anyone.<br>a. The subject of the token, Who has access. This is the public key of the node asking for access<br>b. The issuer of the token, Who gave access. This is the public key central issuer of the network.<br>c. The network they have access to. This can be the public key of the network issuer. Really just the same as the issuer<br>d. The signature of the above facts, signed by the issuers private key.
2. An access list. Essentially a list of all the networks the current node allows connections to. In it's barest form its just a list of public keys of the networks. Can include extra metadata since a public key without context can be hard to audit. This will be checked against when verifying the network tokens.
3. A mechanism for providing the token to prove access after connection establishment. To do this, we will need to rework add middleware to our node-to-node rpc calls to limit calls until a node is authenticated. Similarly, the NodeGraph needs to be modified to only add nodes after authentication as to not add out-of-network nodes. The NodeGraph may also need to have network metadata as to filter out what node connection information to give to a peer. Finally, RPC handlers will have to be made to allow for the request and distribution of authentication tokens from seednodes, as well as allow for authentication between nodes.

We have decided, much like the ACL logic of our nodes, the authentication procedure should be carried out within the application layer. This parallels with the separation of authorization and TLS verification seen in the web. Separation of concerns is good here as the certchain is an immutable and inflexible structure that is purely meant to represent the flow of identity.

As for a node having access to multiple networks. It should be possible with this. It just means a node provides multiple network tokens when connecting and the connected node can decide to accept the connection based on it's access list. The only problem I can see with this is that one network might want to be strict and only allow nodes that use that one network. It would be possible to enforce this by rejecting nodes that provide multiple network tokens. but there's not way to force a node to provide all network tokens it has.

#### Bootstrapping

In order to bootstrap into a network, a node will make an RPC call to a seecnode in order to request a token to be signed in order to prove that the node is a part of the network. Connections to other nodes must not be made before the signed token is returned.

There should be separate RPC calls for requesting the token and receiving the token. The reason for this is so that a node can attempt to join a network and get the signed claim much later even if the node restarts or stops. This allows for nodes to be potentially individually accepted into the network by a human operator. If the node is offline when the seednode attempts to send it the fully signed token, the seednode can schedule a task to attempt to resend the token whenever a connection with the node is established again.

image

##### Trust Anchors

In order for the bootstrapping to work, there needs to be some anchor where trust is derived from.

To solve this, we present Trust Anchor entities that simply hold a public-private key. A token can be generated from the public private keypair that grants authority for seednodes to distribute tokens that prove that nodes are in the network.

An entity can be a node, a person, or some autonomous system. All that matters is that it is able to give the token to seednodes during bootstrapping so that they have a way to distribute network membership tokens.

Network membership tokens should contain the trust anchor's token embedded within. By comparing the facts between the network membership tokens and the trust anchor tokens, we can determine if a peer node is apart of a network.

#### Connection Procedure

When a node joins a network, it receives a token with rules and facts that represent its permissions and memberships.

An example what it could look like would be

```json
{
 "rules": [
   "can_connect_to_network($N, $Network) :- node($N), has_membership($N, $Network)."
 ],
 "facts": [
   "node(\"N1\").",
   "has_membership(\"N1\", \"mainnet.polykey.com\").",
   "has_membership(\"N1\", \"testnet.polykey.com\")."
 ]
}
```

1. When Node A attempts to connect to Node B, it presents its JWT token.
2. Node B verifies the token's signature and evaluates its rules and facts against its own policies.
3. The security policies are evaluated using a Datalog engine or a similar logic programming interpreter.
4. The connection is allowed or denied based on the evaluation results.

There needs to be a way to bootstrap seednodes of a network with a same private key to use to sign tokens to allow this.

### Additional context

[https://github.com/MatrixAI/Polykey-Enterprise/issues/11](<https://github.com/MatrixAI/Polykey-Enterprise/issues/11>)<br>Related: [ENG-370](https://linear.app/matrix-ai/issue/ENG-370) <br>Related: [ENG-371](https://linear.app/matrix-ai/issue/ENG-371) <br>Related: [ENG-372](https://linear.app/matrix-ai/issue/ENG-372) <br>Related: [ENG-373](https://linear.app/matrix-ai/issue/ENG-373) <br>Related: [ENG-374](https://linear.app/matrix-ai/issue/ENG-374)

### Tasks

1. \[ \] Token request from seednodes RPC handler
2. \[ \] Token response from seednodes RPC handler
3. \[ \] Authentication RPC handler between seednodes
4. \[ \] Modify NodeGraph to only add nodes after authentication
5. \[ \] Limit RPC calls whilst peer node is unauthenticated

### LINKED ISSUE #779: Network Segregation/Authentication Token Logic
### Specification

In order to form a network securely, we need some way to verify cryptographically whether a node should be able to access a network. To do this we need to make use of tokens as a portable, scale-able and secure method for proving access to a network. To do this the tokens need to encode details about the network authority, the subject node and provide a signature.

For out needs, we need a single network authority that defines the network itself. The details of this can be decided but all we require is a secure key pair to act as the trust anchor for the network authority. But we need to be able to issue access tokens in a decentralized way. Ideally using seed nodes issues for network access tokens.

We are going to be doing this with 2 types of tokens:

1. `ClaimNetworkAuthority` - This is a token issued by the network authority authorizing the ability for nodes to mint `ClaimnetworkAccess` tokens.
2. `ClaimNetworkAccess` - This token is minted for a specific node to prove access to a network.

Both tokens are required to be provided as a `token chain` to prove access to a network. They need to encode the following facts to prove this. Note that the ids for all purposes will just be `NodeIds` of nodes in the network as a way to identify a node and key interchangeably.

1. The id/publicKey for the network authority. Its use to verify and identify the network
2. The id/publicKey of the minter for the network access token.
3. The id/publicKey of the node that gains access to the network.
4. Signature from the network authority.
5. Signature from the minter node.

#### Minting tokens

The `ClaimNetworkAuthority` token exists to prove that a node can mint tokens on the authority of the network. This means minters can be dynamic and not know ahead of time so long as they hold the authority token and the authority is known. This is important for dynamic scaling of minter nodes. 

In the most basic form this `ClaimNetworkAuthority` needs to encode the following information.

1. type=ClaimNetworkAuthority - indicating that the claim is a `ClainNetworkAuthority`.
1. iss: The id/publicKey of the network authority.
2. sub: the id/publicKey of the minter node.
3. signature created with the privateKey of the network authority (iss).

Succinctly this proves that the subject (sub) has the authority of the network authority (iss) to mint access tokens for that network. This is verified with the signature. This for fills the facts 1, 2 and 4 above.

The `ClaimNetworkAccess` token exists to prove that a node has access to the network. This can be issued by any node that has a valid `ClaimNetworkAuthority` token. 

In its most basic form the `ClaimNetworkAccess` token needs to encode the following information.

1. type=ClaimNetworkAccess - indicating that the claim is a `ClaimNetworkAccess`.
2. iss: The id/publicKey of the minter node.
3. sub: the id/publicKey of the authorised node.
3. signature created with the privateKey of the minter node (iss).

Succinctly this proves that the subject (sub) has access to the network under the authority of the network authority. This is verified with the signature. This for fills the facts 2, 3 and 5. Note that the fact 2 is common between the two claims and helps verify the chain.

The `ClaimNetworkAccess` needs to include extra metadata refers the `ClaimNetworkAuthority` in some way. While this data isn't strictly required to verify the chain. It should provide a way to find all the required information needed to verify the chain. This can take two forms.

1. Embed the `ClaimNetworkAuthority` claim within the `ClaimNetworkAccess` claim. Simple but bloats the information.
2. Include the `claimId` of the `ClaimNetworkAuthority`, The `NodeId` of the node we can get the token from. And optionally a hash of the token.

So long as the verify-er  has the contents of both tokens then it can verify it.

#### Verifying tokens

Much like a certificate chain in TLS, the tokens form a `token chain`. Where verifying the tokens in order can prove a chain of trust that verifies access to a network. To verify all of the following facts.

1. Check the TLS certificate of the node connecting and asking for access to extract the `NodeId`.
2. Check the `sub` of the `ClaimNetworkAccess` token matches the connected node.
3. Check the signature of the `ClaimNetworkAccess` token is signed buy it's `iss`.
4. Check that the `iss` of the `ClaimNetworkAccess` token matches the `sub` of the `ClaimNetworkAuthority` token.
5. Check that the `iss` of the `ClaimNetworkAccess` token matches the network we are authorised under.
6. Check that the signature of the `claimNetworkAccess` token is signed bu the `iss`.

If all these facts are true and the network matches one we allow then we allow the connection. Otherwise we reject and kill the connection.

To expand on this we may want to enforce a strict mode where we only allow the node to access a single network. To do this we can can requests the access tokens as part of their sigchain. If we only see one network access token within their chain then that meets the strict condition.

### Additional context

Parent: https://github.com/MatrixAI/Polykey/issues/770

### Tasks

1. Create a structure for `ClaimNetworkAuthority` and `ClaimNetworkAccess` claims within the `claims` domain.
2. Add logic for minting these claims.
3. Add a utility for verifying these claims.

amydevs:
### Description

This PR aims to implement network segregation via a fractal network pattern enforced through token verification,

### Issues Fixed

* Fixes #779 
* REF ENG-370

### Tasks

- [x] 1. Implement Token Verification Logic into RPC system

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -52,6 +52,7 @@ import * as workersUtils from './workers/utils';
 import * as clientMiddleware from './client/middleware';
 import clientServerManifest from './client/handlers';
 import agentServerManifest from './nodes/agent/handlers';
+
 /**
  * Optional configuration for `PolykeyAgent`.
  */
@@ -61,6 +62,7 @@ type PolykeyAgentOptions = {
   clientServicePort: number;
   agentServiceHost: string;
   agentServicePort: number;
+  network: string;
   seedNodes: SeedNodes;
   workers: number;
   ipv6Only: boolean;
@@ -160,6 +162,7 @@ class PolykeyAgent {
       agentServiceHost: config.defaultsUser.agentServiceHost,
       agentServicePort: config.defaultsUser.agentServicePort,
       seedNodes: config.defaultsUser.seedNodes,
+      network: config.defaultsUser.network,
       workers: config.defaultsUser.workers,
       ipv6Only: config.defaultsUser.ipv6Only,
       keys: {
@@ -687,6 +690,7 @@ class PolykeyAgent {
         groups: Array<string>;
         port: number;
       };
+      network: string;
       seedNodes: SeedNodes;
     }>;
     workers?: number;
@@ -705,6 +709,7 @@ class PolykeyAgent {
           groups: config.defaultsSystem.mdnsGroups,
           port: config.defaultsSystem.mdnsPort,
         },
+        network: config.defaultsUser.network,
         seedNodes: config.defaultsUser.seedNodes,
       });
       // Register event handlers
--- a/src/bootstrap/utils.ts
+++ b/src/bootstrap/utils.ts
@@ -30,7 +30,7 @@ import * as utils from '../utils';
 import * as errors from '../errors';
 
 /**
- * Bootstraps the Node Path
+ * Bootstraps the Node Path`
  */
 async function bootstrapState({
   // Required parameters
--- a/src/claims/payloads/claimNetworkAccess.ts
+++ b/src/claims/payloads/claimNetworkAccess.ts
@@ -1,0 +1,93 @@
+import type { Claim, SignedClaim } from '../types';
+import type { NodeIdEncoded } from '../../ids/types';
+import type { SignedTokenEncoded } from '../../tokens/types';
+import * as tokensSchema from '../../tokens/schemas';
+import * as ids from '../../ids';
+import * as claimsUtils from '../utils';
+import * as tokensUtils from '../../tokens/utils';
+import * as validationErrors from '../../validation/errors';
+import * as utils from '../../utils';
+
+/**
+ * Asserts that a node is apart of a network
+ */
+interface ClaimNetworkAccess extends Claim {
+  typ: 'ClaimNetworkAccess';
+  iss: NodeIdEncoded;
+  sub: NodeIdEncoded;
+  network: string;
+  signedClaimNetworkAuthorityEncoded?: SignedTokenEncoded;
+}
+
+function assertClaimNetworkAccess(
+  claimNetworkAccess: unknown,
+): asserts claimNetworkAccess is ClaimNetworkAccess {
+  if (!utils.isObject(claimNetworkAccess)) {
+    throw new validationErrors.ErrorParse('must be POJO');
+  }
+  if (claimNetworkAccess['typ'] !== 'ClaimNetworkAccess') {
+    throw new validationErrors.ErrorParse(
+      '`typ` property must be `ClaimNetworkAccess`',
+    );
+  }
+  if (
+    claimNetworkAccess['iss'] == null ||
+    ids.decodeNodeId(claimNetworkAccess['iss']) == null
+  ) {
+    throw new validationErrors.ErrorParse(
+      '`iss` property must be an encoded node ID',
+    );
+  }
+  if (
+    claimNetworkAccess['sub'] == null ||
+    ids.decodeNodeId(claimNetworkAccess['sub']) == null
+  ) {
+    throw new validationErrors.ErrorParse(
+      '`sub` property must be an encoded node ID',
+    );
+  }
+  if (
+    claimNetworkAccess['network'] == null ||
+    typeof claimNetworkAccess['network'] !== 'string'
+  ) {
+    throw new validationErrors.ErrorParse(
+      '`network` property must be a string',
+    );
+  }
+  if (
+    claimNetworkAccess['signedClaimNetworkAuthorityEncoded'] != null &&
+    !tokensSchema.validateSignedTokenEncoded(
+      claimNetworkAccess['signedClaimNetworkAuthorityEncoded'],
+    )
+  ) {
+    throw new validationErrors.ErrorParse(
+      '`signedClaimNetworkAuthorityEncoded` property must be an encoded signed token',
+    );
+  }
+}
+
+function parseClaimNetworkAccess(
+  claimNetworkAccessEncoded: unknown,
+): ClaimNetworkAccess {
+  const claimNetworkNode = claimsUtils.parseClaim(claimNetworkAccessEncoded);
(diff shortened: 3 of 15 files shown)


tegefaulkes:
@amydevs avoid `wip` commits. Be sure to add a description to them `wip: did a thing` just so we can get an idea of the changes from the commit message. On top of that, if you know CI wont pass then add `[ci skip]` in the commit message as well.

The PR description needs to be updated as well since we've subdivided the work. This is focusing solely on the authentication logic for the network access claims now. So the description and tasks need to reflect that.

amydevs:
I've added the authentication logic + tests, i'm going to look as to where to slot in the logic that will prevent calls being made to the RPC client when the connection has not yet been authenticated, as well as where to drop the connection if authentication fails.

CMCDragonkai:
Merge this with just mainnet and testnet segregation.

Start with `--network=X` use the `X` is the unique identifier for the network. Then just use 1 single token the `ClaimNetworkAccess` token to indicate what network you're claiming to access to.

Special case for `--network=''` which is an empty string. You could claim with an empty string as the unique identifier, or we can say it's just equivalent to `null` or `undefined` meaning there's no token at all.

Separate out the work for PKE. PKE side of things can be done later.

PR #785, opened 2024-08-06
Implementing general-purpose file list handler

aryanjassal:
### Description

Currently, Polykey can only return a list of secrets within a vault. This PR aims to implement an RPC handler which creates a general-purpose function able to output data similar to the UNIX command `ls`.

### Issues Fixed

* Relates to https://github.com/MatrixAI/Polykey-CLI/issues/245
* Relates to https://github.com/MatrixAI/Polykey-CLI/pull/255
* REF ENG-358

### Tasks

- [x] 1. Write RPC handler for UNIX `ls`
- [x] 2. Remove deprecated `list` command for secrets
- [x] 3. Write tests for `ls` handler
- ~~4. Write handler for streaming file contents~~ will be completed in `cat` command
- ~~5. Write tests for file streaming (`cat`) handler~~ will be completed in `cat` command

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsSecretsList.ts
+++ b/src/client/handlers/VaultsSecretsList.ts
@@ -2,55 +2,69 @@ import type { DB } from '@matrixai/db';
 import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
-  SecretNameMessage,
-  VaultIdentifierMessage,
+  SecretFilesMessage,
+  SecretIdentifierMessage,
 } from '../types';
 import type VaultManager from '../../vaults/VaultManager';
+import path from 'path';
 import { ServerHandler } from '@matrixai/rpc';
 import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
-import * as vaultOps from '../../vaults/VaultOps';
 
 class VaultsSecretsList extends ServerHandler<
   {
     vaultManager: VaultManager;
     db: DB;
   },
-  ClientRPCRequestParams<VaultIdentifierMessage>,
-  ClientRPCResponseResult<SecretNameMessage>
+  ClientRPCRequestParams<SecretIdentifierMessage>,
+  ClientRPCResponseResult<SecretFilesMessage>
 > {
   public async *handle(
-    input: ClientRPCRequestParams<VaultIdentifierMessage>,
-    _cancel,
-    _meta,
-    ctx,
-  ): AsyncGenerator<ClientRPCResponseResult<SecretNameMessage>> {
-    if (ctx.signal.aborted) throw ctx.signal.reason;
+    input: ClientRPCRequestParams<SecretIdentifierMessage>,
+    _cancel: any,
+  ): AsyncGenerator<ClientRPCResponseResult<SecretFilesMessage>, void, void> {
     const { vaultManager, db } = this.container;
-    const secrets = await db.withTransactionF(async (tran) => {
+    const vaultId = await db.withTransactionF(async (tran) => {
       const vaultIdFromName = await vaultManager.getVaultId(
         input.nameOrId,
         tran,
       );
       const vaultId =
         vaultIdFromName ?? vaultsUtils.decodeVaultId(input.nameOrId);
-      if (vaultId == null) {
-        throw new vaultsErrors.ErrorVaultsVaultUndefined();
-      }
-      return await vaultManager.withVaults(
-        [vaultId],
-        async (vault) => {
-          return await vaultOps.listSecrets(vault);
-        },
-        tran,
-      );
+      if (vaultId == null) throw new vaultsErrors.ErrorVaultsVaultUndefined();
+      return vaultId;
     });
-    for (const secret of secrets) {
-      if (ctx.signal.aborted) throw ctx.signal.reason;
-      yield {
-        secretName: secret,
-      };
-    }
+
+    yield* vaultManager.withVaultsG([vaultId], (vault) => {
+      return vault.readG(async function* (fs): AsyncGenerator<
+        SecretFilesMessage,
+        void,
+        void
+      > {
+        let files: Array<string | Buffer>;
+        try {
+          files = await fs.promises.readdir(input.secretName);
+        } catch (e) {
+          if (e.code === 'ENOENT') {
+            throw new vaultsErrors.ErrorSecretsDirectoryUndefined(e.message, {
+              cause: e,
+            });
+          }
+          if (e.code === 'ENOTDIR') {
+            throw new vaultsErrors.ErrorSecretsIsSecret(e.message, {
+              cause: e,
+            });
+          }
+          throw e;
+        }
+        for await (const file of files) {
+          const filePath = path.join(input.secretName, file.toString());
+          const stat = await fs.promises.stat(filePath);
+          const type = stat.isFile() ? 'FILE' : 'DIRECTORY';
+          yield { path: filePath, type: type };
+        }
+      });
+    });
   }
 }
 
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -300,12 +300,11 @@ type VaultsLatestVersionMessage = {
 };
 
 // Secrets
-
-type SecretNameMessage = {
+type SecretPathMessage = {
   secretName: string;
 };
 
-type SecretIdentifierMessage = VaultIdentifierMessage & SecretNameMessage;
+type SecretIdentifierMessage = VaultIdentifierMessage & SecretPathMessage;
 
 // Contains binary content as a binary string 'toString('binary')'
 type ContentMessage = {
@@ -325,6 +324,11 @@ type SecretDirMessage = VaultIdentifierMessage & {
 
 type SecretRenameMessage = SecretIdentifierMessage & {
   newSecretName: string;
+};
+
+type SecretFilesMessage = {
+  path: string;
+  type: 'FILE' | 'DIRECTORY';
 };
 
 // Stat is the 'JSON.stringify version of the file stat
@@ -410,13 +414,14 @@ export type {
   VaultsScanMessage,
   VaultsVersionMessage,
   VaultsLatestVersionMessage,
-  SecretNameMessage,
+  SecretPathMessage,
   SecretIdentifierMessage,
   ContentMessage,
   SecretContentMessage,
   SecretMkdirMessage,
   SecretDirMessage,
   SecretRenameMessage,
+  SecretFilesMessage,
   SecretStatMessage,
   SignatureMessage,
   OverrideRPClientType,
--- a/src/vaults/fileTree.ts
+++ b/src/vaults/fileTree.ts
@@ -542,8 +542,20 @@ function parserTransformStreamFactory(): TransformStream<
     };
     jsonParser.write(initialChunk);
   };
+  /* Check if any chunks have been processed. If the stream is being flushed
+   * without processing any chunks, then something went wrong with the stream.
+   */
+  let processedChunks: boolean = false;
   return new TransformStream<Uint8Array, TreeNode | ContentNode | Uint8Array>({
+    flush: (controller) => {
+      if (!processedChunks) {
+        controller.error(
+          new validationErrors.ErrorParse('Stream ended prematurely'),
+        );
+      }
+    },
     transform: (chunk, controller) => {
+      if (chunk.byteLength > 0) processedChunks = true;
       switch (phase) {
         case 'START': {
           workingBuffer = vaultsUtils.uint8ArrayConcat([workingBuffer, chunk]);
--- a/tests/client/handlers/vaults.test.ts
+++ b/tests/client/handlers/vaults.test.ts
@@ -1587,16 +1587,32 @@ describe('vaultsSecretsNewDir and vaultsSecretsList', () => {
       dirName: secretDir,
     });
     expect(addResponse.success).toBeTruthy();
-    // List secrets
-    const listResponse = await rpcClient.methods.vaultsSecretsList({
+
+    const noFiles = await rpcClient.methods.vaultsSecretsList({
       nameOrId: vaultsIdEncoded,
+      secretName: 'doesntExist',
     });
-    const secrets: Array<string> = [];
-    for await (const secret of listResponse) {
-      secrets.push(secret.secretName);
+
+    await expect(async () => {
+      try {
+        for await (const _ of noFiles);
+      } catch (e) {
+        throw e.cause;
+      }
+    }).rejects.toThrow(vaultsErrors.ErrorSecretsDirectoryUndefined);
+
+    const secrets = await rpcClient.methods.vaultsSecretsList({
+      nameOrId: vaultsIdEncoded,
+      secretName: 'secretDir',
+    });
+
+    // Extract secret file paths
+    const parsedFiles: Array<string> = [];
+    for await (const file of secrets) {
+      parsedFiles.push(file.path);
     }
-    expect(secrets.sort()).toStrictEqual(
-      secretList.map((secret) => path.join('secretDir', secret)).sort(),
+    expect(parsedFiles).toIncludeAllMembers(
+      secretList.map((secret) => path.join('secretDir', secret)),
     );
   });
 });
--- a/src/vaults/VaultManager.ts
+++ b/src/vaults/VaultManager.ts
@@ -1058,13 +1058,14 @@ class VaultManager {
       },
     );
     // Running the function with locking
+    const vaultThis = this;
     return yield* this.vaultLocks.withG(
       ...vaultLocks,
       async function* (): AsyncGenerator<T, Treturn, Tnext> {
         // Getting the vaults while locked
         const vaults = await Promise.all(
           vaultIds.map(async (vaultId) => {
-            return await this.getVault(vaultId, tran);
+            return await vaultThis.getVault(vaultId, tran);
           }),
         );
         return yield* g(...vaults);
--- a/src/vaults/errors.ts
+++ b/src/vaults/errors.ts
@@ -118,6 +118,16 @@ class ErrorSecretsIsDirectory<T> extends ErrorSecrets<T> {
   exitCode = sysexits.USAGE;
 }
 
+class ErrorSecretsIsSecret<T> extends ErrorSecrets<T> {
+  static description = 'Is a secret and not a directory';
+  exitCode = sysexits.USAGE;
+}
+
+class ErrorSecretsDirectoryUndefined<T> extends ErrorSecrets<T> {
+  static description = 'Directory does not exist';
+  exitCode = sysexits.USAGE;
+}
+
 export {
   ErrorVaults,
   ErrorVaultManagerRunning,
@@ -144,4 +154,6 @@ export {
   ErrorSecretsSecretUndefined,
   ErrorSecretsSecretDefined,
   ErrorSecretsIsDirectory,
+  ErrorSecretsIsSecret,
+  ErrorSecretsDirectoryUndefined,
 };


package.json
```diff
@@ -36,6 +36,9 @@
     },
     {
       "name": "Aditya Varma"
+    },
+    {
+      "name": "Aryan Jassal"
```
tegefaulkes: I'd commit this change separately to staging.

src/client/handlers/VaultsSecretsList.ts
```diff
@@ -2,32 +2,29 @@ import type { DB } from '@matrixai/db';
...
+class VaultsSecretsList extends UnaryHandler<
   {
     vaultManager: VaultManager;
     db: DB;
   },
-  ClientRPCRequestParams<VaultIdentifierMessage>,
-  ClientRPCResponseResult<SecretNameMessage>
+  ClientRPCRequestParams<SecretPatternMessage>,
+  ClientRPCResponseResult<SecretFilesList>
 > {
```
tegefaulkes: 1. We need a new name, this will be a generic RPC call for getting the file tree using the globWalk utility.
2. It needs to be a `RawHandler` since the file tree by design needs to be streamed over as raw binary data. Changing this to a `UnaryHandler` is an odd choice here based on what we discussed.

src/client/handlers/VaultsSecretsList.ts
```diff
@@ -40,18 +37,37 @@ class VaultsSecretsList extends ServerHandler<
...
+            const fileTreeGen = fileTree.globWalk({
+              fs: fs,
+              basePath: '.',
+              pattern: input.pattern,
+              yieldStats: true,
+              yieldRoot: false,
+              yieldFiles: true,
+              yieldParents: true,
+              yieldDirectories: true,
+            });
```
tegefaulkes: Only the `fs` and `basePath` should be hard coded here. All other options should be provided by the RPC call.

src/client/handlers/VaultsSecretsList.ts
```diff
@@ -40,18 +37,37 @@ class VaultsSecretsList extends ServerHandler<
...
+            const serializedStream = fileTree.serializerStreamFactory(
+              fs,
+              fileTreeGen,
+              true,
+            );
+            for await (const output of serializedStream) {
+              data.push(output);
+            }
+            return data;
+          });
```
tegefaulkes: This is a bad pattern, everything until now has been designed to be streamed. So we don't want to accumulate the stream as a single array of strings. You turn something that streams parts in piece meal into one gigantic message here.

We want to avoid that so the call can scale. For example what happens if the tree is really big and contains 10gb of data? Do we want to wait the whole time it takes to read all of that? Then store it all in memory while we format the message and send it over in one giant chunk? It's not going to work.

The Handler needs to be a `RawHandler` and `serializedStream` needs to be returned as the output raw stream.

I'm guessing you did this because `withVaults` and `readF` returns a single value. You need to use the generator forms of `withVaultsG` and `readG` to do this properly. If you don't know how js `AsyncGenerator`s work then be sure to read up on them.

src/client/handlers/VaultsSecretsList.ts
aryanjassal: For 1, I guess a good name for this could be `VaultsSecretsGetFileTree`

aryanjassal:
I have encountered an issue in the implementation, where accessing the stream returns a `read 0` error. This needs some time to fix, so I will need to have a chat with @tegefaulkes on Monday to work this out.

CMCDragonkai:
So in https://github.com/MatrixAI/Polykey-CLI/issues/245, there's nowhere that the spec specifies the possibility of a getting file contents.

However this PR appears to create a raw stream that can also stream file contents.

I think file content streaming (if you really want to do a sort of archive streaming), should really be a different RPC call.

The `ls` RPC call should be simplified, and be similar to using a "scandir" in Python, or iterating over directory entries like in `fs.opendir`.

Basically that means this should work using a server streaming call, not a raw stream.

You can rename the raw stream handler to some other thing like `archiveDir` which then focuses on file contents.

I don't think it's always the right thing to have a single flexible god function that can do anything, instead having named functions that do specific things is often much clearer especially between program interfaces.

tegefaulkes:
The way I designed this was in two parts. First the `globWalk` utility that generated the file tree structure using a glob pattern. And secondly the `serializerStreamFactory` which takes the `globWalk` generator and serialises it into a binary stream. Stemming from that we'd have a single RPC handler that wraps that whole functionality of serialising the file tree and it's content.

But you're right, this could be split into two stages. 
1. server stream that wraps the `globWalk` for getting the file tree.
2. The raw stream that serialises the contents of secrets over the raw stream.

The problem with this is we can run into race conditions across the multiple RPC streams since the locks would only apply to each stream separately. To solve this we'd need a RPC call for getting the lock of one or more vaults and allows us to use that lock for the follow up RPC calls. This lock can be held for the duration of the RPC call and any follow up RPC calls. It will need to handle timeouts for a call holding the lock for too long for whatever reason. I was thinking this was needed for some of the more complex unix commands such as `cp` and `mv` anyway.

This leaves us with a few small changes to how I wrote the utilities.

1. Make a `ServerHandler` to wrap the functionality of `globWalk`.
2. rewrite the  `serializerStreamFactory` to only handle the content phase. so the tree phase needs to be cut out.
3. Make a `RawHandler` for wrapping the new `serializerStreamFactory` functionality.
4. Make a vault locking RPC call that will hold the lock for the requested vaults for the duration of the call. This may need to be a  `ClientHandler`. It will need to grab the lock, return an array of numbers to reference the requested locks. These numbers need to be used when making follow up RPC calls that require locking such as the glob walk and the content serialiser.

aryanjassal:
> 1. rewrite the serializerStreamFactory to only handle the content phase. so the tree phase needs to be cut out.
Make a RawHandler for wrapping the new serializerStreamFactory functionality.
> 4. Make a vault locking RPC call that will hold the lock for the requested vaults for the duration of the call. This may need to be a ClientHandler. It will need to grab the lock, return an array of numbers to reference the requested locks. These numbers need to be used when making follow up RPC calls that require locking such as the glob walk and the content serialiser.

I'm not sure how to do these tasks. We will need to have a discussion to help me with understanding and implementing this.

aryanjassal:
In `src/vaults/fileTree.ts:343`, I have modified the `serializerStreamFactory` to return a generator instead of returning a stream. This change was necessary to make it work with `withVaultsG`.

The only changes I made was to replace the `controller.enqueue(data)` to `yield *data` in the actual code to align with making it act as a generator instead of a stream. This, however, broke one of the tests, and I'm unable to understand why or how to fix it.

```
Summary of all failing tests
 FAIL  tests/vaults/fileTree.test.ts
  ● fileTree › serializer › handles invalid data (with seed=-646156050)

    Property failed after 20 tests
    { seed: -646156050, path: "19", endOnFailure: true }
    Counterexample: [Buffer.from([])]
    Shrunk 0 time(s)
    Got error: Should have thrown an error when parsing

      718 |         return;
      719 |       }
    > 720 |       throw Error('Should have thrown an error when parsing');
          |             ^
      721 |     });
      722 |     // TODO: tests for
      723 |     //  - empty files

      at Error (tests/vaults/fileTree.test.ts:720:13)
      at AsyncProperty.run (node_modules/fast-check/lib/check/property/AsyncProperty.generic.js:49:28)
      Hint: Enable verbose mode in order to have the list of all failing values encountered during the run
      at buildError (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:131:15)
      at asyncThrowIfFailed (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:148:11)
```

CMCDragonkai:
> The way I designed this was in two parts. First the `globWalk` utility that generated the file tree structure using a glob pattern. And secondly the `serializerStreamFactory` which takes the `globWalk` generator and serialises it into a binary stream. Stemming from that we'd have a single RPC handler that wraps that whole functionality of serialising the file tree and it's content.
> 
> But you're right, this could be split into two stages.
> 
> 1. server stream that wraps the `globWalk` for getting the file tree.
> 2. The raw stream that serialises the contents of secrets over the raw stream.
> 
> The problem with this is we can run into race conditions across the multiple RPC streams since the locks would only apply to each stream separately. To solve this we'd need a RPC call for getting the lock of one or more vaults and allows us to use that lock for the follow up RPC calls. This lock can be held for the duration of the RPC call and any follow up RPC calls. It will need to handle timeouts for a call holding the lock for too long for whatever reason. I was thinking this was needed for some of the more complex unix commands such as `cp` and `mv` anyway.
> 
> This leaves us with a few small changes to how I wrote the utilities.
> 
> 1. Make a `ServerHandler` to wrap the functionality of `globWalk`.
> 2. rewrite the  `serializerStreamFactory` to only handle the content phase. so the tree phase needs to be cut out.
> 3. Make a `RawHandler` for wrapping the new `serializerStreamFactory` functionality.
> 4. Make a vault locking RPC call that will hold the lock for the requested vaults for the duration of the call. This may need to be a  `ClientHandler`. It will need to grab the lock, return an array of numbers to reference the requested locks. These numbers need to be used when making follow up RPC calls that require locking such as the glob walk and the content serialiser.

You also have to keep in mind how the RPC may be used by third party client sdks in the future. So maintaining a simple IPC interface helps adoption. This is why I prefer having both complex RPC and simple RPC options being compostable.

tegefaulkes:
> > 1. rewrite the serializerStreamFactory to only handle the content phase. so the tree phase needs to be cut out.
> >    Make a RawHandler for wrapping the new serializerStreamFactory functionality.
> > 2. Make a vault locking RPC call that will hold the lock for the requested vaults for the duration of the call. This may need to be a ClientHandler. It will need to grab the lock, return an array of numbers to reference the requested locks. These numbers need to be used when making follow up RPC calls that require locking such as the glob walk and the content serialiser.
> 
> I'm not sure how to do these tasks. We will need to have a discussion to help me with understanding and implementing this.

For 1. 

Right now the `serializerStreamFactory` will send over the tree data and then the content data if it's enabled with `yieldContents` set to true. You need to change it so it only  takes `TreeNode` of type `FILE` for the input and yield the contents of these files. So you're just cutting out some of the functionality it has.

for 2. 

Ignore that for now. Things will *work* without it, we just run into locking and race condition issues without it. I'll write up a new issue for it but we can proceed without it for now.

tegefaulkes:
> In `src/vaults/fileTree.ts:343`, I have modified the `serializerStreamFactory` to return a generator instead of returning a stream. This change was necessary to make it work with `withVaultsG`.
> 
> The only changes I made was to replace the `controller.enqueue(data)` to `yield *data` in the actual code to align with making it act as a generator instead of a stream. This, however, broke one of the tests, and I'm unable to understand why or how to fix it.
> 
> ```
> Summary of all failing tests
>  FAIL  tests/vaults/fileTree.test.ts
>   ● fileTree › serializer › handles invalid data (with seed=-646156050)
> 
>     Property failed after 20 tests
>     { seed: -646156050, path: "19", endOnFailure: true }
>     Counterexample: [Buffer.from([])]
>     Shrunk 0 time(s)
>     Got error: Should have thrown an error when parsing
> 
>       718 |         return;
>       719 |       }
>     > 720 |       throw Error('Should have thrown an error when parsing');
>           |             ^
>       721 |     });
>       722 |     // TODO: tests for
>       723 |     //  - empty files
> 
>       at Error (tests/vaults/fileTree.test.ts:720:13)
>       at AsyncProperty.run (node_modules/fast-check/lib/check/property/AsyncProperty.generic.js:49:28)
>       Hint: Enable verbose mode in order to have the list of all failing values encountered during the run
>       at buildError (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:131:15)
>       at asyncThrowIfFailed (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:148:11)
> ```

I know what's wrong with it but it's a good object lesson in debugging problems so I'll let you work this one out. I can tell you that.

1. The problem existed before your changes.
2. It's a common oversight with fast-check and not considering an edge case.

tegefaulkes:
I made a new issue for tracking the vault locking RPC call at #786

aryanjassal:
> Make a RawHandler for wrapping the new `serializerStreamFactory` functionality.

The specification of the file content streaming RPC call (which I have named `VaultsSecretsGetFileContents`) isn't very well defined. I'm not sure if the `RawHandler` would receive `TreeNodes` directly, or would it receive a pattern, which it would then use to match files and return their contents?

After changing the `serializerStreamFactory`, do I also make the changes in `parserTransformStreamFactory`, or leave it as-is, still supporting everything including `TreeNode`, `ContentNode`, and `Uint8Array`?

Most importantly, what is the use-case of this RPC handler? Is it to implement something like UNIX's `cat` functionality? If so, then this really belongs in a separate PR.

I would need clarification on all these points before I can begin implementation of the raw handler RPC call.

@tegefaulkes

CMCDragonkai:
Some of these function names seem really generic.

aryanjassal:
> Most importantly, what is the use-case of this RPC handler? Is it to implement something like UNIX's cat functionality? If so, then this really belongs in a separate PR.

This new handler will indeed be for other UNIX commands and is not relevant for the `ls` command, so this will not be done in this PR.

After having a chat with @tegefaulkes, I was able to better understand the requirements of this RPC and how to achieve it.

The `serializerStreamFactory` will receive a generator of `TreeNodes` of files we want to be serialised. If the generator does not return a `FILE`, then the value will simply be consumed. Otherwise, those files' contents will be serialized. This way, we don't even need any special headers like `TREE`, etc.

Same thing in the `parserTransformStreamFactory`. In there, we can also remove most of the code handling the different headers, as we are only interested in the `CONTENT` headers.

On that note, the `CONTENT` header can be simplified further by removing the `iNode` field. This is because the files returned will be in the same order that they were sent in. Thus, we don't need to keep track of which file is being returned and in what order.

aryanjassal:
As per our discussion, I have removed all the extra features we didn't need, and trimmed it to the bare minimum an `ls` command should do: list secrets. Advanced functionality can be iterated over in the future, but this will get the command out there.

I am working on more advanced features like parity with GNU's `ls` command, or showing a longer, detailed view of the files and directories, etc. which will not be a part of this PR, but the code for that does exist, and those features might make the cut next iteration.

src/client/types.ts
```diff
@@ -327,6 +330,11 @@ type SecretRenameMessage = SecretIdentifierMessage & {
   newSecretName: string;
 };
 
+type SecretFiles = {
+  path: string;
+  type: 'FILE' | 'DIRECTORY';
+};
+
```
tegefaulkes: Most of the types here are called `XMessage`. Don't deviate from the convention unless it makes sense. So this should be called `SecretFilesMessage`. But since it could be a file or directory the name `SecretFiles` doesn't make sense here. Shouldn't we call it `SecretDirectoryContentsMessage` in stead?

Ah, you didn't make it `XMessage` because that's already used. `SecretDirectoryContentsMessage` would make more sense here.

src/client/types.ts
```diff
@@ -300,6 +300,9 @@ type VaultsLatestVersionMessage = {
 };
 
 // Secrets
+type SecretFilesMessage = VaultIdentifierMessage & {
+  path: string;
+};
```
tegefaulkes: Why did we need a new message here? We already have a message that identifies a vault and secret path. It's called `SecretIdentifierMessage` right below this one. When adding messages you need to make sure that an applicable one already exists.

Maybe `SecretNameMessage` should be renamed as `SecretPathMessage` and `secretName` becomes `secretPath`.

src/vaults/fileTree.ts
```diff
@@ -542,8 +542,17 @@ function parserTransformStreamFactory(): TransformStream<
...
   return new TransformStream<Uint8Array, TreeNode | ContentNode | Uint8Array>({
+    flush: (controller) => {
+      if (!processed) {
+        controller.error(
+          new validationErrors.ErrorParse('Stream ended prematurely'),
+        );
+      }
+    },
     transform: (chunk, controller) => {
+      if (chunk.byteLength > 0) processed = true;
```
tegefaulkes: `processed` is not a clear name. we need a better name and a comment explaining why we have it at all.

tests/client/handlers/vaults.test.ts
```diff
@@ -1587,16 +1588,31 @@ describe('vaultsSecretsNewDir and vaultsSecretsList', () => {
...
+      const files = await rpcClient.methods.vaultsSecretsList({
+        nameOrId: vaultsIdEncoded,
+        path: 'doesntExist',
+      });
+      try {
+        for await (const _ of files); // Consume values
+      } catch (e) {
+        throw e.cause;
+      }
+    }).rejects.toThrow(clientErrors.ErrorClientFSReadFailed);
```
tegefaulkes: it might be a better idea to unwrap this expectation and just do the expect in the try-catch block.

```ts
try {
        for await (const _ of files); // Consume values
        throw Error('failed to throw an error');
      } catch (e) {
        expect(e.cause).toBeInstanceOf(clientErrors.ErrorClientFSReadFailed);
      }
```

or only wrap the `for await (const _ of files);` in the expectation async function.

src/client/handlers/VaultsSecretsList.ts
```diff
@@ -2,55 +2,65 @@ import type { DB } from '@matrixai/db';
...
+        let files: Array<string>;
+        try {
+          // @ts-ignore: While the types don't fully match, it matches enough for our usage.
+          files = await fs.promises.readdir(input.path);
+        } catch (e) {
+          throw new clientErrors.ErrorClientFSReadFailed(
+            `No matches exist for path: ${input.path}`,
+            { cause: e },
+          );
+        }
```
tegefaulkes: Two points.

1. Avoid using the `@ts-ignore` here. Needing a `@ts-ignore` is a last ditch solution. In this case we can take both the outputs of `string` and `Buffer` and convert it back to the `string` format with `.toString()`.
2. I don't think we should wrap the error here? In any case never assume that it will be the error you expect. If you want to wrap an `ENOENT` error then you need to check it's that error. If it isn't then just re-throw the unchanged error.

src/client/handlers/VaultsSecretsList.ts
```diff
@@ -2,55 +2,65 @@ import type { DB } from '@matrixai/db';
...
+        files = files.map((file) => path.join(input.path, file));
+
+        for await (const file of files) {
+          const stat = await fs.promises.stat(file);
+          const type = stat.isFile() ? 'FILE' : 'DIRECTORY';
+          yield { path: file, type: type };
+        }
```
tegefaulkes: No point doing a `files.map` here just to iterate over it to do a bunch of stuff anyway. You can move the `path.join` logic into the `for await` loop.

tegefaulkes (commented):
Some small things need changing.

src/client/types.ts
aryanjassal: Ah, before, I had added other options specific to `globWalk` like yielding stats, etc.

For this PR, everything had to be simplified, so I removed those options. I guess after removing the options, my type became too simple and was already existing. I will change this to use the existing type then.

src/client/types.ts
aryanjassal: Following https://github.com/MatrixAI/Polykey/pull/785#discussion_r1730588770, I've named this the `SecretFilesMessage` and removed `SecretFilesMessage` as `SecretIdentifierMessage` already existed for that purpose.

src/vaults/fileTree.ts
aryanjassal: I have renamed it to `processedChunks` and added a comment briefly explaining the purpose of the boolean and why should it be an error.

src/client/handlers/VaultsSecretsList.ts
aryanjassal: > I don't think we should wrap the error here? In any case never assume that it will be the error you expect. If you want to wrap an ENOENT error then you need to check it's that error. If it isn't then just re-throw the unchanged error.

Okay, I'll just remove the error wrapping here then.

aryanjassal: Apparently this error isnt registered to be thrown as an error in Polykey, so while the RPC will throw, the handler will not interpret it as an error, rather just a `Object`. As such, I will be wrapping the error around something like `vaultErrors.ErrorVaultSecretUndefined`.

tests/client/handlers/vaults.test.ts
aryanjassal: I got the `expect.rejects.toThrow` from another test case, like `tests/vaults/VaultOps.test.ts:226`

```ts
test('updating a secret that does not exist should fail', async () => {
      await expect(
        vaultOps.updateSecret(vault, secretName, secretContentNew),
      ).rejects.toThrow(vaultsErrors.ErrorSecretsSecretUndefined);
    });
```

src/vaults/fileTree.ts
tegefaulkes: Not really chunks, we're just checking that any bytes were processed.

src/client/handlers/VaultsSecretsList.ts
```diff
@@ -2,55 +2,69 @@ import type { DB } from '@matrixai/db';
...
+          if (e.code === 'ENOENT') {
+            throw new vaultsErrors.ErrorSecretsDirectoryUndefined(e.message, {
+              cause: e,
+            });
+          }
+          if (e.code === 'ENOTDIR') {
+            throw new vaultsErrors.ErrorSecretsDirectoryUndefined(e.message, {
+              cause: e,
+            });
+          }
```
tegefaulkes: It should be a `ErrorSecretsIsSecret` error that you need to make. as a complement to the `ErrorSecretsIsDirectory` that already exists

aryanjassal:
Everything looks good. Merging.

PR #803, opened 2024-09-10
Implementing `secrets rm` RPC handler

aryanjassal:
### Description

Implement a RPC handler to delete multiple files. If an option is set, then also delete the specified directory and the contents of that directory from the vaults.

### Issues Fixed

* Relates to https://github.com/MatrixAI/Polykey-CLI/issues/247 ([ENG-360](https://linear.app/matrix-ai/issue/ENG-360/implement-secrets-rm-command))
* Closes ENG-360

### Tasks

- [x] 1. Add new `VaultsSecretsRemove` handler (functionally similar to `VaultsSecretsDelete`)
- [x] 2. Add support for specifying and deleting multiple files
- [x] 3. Add recursive flag to delete directories and their contents
- [x] 4. Add tests

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -30,21 +30,22 @@ class VaultsSecretsDelete extends UnaryHandler<
       );
       const vaultId =
         vaultIdFromName ?? vaultsUtils.decodeVaultId(input.nameOrId);
-      if (vaultId == null) {
-        throw new vaultsErrors.ErrorVaultsVaultUndefined();
-      }
+      if (vaultId == null) throw new vaultsErrors.ErrorVaultsVaultUndefined();
```
tegefaulkes: You also need to validate that the `input.secretNames` is an array of strings that match the path format.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -30,21 +30,22 @@ class VaultsSecretsDelete extends UnaryHandler<
...
       await vaultManager.withVaults(
         [vaultId],
         async (vault) => {
-          await vaultOps.deleteSecret(vault, input.secretName);
+          for (const secretName of input.secretNames) {
+            await vaultOps.deleteSecret(vault, secretName, {
+              recursive: input.options?.recursive,
+            });
+          }
```
tegefaulkes: It's cool that `deleteSecret` handles all the logic for us so far. But using this way will result in a separate vault log message for each deletion. We want to include all of them within the same vault commit.

To do this you need to modify `vaultsOps.deleteSecret` to take the array of paths and delete each one within the same `vault.writeF()` call. It's a simple change.

src/client/types.ts
```diff
@@ -304,8 +304,19 @@ type SecretPathMessage = {
   secretName: string;
 };
 
+type SecretPathsMessage = {
+  secretNames: Array<string>;
+};
+
```
tegefaulkes: No point making a separate `SecretPathsMessage` if it's not used anywhere else. Just roll it into the `SecretRemoveMessage`

tests/client/handlers/vaults.test.ts
```diff
@@ -1470,6 +1477,90 @@ describe('vaultsSecretsNew and vaultsSecretsDelete, vaultsSecretsGet', () => {
...
+      vaultsErrors.ErrorVaultsRecursive,
+    );
+    const deleteResponse = await rpcClient.methods.vaultsSecretsRemove({
+      nameOrId: vaultsIdEncoded,
+      secretNames: ['secretDir'],
+      options: {
+        recursive: true,
+      },
+    });
+    expect(deleteResponse.success).toBeTruthy();
```
tegefaulkes: You still need to check that each secret is removed as expected including the directory.

tegefaulkes (changes requested):
Some small changes.

src/client/handlers/VaultsSecretsRemove.ts
aryanjassal: No other command does any checking for ensuring the `secretName` (or any other relevant parameter) matches the path format or validating the type. I have specified types in the actual parameter, so those types should be upheld.

I should be able to add type checking, but I will probably need to write a utility function for validation of the path formatting, and I'm not sure we should be doing that for this PR.

tests/vaults/VaultOps.test.ts
tegefaulkes, line 1: You need two new tests here.
1. deleting multiple secrets at once.
2. When deleting multiple secrets, only new log message is added for the vault.

tegefaulkes (changes requested):
Looking good, one last thing to add.

tegefaulkes (approved):

tegefaulkes:
Looking good now. Prep it for merge but I wouldn't merge it until the CLI side is done and ready to merge as well.

CMCDragonkai:
Even when removing a secret, it will exist in the vault history which can be checked out.

aryanjassal:
> Even when removing a secret, it will exist in the vault history which can be checked out.

Yes, I believe this is correct. Do I need to remove the file from vault history too?

If the user has access to a secret, they can always just copy the secret and back it up somewhere else, so removing the file from vault history doesn't make much sense to me.

CMCDragonkai:
No.

aryanjassal:
https://github.com/MatrixAI/Polykey-CLI/pull/279#discussion_r1758081601

aryanjassal:
Everything for this PR has been done and approved. Merging.

PR #804, opened 2024-09-10
Implementing network authentication and segregation logic to the nodes domain

### LINKED ISSUE #782: Prevent RPC calls for Unauthenticated `NodeConnection`s (Segregated Network Connections)
### Specification

In order to segregate nodes of different networks using the `ClaimNetworkAccess` tokens defined in https://github.com/MatrixAI/Polykey/issues/779, there needs to be some logic for nodes to prevent accepting RPC calls from nodes that are out-of-network. There will need to be some RPC calls that are whitelisted to enable some level of out of network negotiation such as authenticating into the network or requesting access.

Currently a `NodeConnection` has two stages, the creating stage when you are awaiting the `createNodeConnection` factory. And the connected stage when the `NodeConnection` fully connects and the creation resolve. The problem with this is we need to be able to authenticate without allowing most RPC calls and nominal traffic.

To this end we need a connected but unauthenticated state where the `NodeConnection` is fully operational but not entertaining RPC calls by rejecting them. The `NodeConnection.createNodeConnection` should negotiate it's access to the network before resolving with the completed connection. This means the connecting state and authenticating state is hidden inside the creation of the `NodeConnection`. This will mean minimal changes to how the `NodeConnectionManager` handles these connections. However, the `NodeConnection` will need separate connected, authenticated and created events to reflect these stages. 

1. Connection event should be recorded by the audit domain as a connection
2. Authenticated event should be recorded by the audit domain as well. Probably a new authenticated event. There will be a distinct forward and reverse authentication event.
3. Not actually sure about creation, fills the same role as authenticated so subject to discussion.

While in the authenticating state, all non whitelisted RPC calls need to be rejected outright. That said, the NodeConnection shouldn't be available to make these calls at this stage but we still need to be secure about it. To avoid having to add logic to all of the RPC handlers we should apply this connection rejection logic to the middleware. The middleware will need to refer back to the `NodeConnection` somehow to assess the authenticated state.

To authenticate the `NodeConnection` needs to make an authentication RPC call and provide a valid network token. It's up to the handler of this to decide to reject the authentication with an error and kill the `NodeConnection` if it fails. Note that this needs to be symmetric in the forward and reverse direction. BOTH sides of the connection need to fully authenticate. This opens us to annoying race conditions so we need to be extra careful here.  The handshake has been described within #779.

It is extremely important that the following conditions are met.

1. A `NodeConnection` is only fully created if it fully authenticates.
2. The `NodeConnection` is only added to the `NodeConnectionManager`'s connection map if it fully authenticates.
3. The connection details are only added to the `NodeGraph` if it fully authenticates.
4. No non whitelisted RPC calls are allowed or made during the `Authenticating` state.

### Additional context

Related: #779 - Defines the network access tokens and how they are verified.
Related: #770 - Parent issue

### Tasks
1. Add an authentication state to the `NodeConnection`.
2. Add middleware logic to RPC to only allow non whitelisted RPC calls to be handled if we are in the authenticated state. Otherwise kill the stream involved.
3. Add an RPC handler authentication the connections access to the network. This will resolve with no message, or throw an error for why it failed to authenticate. This should be whitelisted.
4. Review and expand on the creation events to include separate connection, authentication and creation events.
5. `NodeConnection` must switch to the `authenticated` state and fully create only after both the authentication handler and call succeeds and resolves.
6. We need to make sure timeouts work here. There are two levels of timeouts. It's up to the handler side to kill the connection if it fails, but the calling side can timeout and kill the connection as a fail-safe. We also need an overall timeout for the whole process so if it fails to authenticate fully before the timeout then we just give up and kill the connection.

tegefaulkes:
### Description

This PR focuses on adding logic to the nodes domain to handle checking if a connection is part of a certain network. If the check passes then the connection is added to the usual connection map. If it fails the connection is rejected and forced to close. 

During this authentication period, only certain RPC calls are allowed to be made. These are calls that...
1. Are part of the network authentication process.
2. Are used to negotiate joining a network.

Both sides of the connection need to authenticate each other before the connection can be used for normal RPC traffic.

### Issues Fixed

* Fixes #782 
* Related #779 

### Tasks

- [x] 1. Add a new map that tracks connections that have not authenticated their network access yet. Add in the logic for negotiating authentication to the network. And only move the connection from this map to the usual connection map once the connection's network access has been checked in both directions. It may be simpler to only track this for the reverse direction.
- [x] 2. Create a middleware utility that rejects RPC calls from being made in the forward and reverse direction unless the connection has passed the network authentication.
- [x] 3. create tests showcasing this new functionality. both the normal case and connections being rejected. 3 types of rejection, forward rejected, reverse rejected and both.
- [x] 4. Update existing tests to work with the logic change.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -386,6 +386,14 @@ class PolykeyAgent {
           optionsDefaulted.nodes.connectionInitialMaxStreamsBidi,
         rpcParserBufferSize: optionsDefaulted.nodes.rpcParserBufferSize,
         rpcCallTimeoutTime: optionsDefaulted.nodes.rpcCallTimeoutTime,
+        authenticateNetworkForwardCallback:
+          nodesUtils.nodesAuthenticateConnectionForwardBasicPublicFactory(
+            optionsDefaulted.network,
+          ),
+        authenticateNetworkReverseCallback:
+          nodesUtils.nodesAuthenticateConnectionReverseBasicPublicFactory(
+            optionsDefaulted.network,
+          ),
         logger: logger.getChild(NodeConnectionManager.name),
       });
       nodeManager = new NodeManager({
--- a/src/client/ClientService.ts
+++ b/src/client/ClientService.ts
@@ -34,8 +34,9 @@ class ClientService {
     const conn = evt.detail;
     const streamHandler = (evt: wsEvents.EventWebSocketConnectionStream) => {
       const stream = evt.detail;
+      // If the RPCServer is stopping or stopped then we want to reject new streams outright
       if (!this.rpcServer[running] || this.rpcServer[status] === 'stopping') {
-        stream.cancel(Error('TMP RPCServer not running'));
+        stream.cancel(new errors.ErrorClientServiceNotRunning());
         return;
       }
       this.rpcServer.handleStream(stream);
--- a/src/nodes/NodeConnectionManager.ts
+++ b/src/nodes/NodeConnectionManager.ts
@@ -1,6 +1,15 @@
 import type { ResourceAcquire } from '@matrixai/resources';
 import type { ContextTimed, ContextTimedInput } from '@matrixai/contexts';
 import type { QUICConnection } from '@matrixai/quic';
+import type { JSONRPCRequest, JSONRPCResponse } from '@matrixai/rpc';
+import type {
+  AuthenticateNetworkForwardCallback,
+  AuthenticateNetworkReverseCallback,
+  NodeId,
+  NodeIdString,
+} from './types';
+import type { NodesAuthenticateConnectionMessage } from './agent/types';
+import type { AgentServerManifest } from './agent/handlers';
 import type KeyRing from '../keys/KeyRing';
 import type { CertificatePEM } from '../keys/types';
 import type {
@@ -10,8 +19,8 @@ import type {
   Port,
   TLSConfig,
 } from '../network/types';
-import type { AgentServerManifest } from './agent/handlers';
-import type { NodeId, NodeIdString } from './types';
+import type { JSONValue } from '../types';
+import { TransformStream } from 'stream/web';
 import {
   events as quicEvents,
   QUICServer,
@@ -42,6 +51,7 @@ import agentClientManifest from './agent/callers';
 import * as nodesUtils from './utils';
 import * as nodesErrors from './errors';
 import * as nodesEvents from './events';
+import * as agentUtils from './agent/utils';
 import * as keysUtils from '../keys/utils';
 import * as networkUtils from '../network/utils';
 import * as utils from '../utils';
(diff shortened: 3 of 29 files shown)


tegefaulkes:
I'm refreshing my memory and digging through the code. It seems simple enough to add tracking the authentication to the `NodeConnectionManager`. The only real change is adding `authenticatedForward` and `authenticatedReverse` booleans to the `newConnAndTimer` object. On top of that, we should only update the `activeConnection` field of the `ConnectionsEntry` in the `connections` map once the authentication has been completed. 

So.
1. Add booleans for tracking the authentication state for the forward and reverse direction.
2. Only add or switch out the primary connection when the network authentication of a connection has completed.

There is one stumbling block however. Current there is an assumption that a connection that has been added with `addConnection` is already available to be used. But these changes will create a period of time where a connection exists but can't be used.

So when we create a new connection  we not only have to wait for the connection to be established but also wait for the network authentication to be completed before it can be used. So we need to update `createConnection` to wait for this condition before returning.

I'll need to think on this some more.

CMCDragonkai:
Status of this @tegefaulkes?

tegefaulkes:
Still no progress with me focusing on PKE.

CMCDragonkai:
Reuqires rebase. And ETA on merging this?

CMCDragonkai:
As a suggestion, don't go too deep in the technicalities of this, until we have a PKE prototype of the entire login process. I want to see `polykey agent start --network matrixai.enterprise.polykey.com` trigger a browser launch and a login process. It can be flaky, but we need to go through the entire LOOP. All the way to the bottom.

tegefaulkes:
If I can focus solely on this? 2-3 days. The core of the implementation is done I just need to do some edge case logic and clean up. The main thing missing right now is the timeout for authenticating and testing.

Right now I'm focusing on prototyping the real time data for PKE and specing out work for Matthew.

CMCDragonkai:
Should start squashing.

tegefaulkes:
Pretty much done now. I've done a personal review and I just need to do the final checklist before merging.

With the network authentication, I've set it up so that it works with the public networks. So nodes will only connect to nodes within the same network identified by the network URL. This isn't verified in any way besides just checking what network a node claims to be in. This will be specified by the `network` parameter when starting a Polykey node. Which is set by the `--network` parameter in CLI.

tegefaulkes:
Done and merging.

CMCDragonkai:
> Pretty much done now. I've done a personal review and I just need to do the final checklist before merging.
> 
> With the network authentication, I've set it up so that it works with the public networks. So nodes will only connect to nodes within the same network identified by the network URL. This isn't verified in any way besides just checking what network a node claims to be in. This will be specified by the `network` parameter when starting a Polykey node. Which is set by the `--network` parameter in CLI.

For the public networks this is all that is required since they are public. They are self-declared.

CMCDragonkai:
@aryanjassal can we verify this is working by monitoring the node connections. Testnet and mainnet shouldn't be connecting to each other.

CMCDragonkai:
Ideally they shouldn't even try to connect to each other as the node discovery algos should ignore telling the other node about them.

PR #805, opened 2024-09-13
Updating behaviour of `VaultsSecretsGet` to align with requirements of `secrets cat`

aryanjassal:
### Description

The current `VaultsSecretsGet` only gets a single secret from the vault, and returns it in the form of a `UnaryHandler`, meaning for large files, the RPC call will timeout. It also fetches only one file at a time, so behaviour like UNIX's `cat` command isn't possible.

This PR aims to fix that issue by switching over the RPC handler to a `ServerHandler`, supporting larger files. This PR also adds support for listing the contents of multiple files in order, like what `cat` does.

### Issues Fixed

* Relates to https://github.com/MatrixAI/Polykey-CLI/issues/243
* Closes ENG-356 ([link](https://linear.app/matrix-ai/issue/ENG-356/implement-secrets-cat-command)) 

### Tasks

- [x] 1. Switch `UnaryHandler` to `StreamHandler`
- [x] 2. Implement support for displaying contents of multiple files
- [x] 3. Implement support for listing multiple files across multiple vaults
- [x] 4. Update old test and add new tests

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsSecretsGet.ts
+++ b/src/client/handlers/VaultsSecretsGet.ts
@@ -2,47 +2,71 @@ import type { DB } from '@matrixai/db';
 import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
-  ContentMessage,
+  ContentWithErrorMessage,
   SecretIdentifierMessage,
 } from '../types';
 import type VaultManager from '../../vaults/VaultManager';
-import { UnaryHandler } from '@matrixai/rpc';
+import { DuplexHandler } from '@matrixai/rpc';
 import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
 import * as vaultOps from '../../vaults/VaultOps';
 
-class VaultsSecretsGet extends UnaryHandler<
+class VaultsSecretsGet extends DuplexHandler<
   {
     vaultManager: VaultManager;
     db: DB;
   },
   ClientRPCRequestParams<SecretIdentifierMessage>,
-  ClientRPCResponseResult<ContentMessage>
+  ClientRPCResponseResult<ContentWithErrorMessage>
 > {
-  public handle = async (
-    input: ClientRPCRequestParams<SecretIdentifierMessage>,
-  ): Promise<ClientRPCResponseResult<ContentMessage>> => {
+  public handle = async function* (
+    input: AsyncIterable<ClientRPCRequestParams<SecretIdentifierMessage>>,
+    _cancel,
+    _meta,
+    ctx,
+  ): AsyncGenerator<ClientRPCResponseResult<ContentWithErrorMessage>> {
+    if (ctx.signal.aborted) throw ctx.signal.reason;
     const { vaultManager, db } = this.container;
-    return await db.withTransactionF(async (tran) => {
-      const vaultIdFromName = await vaultManager.getVaultId(
-        input.nameOrId,
-        tran,
-      );
-      const vaultId =
-        vaultIdFromName ?? vaultsUtils.decodeVaultId(input.nameOrId);
-      if (vaultId == null) {
-        throw new vaultsErrors.ErrorVaultsVaultUndefined();
+    yield* db.withTransactionG(async function* (tran): AsyncGenerator<
+      ClientRPCResponseResult<ContentWithErrorMessage>
+    > {
+      if (ctx.signal.aborted) throw ctx.signal.reason;
+      // As we need to preserve the order of parameters, we need to loop over
+      // them individually, as grouping them would make them go out of order.
+      let metadata: any = undefined;
+      for await (const secretIdentiferMessage of input) {
+        if (ctx.signal.aborted) throw ctx.signal.reason;
+        if (metadata == null) metadata = secretIdentiferMessage.metadata ?? {};
+        const { nameOrId, secretName } = secretIdentiferMessage;
+        const vaultIdFromName = await vaultManager.getVaultId(nameOrId, tran);
+        const vaultId = vaultIdFromName ?? vaultsUtils.decodeVaultId(nameOrId);
+        if (vaultId == null) throw new vaultsErrors.ErrorVaultsVaultUndefined();
+        yield await vaultManager.withVaults(
+          [vaultId],
+          async (vault) => {
+            try {
+              const content = await vaultOps.getSecret(vault, secretName);
+              return { secretContent: content.toString('binary') };
+            } catch (e) {
+              if (metadata?.options?.continueOnError === true) {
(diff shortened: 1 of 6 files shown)


aryanjassal:
I have implemented basic functionality for displaying file contents for multiple files, and updated the tests in `tests/client/handlers/vaults.test.ts`.

I still need to review the issue spec and all the code to ensure everything's ready for review.

aryanjassal:
https://github.com/MatrixAI/Polykey-CLI/issues/243#issuecomment-2348535798

I'm going to go ahead with the assumption that we need to implement support for listing contents for files from multiple vaults concurrently, as that makes sense and retains consistency with UNIX commands and `secrets rm`.

I will wait for review of my approach from `secrets rm`, and if my implementation seems correct, I will implement that in `secrets cat` too.

aryanjassal:
The naming of secret commands has changed from what it was. The intention was to keep using the pattern of `SecretsGet`, `SecretsPut`, `SecretsDel`, but to better align with the new commands, these names have been changed to `SecretsGet` (I haven't changed this name yet), `SecretsEdit`, `SecretsRemove`.

We are kinda moving away from the pattern. This was hard to detect as this change was made one-at-a-time across multiple PRs. We might need to look into this and set a standard pattern to follow for ensuring consistency.

src/client/handlers/VaultsSecretsGet.ts
```diff
@@ -3,48 +3,51 @@ import type {
...
+        const vaultId = vaultIdFromName ?? vaultsUtils.decodeVaultId(vaultName);
+        if (vaultId == null) throw new vaultsErrors.ErrorVaultsVaultUndefined();
+        yield* vaultManager.withVaultsG(
+          [vaultId],
+          async function* (vault): AsyncGenerator<ContentMessage, void, void> {
+            yield {
+              secretContent: (
+                await vaultOps.getSecret(vault, secretName)
+              ).toString('binary'),
+            };
```
tegefaulkes: Using a generator here isn't necessary since we only get one file and all of it at once. Since we're not implementing streaming here yet you can just use a `withVaultsF` instead.

src/client/types.ts
```diff
@@ -304,10 +304,15 @@ type SecretPathMessage = {
   secretName: string;
 };
 
+// The secrets must be in a format like:
+// [ [vaultName, secretPath], [vaultName, secretPath ] ]
+type SecretManyPathMessage = {
+  secretNames: Array<Array<string>>;
+};
```
tegefaulkes: A type of `Array<Array<string>>` is very confusing and doesn't convey the data properly. We want to be consistent with how other commands work. For example the SecretsEnv command functions in a similar way where we get the contents of a bunch of secrets. So we want to mimic how that works.

In that case we just need to use the `SecretIdentifierMessage` as stream them over. So the cat RPC call needs to be a duplex stream. The `VaultsSecretsEnv.ts` already does this. We just need a stripped down version of it for `secrets cat`.

tegefaulkes (changes requested):

src/client/types.ts
aryanjassal: I've updated both `VaultsSecretsGet` and `VaultsSecretsRemove` to allow streaming of file paths, and also updated the tests accordingly.

src/client/handlers/VaultsSecretsGet.ts
aryanjassal: I have changed the `withVaultsG` to simply use `withVaults` to get the secret contents. I now yield the fetched contents outside the `withVaults` block.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -3,34 +3,46 @@ import type {
...
+    let metadata: any = {};
+    let first = true;
+    for await (const secretRemoveMessage of input) {
+      if (first) metadata = secretRemoveMessage.metadata;
+      secretNames.push([
+        secretRemoveMessage.nameOrId,
+        secretRemoveMessage.secretName,
+      ]);
+      first = false;
+    }
```
tegefaulkes: You don't need a boolean here, you can use the fact that `metadata` is defined to deal with it.

`if (metadata == null) metadata = secretRemoveMessage.metadata ?? {};`

tests/client/handlers/vaults.test.ts
```diff
@@ -1462,163 +1462,290 @@ describe('vaultsSecretsNew and vaultsSecretsDelete, vaultsSecretsGet', () => {
...
+        });
+      }
+      await writer.close();
+    })();
+    const secretContent: Array<string> = [];
+    for await (const data of getStream.readable) {
+      secretContent.push(data.secretContent);
+    }
+    expect(secretContent.join('')).toStrictEqual(secretNames.join(''));
+  });
```
tegefaulkes: name of the test doesn't make sense here. We use `vaultsSecretGet` within the new `cat` command. but that handler doesn't do the cat action. So you want to make sure that tests have clear names explaining exactly what you're testing. In this case we're testing if you can get multiple secrets at once, theyre in order and correct.

tegefaulkes (commented):
Small changes, just check your tests are only testing the specific condition we want.

aryanjassal:
While I am writing the tests, I can see that we use RPC calls everywhere, even when we don't need to. For example, if we are testing `vaultsSecretsGet` and `vaultsSecretsRemove`, then we will have RPC handlers for `vaultsSecretsNew`, `vaultsSecretsMkdir`, `vaultsSecretsStat`, etc. as other RPC handlers we use to help perform vault operations in the tests.

This introduces slowdowns as each RPC call takes time to complete. A faster method of doing the same would be to acquire a vault, and perform `writeF` on the vault filesystem directly. It should speed up all our tests.

@tegefaulkes and I had a chat about this, and he was also of the same opinion as I, but he mentioned how all the tests would need to be updated to ensure consistency.

How should I deal with this? Do I make a new issue tracking this? It should be relatively straightforward, so maybe this can be something @brynblack could attempt once she gets into Polykey Core development.

tegefaulkes:
It's some thing we can address later. I think we should make a new issue for it. It should be a general review and refactor of the vaults domain tests. I think one for the CLI and one for Polykey. You mentioned that the vaults tests is one big test file. We should look into splitting that out again but we need to check if it still crashes the CI. The other part is dealing with the `vaultOps.test.ts` testfile. Currently its the longest running tests by far so if we can speed it up or split it into parts that would help as well.

tegefaulkes (approved):

tegefaulkes:
Looks good to me.

aryanjassal:
All tasks have been completed for this PR, all checks are passing, and reviews has also been approved. Merging.

PR #814, opened 2024-09-27
Implement handler to wrap `fs.writeFile` functionality

aryanjassal:
### Description

Currently, the `VaultsSecretsEdit` handler uses `vaultOps.updateSecret()` to update a secret's contents if it exists. Otherwise, it throws an error. This behaviour convolutes the implementation for `secrets edit` and `secrets write` as a lot of needless checks are run at the client side to work around a small oversight.

As such, this PR will aim on laying the groundwork for `secrets write` by creating a handler which will write secret contents to a secret. If the secret doesn't exist, it will automatically get created.

### Issues Fixed

* Relates to <b>[Polykey-CLI#242](https://github.com/MatrixAI/Polykey-CLI/issues/242)</b>
* REF ENG-355

### Tasks

- [x] 1. Remove `VaultsSecretsEdit`
- [x] 2. Add `VaultsSecretsWriteFile`
- [x] 3. Create `vaultOps.writeSecret` as a wrapper around `fs.writeFile`
- [x] 4. Make tests for `VaultsSecretsWriteFile`
- [x] 5. Make tests for `vaultOps.writeSecret`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


aryanjassal:
Currently, we are using a `UnaryHandler` to update secrets. For consistency, this behaviour needs to be replicated over to the `VaultsSecretsWriteFile` handler too.

However, `UnaryHandler` is inherently not a good choice for streaming file contents. A better choice would be a `RawStream`, but it needs more work and discussions to set up. As such, it will be done in later iterations. For now, the new handler will keep using `UnaryHandler` but use the new `vaultOps` function.

src/client/handlers/VaultsSecretsWriteFile.ts
```diff
@@ -11,7 +11,7 @@ import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
 import * as vaultOps from '../../vaults/VaultOps';
```
aryanjassal, line 12: This line imports everything from `vaultOps`, but we are only using `writeSecret`. We are needlessly importing things we don't need. While this is consistent with other handlers, and I understand that this is putting `vaultOps` functions behind its namespace, I still think that it is unnecessary. Should I not do that and only import the singular function I need, or do I keep it as-is?

tegefaulkes: We usually just import the whole namespace unless we're importing types. Just keep it consistent. AFAIK it doesn't make a difference as far as importing works.

tegefaulkes (approved):
I skimmed through it and nothing stands out. Looks good.

aryanjassal:
Everything for this PR has been done and approval has also been granted. Merging.

PR #819, opened 2024-10-01
Make VaultsSecretsMkdir support specifying multiple secret paths

aryanjassal:
### Description

Following the behaviour of UNIX's `mkdir`, the behaviour of `VaultsSecretsMkdir` is being tweaked to function more seamlessly.

The `VaultsSecretsMkdir` handler now supports creating multiple directories in multiple vaults, all in a single commit message. If a directory fails to be created, the whole command will no longer fail, merely returning an error.

### Issues Fixed

* Relates to <b>[Polykey-CLI#246](https://github.com/MatrixAI/Polykey-CLI/issues/246)</b>
* REF ENG-359

### Tasks

- [x] 1. Make new vaultOps function for making directories without returning on error
- [x] 2. Update `VaultsSecretsMkdir` handler to make multiple directories across multiple vaults in a single vaults log commit
- [x] ~~3. Allow serialisation and deserialisation of errors~~ We are returning plain error strings. We won't be serialising and deserialising errors.
- [x] 4. Merge behaviour of `vaultOps.makeDirectories` into `vaultOps.mkdir`
- [x] 5. Update tests for `vaultOps.mkdir`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsSecretsMkdir.ts
+++ b/src/client/handlers/VaultsSecretsMkdir.ts
@@ -2,50 +2,61 @@ import type { DB } from '@matrixai/db';
 import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
-  SecretMkdirMessage,
-  SuccessMessage,
+  SecretDirMessage,
+  SuccessOrErrorMessage,
 } from '../types';
 import type VaultManager from '../../vaults/VaultManager';
-import { UnaryHandler } from '@matrixai/rpc';
+import type { POJO } from '../../types';
+import { DuplexHandler } from '@matrixai/rpc';
 import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
 import * as vaultOps from '../../vaults/VaultOps';
 
-class VaultsSecretsMkdir extends UnaryHandler<
+class VaultsSecretsMkdir extends DuplexHandler<
   {
-    vaultManager: VaultManager;
     db: DB;
+    vaultManager: VaultManager;
   },
-  ClientRPCRequestParams<SecretMkdirMessage>,
-  ClientRPCResponseResult<SuccessMessage>
+  ClientRPCRequestParams<SecretDirMessage>,
+  ClientRPCResponseResult<SuccessOrErrorMessage>
 > {
-  public handle = async (
-    input: ClientRPCRequestParams<SecretMkdirMessage>,
-  ): Promise<ClientRPCResponseResult<SuccessMessage>> => {
-    const { vaultManager, db } = this.container;
-    await db.withTransactionF(async (tran) => {
-      const vaultIdFromName = await vaultManager.getVaultId(
-        input.nameOrId,
-        tran,
-      );
-      const vaultId =
-        vaultIdFromName ?? vaultsUtils.decodeVaultId(input.nameOrId);
-      if (vaultId == null) {
-        throw new vaultsErrors.ErrorVaultsVaultUndefined();
-      }
-      await vaultManager.withVaults(
-        [vaultId],
-        async (vault) => {
-          await vaultOps.mkdir(vault, input.dirName, {
-            recursive: input.recursive,
-          });
-        },
-        tran,
-      );
-    });
-    return {
-      success: true,
-    };
+  public handle = async function* (
+    input: AsyncIterable<ClientRPCRequestParams<SecretDirMessage>>,
+  ): AsyncGenerator<ClientRPCResponseResult<SuccessOrErrorMessage>> {
+    const { db, vaultManager }: { db: DB; vaultManager: VaultManager } =
+      this.container;
+    let metadata: POJO;
+    yield* db.withTransactionG(
+      async function* (tran): AsyncGenerator<SuccessOrErrorMessage> {
+        for await (const secretDirMessage of input) {
+          // Unpack input
+          if (metadata == null) metadata = secretDirMessage.metadata ?? {};
+          const nameOrId = secretDirMessage.nameOrId;
+          const dirName = secretDirMessage.dirName;
+          // Get vaultId
+          const vaultIdFromName = await vaultManager.getVaultId(nameOrId, tran);
+          const vaultId =
+            vaultIdFromName ?? vaultsUtils.decodeVaultId(nameOrId);
+          if (vaultId == null) {
+            throw new vaultsErrors.ErrorVaultsVaultUndefined();
+          }
+          // Write directories. This doesn't need to be grouped by vault names,
+          // as no commit is created for empty directories anyways. The
+          // vaultOps.mkdir() method also returns an object of type
+          // SuccessOrErrorMessage. As such, we can return the result without
+          // doing any type conversion or extra processing.
+          yield await vaultManager.withVaults(
+            [vaultId],
+            async (vault) => {
+              return await vaultOps.mkdir(vault, dirName, {
+                recursive: metadata?.options?.recursive,
+              });
+            },
+            tran,
+          );
+        }
+      },
+    );
   };
 }
 
--- a/src/vaults/VaultOps.ts
+++ b/src/vaults/VaultOps.ts
@@ -4,6 +4,7 @@
 import type Logger from '@matrixai/logger';
 import type { Vault } from './Vault';
 import type { Stat } from 'encryptedfs';
+import type { SuccessOrErrorMessage } from '../client/types';
 import path from 'path';
 import * as vaultsErrors from './errors';
 import * as vaultsUtils from './utils';
@@ -173,29 +174,35 @@ async function mkdir(
   dirPath: string,
   fileOptions?: FileOptions,
   logger?: Logger,
-): Promise<void> {
+): Promise<SuccessOrErrorMessage> {
   const recursive = fileOptions?.recursive ?? false;
-
-  await vault.writeF(async (efs) => {
-    try {
+  // Technically, writing an empty directory won't make a commit, and doesn't
+  // need a write resource as git doesn't track empty directories. It is
+  // still being used to allow concurrency.
+  try {
+    await vault.writeF(async (efs) => {
       await efs.mkdir(dirPath, fileOptions);
-    } catch (e) {
-      if (e.code === 'ENOENT' && !recursive) {
-        throw new vaultsErrors.ErrorVaultsRecursive(
-          `Could not create directory '${dirPath}' without recursive option`,
-          { cause: e },
-        );
-      }
-      if (e.code === 'EEXIST') {
-        throw new vaultsErrors.ErrorSecretsSecretDefined(
-          `${dirPath} already exists`,
-          { cause: e },
-        );
-      }
-      throw e;
+      logger?.info(`Created secret directory at '${dirPath}'`);
+    });
+    return { type: 'success', success: true };
+  } catch (e) {
+    logger?.error(`Failed to create directory '${dirPath}'. Reason: ${e.code}`);
+    if (e.code === 'ENOENT' && !recursive) {
+      return {
+        type: 'error',
+        code: e.code,
+        reason: dirPath,
+      };
     }
-    logger?.info(`Created secret directory at '${dirPath}'`);
-  });
+    if (e.code === 'EEXIST') {
+      return {
+        type: 'error',
+        code: e.code,
+        reason: dirPath,
+      };
+    }
+    throw e;
+  }
 }
 
 /**
--- a/tests/client/handlers/vaults.test.ts
+++ b/tests/client/handlers/vaults.test.ts
@@ -3,6 +3,7 @@ import type { FileSystem } from '@/types';
 import type { VaultId } from '@/ids';
 import type NodeManager from '@/nodes/NodeManager';
 import type {
+  ErrorMessage,
   LogEntryMessage,
   SecretContentMessage,
   VaultListMessage,
@@ -1341,15 +1342,153 @@ describe('vaultsSecretsMkdir', () => {
     const vaultName = 'test-vault';
     const vaultId = await vaultManager.createVault(vaultName);
     const dirPath = 'dir/dir1/dir2';
-    const response = await rpcClient.methods.vaultsSecretsMkdir({
-      recursive: true,
+    const response = await rpcClient.methods.vaultsSecretsMkdir();
+    const writer = response.writable.getWriter();
+    await writer.write({
       nameOrId: vaultsUtils.encodeVaultId(vaultId),
       dirName: dirPath,
+      metadata: { options: { recursive: true } },
     });
-    expect(response.success).toBeTruthy();
+    await writer.close();
+
+    for await (const data of response.readable) {
+      expect(data.type).toEqual('success');
+    }
     await vaultManager.withVaults([vaultId], async (vault) => {
       await vault.readF(async (efs) => {
         expect(await efs.exists(dirPath)).toBeTruthy();
+      });
+    });
+  });
+  test('fails to make directories without recursive', async () => {
+    const vaultName = 'test-vault';
+    const vaultId = await vaultManager.createVault(vaultName);
+    const encodeVaultId = vaultsUtils.encodeVaultId(vaultId);
+    const dirPath = 'dir/dir1/dir2';
+    const response = await rpcClient.methods.vaultsSecretsMkdir();
+    const writer = response.writable.getWriter();
+    await writer.write({ nameOrId: encodeVaultId, dirName: dirPath });
+    await writer.close();
+    for await (const data of response.readable) {
+      expect(data.type).toEqual('error');
+      // TS cannot properly evaluate a type as nested as this, so we use the
+      // as keyword to help it. Inside this block, the type of data is 'error'.
+      const error = data as ErrorMessage;
+      expect(error.code).toEqual('ENOENT');
+      expect(error.reason).toEqual(dirPath);
+    }
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.readF(async (efs) => {
+        expect(await efs.exists(dirPath)).toBeFalsy();
+      });
+    });
+  });
+  test('makes directories across multiple vaults', async () => {
+    const vaultName1 = 'test-vault1';
+    const vaultName2 = 'test-vault2';
+    const vaultId1 = await vaultManager.createVault(vaultName1);
+    const vaultId2 = await vaultManager.createVault(vaultName2);
+    const vaultIdEncoded1 = vaultsUtils.encodeVaultId(vaultId1);
+    const vaultIdEncoded2 = vaultsUtils.encodeVaultId(vaultId2);
+    const dirPath1 = 'dir-1';
+    const dirPath2 = 'dir-2';
+    const dirPath3 = 'dir-3';
+    // Attempt to make directories
+    const response = await rpcClient.methods.vaultsSecretsMkdir();
+    const writer = response.writable.getWriter();
+    await writer.write({
+      nameOrId: vaultIdEncoded1,
+      dirName: dirPath1,
+      metadata: { options: { recursive: true } },
+    });
+    await writer.write({ nameOrId: vaultIdEncoded2, dirName: dirPath2 });
+    await writer.write({ nameOrId: vaultIdEncoded1, dirName: dirPath3 });
+    await writer.close();
+    // Check if the operation concluded as expected
+    for await (const data of response.readable) {
+      expect(data.type).toEqual('success');
+    }
+    await vaultManager.withVaults(
+      [vaultId1, vaultId2],
+      async (vault1, vault2) => {
+        await vault1.readF(async (efs) => {
+          expect(await efs.exists(dirPath1)).toBeTruthy();
+          expect(await efs.exists(dirPath3)).toBeTruthy();
+        });
+        await vault2.readF(async (efs) => {
+          expect(await efs.exists(dirPath2)).toBeTruthy();
+        });
+      },
+    );
+  });
+  test('continues on error', async () => {
+    const vaultName1 = 'test-vault1';
+    const vaultName2 = 'test-vault2';
+    const vaultId1 = await vaultManager.createVault(vaultName1);
+    const vaultId2 = await vaultManager.createVault(vaultName2);
+    const vaultIdEncoded1 = vaultsUtils.encodeVaultId(vaultId1);
+    const vaultIdEncoded2 = vaultsUtils.encodeVaultId(vaultId2);
+    const dirPath1 = 'dir-1';
+    const dirPath2 = 'dir-2';
+    const dirPath3 = 'nodir/dir-3';
+    // Attempt to make directories
+    const response = await rpcClient.methods.vaultsSecretsMkdir();
+    const writer = response.writable.getWriter();
+    await writer.write({ nameOrId: vaultIdEncoded1, dirName: dirPath1 });
+    await writer.write({ nameOrId: vaultIdEncoded2, dirName: dirPath2 });
+    await writer.write({ nameOrId: vaultIdEncoded1, dirName: dirPath3 });
+    await writer.close();
+    // Check if the operation concluded as expected
+    for await (const data of response.readable) {
+      if (data.type === 'error') {
+        // TS cannot properly evaluate a type as nested as this, so we use the
+        // as keyword to help it. Inside this block, the type of data is 'error'.
+        const error = data as ErrorMessage;
+        expect(error.code).toEqual('ENOENT');
+        expect(error.reason).toEqual(dirPath3);
+      }
+    }
+    await vaultManager.withVaults(
+      [vaultId1, vaultId2],
+      async (vault1, vault2) => {
+        await vault1.readF(async (efs) => {
+          expect(await efs.exists(dirPath1)).toBeTruthy();
+          expect(await efs.exists(dirPath3)).toBeFalsy();
+        });
+        await vault2.readF(async (efs) => {
+          expect(await efs.exists(dirPath2)).toBeTruthy();
+        });
+      },
+    );
+  });
+  test('fails if secret with same name exists', async () => {
+    const vaultName = 'test-vault';
+    const vaultId = await vaultManager.createVault(vaultName);
+    const vaultIdEncoded = vaultsUtils.encodeVaultId(vaultId);
+    const dirPath = 'secret-first';
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.writeF(async (efs) => {
+        await efs.writeFile(dirPath, dirPath);
+      });
+    });
+    // Attempt to make directory
+    const response = await rpcClient.methods.vaultsSecretsMkdir();
+    const writer = response.writable.getWriter();
+    await writer.write({ nameOrId: vaultIdEncoded, dirName: dirPath });
+    await writer.close();
+    // Check if the operation concluded as expected
+    for await (const data of response.readable) {
+      expect(data.type).toEqual('error');
+      // TS cannot properly evaluate a type as nested as this, so we use the
+      // as keyword to help it. Inside this block, the type of data is 'error'.
+      const error = data as ErrorMessage;
+      expect(error.code).toEqual('EEXIST');
+      expect(error.reason).toEqual(dirPath);
+    }
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.readF(async (efs) => {
+        const stat = await efs.stat(dirPath);
+        expect(stat.isFile).toBeTruthy();
       });
     });
   });
@@ -1377,35 +1516,45 @@ describe('vaultsSecretsNew and vaultsSecretsDelete, vaultsSecretsGet', () => {
   }>;
   let vaultManager: VaultManager;
   // Helper function to create secrets in a vault
-  const createVaultSecret = async (vaultName: string, secretName: string) => {
-    const createResponse = await rpcClient.methods.vaultsSecretsNew({
-      nameOrId: vaultName,
-      secretName: secretName,
-      secretContent: Buffer.from(secretName).toString('binary'),
+  const createVaultSecret = async (
+    vaultId: VaultId,
+    secretName: string,
+    content: string,
+  ) => {
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.writeF(async (efs) => {
+        await efs.writeFile(secretName, content);
+        expect(await efs.exists(secretName)).toBeTruthy();
+      });
     });
-    expect(createResponse.success).toBeTruthy();
   };
   // Helper function to ensure each file path was deleted
-  const checkSecretIsDeleted = async (
-    vaultName: string,
-    secretName: string,
-  ) => {
-    await testsUtils.expectRemoteError(
-      rpcClient.methods.vaultsSecretsStat({
-        nameOrId: vaultName,
-        secretName: secretName,
-      }),
-      vaultsErrors.ErrorSecretsSecretUndefined,
-    );
+  const checkSecretIsDeleted = async (vaultId: VaultId, secretName: string) => {
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.readF(async (efs) => {
+        expect(await efs.exists(secretName)).toBeFalsy();
+      });
+    });
   };
   // Helper function to ensure each file path exists in the vault
-  const checkSecretExists = async (vaultName: string, secretName: string) => {
-    await expect(
-      rpcClient.methods.vaultsSecretsStat({
-        nameOrId: vaultName,
-        secretName: secretName,
-      }),
-    ).toResolve();
+  const checkSecretExists = async (vaultId: VaultId, secretName: string) => {
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.readF(async (efs) => {
+        expect(await efs.exists(secretName)).toBeTruthy();
+      });
+    });
+  };
+  // Helper function to create a directory
+  const createVaultDir = async (
+    vaultId: VaultId,
+    dirName: string,
+    recursive: boolean = false,
+  ) => {
+    await vaultManager.withVaults([vaultId], async (vault) => {
+      await vault.writeF(async (efs) => {
+        await efs.mkdir(dirName, { recursive: recursive });
+      });
+    });
   };
   beforeEach(async () => {
     dataDir = await fs.promises.mkdtemp(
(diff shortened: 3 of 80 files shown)


src/vaults/VaultOps.ts
```diff
@@ -198,6 +206,36 @@ async function mkdir(
...
+        await efs.mkdir(dirPath, fileOptions);
+        yield { success: true };
+      } catch (e) {
+        if (e.code === 'ENOENT' && !recursive) {
+          yield { success: false, error: { code: 'ENOENT', path: dirPath } };
+        }
+        if (e.code === 'EEXIST') {
+          yield { success: false, error: { code: 'EEXIST', path: dirPath } };
+        }
+        throw e;
```
aryanjassal: I'm not sure how to handle catching and throwing errors without exiting the loop. So far, I've tried creating a `vaultsErrors` error object like how `mkdir` does but yielding the error instead, but that didn't work and I was unable to deserialise the object back into an actual error. It remained as a serialised JSON object and gave me errors when I was trying to access fields sometimes.

I've tried returning error strings directly, formatted exactly as they should be displayed on the client side, but this seems unoptimal, so I didn't go with that.

I need some help in figuring how to properly pass errors back to the caller without returning or throwing anything. @tegefaulkes

src/client/handlers/VaultsSecretsMkdir.ts
```diff
@@ -2,50 +2,77 @@ import type { DB } from '@matrixai/db';
...
+                recursive: metadata?.options?.recursive,
+              });
+              for await (const data of response) {
+                if (data.success) yield { success: true };
+                if (data.error != null) {
+                  yield {
+                    success: false,
+                    error: { code: data.error.code, path: data.error.path },
+                  };
+                }
```
aryanjassal: This current approach that I have highlighted actually either gets `read 0` or `ErrorRPCParse`. This is probably not even the right solution here, so I haven't put in much effort in getting this prototype error handler working.

src/vaults/VaultOps.ts
aryanjassal: The command `secrets touch` will have the same issues, as it also needs to do this. Previously, I literally returned the error we need to print, like in my second suggested method. After a solution is found, I will update the relevant handler (`VaultsSecretsGet`) to instead return properly serialised errors and not just the actual error strings. The error strings need to be the responsibility of the client, not the handler itself.

aryanjassal: After discussing this with Brian, the conclusion we have arrived at is to keep it as-is. Don't change the behaviour of returning a hard-coded string for `VaultsSecretGet` and implement the same for `VaultsSecretsMkdir`. We can update it later as needed, but we need to keep it simple and roll out features quickly.

tests/client/handlers/vaults.test.ts
```diff
@@ -1341,18 +1342,195 @@ describe('vaultsSecretsMkdir', () => {
...
+  //   await vaultManager.withVaults(
+  //     [vaultId1, vaultId2],
+  //     async (vault1, vault2) => {
+  //       const log1 = await vault1.log();
+  //       const log2 = await vault2.log();
+  //       expect(log1.length).toEqual(logLength1 + 1);
+  //       expect(log2.length).toEqual(logLength2 + 1);
+  //     },
+  //   );
+  // });
```
aryanjassal: I would assume that creating a directory would be modifying the vaults in some way, and would create a commit and increase the length of the log. However, I have noticed that the log length isn't changed, yet the directories still exist, weirdly enough.

I have looked at the code and confirmed that a commit is created after the callback function runs, but it clearly doesn't do anything in this case. If this shouldn't happen, I will keep on researching this issue. Otherwise, if this is intentional, I can just remove this test.

src/vaults/VaultOps.ts
```diff
@@ -168,33 +173,36 @@ async function deleteSecret(
...
+            error: `${e.code}: cannot create directory ${dirPath}: No such secret or directory`,
+          };
+        } else if (e.code === 'EEXIST') {
+          yield {
+            success: false,
+            error: `${e.code}: cannot create directory ${dirPath}: Secret or directory exists`,
+          };
+        } else {
+          throw e;
+        }
```
aryanjassal: The new `mkdir` function is an async generator and doesn't return a promise. All the tests treat `mkdir` as a promise. Do I somehow convert this into a promise, or keep it as-is, and rely on consuming values as the 'await' part of the method?

tests/client/handlers/vaults.test.ts
tegefaulkes: That's a bug. Dig into the `writeF` code of `VaultInternal` and see what's happening.  We should see a commit if a directory is created. The fact that is is created without a new commit is doubly weird since that means it's in a dirty state as well and likely to be removed when it attempts a clean up.

src/vaults/VaultOps.ts
tegefaulkes: I'm not sure what you mean. but if you converted `mkdir` from a `Promise` to an `AsyncGenerator` then you need to update everything that used it including the tests.

tests/client/handlers/vaults.test.ts
aryanjassal: https://chatgpt.com/share/e/66ff4177-c57c-800d-a89b-50c9597d155d

Apparently git does not keep track of empty directories that it creates. They will remain invisible until a file is created within it, which will force the directory to also be recognised by git.

So, this is not an issue with `writeF`, but an issue with how git works itself. The solution is to make a placeholder file inside the directory, forcing git to stage that directory.

tegefaulkes: This is possibly a problem that would allow a disconnect between the vault state and the clean vault history. We'll need to scope the problem out some more to see how it affects things. 

- Does the empty directory just exist within the vault? 
- What happens if we checkout a previous version? 
- Or checkout back to the latest version? 
- Does it get cloned when shared? 

Can we live with any of this? Or should we only allow directory creation when creating a new secret?

aryanjassal: > Does the empty directory just exist within the vault?

The empty directory exists within the vault file system. When running `pk secrets ls vault`, the newly created directory does show up in the output as a directory.

> What happens if we checkout a previous version?
> Or checkout back to the latest version?

If we checkout a previous version, the directory still exists in the vault fs. I tried rolling back to a previous version, and even back to the initial commit of the vault. The directory existed even there. This is most likely because the directory is outside git's version control tree, so it doesn't get deleted or modified when vault's version changes. Checking out to the latest version again yields a similar result of the empty directory being untouched.

> Does it get cloned when shared?

No, it doesn't. As the directory exists outside the git tree, it is not shared, and is not present when someone else is cloning the vault. I make a new local node, and tried to clone the vault with the empty directory.* It cloned everything but the empty directory. So, the empty directory only exists for the vault where it was initially created. 

This could pose issues where, if a vault's receiver manually made a similar directory in their vault, and the original directory got an update, then the incoming changes will conflict with the existing changes, causing all sorts of bad things.

\* The original vault which I use for testing purposes, `vault`, was unable to be shared via RPC cloning, even when both nodes were on the same device, and the vault only contains around 8 text files of a reasonable size. To increase performance, we may need to use the `tar` compression to zip the vault fs for streaming, and converting the RPC over to, say, `RawHandler`.

tegefaulkes: Yeah, I'm not a fan of this. There seems to be 3 options here. 

1. Just let there be a directory that's not part of the git tree. Not really good since its technically in a dirty vault state that we can't really manage.
2. Use a dummy special cased file in the directory when creating it. This would force the directory to be tracked by git but  we'd have to add a bunch of logic everywhere just to special case it and handle it. I really don't like it since it's prone to a buttload of edge cases and problems.
3. We scrap the `mkdir` command and only allow creating directories as part of creating secrets. So If I wanted a `newDir` i'd have to `secrets write vault:newdir/secret --recursive`, `secrets edit vault:newdir/secret --recursive` etc etc. This would be the best way to prevent the dirty vault state. `--recursive` would be needed because we generally don't implicitly create directories like this.

@CMCDragonkai Got any thoughts?

CMCDragonkai: Making an empty directory in git has no effect. This is part of the design of `git`.

It's an interesting conundrum here cause usually you'd create a directory and put some files in there AND then make the commit. But making a commit for just an empty directory is not possible.

So basically the issue is that empty directories basically do not exist as a concept in git, you may mutate the underlying vault and have these empty directories but they would not be committed to the log, nor would they be shared, if you have empty directories, it's as if they don't exist at all when you share the vault.

It's not a dirty vault state, it's not considered dirty at all. They are just invisible with respect to git and thus invisible with respect to git sharing/git log.

I think we keep the `mkdir`, but just that it doesn't by itself update the log. In that case you may need to skip the log commit system. It's an untracked invisible mutation basically. You keep it around, for as long as the the local vault exists, so I think solution 1. is the way to go.

I think you want to special case write a comment for this operation and indicate that empty are effectively non-existent in the vaults.

aryanjassal: > I think you want to special case write a comment for this operation and indicate that empty are effectively non-existent in the vaults.

I would say that the warning would be implemented on the CLI side, as the library itself shouldn't have this warning hard-coded. This must be up to the developers of the client.

As such, when I open a PR in Polykey CLI, I will have a reference to this comment, and add the warning/notice in that PR.

tests/vaults/VaultOps.test.ts
```diff
@@ -323,32 +323,86 @@ describe('VaultOps', () => {
...
       const dirPath = path.join(dirName, dirName);
-      await vaultOps.mkdir(vault, dirPath, { recursive: true });
+      const response = vaultOps.mkdir(vault, [dirPath], {
+        recursive: true,
+      });
+      // For await (const _ of response);
+      for await (const data of response) {
+        expect(data.success).toBeTruthy();
+        expect(data.error).toBeUndefined();
+      }
```
tegefaulkes: Errant comments here? If they're not needed then you should remove them.

tests/client/handlers/vaults.test.ts
aryanjassal: > It's not a dirty vault state, it's not considered dirty at all. They are just invisible with respect to git and thus invisible with respect to git sharing/git log.

The `VaultsSecretsRemove` handler can also modify the vault state in a way such that we can be left with an empty directory. Do I need to warn users in that command too, that empty directories will not be a part of the source tree but will show up in the vault's file system when performing operations like listing contents, etc.? @CMCDragonkai

tegefaulkes (commented):
besides that it looks fine. Though the handler is duplex but accumulates all the messages  to do the mkdir all at once. I would've thought we'd keep it streaming all the way through but I'm not sure it's worth changing here.

you could do the grouping on the CLI side and then we can have end to end streaming for this handler.

tests/vaults/VaultOps.test.ts
aryanjassal: Ah, oops.

aryanjassal:
> you could do the grouping on the CLI side and then we can have end to end streaming for this handler.

I believe this was discussed when we were talking about managing this for `secrets rm` command. There, you mentioned that you prefer having all the logic at one place, and I moved the grouping logic to the handler side.

If we want to update this, then we can, but then this would be inconsistent with the behaviour of `VaultsSecretsRemove`. The `DuplexHandler` also expects each streamed object to be "self-contained", so we would also need to pass the vault information using the metadata, which would be weird usage-wise. I already think that us using the metadata field for options like `recursive` is weird enough (as it is usually used for auth, and it provides no autocompletion/LSP integration), and using it for the vault name might not be best for development or maintenance given the lack of intrinsic documentation.

On the other hand, it would enable us to easily implement local file system integration if the grouping is done on CLI side, as the developer can detect the absence of a vault path, infer that as an operation on the user's file system, and perform the command on the local file system. With the current approach, we would need to filter paths which don't contain a vault path, then send over the rest which do, to be looped over in the handler instead for grouping.

While both approaches have their own advantages and disadvantages, we can work with either. We need to figure out a "standard" and stick with it for further commands.

tests/client/handlers/vaults.test.ts
CMCDragonkai: I think if our vaults operate like a regular git repo locally on a Unix system, then it's good enough. Basically people just need to realize that it's the same like a git repo.

aryanjassal: So, no need for a warning for other commands?

tegefaulkes:
By grouping I really mean sorting. So we apply the constraint that the streamed directory paths must be sorted so that means directories within a single vault is grouped together. Then we can easily create each directory for a single vault all at once within the same writeF. But all that is moot since creating directories doesn't actually create commits within the version history so it doesn't even need to be ordered.

`I believe this was discussed when we were talking about managing this for secrets rm command. There, you mentioned that you prefer having all the logic at one place, and I moved the grouping logic to the handler side.` That was more about avoiding composing RPC calls within the CLI command if possible. We prefer to do our stuff within a single RPC call but that doesn't mean we can't do *any* CLI side processing.

Also there's no reason to pass the vault information in the metadata. You can keep it in the message.

aryanjassal:
> But all that is moot since creating directories doesn't actually create commits within the version history so it doesn't even need to be ordered.

As such, I guess I can revert the changes I made to `vaultOps.mkdir` as ensuring a single write transaction isn't necessary anymore. This would make the command support end-to-end streaming.

Another concern I have is that on UNIX, the `touch` command also works like this, touching files in order that they were provided in. If we perform grouping, then it will break the constraint of UNIX's order and might yield unexpected behaviour. On the other hand, if we don't group the commands, then the vault will see multiple commits for creating multiple secrets, which isn't optimal. How should the `touch` command handle this?

> That was more about avoiding composing RPC calls within the CLI command if possible. We prefer to do our stuff within a single RPC call but that doesn't mean we can't do any CLI side processing.

I see. However, I still have the question about the "standard" for this. What do we follow? Grouping on the client side, and the handler expects the inputs to be sorted by vaults or doing it on the handler side as we currently are?

In my opinion, we should do the handling on the client side. This would make the most sense as it would allow the most freedom. For example, in the `touch` command, the user can specify multiple touched files within the same vault (which would be the use case for most users), and the sorted handler will group them together in a single `writeF` command. When the handler encounters a different vault name in the input stream, then it can start a new write transaction for the new vault. For example, this matches the behaviour of the UNIX commands very closely, which might be a good tradeoff between consistency and reducing commit redundancy.

```console
[aryanj@matrix-34xx:~]$ pk secrets touch v1:file1 v1:file2 v2:file3
 > making file1 and file2 in v1
 > making file3 in v2

[aryanj@matrix-34xx:~]$ pk secrets touch v1:file1 v1:file2 v2:file3 v1:file4
 > making file1 and file2 in v1
 > making file3 in v2
 > making file4 in v1
```

CMCDragonkai:
> > But all that is moot since creating directories doesn't actually create commits within the version history so it doesn't even need to be ordered.
> 
> As such, I guess I can revert the changes I made to `vaultOps.mkdir` as ensuring a single write transaction isn't necessary anymore. This would make the command support end-to-end streaming.
> 
> Another concern I have is that on UNIX, the `touch` command also works like this, touching files in order that they were provided in. If we perform grouping, then it will break the constraint of UNIX's order and might yield unexpected behaviour. On the other hand, if we don't group the commands, then the vault will see multiple commits for creating multiple secrets, which isn't optimal. How should the `touch` command handle this?
> 
> > That was more about avoiding composing RPC calls within the CLI command if possible. We prefer to do our stuff within a single RPC call but that doesn't mean we can't do any CLI side processing.
> 
> I see. However, I still have the question about the "standard" for this. What do we follow? Grouping on the client side, and the handler expects the inputs to be sorted by vaults or doing it on the handler side as we currently are?
> 
> In my opinion, we should do the handling on the client side. This would make the most sense as it would allow the most freedom. For example, in the `touch` command, the user can specify multiple touched files within the same vault (which would be the use case for most users), and the sorted handler will group them together in a single `writeF` command. When the handler encounters a different vault name in the input stream, then it can start a new write transaction for the new vault. For example, this matches the behaviour of the UNIX commands very closely, which might be a good tradeoff between consistency and reducing commit redundancy.
> 
> ```
> [aryanj@matrix-34xx:~]$ pk secrets touch v1:file1 v1:file2 v2:file3
>  > making file1 and file2 in v1
>  > making file3 in v2
> 
> [aryanj@matrix-34xx:~]$ pk secrets touch v1:file1 v1:file2 v2:file3 v1:file4
>  > making file1 and file2 in v1
>  > making file3 in v2
>  > making file4 in v1
> ```

Did you consider that CLI and agent RPC is concurrent. You have to test under the conditions that multiple concurrent calls are made. That's why transactions are still important sometimes.

CMCDragonkai:
> > That was more about avoiding composing RPC calls within the CLI command if possible. We prefer to do our stuff within a single RPC call but that doesn't mean we can't do any CLI side processing.

So there is a thing call promise pipelining which makes RPC composition more efficient.

We don't support it just yet https://capnproto.org/rpc.html, but it would be possible under js-rpc to do so. And https://gist.github.com/crabmusket/6a63eec5aa9d92649f66614f09777c6a

Each call right now opens up a stream, but the underlying transport is flexible enough to support this concept.

It also aligns with our ability to do RPC conversations efficiently.

Generally speaking this is kind of important for any sort of functional composition across RPC boundaries.

CMCDragonkai:
> > But all that is moot since creating directories doesn't actually create commits within the version history so it doesn't even need to be ordered.
> 
> As such, I guess I can revert the changes I made to `vaultOps.mkdir` as ensuring a single write transaction isn't necessary anymore. This would make the command support end-to-end streaming.
> 
> Another concern I have is that on UNIX, the `touch` command also works like this, touching files in order that they were provided in. If we perform grouping, then it will break the constraint of UNIX's order and might yield unexpected behaviour. On the other hand, if we don't group the commands, then the vault will see multiple commits for creating multiple secrets, which isn't optimal. How should the `touch` command handle this?
> 
> > That was more about avoiding composing RPC calls within the CLI command if possible. We prefer to do our stuff within a single RPC call but that doesn't mean we can't do any CLI side processing.
> 
> I see. However, I still have the question about the "standard" for this. What do we follow? Grouping on the client side, and the handler expects the inputs to be sorted by vaults or doing it on the handler side as we currently are?
> 
> In my opinion, we should do the handling on the client side. This would make the most sense as it would allow the most freedom. For example, in the `touch` command, the user can specify multiple touched files within the same vault (which would be the use case for most users), and the sorted handler will group them together in a single `writeF` command. When the handler encounters a different vault name in the input stream, then it can start a new write transaction for the new vault. For example, this matches the behaviour of the UNIX commands very closely, which might be a good tradeoff between consistency and reducing commit redundancy.
> 
> ```
> [aryanj@matrix-34xx:~]$ pk secrets touch v1:file1 v1:file2 v2:file3
>  > making file1 and file2 in v1
>  > making file3 in v2
> 
> [aryanj@matrix-34xx:~]$ pk secrets touch v1:file1 v1:file2 v2:file3 v1:file4
>  > making file1 and file2 in v1
>  > making file3 in v2
>  > making file4 in v1
> ```

Client IO vs Agent IO. This is important design decision. In the case of PK CLI, you must assume that the agent could be remote. That means if I touch a local file, only the CLI client can be one doing that job, cannot expect the agent to do that work because it might not be on the same filesystem.

CMCDragonkai:
> Another concern I have is that on UNIX, the `touch` command also works like this, touching files in order that they were provided in. If we perform grouping, then it will break the constraint of UNIX's order and might yield unexpected behaviour. On the other hand, if we don't group the commands, then the vault will see multiple commits for creating multiple secrets, which isn't optimal. How should the `touch` command handle this?

This requires understanding what an atomic commit is.

The right abstraction here is to define what a "Vault Operation" means. That's why we have a vault ops abstaction.

1 vault op should equal 1 commit.

A vault op however doesn't mean you cannot create multiple files or delete multiple files together.

In fact it appears the right way to do this is to ensure that a vault operation represents a transaction. And that is what a commit is.

Pass through the transaction context through however many vault operations, and then when it commits, that's a single commit.

This allows you do to multiple touch operations that are combined together into a single commit using the same transaction abstraction that we do in the js-db.

In fact this is possible across RPC boundaries as well, as it means you create a stream and as long as the stream is kept open, that represents the transaction context for the commit. Thus even across the RPC boundary you can create a transaction.

The other issue is concurrent commits. In the case of js-db we use a combination of OCC and PCC. OCC is the default and PCC you have to elect into it.

The design of git means that you can't really do OCC, and you're not really taking snapshots of vault state. You're always getting the full state of the vault at any time to make your commit. And a commit just means whatever is still a dirty state.

To simplify this you could state that a vault op must be PCC forcing a vault level lock on commit sequencing. However this isn't technically required by git, it does allow one to make separate edits to the shared files within a git repository, and you can in fact commit them separately. However we don't really do this because we track the files that we need to commit anyway. And each vault operation takes full mutable control of the vault.

If this aspect hasn't been specced out, you need to spec out the trade-offs on the approach.

aryanjassal:
> Another concern I have is that on UNIX, the `touch` command also works like this, touching files in order that they were provided in. If we perform grouping, then it will break the constraint of UNIX's order and might yield unexpected behaviour. On the other hand, if we don't group the commands, then the vault will see multiple commits for creating multiple secrets, which isn't optimal. How should the `touch` command handle this?

Brian and I had a small discussion regarding this. We were only able to brainstorm ideas, but his idea stipulated the following.

Instead of using existing `writeF` commands, what we can do is rely on manually acquiring resource locks. Then, when we encounter a new vault name, acquire a writable transaction for that vault, then make the file in the vault. This will ensure that the files are created in order and the files all come under a single transaction. We can change the behaviour of `writeF` to return the resource lock instead of managing it internally, and we will be able to manually manage the resource, unlocking the resource when the incoming stream of filenames has all ended. This approach will allow us to adhere to both requirements and also allow us to filter for file paths pointing to the local file system.

This approach will entail modifications to previous commands like the `VaultsSecretsRemove` handler, as that groups the paths by vault name to delete all files in a single transaction, but it's not 1:1 replica of UNIX's `rm` behaviour.

We will discuss this more and expand on this idea in a separate meeting today.

src/client/handlers/VaultsSecretsMkdir.ts
```diff
@@ -2,50 +2,60 @@ import type { DB } from '@matrixai/db';
...
-    });
-    return {
-      success: true,
-    };
+  public handle = async function* (
+    input: AsyncIterable<ClientRPCRequestParams<SecretDirMessage>>,
+  ): AsyncGenerator<ClientRPCResponseResult<SuccessWithErrorMessage>> {
+    const { vaultManager, db }: { vaultManager: VaultManager; db: DB } =
+      this.container;
+    let metadata: any = undefined;
```
aryanjassal: Usually, in TypeScript, the `any` type is avoided, as it bypasses type checking, defeating the purpose of types in the first place. Here, however, we need to use the any type as we are parsing the metadata directly.

Should we somehow simplify this to not include `any` type, or is this usage acceptable?

src/client/handlers/VaultsSecretsMkdir.ts
```diff
@@ -2,50 +2,60 @@ import type { DB } from '@matrixai/db';
...
+          // Get vaultId
+          const vaultIdFromName = await vaultManager.getVaultId(nameOrId, tran);
+          const vaultId =
+            vaultIdFromName ?? vaultsUtils.decodeVaultId(nameOrId);
+          if (vaultId == null) {
+            throw new vaultsErrors.ErrorVaultsVaultUndefined();
+          }
+          // Write directories. This doesn't need to be grouped by vault names,
+          // as no commit is created for empty directories anyways.
+          let response: SuccessWithErrorMessage = { success: false };
```
aryanjassal: My LSP was throwing an error if I assigned `undefined` when assigning a value, so I assigned a dummy value as it would be overridden in the `withVaults` bit anyways. Should this be changed, or is it fine as-is?

src/client/handlers/VaultsSecretsMkdir.ts
```diff
@@ -2,50 +2,60 @@ import type { DB } from '@matrixai/db';
...
-      );
-    });
-    return {
-      success: true,
-    };
+  public handle = async function* (
+    input: AsyncIterable<ClientRPCRequestParams<SecretDirMessage>>,
+  ): AsyncGenerator<ClientRPCResponseResult<SuccessWithErrorMessage>> {
+    const { vaultManager, db }: { vaultManager: VaultManager; db: DB } =
+      this.container;
```
aryanjassal: I had to do this as my LSP couldn't detect the types of the values inside `this.container`. The issue can be observed in the `VaultsSecretsGet` handler, which is a `DuplexStream` and suffers with the same issue of the types not being recognised. This interestingly doesn't happen with `UnaryStream` or `ServerStream`, only in `DuplexStream` and `ClientStream`.

This is important as without the types being recognised, objects like `vaultManager` has the `any` type and no autocomplete support exists. It is also really hard to write correct code, as we could be relying on a missing property/attribute, but because the default type is `any`, no errors will be shown while writing the code.

If this issue is not reproducible (eg. JetBrains or VSCode can infer the type properly), then should adding types like this be a standard for all commands, or should this be excluded as it is not a big enough issue?

If this issue can be reproduced (same behaviour on other IDEs), then should this be looked into, or do we use this approach to properly define types for injected dependencies, or do we ignore the issue completely?

src/client/handlers/VaultsSecretsMkdir.ts
tegefaulkes: Metadata has a type already that isn't `any`, just reuse that. I think it's `Record<string, JSONValue>` but you'll have to check.

src/client/handlers/VaultsSecretsMkdir.ts
tegefaulkes: You can do `let response: type | undefined;` if you want it to start unininialized. I think `let response: Type;` *should* also be valid.

src/client/handlers/VaultsSecretsMkdir.ts
tegefaulkes: From what I recall, anonymous generators get weird with `this`. Look at other handlers like `VaultsSecretsList` that seems to be how to handle it. Make sure all the handlers that function like generators are consistent in this.

tests/client/handlers/vaults.test.ts
aryanjassal: I'm assuming that no warning is needed for other commands which might get the directories to a state as such.

src/client/handlers/VaultsSecretsMkdir.ts
aryanjassal: `VaultsSecretsList` is a `ServerHandler` and doesn't have this issue, even though it is an anonymous generator. So, it only applies when we are getting an input stream. For consistency, I can do this to all the `VaultsSecrets` command, but I think at least we should try to understand why this could be happening.

src/client/handlers/VaultsSecretsMkdir.ts
aryanjassal: Technically, I can just do `let response: type;`, but my LSP still complains, saying that the variable is potentially unbound. The LSP I am using is the same one used by VS Code, so this issue isn't specific to my LSP. I have done the following to tell my LSP that I know better: `if (response!.success) {...}`.

src/client/handlers/VaultsSecretsMkdir.ts
aryanjassal: I tried this and it didn't work. I checked the type and it is more complex than that, containing multiple unions between types.

Instead, a `JSONObject` type worked. The types were still kinda borked, as my LSP complained that not all optional types have the `options` and `recursive` fields. To fix this, I had to do the following:
```ts
const options = metadata?.options as { recursive?: boolean };
response = await vaultOps.mkdir(vault, dirName, {
  recursive: options.recursive,
});
```
This is kind of a janky solution, but if I don't have the `as type`, then TS will check if the attribute `recursive` exists on all the types, which of course it doesn't, and it will raise an error in the linter.

src/client/handlers/VaultsSecretsMkdir.ts
tegefaulkes: That doesn't make sense.

I just had a look at it and it is a normal generator function, not a anonymous one. That alone makes it inconsistent with the other handlers. In fact there seems to be some amount of inconsistency with how we create handlers as generators.

Some are using `public async *handle()` and others `public handle = async function* ()`. I think the 2nd version is the one we want and we should make sure all the generator handlers are written that way.

aryanjassal: Yeah, I see that now. The functions which are written in the first way have full type hinting, and the ones written with the second approach don't.

I will make sure all `VaultsSecrets` handlers use the `public handle = async function* ()` syntax. I will also use the type definition for unpacking the container to get type hints, too.

I will resolve this when I finish implementing this.

aryanjassal: Should I go over all the places where we use the first approach and switch it out for the second one, or only in the client domain's handlers? Or only in `VaultsSecrets` handlers? What is the scope of this change?

aryanjassal:
During development, I have noticed that some tests take exceptionally long to run. The longest running test by far is `client/vaults.test.ts` taking approximately 300 seconds if testing the entire file. Brian mentioned that all the client tests needed to be collapsed into a single file, because otherwise it would cause memory leaks and crash the CI.

There is another test file which also takes pretty long to run; the `vaults/VaultOps.test.ts` file. It takes over 100 seconds to run through all the tests. As tests in only separate files can be run in parallel, I believe the tests need to be separated into smaller chunks.

Is there an issue where we are tracking this? Or should I make a new issue to track this?

src/client/handlers/VaultsSecretsMkdir.ts
aryanjassal: I have discovered some issues with the code while doing this change.

In `src/client/handlers/NodesGetAll.ts`, we pass in `nodeGraph` and `keyRing` into the container, but only take out `nodeGraph`. Why do we have `keyRing` if we are not going to use it?

tegefaulkes: If we don't use it there then remove it.

src/vaults/VaultOps.ts
```diff
@@ -173,29 +174,35 @@ async function mkdir(
...
+      };
+    } else if (e.code === 'EEXIST') {
+      return {
+        type: 'error',
+        code: e.code,
+        reason: dirPath,
+      };
+    } else {
       throw e;
     }
```
tegefaulkes: If each if statement is its own return path then you can simplify a block like this to just 
```
if(a) return {};
if(b) return {};
throw e
```

tests/vaults/VaultOps.test.ts
```diff
@@ -323,32 +324,40 @@ describe('VaultOps', () => {
...
     test('creating existing secret should fail', async () => {
       await writeSecret(secretName, secretContent);
-      await expect(vaultOps.mkdir(vault, secretName)).rejects.toThrow(
-        vaultsErrors.ErrorSecretsSecretDefined,
-      );
+      const response = await vaultOps.mkdir(vault, secretName);
+      expect(response.type).toEqual('error');
+      const error = response as ErrorMessage;
+      expect(error.code).toEqual('EEXIST');
+      await expectSecret(secretName, secretContent);
```
tegefaulkes, line 360: Don't use the as keyword here. Use the utils.never() to enforce the type by checking it. Annoyingly the `expect` doesn't do it for you. So do 

```ts
expect(result.type).toBe('error');
if (result.type !='errror') utils.never('should be an error')
expect(result.code).toBe('EEXISTS')
```

aryanjassal: This is probably something that your IDE does for you. My IDE does not do this.

```ts
  for await (const data of response.readable) {
    if (data.type !== 'success') {
      if (data.type !== 'error') utils.never();
      expect(data.code).toEqual('ENOENT');
      expect(data.reason).toEqual(dirPath3);
    }
  }
```

This gives me the same error that `code` and `reason` doesn't exist on `SuccessOrErrorMessage`.

image

aryanjassal: Interestingly enough, for simpler cases with more evident types, my LSP seems to work just fine. This is an issue with my local environment and does not impact Polykey in any way. I will have to resolve this issue myself locally, but this thread can be resolved now that I have implemented these changes.

src/client/handlers/VaultsSecretsMkdir.ts
aryanjassal: This is not the correct solution, as I am using the `as` keyword here. After having a talk with Brian regarding this, using the `as` keyword in all but the most critical cases is discouraged, so I will find an alternative solution for this.

aryanjassal: My solution was to use the `POJO` type, as it is defined as `[key: string]: any`. This is not strict type checking, but using the exact type as of the metadata field makes it challenging to access potentially deeply nested fields. The `POJO` type retains some type checking while still allowing the freedom of `any`.

This is not a perfect solution, but this is better than using `any`. These changes will need to be made in other handlers too which are expecting values in `metadata`, but that can be handled in future PRs as this one is already getting pretty large.

tests/client/handlers/vaults.test.ts
```diff
@@ -1341,18 +1342,149 @@ describe('vaultsSecretsMkdir', () => {
...
+    const response = await rpcClient.methods.vaultsSecretsMkdir();
+    const writer = response.writable.getWriter();
+    await writer.write({ nameOrId: encodeVaultId, dirName: dirPath });
+    await writer.close();
+    for await (const data of response.readable) {
+      expect(data.type).toEqual('error');
+      if (data.type !== 'error') utils.never();
+      expect(data.code).toEqual('ENOENT');
+      expect(data.reason).toEqual(dirPath);
+    }
```
aryanjassal: Apparently the LSP complaints of it not being able to properly narrow down the type wasn't an issue in my LSP, but rather, the code.

https://github.com/typescript-language-server/typescript-language-server/issues/924

I raised this issue on my LSP's issue page, and there, I was informed that `ts-node` doesn't actually use typescript, and if I ran `npx tsc` then it should give me the same errors, which was the case.

As such, while this works on `ts-node`, if we ever switched compilers, this would just fail to build. In my opinion, this needs to be resolved rather than ignoring this as it isn't cropping up for now.

aryanjassal: After talking about this a bit more with Brian, we have realised that the types are correct, but typescript cannot evaluate this type properly as it is too complicated. As such, while not ideal, but this is one time that I will be allowed to use the `as` keyword to explicitly change the type of an object.

tegefaulkes: The problem is the utility type in `ClientRPCResponseResult` that's wrapping the `SuccessOrErrorMessage`. Its preventing typescripts ability to narrow the type down. We've had issues with utility types and  type inference like this before.

aryanjassal:
All the tasks have been completed for this PR, and all the reviews are also approving. Merging.

CMCDragonkai:
> During development, I have noticed that some tests take exceptionally long to run. The longest running test by far is `client/vaults.test.ts` taking approximately 300 seconds if testing the entire file. Brian mentioned that all the client tests needed to be collapsed into a single file, because otherwise it would cause memory leaks and crash the CI.
> 
> There is another test file which also takes pretty long to run; the `vaults/VaultOps.test.ts` file. It takes over 100 seconds to run through all the tests. As tests in only separate files can be run in parallel, I believe the tests need to be separated into smaller chunks.
> 
> Is there an issue where we are tracking this? Or should I make a new issue to track this?

Hmm I'm not sure why client tests have to be in one a single file? It seems better isolation is needed, but also it's possibly due to the concurrent test runner, maybe there's a way to share resources between the client tests, or make them grouped up in some way.

tegefaulkes:
At the time they were moved into a single test file because the CI was running out of memory when they were all separated. It was a quick fix at the time.

I suspect the memory leak was some part of the networking code keeping a reference in the global scope. Which fails to clean up properly between tests. We'll have to fix that before splitting up the tests again.

Issue #832, opened 2024-10-18
Review and refactor RPC handlers to handle cancellation properly


tegefaulkes:
### Specification

Some RPC handlers are preforming long running async tasks. In these cases they need to make use of the provided `ctx` and handle cancellation gracefully to prevent any resource leaks. A clear example of this is the vaults pull and clone handlers.

So we need to review handlers and update the ones that need CTX handling applied.

### Additional context

- Related https://github.com/MatrixAI/Polykey-CLI/issues/264 REF ENG-397

### Tasks

1. Review all RPC handlers and identify ones that preform complex async  or long running tasks.
2. Refactor theses handlers to handle cancellation in a responsive and graceful manor.
3. Apply testing to see if timing out during these tasks causes any problems. Possibly apply regression testing for this.

tegefaulkes:
@aryanjassal You'll need to take this over. There is existing work done in branch `feature-eng-432-review-and-refactor-rpc-handlers-to-handle-cancellation` relating to this.

aryanjassal:
For handlers like `VaultsSecretsGet` (as of [**Polykey#838**](https://github.com/MatrixAI/Polykey/pull/838)), it is a `ServerHandler`, but it streams back the file contents. As such, there is no loop anywhere. For cases like that, where should the context handling be put?

Similarly, I don't think we can actually handle cancellation for `ClientHandlers` or `UnaryHandlers` as they don't send back streamed data, so cancellaton doesn't make much sense.

tegefaulkes:
Cancellation applies to async operations and not just streamed data. For the most part the streaming handlers are the biggest targets for handling cancellation. But anything that waits for locks or takes a little while to do a thing can be cancelled.

For example, the `fetch()` call is considered a unary function, but it has cancellation via an `AbortSignal` because a fetch operation could take a long time.

CMCDragonkai:
> For handlers like `VaultsSecretsGet` (as of [**Polykey#838**](https://github.com/MatrixAI/Polykey/pull/838)), it is a `ServerHandler`, but it streams back the file contents. As such, there is no loop anywhere. For cases like that, where should the context handling be put?
> 
> Similarly, I don't think we can actually handle cancellation for `ClientHandlers` or `UnaryHandlers` as they don't send back streamed data, so cancellaton doesn't make much sense.

Anything asynchronous can be cancelled. Streaming is orthogonal.

CMCDragonkai:
[...]

PR #838, opened 2024-11-01
Wrapping `VaultOps` errors in `ErrorPolykey` errors and updating tests

### LINKED ISSUE #311: Some `secrets` commands throw an `undefined` error
### Describe the bug

When writing to a directory, the command should provide appropriate feedback bail out. This should either be done immediately (by preemptively checking if the target is a valid file path or not), or later, by raising an appropriate error. 

Currently, it just returns this output, which is not really explicit or apparent in what went wrong:

```console
[aryanj@matrix-34xx:~]$ pk secrets write vault:dir
writing to dir <Ctrl-D>
undefined

[aryanj@matrix-34xx:~]$ echo $?
255
```

### To Reproduce

1. Try writing to a directory using stdin
2. Observe issue

### Expected behavior

As in `secrets cat`, an error could be raised:
```console
[aryanj@matrix-34xx:~]$ pk secrets cat vault:dir     
ErrorSecretsIsDirectory: dir: Is a directory
```

### Platform
 - Device: Dell Precision 3480
 - OS: NixOS
 - Version: `["0.10.0","1.14.0","1","1"]`

### Notify maintainers

@aryanjassal

aryanjassal:
### Description

Any non-Polykey error, when sent through the RPC, tries to throw raw JSON. In Polykey CLI, this leads to `undefined` errors.

This PR fixes that by wrapping all errors thrown by the `VaultOps` operation in an appropriate `ErrorPolykey` error.

### Issues Fixed

* Partially resolves https://github.com/MatrixAI/Polykey-CLI/issues/311 (ENG-434)

### Tasks

- [x] 1. Update all `VaultOps` to wrap errors in `ErrorPolykey`
- ~~[ ] 2. Return `SuccessOrErrorMessage` instead of string on error in related `VaultsSecrets` handlers~~ will be done in another PR.
- ~~[ ] 3. Update tests to be more stringent for all `VaultsSecrets` handlers~~ will be done in another PR.
- [x] 4. Make sure all errors are wrapped in a polykey error. This is something the API should guarantee.
- [x] 5. Spilt `VaultsSecretsGet` and `VaultsSecretsCat`
- ~~[ ] 6. Add tests for error serialisation~~ we don't test utils functions

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsSecretsGet.ts
+++ b/src/client/handlers/VaultsSecretsGet.ts
@@ -2,72 +2,44 @@ import type { DB } from '@matrixai/db';
 import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
-  ContentWithErrorMessage,
+  ContentMessage,
   SecretIdentifierMessage,
 } from '../types';
 import type VaultManager from '../../vaults/VaultManager';
-import { DuplexHandler } from '@matrixai/rpc';
+import { ServerHandler } from '@matrixai/rpc';
 import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
 import * as vaultOps from '../../vaults/VaultOps';
 
-class VaultsSecretsGet extends DuplexHandler<
+// This method only returns the contents of a single secret, and throws an error
+// if the secret couldn't be read. To read multiple secrets, refer to
+// `VaultsSecretsCat`.
+class VaultsSecretsGet extends ServerHandler<
   {
     db: DB;
     vaultManager: VaultManager;
   },
   ClientRPCRequestParams<SecretIdentifierMessage>,
-  ClientRPCResponseResult<ContentWithErrorMessage>
+  ClientRPCResponseResult<ContentMessage>
 > {
   public handle = async function* (
-    input: AsyncIterable<ClientRPCRequestParams<SecretIdentifierMessage>>,
-    _cancel,
-    _meta,
-    ctx,
-  ): AsyncGenerator<ClientRPCResponseResult<ContentWithErrorMessage>> {
-    if (ctx.signal.aborted) throw ctx.signal.reason;
+    input: ClientRPCRequestParams<SecretIdentifierMessage>,
+  ): AsyncGenerator<ClientRPCResponseResult<ContentMessage>> {
     const { db, vaultManager }: { db: DB; vaultManager: VaultManager } =
       this.container;
-    yield* db.withTransactionG(async function* (tran): AsyncGenerator<
-      ClientRPCResponseResult<ContentWithErrorMessage>
-    > {
-      if (ctx.signal.aborted) throw ctx.signal.reason;
-      // As we need to preserve the order of parameters, we need to loop over
-      // them individually, as grouping them would make them go out of order.
-      let metadata: any = undefined;
-      for await (const secretIdentiferMessage of input) {
-        if (ctx.signal.aborted) throw ctx.signal.reason;
-        if (metadata == null) metadata = secretIdentiferMessage.metadata ?? {};
-        const { nameOrId, secretName } = secretIdentiferMessage;
-        const vaultIdFromName = await vaultManager.getVaultId(nameOrId, tran);
-        const vaultId = vaultIdFromName ?? vaultsUtils.decodeVaultId(nameOrId);
-        if (vaultId == null) throw new vaultsErrors.ErrorVaultsVaultUndefined();
-        yield await vaultManager.withVaults(
-          [vaultId],
-          async (vault) => {
-            try {
-              const content = await vaultOps.getSecret(vault, secretName);
-              return { secretContent: content.toString('binary') };
-            } catch (e) {
-              if (metadata?.options?.continueOnError === true) {
-                if (e instanceof vaultsErrors.ErrorSecretsSecretUndefined) {
-                  return {
-                    secretContent: '',
-                    error: `${e.name}: ${secretName}: No such secret or directory\n`,
-                  };
-                } else if (e instanceof vaultsErrors.ErrorSecretsIsDirectory) {
-                  return {
-                    secretContent: '',
-                    error: `${e.name}: ${secretName}: Is a directory\n`,
-                  };
-                }
-              }
-              throw e;
-            }
-          },
-          tran,
-        );
-      }
+    yield await db.withTransactionF(async (tran) => {
+      const vaultIdFromName = await vaultManager.getVaultId(
+        input.nameOrId,
+        tran,
+      );
+      const vaultId =
+        vaultIdFromName ?? vaultsUtils.decodeVaultId(input.nameOrId);
+      if (vaultId == null) throw new vaultsErrors.ErrorVaultsVaultUndefined();
+      // Get the contents of the file
+      return await vaultManager.withVaults([vaultId], async (vault) => {
+        const content = await vaultOps.getSecret(vault, input.secretName);
+        return { secretContent: content.toString('binary') };
+      });
     });
   };
 }
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -11,6 +11,11 @@ class ErrorPolykeyUnknown<T> extends ErrorPolykey<T> {
   exitCode = sysexits.PROTOCOL;
 }
 
+class ErrorPolykeyUnexpected<T> extends ErrorPolykey<T> {
+  static description = 'An error originating outside Polykey was thrown';
+  exitCode = sysexits.UNKNOWN;
+}
+
 class ErrorPolykeyAgentRunning<T> extends ErrorPolykey<T> {
   static description = 'PolykeyAgent is running';
   exitCode = sysexits.USAGE;
@@ -56,6 +61,7 @@ export {
   ErrorPolykey,
   ErrorPolykeyUnimplemented,
   ErrorPolykeyUnknown,
+  ErrorPolykeyUnexpected,
   ErrorPolykeyAgentRunning,
   ErrorPolykeyAgentNotRunning,
   ErrorPolykeyAgentDestroyed,
--- a/src/network/utils.ts
+++ b/src/network/utils.ts
@@ -450,7 +450,14 @@ function fromError(error: any) {
   if (error instanceof Error) {
     const cause = fromError(error.cause);
     const timestamp: string = ((error as any).timestamp ?? new Date()).toJSON();
-    if (error instanceof AbstractError) {
+    // ErrorParse is an exception to the Polykey error tree, where the error is
+    // not actually derived from ErrorPolykey, but rather an AbstractError. We
+    // cannot simply check for AbstractError because other libraries can also
+    // throw that, and those errors should be wrapped in ErrorPolykeyUnexpected.
+    if (
+      error instanceof ErrorPolykey ||
+      error instanceof validationErrors.ErrorParse
+    ) {
       return error.toJSON();
     } else if (error instanceof AggregateError) {
       // AggregateError has an `errors` property
@@ -466,18 +473,55 @@ function fromError(error: any) {
       };
     }
 
-    // If it's some other type of error then only serialise the message and
-    // stack (and the type of the error)
-    return {
-      type: error.name,
-      message: error.message,
-      data: {
-        stack: error.stack,
-        timestamp,
-        cause,
-      },
-    };
+    // If it's some other type of error then wrap it in ErrorPolykeyUnexpected,
+    // serialising only the error type, message and its stack.
+    const wrappedError = new errors.ErrorPolykeyUnexpected(
+      `Unexpected error occurred: ${error.name}`,
+      { cause: error },
+    );
+    return wrappedError.toJSON();
   }
+
+  // Use the properties of the error to create an appropriate error message.
+  let message = '';
+  switch (typeof error) {
+    case 'boolean':
+    case 'number':
+    case 'string':
+    case 'bigint':
+    case 'symbol':
+      message = `Non-error literal ${String(error)} was thrown`;
+      break;
+    case 'object':
+      // Let the fallback handler catch null values.
+      if (error == null) break;
+      // If we have an error message defined, then return that.
+      if ('message' in error && typeof error.message === 'string') {
+        message = error.message;
+      }
+      // If present, mention the constructor name in the message.
+      if (error.constructor?.name != null) {
+        message = `Non-error object ${error.constructor.name} was thrown`;
+      }
+      // Any other values should be handled by the fallback handler.
+      break;
+  }
+
+  // Handle cases where the error is not serialisable, like objects created
+  // using Object.create(null). Trying to serialise this throws a TypeError.
+  try {
+    message = `Non-error value ${String(error)} was thrown`;
+  } catch (e) {
+    if (e instanceof TypeError) message = 'Non-error value was thrown';
+    else throw e;
+  }
+
+  // If the error was not an Error, then wrap it inside ErrorPolykeyUnexpected
+  // and return that.
+  const wrappedError = new errors.ErrorPolykeyUnexpected(message, {
+    cause: error,
+  });
+  return wrappedError.toJSON();
 }
 
 const standardErrors: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@matrixai/mdns": "^1.2.6",
         "@matrixai/quic": "^1.2.10",
         "@matrixai/resources": "^1.1.5",
-        "@matrixai/rpc": "^0.5.1",
+        "@matrixai/rpc": "^0.6.0",
         "@matrixai/timer": "^1.1.3",
         "@matrixai/workers": "^1.3.7",
         "@matrixai/ws": "^1.1.7",
@@ -1696,9 +1696,9 @@
       "integrity": "sha512-m/DEZEe3wHqWEPTyoBtzFF6U9vWYhEnQtGgwvqiAlTxTM0rk96UBpWjDZCTF/vYG11ZlmlQFtg5H+zGgbjaB3Q=="
     },
     "node_modules/@matrixai/rpc": {
-      "version": "0.5.1",
-      "resolved": "https://registry.npmjs.org/@matrixai/rpc/-/rpc-0.5.1.tgz",
-      "integrity": "sha512-crqC2J7jGSQhPwHOoO0dWmUXx5JVb1Cfl4bXB9dcc9JobetU3k0rf5iC4BF7JGCtIxQPRJnT63y9+6EGzozZgA==",
+      "version": "0.6.0",
+      "resolved": "https://registry.npmjs.org/@matrixai/rpc/-/rpc-0.6.0.tgz",
+      "integrity": "sha512-ENjJO2h7CmPLaHhObHs2nvpv98YZPxa79/jf+TqEPEfbhE1BkNCys9pXDE/CYDP9vxb4seS39WkR9cNivQU50A==",
       "dependencies": {
         "@matrixai/async-init": "^1.10.0",
         "@matrixai/contexts": "^1.2.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "@matrixai/mdns": "^1.2.6",
     "@matrixai/quic": "^1.2.10",
     "@matrixai/resources": "^1.1.5",
-    "@matrixai/rpc": "^0.5.1",
+    "@matrixai/rpc": "^0.6.0",
     "@matrixai/timer": "^1.1.3",
     "@matrixai/workers": "^1.3.7",
     "@matrixai/ws": "^1.1.7",
--- a/src/client/callers/index.ts
+++ b/src/client/callers/index.ts
@@ -63,6 +63,7 @@ import vaultsPermissionUnset from './vaultsPermissionUnset';
 import vaultsPull from './vaultsPull';
 import vaultsRename from './vaultsRename';
 import vaultsScan from './vaultsScan';
+import vaultsSecretsCat from './vaultsSecretsCat';
 import vaultsSecretsEnv from './vaultsSecretsEnv';
 import vaultsSecretsGet from './vaultsSecretsGet';
 import vaultsSecretsList from './vaultsSecretsList';
@@ -144,6 +145,7 @@ const clientManifest = {
   vaultsPull,
   vaultsRename,
   vaultsScan,
+  vaultsSecretsCat,
   vaultsSecretsEnv,
   vaultsSecretsGet,
   vaultsSecretsList,
@@ -224,6 +226,7 @@ export {
   vaultsPull,
   vaultsRename,
   vaultsScan,
+  vaultsSecretsCat,
   vaultsSecretsEnv,
   vaultsSecretsGet,
   vaultsSecretsList,
--- a/src/client/callers/vaultsSecretsCat.ts
+++ b/src/client/callers/vaultsSecretsCat.ts
@@ -1,0 +1,12 @@
+import type { HandlerTypes } from '@matrixai/rpc';
+import type VaultsSecretsCat from '../handlers/VaultsSecretsCat';
+import { DuplexCaller } from '@matrixai/rpc';
+
+type CallerTypes = HandlerTypes<VaultsSecretsCat>;
+
+const vaultsSecretsCat = new DuplexCaller<
+  CallerTypes['input'],
+  CallerTypes['output']
+>();
+
+export default vaultsSecretsCat;
--- a/src/client/callers/vaultsSecretsGet.ts
+++ b/src/client/callers/vaultsSecretsGet.ts
@@ -1,10 +1,10 @@
 import type { HandlerTypes } from '@matrixai/rpc';
 import type VaultsSecretsGet from '../handlers/VaultsSecretsGet';
-import { DuplexCaller } from '@matrixai/rpc';
+import { ServerCaller } from '@matrixai/rpc';
 
 type CallerTypes = HandlerTypes<VaultsSecretsGet>;
 
-const vaultsSecretsGet = new DuplexCaller<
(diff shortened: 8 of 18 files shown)


aryanjassal:
This `VaultOps` method still exists, which was previously powering the `secrets list` command, but right now, is not being used anywhere.

What do we do with outdated `VaultOps`? Do we remove this, or keep this in, just in case this is needed sometime in the future?

```ts
/**
 * Retrieves a list of the secrets in a vault
 */
async function listSecrets(vault: Vault): Promise<string[]> {
  return await vault.readF(async (efs) => {
    const secrets: string[] = [];
    for await (const secret of vaultsUtils.readDirRecursively(efs)) {
      secrets.push(secret);
    }
    return secrets;
  });
}
```

aryanjassal:
We need to make sure that all errors, Polykey or otherwise, are being wrapped in a Polykey error. All Polykey errors which happen during a RPC call are wrapped in `ErrorPolykeyRemote`. However, non-Polykey errors are propagated as-is. This was also why we had the `undefined` errors in the first place.

This needs to be reviewed. *All* errors must be wrapped in `ErrorPolykeyRemote` to ensure integrity after passing through the RPC. The change for this is needed inside the `toError` and `fromError` within the `networks` domain.

Thoughts, @tegefaulkes ?

aryanjassal:
The `VaultsSecretsGet` handler is becoming too complex to use with the `continueOnError` option. It was introduced to retain backwards compatibility, and to ensure that the command behaves like `VaultOps.getSecret`, but that use case is not used anymore.

However, it is a very common use case, so I will split it into `VaultsSecretsGet` and `VaultsSecretsCat`. The `VaultsSecretsGet` will behave like `VaultOps.getSecret`, and `VaultsSecretsCat` will behave like the current `VaultsSecretsGet` with `continueOnError` set, which ignores any errors.

Brian mentioned this before, when we were first implementing this command, but at the time, it was an acceptable trade-off. However, I now realise that this introduces a vector where new bugs can very easily be introduced, and it is also inconsistent with every other `VaultsSecrets` handler, so I could make this change here, but is it a good idea?

Thoughts, @tegefaulkes ?

aryanjassal:
Now when trying to serialise a non-Polykey error, it gets wrapped in `ErrorPolykeyUnexpected` like this before propagating through the RPC.

```
    ErrorPolykeyRemote: 
        at toError (/home/aryanj/Projects/polykey/src/network/utils.ts:597:27)
        at Object.transform (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/utils.ts:452:19)
        ... 64 lines matching cause stack trace ...
        at processTicksAndRejections (node:internal/process/task_queues:95:5) {
      data: {},
      cause: ErrorPolykeyUnexpected: Unexpected error of type ErrorEncryptedFSError occured
          at toError (/home/aryanj/Projects/polykey/src/network/utils.ts:589:13)
          at Object.transform (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/utils.ts:452:19)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at transformStreamDefaultControllerPerformTransform (node:internal/webstreams/transformstream:515:18)
          at transformStreamDefaultSinkWriteAlgorithm (node:internal/webstreams/transformstream:565:10)
          at Object.write (node:internal/webstreams/transformstream:360:14)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at writableStreamDefaultControllerProcessWrite (node:internal/webstreams/writablestream:1112:5)
          at writableStreamDefaultControllerAdvanceQueueIfNeeded (node:internal/webstreams/writablestream:1227:5)
          at writableStreamDefaultControllerWrite (node:internal/webstreams/writablestream:1101:3)
          at writableStreamDefaultWriterWrite (node:internal/webstreams/writablestream:991:3)
          at PipeToReadableStreamReadRequest.[kChunk] (node:internal/webstreams/readablestream:1554:31)
          at readableStreamFulfillReadRequest (node:internal/webstreams/readablestream:2096:24)
          at readableStreamDefaultControllerEnqueue (node:internal/webstreams/readablestream:2287:5)
          at transformStreamDefaultControllerEnqueue (node:internal/webstreams/transformstream:496:5)
          at TransformStreamDefaultController.enqueue (node:internal/webstreams/transformstream:306:5)
          at Object.transform (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/middleware.ts:143:20)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at transformStreamDefaultControllerPerformTransform (node:internal/webstreams/transformstream:515:18)
          at transformStreamDefaultSinkWriteAlgorithm (node:internal/webstreams/transformstream:565:10)
          at Object.write (node:internal/webstreams/transformstream:360:14)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at writableStreamDefaultControllerProcessWrite (node:internal/webstreams/writablestream:1112:5)
          at writableStreamDefaultControllerAdvanceQueueIfNeeded (node:internal/webstreams/writablestream:1227:5)
          at writableStreamDefaultControllerWrite (node:internal/webstreams/writablestream:1101:3)
          at writableStreamDefaultWriterWrite (node:internal/webstreams/writablestream:991:3)
          at PipeToReadableStreamReadRequest.[kChunk] (node:internal/webstreams/readablestream:1554:31)
          at readableStreamFulfillReadRequest (node:internal/webstreams/readablestream:2096:24)
          at readableStreamDefaultControllerEnqueue (node:internal/webstreams/readablestream:2287:5)
          at transformStreamDefaultControllerEnqueue (node:internal/webstreams/transformstream:496:5)
          at defaultTransformAlgorithm (node:internal/webstreams/transformstream:345:3)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at transformStreamDefaultControllerPerformTransform (node:internal/webstreams/transformstream:515:18)
          at transformStreamDefaultSinkWriteAlgorithm (node:internal/webstreams/transformstream:565:10)
          at Object.write (node:internal/webstreams/transformstream:360:14)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at writableStreamDefaultControllerProcessWrite (node:internal/webstreams/writablestream:1112:5)
          at writableStreamDefaultControllerAdvanceQueueIfNeeded (node:internal/webstreams/writablestream:1227:5)
          at writableStreamDefaultControllerWrite (node:internal/webstreams/writablestream:1101:3)
          at writableStreamDefaultWriterWrite (node:internal/webstreams/writablestream:991:3)
          at PipeToReadableStreamReadRequest.[kChunk] (node:internal/webstreams/readablestream:1554:31)
          at readableStreamFulfillReadRequest (node:internal/webstreams/readablestream:2096:24)
          at readableStreamDefaultControllerEnqueue (node:internal/webstreams/readablestream:2287:5)
          at transformStreamDefaultControllerEnqueue (node:internal/webstreams/transformstream:496:5)
          at TransformStreamDefaultController.enqueue (node:internal/webstreams/transformstream:306:5)
          at TokenParser.parser.onValue (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/middleware.ts:46:20)
          at TokenParser.emit (/home/aryanj/Projects/polykey/node_modules/@streamparser/json/src/tokenparser.ts:144:12)
          at TokenParser.pop (/home/aryanj/Projects/polykey/node_modules/@streamparser/json/src/tokenparser.ts:130:10)
          at TokenParser.write (/home/aryanj/Projects/polykey/node_modules/@streamparser/json/src/tokenparser.ts:276:16)
          at Tokenizer.write (/home/aryanj/Projects/polykey/node_modules/@streamparser/json/src/tokenizer.ts:193:20)
          at JSONParser.write (/home/aryanj/Projects/polykey/node_modules/@streamparser/json/src/jsonparser.ts:34:20)
          at Object.transform (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/middleware.ts:53:16)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at transformStreamDefaultControllerPerformTransform (node:internal/webstreams/transformstream:515:18)
          at transformStreamDefaultSinkWriteAlgorithm (node:internal/webstreams/transformstream:565:10)
          at Object.write (node:internal/webstreams/transformstream:360:14)
          at ensureIsPromise (node:internal/webstreams/util:185:19)
          at writableStreamDefaultControllerProcessWrite (node:internal/webstreams/writablestream:1112:5)
          at writableStreamDefaultControllerAdvanceQueueIfNeeded (node:internal/webstreams/writablestream:1227:5)
          at writableStreamDefaultControllerWrite (node:internal/webstreams/writablestream:1101:3)
          at writableStreamDefaultWriterWrite (node:internal/webstreams/writablestream:991:3)
          at PipeToReadableStreamReadRequest.[kChunk] (node:internal/webstreams/readablestream:1554:31)
          at readableStreamFulfillReadRequest (node:internal/webstreams/readablestream:2096:24)
          at readableStreamDefaultControllerEnqueue (node:internal/webstreams/readablestream:2287:5)
          at ReadableStreamDefaultController.enqueue (node:internal/webstreams/readablestream:1075:5)
          at constructor_.readablePull (/home/aryanj/Projects/polykey/node_modules/@matrixai/ws/src/WebSocketStream.ts:380:16)
          at processTicksAndRejections (node:internal/process/task_queues:95:5) {
        data: {},
        cause: { type: 'ErrorEncryptedFSError', data: [Object] },
        timestamp: 2024-11-04T06:11:14.216Z,
        exitCode: 76
      },
      timestamp: 2024-11-04T06:11:14.216Z,
      exitCode: 76,
      metadata: {
        localHost: '127.0.0.1',
        localPort: 50070,
        remoteHost: '127.0.0.1',
        remotePort: 45015,
        command: 'vaultsSecretsWriteFile'
      }
    }
```

Should the error only be wrapped in `ErrorPolykeyRemote`, or this additional layer fine for wrapping unexpected errors?

@tegefaulkes @CMCDragonkai

tegefaulkes:
- Yes, to properly deseralise the error it needs to be an error that is exported by the root `errors.ts` in polykey. So we need to wrap any errors not in that tree with one that is. Make sure that the error doing the wrapping has a description that describes the error that it wraps.
- All errors that come through the RPC are wrapped in an `ErrorPolykeyRemote`. I forget when this is done exactly so you'll need to check. I'm guessing it happens in the `toError` after the error has been deserialized.
- If `VaultsSecretsGet` is getting too complex by trying to address to many cases then you need to split it up into two methods. Sure they do similar things but the behaviour differs too much now. You need to split it up like we discussed before. Make comments having the two methods refer to each other in case an similar behaviour needs to be changed later down the line.

src/network/utils.ts
```diff
@@ -572,33 +572,52 @@ function toError(
       }
     }
     const eClass = errors[errorData.type];
+    let e: any;
```
tegefaulkes: Never use the `any` type. There's a `unknown` type if you don't know what the type actually is.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

src/network/utils.ts
```diff
@@ -572,33 +572,52 @@ function toError(
...
-            localHost: clientMetadata.localHost,
-            localPort: clientMetadata.localPort,
-            remoteHost: clientMetadata.remoteHost,
-            remotePort: clientMetadata.remotePort,
-            command: clientMetadata.command,
-          },
-          undefined,
-          {
-            cause: e,
+            cause: toError(errorData, clientMetadata, false),
```
tegefaulkes: We should be wrapping non polykey tree errors as a `ErrorPolykeyUnexpected` before serialising it in the `fromError`.

src/client/handlers/VaultsSecretsGet.ts
tegefaulkes, line 1: Why are you removing cancellation support here?

aryanjassal:
> All errors that come through the RPC are wrapped in an `ErrorPolykeyRemote`. I forget when this is done exactly so you'll need to check. I'm guessing it happens in the `toError` after the error has been deserialized.

I found the relevant part and modified the code to wrap an error inside `ErrorPolykeyUnexpected` if a non-Polykey error is thrown at the top of the stack, as that would be an unwrapped error. My question is, should the error be wrapped in a new error (like `ErrorPolykeyRemote > ErrorPolykeyUnexpected > ErrorEFS`) or be wrapped in only `ErrorRemotePolykey` (like `ErrorRemotePolykey > ErrorEFS`)?

src/errors.ts
```diff
@@ -11,6 +11,11 @@ class ErrorPolykeyUnknown<T> extends ErrorPolykey<T> {
   exitCode = sysexits.PROTOCOL;
 }
 
+class ErrorPolykeyUnexpected<T> extends ErrorPolykey<T> {
+  static description = 'Unknown error occurred';
+  exitCode = sysexits.PROTOCOL;
+}
+
```
tegefaulkes: `sysexits.PROTOCOL` is a bad code for this. Unknow errors are always `255` so it should use sysexits.UNKNOWN`.

tegefaulkes (changes requested):

src/client/handlers/VaultsSecretsGet.ts
aryanjassal: I believed that it was done incorrectly here, so I removed it. Now that I am going through and separating it into two functions, I can go in and properly implement it.

tegefaulkes: Just make sure that the handlers make use of the `ctx` and handle cancellation.

CMCDragonkai:
I think the `ErrorPolykeyRemote` is a generic rpc error that wraps any error coming from the other side. This is supposed to apply for any errors sent to the server side but this hasn't been implemented. See js-rpc issues.

Then what you're doing is wrapping something unknown. Which is not quite the same thing.

Point is an error can technically be anything. Any value of any type can be thrown in JS. Therefore ErrorPolykeyRemote can just wrap it as the cause.

The processor of ErrorPolykeyRemote can interpret anything and therefore should be able to deal with truly unknown things.

Then I believe we already have some sort of Unknown/Undefined/Unexpected sort of error class already defined in the errors.ts in the PK core library.

CMCDragonkai:
So always check what we already have before reinventing it. Reinvention leads to entropy.

aryanjassal:
> Then I believe we already have some sort of Unknown/Undefined/Unexpected sort of error class already defined in the errors.ts in the PK core library.

Yes, we do have an `ErrorPolykeyUnknown` error implemented, but it's description says `'Unable to deserialise to known error'`, so it's function is slightly different than what is being handled here, where the error was unexpected.

src/network/utils.ts
aryanjassal: So, basically, I would need to remove the special handling from `toError`, and instead catch the error being thrown in `fromError`, and if the error isn't a Polykey error, I wrap it in `ErrorPolykeyUnexpected`. That makes sense and keeps things simple.

CMCDragonkai:
Yea I think I meant that to mean I don't know what exception to deserialize to... that is over the RPC system, if you do `throw 123;` then `123` is the error that is unknown.

aryanjassal:
I have encountered an interesting issue with wrapping `throw 123`. When I throw regular errors, they parse properly, but when I throw a non-error (like `123`), I get this error:

```
    Received constructor: ErrorRPCParse

    Received message: "structure did not match a `JSONRPCResponse`"

          at parseJSONRPCResponse (node_modules/@matrixai/rpc/src/utils.ts:190:10)
          at TokenParser.parser.onValue (node_modules/@matrixai/rpc/src/middleware.ts:45:29)
          at TokenParser.emit (node_modules/@streamparser/json/src/tokenparser.ts:144:12)
          at TokenParser.pop (node_modules/@streamparser/json/src/tokenparser.ts:130:10)
          at TokenParser.write (node_modules/@streamparser/json/src/tokenparser.ts:276:16)
          at Tokenizer.write (node_modules/@streamparser/json/src/tokenizer.ts:193:20)
          at JSONParser.write (node_modules/@streamparser/json/src/jsonparser.ts:34:20)
          at Object.transform (node_modules/@matrixai/rpc/src/middleware.ts:54:8)
          at constructor_.readablePull (node_modules/@matrixai/ws/src/WebSocketStream.ts:380:16)
```

I am investigating the RPC code as I'm not sure what can be causing this from Polykey side, but it could be as easily caused by Polykey, too.

After analysing the incoming JSON responses, I got this:

(for `123`)
```json
{
  "jsonrpc": "2.0",
  "error": { 
    "code": -32006,
    "data": {
      "type": "ErrorPolykeyUnexpected",
      "data": {
        "message": "Unexpected error occured: 123",
        "timestamp":"2024-11-05T08:14:47.221Z",
        "data": { "reason": 123 },
        "stack": "ErrorPolykeyUnexpected: Unexpected error occured: 123\n    at constructor_.fromError (/home/aryanj/Projects/polykey/src/network/utils.ts:498:26)\n    at Object.pull (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/RPCServer.ts:331:28)",
        "description": "An error originating outside Polykey was thrown",
        "exitCode": 255
      }
    }
  },
  "id": null
}
```
(reason for error)
```
{
    jsonrpc: "2.0",
    error: {
      code: -32006,
      data: { type: "ErrorPolykeyUnexpected", data: [Object] }
  },
  id: null
}
```

(for `ErrorEFS`)
```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32006,
    "message": "ENOENT: no such file or directory, dir/secret-name",
    "data": {
      "type": "ErrorPolykeyUnexpected",
      "data": {
        "message": "Unexpected error occurred: ErrorEncryptedFSError",
        "timestamp": "2024-11-05T08:04:32.830Z",
        "data": {},
        "cause": {
          "type": "ErrorEncryptedFSError",
            "data": {
              "message": "ENOENT: no such file or directory, dir/secret-name",
              "timestamp":"2024-11-05T08:04:32.724Z",
              "data": {},
              "stack": "ErrorEncryptedFSError: ENOENT: no such file or directory, dir/secret-name\n    at constructor_._open (/home/aryanj/Projects/polykey/node_modules/encryptedfs/src/EncryptedFS.ts:2151:17)\n    at /home/aryanj/Projects/polykey/node_modules/encryptedfs/src/EncryptedFS.ts:1930:15\n    at Object.maybeCallback (/home/aryanj/Projects/polykey/node_modules/encryptedfs/src/utils.ts:405:12)\n    at /home/aryanj/Projects/polykey/node_modules/encryptedfs/src/EncryptedFS.ts:3421:19\n    at Object.maybeCallback (/home/aryanj/Projects/polykey/node_modules/encryptedfs/src/utils.ts:405:12)\n    at /home/aryanj/Projects/polykey/src/vaults/VaultOps.ts:306:5\n    at /home/aryanj/Projects/polykey/src/vaults/VaultInternal.ts:475:9\n    at withF (/home/aryanj/Projects/polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)\n    at withF (/home/aryanj/Projects/polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)\n    at Object.writeSecret (/home/aryanj/Projects/polykey/src/vaults/VaultOps.ts:305:3)\n    at /home/aryanj/Projects/polykey/src/client/handlers/VaultsSecretsWriteFile.ts:41:11\n    at /home/aryanj/Projects/polykey/src/vaults/VaultManager.ts:1032:14\n    at withF (/home/aryanj/Projects/polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)\n    at constructor_.withVaults (/home/aryanj/Projects/polykey/src/vaults/VaultManager.ts:1025:12)\n    at /home/aryanj/Projects/polykey/src/client/handlers/VaultsSecretsWriteFile.ts:38:7\n    at withF (/home/aryanj/Projects/polykey/node_modules/@matrixai/resources/src/utils.ts:24:12)\n    at handle (/home/aryanj/Projects/polykey/src/client/handlers/VaultsSecretsWriteFile.ts:27:5)\n    at wrapperDuplex (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/RPCServer.ts:398:15)\n    at outputGen (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/RPCServer.ts:304:26)\n    at Object.pull (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/RPCServer.ts:320:37)",
              "_errno": 34,
              "_code": "ENOENT",
              "_description": "no such file or directory",
              "_syscall":"open"
            }
          },
          "stack": "ErrorPolykeyUnexpected: Unexpected error occurred: ErrorEncryptedFSError\n    at constructor_.fromError (/home/aryanj/Projects/polykey/src/network/utils.ts:480:26)\n    at Object.pull (/home/aryanj/Projects/polykey/node_modules/@matrixai/rpc/src/RPCServer.ts:331:28)",
          "description": "An error originating outside Polykey was thrown",
          "exitCode": 255
        }
      }
    },
    "id": null
}
```

(reason for passing)
```
{
    jsonrpc: "2.0",
    error: {
      code: -32006,
      message: "ENOENT: no such file or directory, dir/secret-name",
      data: { type: "ErrorPolykeyUnexpected", data: [Object] }
    },
    id: null
  }
```

After investigating, for some reason, the `message` field is missing when I try to wrap non-errors in `ErrorPolykeyUnexpected`.

@tegefaulkes can you hint me in the right direction to resolve this? I have pushed up the latest code for review if required.

tegefaulkes:
I'm not sure if you need my direction on this. If `message` is required but missing in the RPC message then that's the problem. Look into why it's missing. If I had to guess, the `message` is expected to be defined but through some shenanigans its `undefined`. When serialised to JSON the field is just removed from the message.

CMCDragonkai:
Yes if a field is `undefined` value, JSON serialisation removes the key entirely.

The real question is why is it producing `undefined` at all, BEFORE it gets to the JSON serialisation.

aryanjassal:
```ts
// Extract from js-rpc (src/RPCServer.ts:317-346)
const reverseMiddlewareStream = new ReadableStream<JSONRPCResponse>({
  pull: async (controller) => {
    try {
      const { value, done } = await outputGenerator.next();
      if (done) {
        controller.close();
        return;
      }
      controller.enqueue(value);
    } catch (e) {
      try {
        const rpcError: JSONRPCResponseError = {
          code: errors.JSONRPCResponseErrorCode.RPCRemote,
          message: e.message,
          data: this.fromError(e),
        };
        const rpcErrorMessage: JSONRPCResponseFailed = {
          jsonrpc: '2.0',
          error: rpcError,
          id,
        };
        controller.enqueue(rpcErrorMessage);
      } catch (e) {
        this.dispatchEvent(
          new events.RPCErrorEvent({
            detail: e,
          }),
        );
        controller.error(e);
      }
    }
  }
});
```

I finally pinpointed the reason for this issue. And the issue is in `js-rpc` itself and not in my code.

In the snippet I have provided, you can see that there is a `try-catch` block to try and get the next value to read, or if an error occurs, then do the error serialisation. The issue is in this part.

When I do `throw 123`, it is caught by the catch block. As such, the value of `e` is `123`. When trying to serialise it, for the `message`, we assign it `e.message`. It would normally be defined for all `Error` instances, but this is not the case if we throw non-`Error` objects. This results in `message` being `undefined`, and being omitted from the message, and breaking the parsing on the other side.

I tried an easy fix here by adding a check for undefined values. It worked perfectly.

```ts
const rpcError: JSONRPCResponseError = {
  code: errors.JSONRPCResponseErrorCode.RPCRemote,
  message: e.message ?? '', // Use an empty string if the message is undefined
  data: this.fromError(e),
};
```

I will make a commit to `js-rpc` with this fix.

@CMCDragonkai @tegefaulkes

src/network/utils.ts
aryanjassal: Ah, I'll keep this in mind for the future. Right now, I have moved the wrapping logic to `fromError` so I don't need to worry about this anymore.

tegefaulkes:
Ah I see. The undefined message will need to be fixed in `rpc` and the `fromError` method. We don't want to just stub it as an empty string if there is no message. Provide a more descriptive message in that case. Like `thrown error was a literal ${e}` for primitive values.

src/client/handlers/VaultsSecretsGet.ts
aryanjassal: I'll just get this done as a part of #832

aryanjassal:
This PR kinda relies on https://github.com/MatrixAI/js-rpc/pull/69

CMCDragonkai:
Should the message be an empty string if the e.message doesn't exist? Shouldn't it be undefined? The e is therefore not an exception you'd have to handle this somewhere.

aryanjassal:
> Should the message be an empty string if the e.message doesn't exist? Shouldn't it be undefined? The e is therefore not an exception you'd have to handle this somewhere.

In cases where `e` isn't an actual exception, in `js-rpc`, the message is indeed `undefined`, which results in the field being omitted from being transferred, and the message not being properly parsable.

To deal with this, the PR in `js-rpc` actually intelligently creates a message if it doesn't exist on the actual error. For example, if we do `throw 123`, the generated message would be `'Non-error literal 123 was thrown'`. And if we do `throw new ReadableStream()`, then the error message reads, `'Non-error object ReadableStream was thrown'`.

This was done to support properly wrapping `throw 123` from the handler itself, as otherwise, doing this would result in an error on the library's side.

I can also implement something similar in Polykey itself, where it can intelligently synthesise an error message based on the error type and value.

aryanjassal:
This PR is basically done. This needs a final review, and if nothing crazy crops up in this, then we can get this merged. Merging this will allow resuming work on the CLI side, although the work there is also mostly done. All I had to do in the CLI is update the way these errors are rendered. Maybe that also needs a bit of discussion?

tegefaulkes (approved):

tegefaulkes:
I had a look. Nothing jumps out at me. Should be good to merge.

aryanjassal:
This PR has been approved and all the CI is also passing. Once https://github.com/MatrixAI/Polykey-CLI/pull/320 has also been approved, we can merge this.

PR #840, opened 2024-11-11
Fix undefined behaviour for `GestaltId` when starting agent

### LINKED ISSUE #797: Fix undefined behaviour error message coming from background tasks
### Specification

When starting an existing agent that hasn't run for a little while we get the following warning logger message.

image

We'll need to find the cause for this and fix it. There aren't many tasks that persist between restarts so it should be simple to narrow down. But in that vein we need to make the following changes to make these problems more clear.

1. The message should mention which task handler failed.
2. Any usage of `utils.never()` should include a descriptive message of what shouldn't be happening. so we need to update all usage of never in a task handler.

### Tasks

1. Update the warning log for a failed task handler to include more information about what task was running.
2. Update the `utils.never()` usages in task handlers to include proper failure descriptions.
3. Find and determine the failure occurring here and fix it.

aryanjassal:
### Description

When starting the agent after approximately 1 hour, an error is thrown like such:

```
WARN:polykey.PolykeyAgent.task v0ppdq4uibpo01417n072jft4io:Failed - Reason: ErrorUtilsUndefinedBehaviour("failed to decode vertex GestaltId "undefined"")
```

This is caused due to a minor typo in the discovery domain. This has been resolved in this PR, and similar messages should no longer happen.

### Issues Fixed

* Fixes #797 (REF ENG-400)

### Tasks

- [x] 1. Fix the undefined behaviour error

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -211,7 +211,7 @@ class Discovery {
       Date.now() - this.rediscoverVertexThresholdTime,
     );
     await this.taskManager.scheduleTask({
-      handlerId: this.discoverVertexHandlerId,
+      handlerId: this.checkRediscoveryHandlerId,
       path: [this.constructor.name, this.checkRediscoveryHandlerId],
       lazy: true,
       delay: this.rediscoverCheckIntervalTime,
@@ -285,7 +285,7 @@ class Discovery {
     );
     // Start up rediscovery task
     await this.taskManager.scheduleTask({
-      handlerId: this.discoverVertexHandlerId,
+      handlerId: this.checkRediscoveryHandlerId,
       path: [this.constructor.name, this.checkRediscoveryHandlerId],
       lazy: true,
       delay: this.rediscoverCheckIntervalTime,


aryanjassal:
This PR depends on https://github.com/MatrixAI/Polykey/pull/838.

After that PR has been merged, this can be too.

CMCDragonkai:
That's great how can we prevent regressions of this nature in the future? Types or fast check tests?

aryanjassal:
> That's great how can we prevent regressions of this nature in the future? Types or fast check tests?

https://github.com/MatrixAI/Polykey/issues/797#issuecomment-2467292822

tegefaulkes (approved):

aryanjassal:
I ran manual sanity checks for this change by printing the status of the actual task in `TaskManager.startTask()` and by setting the time interval to 1 second so it will trigger continuously. That resulted in zero errors and the rediscovery triggering each second.

This means that the changes made was successful in fixing this issue.

aryanjassal:
Now that the CI is passing and the changes are approved, I will be merging this and making a release.

PR #847, opened 2024-11-25
Allow vault `efs` resource acquisition to operate on multiple vaults in parallel

### LINKED ISSUE #821: Manage multiple transactions dynamically in the vault domain
### Specification

Currently, we have `VaultInternal.writeF` or `VaultInternal.writeG`, both of which are used to acquire transaction locks, and have a callback function which is executed. After that, the function performs some cleanup and releases the resource. Multiple resources can be requested at the same time, but they need to be known ahead of time.

This is not ideal. If we are streaming in multiple vault-name pairs, then multiple commits will be created for simple operations per vault. For example, if we open a new transaction to create a new file in a vault for each vault, then each file would make a new commit in each vault.

Ideally, all the required files should be created in a single commit per vault. This, however, is not possible as we don't know the vaults at the beginning if the vault-name pairs are being streamed in.

To solve this problem, we can group the inputs by vault names and acquire a single resource per vault, performing all operations in one go. However, this results in the input stream needing to be accumulated in-memory, and the modified in potentially unpredictable way (to the user/caller). This shuffling of inputs could cause issues down the line.

Rather, a new method needs to be added to the vaults domain which allows dynamically creating transactions and closing them all together. This will allow each vault to open a new transaction if it doesn't already exist, or operate on an existing transaction, allowing the input stream to be directly consumed instead of needing to be accumulated or sorted/grouped.

Note that reading from vaults like this isn't an issue, as a commit isn't created for reading, and we can safely spam `vault.readF` or `vault.readG` where needed. (see [**Polykey:src/client/handlers/VaultsSecretsGet:44-68**](https://github.com/MatrixAI/Polykey/blob/ab666368d9f1a46b226813c1c662a601840c2cdb/src/client/handlers/VaultsSecretsGet.ts#L44-L68)). As such, technically the reading version is not required, but can be implemented for completeness.

### Additional context

* Discussion about this problem and it's potential solutions [**Polykey#819 (comment)**](https://github.com/MatrixAI/Polykey/pull/819#issuecomment-2397892930)

### Tasks

1. Create dynamic `vault.write` method
2. Close all transactions elegantly on error or otherwise
3. Make sure RPC handlers like `VaultsSecretsRm` and `VaultsSecretsTouch` can utilise them

### LINKED ISSUE #837: `VaultsSecretsRemove` incorrectly reorders directories to be removed
### Describe the bug

In the handler, `VaultsSecretsRemove` performs grouping, which would result in the order of files being changed. This is done to ensure all files from a single vault is removed in one vault commit, but doesn't align with how Unix `rm` works.

Instead, we want to preserve the order of the files while removing them in a single commit. This is blocked by [Polykey#821](https://github.com/MatrixAI/Polykey/issues/821).

### To Reproduce

1. Call `VaultsSecretsRemove` with directory paths and mixed vault names
2. `VaultsSecretsRemove` groups them by vault, changing the order

### Expected behavior

The order should be preserved, still ensuring the operation finishes in one commit.

### Platform
 - Device: Dell Precision 3480
 - OS: NixOS 24.05
 - Version `0.12.0`

### Additional context

* Blocked by [Polykey#821](https://github.com/MatrixAI/Polykey/issues/821)

### Notify maintainers

@aryanjassal

aryanjassal:
### Description

For handlers like `VaultsSecretsCopy` and `VaultsSecretsMove`, we need to acquire a transaction across multiple vaults. This allows for all the given operations to be completed in a single transaction, avoiding polluting the commit space.

This PR creates such pattern, to acquire the resources from a vault dynamically in a way which works across multiple vaults at the same time.

### Issues Fixed

* Fixes #821 (FIX ENG-423)
* Fixes #837 (FIX ENG-450)

### Tasks

- [x] 1. Add `VaultInternal.acquireRead`
- [x] 2. Add `VaultInternal.acquireWrite`
- [x] 3. Add tests
- [x] 4. Update `VaultsSecretsRemove` to use this instead of grouping paths by vault name.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsSecretsRemove.ts
+++ b/src/client/handlers/VaultsSecretsRemove.ts
@@ -1,103 +1,204 @@
 import type { DB } from '@matrixai/db';
+import { ResourceAcquire, withG } from '@matrixai/resources';
 import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
-  SecretIdentifierMessage,
+  SecretsRemoveHeaderMessage,
+  SecretIdentifierMessageTagged,
   SuccessOrErrorMessage,
 } from '../types';
 import type VaultManager from '../../vaults/VaultManager';
 import { DuplexHandler } from '@matrixai/rpc';
 import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
+import * as clientErrors from '../errors';
+import { FileSystemWritable } from '../../vaults/types';
+import * as utils from '../../utils';
 
 class VaultsSecretsRemove extends DuplexHandler<
   {
     db: DB;
     vaultManager: VaultManager;
   },
-  ClientRPCRequestParams<SecretIdentifierMessage>,
+  ClientRPCRequestParams<
+    SecretsRemoveHeaderMessage | SecretIdentifierMessageTagged
+  >,
   ClientRPCResponseResult<SuccessOrErrorMessage>
 > {
   public handle = async function* (
-    input: AsyncIterable<ClientRPCRequestParams<SecretIdentifierMessage>>,
+    input: AsyncIterable<
+      ClientRPCRequestParams<
+        SecretsRemoveHeaderMessage | SecretIdentifierMessageTagged
+      >
+    >,
   ): AsyncGenerator<ClientRPCResponseResult<SuccessOrErrorMessage>> {
     const { db, vaultManager }: { db: DB; vaultManager: VaultManager } =
       this.container;
-    // Create a record of secrets to be removed, grouped by vault names
-    const vaultGroups: Record<string, Array<string>> = {};
-    const secretNames: Array<[string, string]> = [];
-    let metadata: any = undefined;
-    for await (const secretRemoveMessage of input) {
-      if (metadata == null) metadata = secretRemoveMessage.metadata ?? {};
-      secretNames.push([
-        secretRemoveMessage.nameOrId,
-        secretRemoveMessage.secretName,
-      ]);
-    }
-    secretNames.forEach(([vaultName, secretName]) => {
-      if (vaultGroups[vaultName] == null) {
-        vaultGroups[vaultName] = [];
+    const vaultAcquires: Array<ResourceAcquire<FileSystemWritable>> = [];
+    // Extracts the header message from the iterator
+    const headerMessage = await (async () => {
+      let header: SecretsRemoveHeaderMessage | undefined;
+      for await (const value of input) {
+        // TS cannot properly narrow down this type as it is too deeply wrapped.
+        // The as keyword is used to help it with type narrowing.
+        const message = value as
+          | SecretIdentifierMessageTagged
+          | SecretsRemoveHeaderMessage;
+        if (message.type === 'VaultNamesHeaderMesage') header = message;
+        break;
       }
-      vaultGroups[vaultName].push(secretName);
+      // The header message is mandatory
+      if (header == null) throw new clientErrors.ErrorClientInvalidHeader();
+      return header;
+    })();
+    // Create an array of write acquires
+    await db.withTransactionF(async (tran) => {
+      for (const vaultName of headerMessage!.vaultNames) {
+        const vaultIdFromName = await vaultManager.getVaultId(vaultName, tran);
+        const vaultId = vaultIdFromName ?? vaultsUtils.decodeVaultId(vaultName);
+        if (vaultId == null) {
+          throw new vaultsErrors.ErrorVaultsVaultUndefined(
+            `Vault ${vaultName} does not exist`,
+          );
+        }
+        await vaultManager.withVaults([vaultId], async (vault) => {
+          vaultAcquires.push(vault.acquireWrite());
+        });
+      }
     });
-    // Now, all the paths will be removed for a vault within a single commit
-    yield* db.withTransactionG(
-      async function* (tran): AsyncGenerator<SuccessOrErrorMessage> {
-        for (const [vaultName, secretNames] of Object.entries(vaultGroups)) {
-          const vaultIdFromName = await vaultManager.getVaultId(
-            vaultName,
-            tran,
-          );
-          const vaultId =
-            vaultIdFromName ?? vaultsUtils.decodeVaultId(vaultName);
-          if (vaultId == null) {
-            throw new vaultsErrors.ErrorVaultsVaultUndefined();
-          }
-          yield* vaultManager.withVaultsG(
-            [vaultId],
-            async function* (vault): AsyncGenerator<SuccessOrErrorMessage> {
-              yield* vault.writeG(
-                async function* (efs): AsyncGenerator<SuccessOrErrorMessage> {
-                  for (const secretName of secretNames) {
-                    try {
-                      const stat = await efs.stat(secretName);
-                      if (stat.isDirectory()) {
-                        await efs.rmdir(secretName, {
-                          recursive: metadata?.options?.recursive,
-                        });
-                      } else {
-                        await efs.unlink(secretName);
-                      }
-                      yield {
-                        type: 'success',
-                        success: true,
-                      };
-                    } catch (e) {
-                      if (
-                        e.code === 'ENOENT' ||
-                        e.code === 'ENOTEMPTY' ||
-                        e.code === 'EINVAL'
-                      ) {
-                        // INVAL can be triggered if removing the root of the
-                        // vault is attempted.
-                        yield {
-                          type: 'error',
-                          code: e.code,
-                          reason: secretName,
-                        };
-                      } else {
-                        throw e;
-                      }
-                    }
-                  }
-                },
+    // Acquire all locks in parallel and perform all operations at once
+    yield* withG(
+      vaultAcquires,
+      async function* (efses): AsyncGenerator<SuccessOrErrorMessage> {
+        // Creating the vault name to efs map for easy access
+        const vaultMap = new Map<string, FileSystemWritable>();
+        for (let i = 0; i < efses.length; i++) {
+          vaultMap.set(headerMessage!.vaultNames[i], efses[i]);
+        }
+        for await (const value of input) {
+          // TS cannot properly narrow down this type as it is too deeply wrapped.
+          // The as keyword is used to help it with type narrowing.
+          const message = value as
+            | SecretIdentifierMessageTagged
+            | SecretsRemoveHeaderMessage;
+          // Ignoring any header messages
+          if (message.type === 'SecretIdentifierMessage') {
+            const efs = vaultMap.get(message.nameOrId);
+            if (efs == null) {
+              throw new vaultsErrors.ErrorVaultsVaultUndefined(
+                `Vault ${message.nameOrId} was not present in the header message`,
               );
-            },
-            tran,
-          );
+            }
+            try {
+              const stat = await efs.stat(message.secretName);
+              if (stat.isDirectory()) {
+                await efs.rmdir(message.secretName, {
+                  recursive: headerMessage.recursive,
+                });
+              } else {
+                await efs.unlink(message.secretName);
+              }
+              yield {
+                type: 'success',
+                success: true,
+              };
+            } catch (e) {
+              if (
+                e.code === 'ENOENT' ||
+                e.code === 'ENOTEMPTY' ||
+                e.code === 'EINVAL'
+              ) {
+                // INVAL can be triggered if removing the root of the
+                // vault is attempted.
+                yield {
+                  type: 'error',
+                  code: e.code,
+                  reason: message.secretName,
+                };
+              } else {
+                throw e;
+              }
+            }
+          }
         }
       },
     );
+
+    // Create a record of secrets to be removed, grouped by vault names
+    // const vaultGroups: Record<string, Array<string>> = {};
+    // const secretNames: Array<[string, string]> = [];
+    // let metadata: any = undefined;
+    // for await (const secretRemoveMessage of input) {
+    //   if (metadata == null) metadata = secretRemoveMessage.metadata ?? {};
+    //   secretNames.push([
+    //     secretRemoveMessage.nameOrId,
+    //     secretRemoveMessage.secretName,
+    //   ]);
+    // }
+    // secretNames.forEach(([vaultName, secretName]) => {
+    //   if (vaultGroups[vaultName] == null) {
+    //     vaultGroups[vaultName] = [];
+    //   }
+    //   vaultGroups[vaultName].push(secretName);
+    // });
+    // Now, all the paths will be removed for a vault within a single commit
+    // yield* db.withTransactionG(
+    //   async function* (tran): AsyncGenerator<SuccessOrErrorMessage> {
+    //     for (const [vaultName, secretNames] of Object.entries(vaultGroups)) {
+    //       const vaultIdFromName = await vaultManager.getVaultId(
+    //         vaultName,
+    //         tran,
+    //       );
+    //       const vaultId =
+    //         vaultIdFromName ?? vaultsUtils.decodeVaultId(vaultName);
+    //       if (vaultId == null) {
+    //         throw new vaultsErrors.ErrorVaultsVaultUndefined();
+    //       }
+    //       yield* vaultManager.withVaultsG(
+    //         [vaultId],
+    //         async function* (vault): AsyncGenerator<SuccessOrErrorMessage> {
+    //           yield* vault.writeG(
+    //             async function* (efs): AsyncGenerator<SuccessOrErrorMessage> {
+    //               for (const secretName of secretNames) {
+    //                 try {
+    //                   const stat = await efs.stat(secretName);
+    //                   if (stat.isDirectory()) {
+    //                     await efs.rmdir(secretName, {
+    //                       recursive: metadata?.options?.recursive,
+    //                     });
+    //                   } else {
+    //                     await efs.unlink(secretName);
+    //                   }
+    //                   yield {
+    //                     type: 'success',
+    //                     success: true,
+    //                   };
+    //                 } catch (e) {
+    //                   if (
+    //                     e.code === 'ENOENT' ||
+    //                     e.code === 'ENOTEMPTY' ||
+    //                     e.code === 'EINVAL'
+    //                   ) {
+    //                     // INVAL can be triggered if removing the root of the
+    //                     // vault is attempted.
+    //                     yield {
+    //                       type: 'error',
+    //                       code: e.code,
+    //                       reason: secretName,
+    //                     };
+    //                   } else {
+    //                     throw e;
+    //                   }
+    //                 }
+    //               }
+    //             },
+    //           );
+    //         },
+    //         tran,
+    //       );
+    //     }
+    //   },
+    // );
   };
 }
 
--- a/src/client/handlers/VaultsSecretsRemove.ts
+++ b/src/client/handlers/VaultsSecretsRemove.ts
@@ -35,11 +35,13 @@ class VaultsSecretsRemove extends DuplexHandler<
     const { db, vaultManager }: { db: DB; vaultManager: VaultManager } =
       this.container;
     // Extract the header message from the iterator
+    const headerMessagePair = await input.next();
     const headerMessage:
       | SecretsRemoveHeaderMessage
-      | SecretIdentifierMessageTagged = (await input.next()).value;
+      | SecretIdentifierMessageTagged = headerMessagePair.value;
+    // Testing if the header is of the expected format
     if (
-      headerMessage == null ||
+      headerMessagePair.done ||
       headerMessage.type !== 'VaultNamesHeaderMessage'
     ) {
       throw new clientErrors.ErrorClientInvalidHeader();
@@ -72,10 +74,12 @@ class VaultsSecretsRemove extends DuplexHandler<
         for (let i = 0; i < efses.length; i++) {
           vaultMap.set(headerMessage!.vaultNames[i], efses[i]);
         }
+        let loopRan = false;
         for await (const message of input) {
-          // Ignoring any header messages
+          loopRan = true;
+          // Header messages should not be seen anymore
           if (message.type === 'VaultNamesHeaderMessage') {
-            throw new clientErrors.ErrorClientInvalidHeader(
+            throw new clientErrors.ErrorClientProtocolError(
               'The header message cannot be sent multiple times',
             );
           }
@@ -115,6 +119,12 @@ class VaultsSecretsRemove extends DuplexHandler<
               throw e;
             }
           }
+        }
+        // Content messages must follow header messages
+        if (!loopRan) {
+          throw new clientErrors.ErrorClientProtocolError(
+            'No content messages followed header message',
+          );
         }
       },
     );
--- a/src/client/handlers/VaultsSecretsRemove.ts
+++ b/src/client/handlers/VaultsSecretsRemove.ts
@@ -1,5 +1,5 @@
 import type { DB } from '@matrixai/db';
-import { ResourceAcquire, withG } from '@matrixai/resources';
+import type { ResourceAcquire } from '@matrixai/resources';
 import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
@@ -8,12 +8,12 @@ import type {
   SuccessOrErrorMessage,
 } from '../types';
 import type VaultManager from '../../vaults/VaultManager';
+import type { FileSystemWritable } from '../../vaults/types';
+import { withG } from '@matrixai/resources';
 import { DuplexHandler } from '@matrixai/rpc';
 import * as vaultsUtils from '../../vaults/utils';
 import * as vaultsErrors from '../../vaults/errors';
 import * as clientErrors from '../errors';
-import { FileSystemWritable } from '../../vaults/types';
-import * as utils from '../../utils';
 
 class VaultsSecretsRemove extends DuplexHandler<
   {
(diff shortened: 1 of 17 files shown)


tests/client/handlers/vaults.test.ts
```diff
@@ -2371,7 +2374,19 @@ describe('vaultsSecretsRemove', () => {
...
+    await writer.write({
+      type: 'VaultNamesHeaderMesage',
+      vaultNames: ['invalid'],
+    });
+    // Content messages
+    await writer.write({
+      type: 'SecretIdentifierMessage',
+      nameOrId: 'invalid',
+      secretName: 'invalid',
+    });
```
aryanjassal: While trying to use the new resource acquisition pattern to update `VaultsSecretsMkdir`, I ran into an issue.

I need to send over the vault names first, before sending any data. To do this, I send a header message before sending the data. This is reflected as the type `SecretsRemoveHeaderMessage | SecretIdentifierMessage`. However, similar to the issue with `SuccessOrErrorMessage`, when the type is wrapped inside `ClientRPCResponseResult` or `ClientRPCRequestParams`, the type narrowing doesn't work properly, and only the common types are parsed into the resultant type. In this case, only the `metadata` and the `type` fields will exist on the wrapped type. When trying to pass a value into the writer, it complains of invalid type.

This was handled before using the `as` keyword to help typescript with type narrowing. Here, we can define a variable with the relevant type, like `SecretIdentifierMessage`, and pass that into the writer. That works, but is far from ideal and is unmaintainable. The type wrapper is used only to inject a `metadata` field into the input and the output. If we remove the type wrapper, everything works perfectly and typescript can properly narrow the types down, too.

This is a major issue, as in the future when we implement progress packets for long running async calls, then we will run into this issue, too. Previously, as this issue had limited impact, workarounds were used, but now that this issue can potentially impact a major part of the codebase, this approach needs to be reevaluated.

After having a discussion with @tegefaulkes, we happened upon some potential solutions. 

A solution is to remove the wrapper and just pass in the types as-is. This will ensure the type narrowing works perfectly, but might need some additional work and casting or using the `as` keyword if we need to access the `metadata` field, and as Polykey CLI relies on the `metadata` to pass in the authentication information, that might break, too. This would also mean a refactor of all RPC handlers to align with this in both Polykey and Polykey CLI (where relevant). This will be a very annoying and lengthy change.

Another solution could be to update the way we define the utility wrapper. For example, instead of injecting the metadata field, we can do a type union between whatever type is expected and a type which provides the `metadata` field. This would help us retain the current behaviour and potentially fix the issues we are having with type narrowing.

What do you propose we do about this, @CMCDragonkai?

CMCDragonkai: Describe the entire problem in isolation - like give an self contained example of this problem, and there should be some type assertions that can solve this.

aryanjassal: The first line in the reviewed code is basically self-contained example of this problem.

```ts

type SecretIdentifierMessageTagged = {
  type: 'SecretIdentifierMessage';
  secretName: string;
}

type VaultNamesHeaderMessage = {
  type: 'VaultNamesHeaderMessage';
  vaultNames: Array<string>;
};

// This will fail because the only fields that exist on
// the new type is `metadata` and `type`, which are
// common to both types. The field `vaultNames` is
// invalid from this perspective.
let variable1: ClientRPCRequestParams<
  | SecretsRemoveHeaderMessage
  | SecretIdentifierMessage
> = {
  type: 'VaultNamesHeaderMesage', 
  vaultNames: ['invalid']
};

// This will succeed. The types are correctly formed.
let variable2:
  | SecretsRemoveHeaderMessage
  | SecretIdentifierMessage
= {
  type: 'VaultNamesHeaderMesage', 
  vaultNames: ['invalid']
};

// So, writing the valid JSON object via the writer 
// will complain, but writing the second variable 
// won't. This means that we need to update a variable 
// and pass in that variable to the writer instead.
writer.write({
  type: 'VaultNamesHeaderMesage',
  vaultNames: ['invalid'],
}); // Complains of incorrect types
writer.write(variable2); // Works
```

aryanjassal: # Approaches

## Removing `ClientRPC*` wrapper

The `ClientRPC*` type is a utility type which injects a `metadata` field into the request and response. This solution proposes removal of that wrapper to directly return the required type without wrapping. This would clean up the type and resolve the type narrowing issue.

However, doing so would see a complete inability to pass or obtain the metadata. Polykey CLI relies on this metadata to pass on authorisation data, and this would basically brick the commands, unless there is an alternative way to provide authentication. It would also involve massive refactor of the entire code base for all RPC handlers, which also seems kind of excessive for the benefits provided by this approach.

## Type Unions

As the purpose of `ClientRPC*` is just to attach the metadata field, the request and response type itself can be unified with a type which contains the metadata.

```ts
// I'm not sure exactly how this would work. 
// This is just an example.
type ClientRPCRequestParams extends JSONObject = {
  metadata?: JSONObject;
}

type SuccessMessage = {
  success: boolean;
}

type SecretsRemoveMessage = ClientRPCRequestParams & SuccessMessage;
```

While this would still require the refactor in most of the code base, this seems to be the more permanent and lasting solution as compared to ignoring the metadata altogether. However, I'm not sure if this idea will even work, or how effective it would be.

## Modification of the `JSONRPC*` type

The way `js-rpc` has structured its `JSONRPC*` wrapper type disallows for this complex type nesting. That type can be modified to something more functional. This is what I could come up with ChatGPT's help.

```ts
type JSONRPCResponseResultNew<
  T extends JSONObject,
  M extends JSONObject,
> = T extends infer U
  ? U extends { metadata?: JSONObject }
    ? {
        metadata?: JSONObject &
          JSONRPCResponseMetadata &
          Omit<U['metadata'] & M, keyof JSONRPCResponseMetadata>;
      } & Omit<U, 'metadata'>
    : U
  : never;
```

This new and revised type allows for proper type narrowing and can replace the `JSONRPCResponseResult` from `js-rpc` without any additional setup. However, this is very untested, and the changes might be catastrophic to some other commands or use cases. More investigation needs to be done for this solution if this is to be used.

However, this solution is the most complete and robust solution, ensuring robust typing for the RPC. This solution also needs minimal changes, as the only change is required in the `js-rpc` side, making this easy to implement and maintain.

## Unions of `ClientRPC*`

While testing, I realised that if two `ClientRPC*` types are unioned with a single wrapped type, then type narrowing works perfectly.

```ts
// This does not work
let var1: ClientRPCRequestParams<
  | SuccessMessage 
  | ErrorMessage
> = { type: 'success', success: true };

// This works just fine
let var1: 
  | ClientRPCRequestParams<SuccessMessage>
  | ClientRPCRequestParams<ErrorMessage>
> = { type: 'success', success: true };
```

This solution does not require significant changes and this can be used wherever multiple types are required, making this solution easy to implement and use. This does not require massive changes to the entire code base, and only RPC handlers which take in or return multiple types can use this approach selectively.

However, this solution needs all the wrapped types to not be a union themselves. In this case, the type is simple, but in other cases where the type is more complex and defined in `types.ts`, this solution will fail to work. In the client domain, we currently have a `SuccessOrErrorMessage` type which is a union between `SuccessMessage` and `ErrorMessage`. That type will not work anymore because the types need to be isolated. This will cause a lot of code repetition, less maintainability, and higher changes of mistakes.

# Conclusion

All approaches have their own pros and cons. As a temporary solution, manual unions between `ClientRPC*` types can be done, but for a permanent, long-term, and robust solution, the `ClientRPC*` wrapper type itself should be modified to support complex types.

We need to figure out a solution and implement it @tegefaulkes. Which solution do you think I should focus on?

CMCDragonkai: In the case of RPC level things where things are not self-contained, it is not a good idea to choose temporary fixes, you must always consider long term robust solutions as these are not self-contained situations, and it becomes very difficult to fix later in the future.

aryanjassal: Currently, I have switched to using this type instead of `js-rpc` provided one.

```ts
type JSONRPCResponseResult<
  T extends JSONObject = JSONObject,
  M extends JSONObject = JSONObject,
> = T & {
  metadata?: JSONRPCResponseMetadata &
    M &
    (T extends { metadata: infer U } ? U : {}) &
    JSONObject;
};
```
Writing it this way solves all the issues and simplifies the code flow. Now, there is no need to do `as` casting for union types because this version can automatically narrow them down, too. This type is used as follows.
```ts
type ClientRPCRequestParams<T extends JSONObject = JSONObject> =
  JSONRPCResponseResult<
    T,
    Partial<{
      authorization: string;
    }>
  >;

type ClientRPCResponseResult<T extends JSONObject = JSONObject> =
  JSONRPCResponseResult<
    T,
    Partial<{
      authorization: string;
    }>
  >;
```
For reference, this was the old type.
```ts
type JSONRPCRequestParams<
  T extends JSONObject = JSONObject,
  M extends JSONObject = JSONObject,
> = {
  metadata?: JSONObject &
    JSONRPCRequestMetadata &
    Omit<T['metadata'] & M, keyof JSONRPCRequestMetadata>;
} & Omit<T, 'metadata'>;
```
Note that the latest type is work-in-progress, and does not properly handle field conflicts from the message type. This needs to be changed and updated before this change can be merged.

aryanjassal:
The tasks in this PR have been completed except for deciding upon a proper fix for the RPC wrapper type. For now, I have written a custom type which adds metadata field while allowing for better type narrowing. (see https://github.com/MatrixAI/Polykey/pull/847#discussion_r1860030124). However, this is a temporary fix and not final.

I will need to have a discussion if there are any obvious caveats or things I am missing. Once that is no longer an issue, either the type update can be made local to Polykey (not recommended), or made available globally to `js-rpc` (recommended, as this type might be optionally used by Polykey CLI, and keeping a RPC type inside Polykey doesn't make sense)

tegefaulkes:
If the RPC type problem is just a usage problem for now then just deal with it using the `as` keyword and comment why. Move discussion to a new issue directly addressing it.

aryanjassal:
> If the RPC type problem is just a usage problem for now then just deal with it using the `as` keyword and comment why. Move discussion to a new issue directly addressing it.

I believe we discussed why this issue can't be resolved by usage of the `as` keyword. I am asking because I have updated the RPC type to the new one which works in this case, but might not be consistent with other RPCs. SO, I just want to find a solution to this which can be applied globally instead of a targeted fix.

aryanjassal:
I ran a bunch of tests to see if there was a difference between the new type and the old type.

```ts
// Original type currently in RPC
type ParamsTypeA<
  T extends JSONObject = JSONObject,
  M extends JSONObject = JSONObject,
> = {
  metadata?: JSONObject &
    RPCMetadata &
    Omit<T["metadata"] & M, keyof RPCMetadata>;
} & Omit<T, "metadata">;

// My type updated type which works on complex types
type ParamsTypeB<
  T extends JSONObject = JSONObject,
  M extends JSONObject = JSONObject,
> = T & {
  metadata?: RPCMetadata &
    M &
    (T extends { metadata: infer U } ? U : JSONObject);
};

// Types being tested

type NoMetadata = {
  text1: string;
  text2: number;
};

type NonConflictingMetadata = {
  text1: boolean;
  metadata: {
    available: boolean;
  };
};

type ConflictingMetadataDifferentType = {
  text1: number;
  metadata: {
    authorization: number;
  };
};

type ConflictingMetadataSameType = {
  text1: number;
  metadata: {
    authorization: string;
  };
};

type TA = {
  type: true;
  field: string;
};

type TB = {
  type: false;
  value: number;
};

type CombinedType = TA | TB;

// Actual tests

// UsingTypeA is old type currently in RPC
// UsingTypeB is new type I'm working on for this PR

let a1: UsingTypeA<NoMetadata> = { text1: "123", text2: 456 };
let a2: UsingTypeB<NoMetadata> = { text1: "123", text2: 456 };

let b1: UsingTypeA<NonConflictingMetadata> = {
  text1: true,
  metadata: { available: true },
};
let b2: UsingTypeB<NonConflictingMetadata> = {
  text1: true,
  metadata: { available: true },
};

let c1: UsingTypeA<ConflictingMetadataDifferentType> = {
  text1: 1,
  metadata: { authorization: "bearer" },  // Type string is not assignable to never
};
let c2: UsingTypeB<ConflictingMetadataDifferentType> = {
  text1: 1,
  metadata: { authorization: "bearer" },  // Type string is not assignable to never
};

let d1: UsingTypeA<ConflictingMetadataSameType> = {
  text1: 1,
  metadata: { authorization: "bearer" },
};
let d2: UsingTypeB<ConflictingMetadataSameType> = {
  text1: 1,
  metadata: { authorization: "bearer" },
};

let e11: UsingTypeA<CombinedType> = { type: true, field: "123" };  // Fails because type narrowing fails
let e12: UsingTypeA<CombinedType> = { type: false, value: 123 };  // Fails because type narrowing fails
let e21: UsingTypeB<CombinedType> = { type: true, field: "123" };
let e22: UsingTypeB<CombinedType> = { type: false, value: 123 };
```

The results of all the tests are the same (ie, `*1` and `*2` match in their outputs) except for `e*` tests. As we can see here, the behaviour is basically similar except for having complex unified types, which is when `UsingTypeB` (which is the new implementation) shines.

This is the full test file for reference
[test.ts](https://github.com/user-attachments/files/17954282/test.txt)

aryanjassal:
I ran a local update on `js-rpc` and tried building it. It built just fine. I tried to modify the `types.d.ts` in the distributable in both Polykey and Polykey CLI, and they both also built just fine.

Side note, however, is that Brian's IDE doesn't use the simple TS compiler for warnings/errors. As such, some errors are able to fall through. These errors also don't block the building, so the package is built and run just fine, but the errors does exist.

```
polykey • feature-multiple-vault-resource • nix-shell-env
-> tsc
tests/audit/utils.test.ts:50:48 - error TS2345: Argument of type 'string[][]' is not assignable to parameter of type 'TopicSubPath[]'.
  Type 'string[]' is not assignable to type 'TopicSubPath'.
    Type 'string[]' is not assignable to type '["discovery", "checkRediscovery"]'.
      Target requires 2 element(s) but source may have fewer.

50     const filtered = auditUtils.filterSubPaths(data).map((v) => v.join('.'));
                                                  ~~~~

tests/client/handlers/nodes.test.ts:828:11 - error TS2345: Argument of type '{ nodeGraph: NodeGraph; keyRing: KeyRing; }' is not assignable to parameter of type '{ nodeGraph: NodeGraph; }'.
  Object literal may only specify known properties, and 'keyRing' does not exist in type '{ nodeGraph: NodeGraph; }'.

828           keyRing,
              ~~~~~~~

tests/git/utils.ts:128:85 - error TS2339: Property 'type' does not exist on type 'never'.

128         `data.type must be "want", "have", "SEPARATOR", "done", "none", got "${data.type}"`,
                                                                                        ~~~~


Found 3 errors in 3 files.

Errors  Files
     1  tests/audit/utils.test.ts:50
     1  tests/client/handlers/nodes.test.ts:828
     1  tests/git/utils.ts:128

polykey • feature-multiple-vault-resource • nix-shell-env
-> npm run build

> polykey@1.16.0 build
> shx rm -rf ./dist && tsc -p ./tsconfig.build.json


> polykey@1.16.0 postbuild
> shx cp -f src/status/*.json dist/status/


polykey • feature-multiple-vault-resource • nix-shell-env
-> 
```

When I open the relevant files, I do see the errors just fine. This can be replicated on both Neovim and VS Code. We might need another step in our development scripts which runs `tsc` as a step to confirm no warnings or errors exist.

@tegefaulkes @CMCDragonkai

aryanjassal:
Oh shoot, I completely forgot `js-rpc` does not have automatic publishing for its CI. @tegefaulkes you need to do a manual release on Monday for this.

CMCDragonkai:
Why doesn't it? Get it sorted!

aryanjassal:
> Oh shoot, I completely forgot `js-rpc` does not have automatic publishing for its CI. @tegefaulkes you need to do a manual release on Monday for this.

https://github.com/MatrixAI/js-rpc/issues/70

The CI isn't working properly yet, as some tests were failing, so they were commented out. This needs to be resolved on high priority. Moreover, the CI isn't automatically publishing a release. This needs to be investigated, @brynblack.

tests/client/handlers/vaults.test.ts
aryanjassal: This change has been tested in `js-rpc` and it works. As such, this thread is now resolved.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -1,100 +1,114 @@
...
-      vaultGroups[vaultName].push(secretName);
-    });
-    // Now, all the paths will be removed for a vault within a single commit
-    yield* db.withTransactionG(
-      async function* (tran): AsyncGenerator<SuccessOrErrorMessage> {
-        for (const [vaultName, secretNames] of Object.entries(vaultGroups)) {
-          const vaultIdFromName = await vaultManager.getVaultId(
-            vaultName,
-            tran,
+    })();
```
tegefaulkes: This is a very weird way of doing this. The input is type of `AsyncIterableIterator` so you don't need to do `const iterator = input[Symbol.asyncIterator]();` to be able to use `input.next()`. 

You don't need to wrap this in an async arrow function and call it like you're doing here. It's completely unnecessary.

Also don't check if `header == null`, you should be checking the `next.done` if the iterator finished too early.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -1,100 +1,114 @@
...
-          const vaultId =
-            vaultIdFromName ?? vaultsUtils.decodeVaultId(vaultName);
-          if (vaultId == null) {
-            throw new vaultsErrors.ErrorVaultsVaultUndefined();
+        }
+        const acquire = await vaultManager.withVaults(
+          [vaultId],
+          async (vault) => vault.acquireWrite(),
+        );
+        vaultAcquires.push(acquire);
```
tegefaulkes, line 64: IT's better to have `vaultsAcquires` initialised and returned buy this `withTransactionF` callback.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -1,100 +1,114 @@
...
+      vaultAcquires,
+      async function* (efses): AsyncGenerator<SuccessOrErrorMessage> {
+        // Creating the vault name to efs map for easy access
+        const vaultMap = new Map<string, FileSystemWritable>();
+        for (let i = 0; i < efses.length; i++) {
+          vaultMap.set(headerMessage!.vaultNames[i], efses[i]);
+        }
+        for await (const message of input) {
+          // Ignoring any header messages
+          if (message.type !== 'SecretIdentifierMessage') continue;
```
tegefaulkes: If we get a header message out of order then that's an explicit deviation from expected protocol and should be an error.

src/vaults/VaultInternal.ts
```diff
@@ -536,6 +537,86 @@ class VaultInternal {
...
+      const [release] = await acquire();
+      if (tran == null) {
+        const acquireTran = this.db.transaction();
+        [releaseTran, tran] = await acquireTran();
+      }
+      // The returned transaction can be undefined, too. We won't handle those
+      // cases.
+      if (tran == null) utils.never();
+      await tran.lock(
+        [...this.vaultMetadataDbPath, VaultInternal.dirtyKey].join(''),
```
tegefaulkes: Always give `never()` a message. I should change the types to enforce it.

tests/client/handlers/vaults.test.ts
```diff
@@ -1483,11 +1482,9 @@ describe('vaultsSecretsMkdir', () => {
     await writer.close();
     for await (const data of response.readable) {
       expect(data.type).toEqual('error');
-      // TS cannot properly evaluate a type as nested as this, so we use the
-      // as keyword to help it. Inside this block, the type of data is 'error'.
-      const error = data as ErrorMessage;
-      expect(error.code).toEqual('ENOENT');
-      expect(error.reason).toEqual(dirPath);
+      if (data.type !== 'error') utils.never();
+      expect(data.code).toEqual('ENOENT');
```
tegefaulkes: Same here, give your `never()` a message. Saves a lot of headache when debugging.

tests/client/handlers/vaults.test.ts
```diff
@@ -1553,11 +1550,8 @@ describe('vaultsSecretsMkdir', () => {
...
     for await (const data of response.readable) {
       if (data.type === 'error') {
-        // TS cannot properly evaluate a type as nested as this, so we use the
-        // as keyword to help it. Inside this block, the type of data is 'error'.
-        const error = data as ErrorMessage;
-        expect(error.code).toEqual('ENOENT');
-        expect(error.reason).toEqual(dirPath3);
+        expect(data.code).toEqual('ENOENT');
+        expect(data.reason).toEqual(dirPath3);
       }
```
tegefaulkes: This test fill fail to catch anything wrong if an error was never returned. You should explicitly check if an error is returned and then explicitly check that it continues as expected.

tegefaulkes (changes requested):

src/client/handlers/VaultsSecretsRemove.ts
aryanjassal: I'm pretty sure the input is of type `AsyncIterable`, and it does not have the `next()` method on it. This is why I had to do it in such a roundabout way.

image

(`Text` is not actual code autocomplete. It's just picking up relevant words from the file)

aryanjassal: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol

Only iterators have the `next` method. Iterables are meant to be consumed via a `for await ... of ...` loop, but I got backpressure and buffer overflow if I did this, as I had to stop reading from the stream prematurely.

To avoid this issue, I only read the first value using the async iterator symbol which circumvented this issue.

tegefaulkes: `AsyncIterableIterator` has the `next` method. You;'re using `AsyncIterable` which does not. But under the hood it's just an `AsyncGenerator` so the `next` method is avaliable.

aryanjassal: Ohh, that makes sense. Basically, the type is an `AsyncGenerator` which can be cast to other types like this.

However, I'm pretty sure in other RPC calls, we have kept the type as `AsyncIterator`, which is where I copied the boilerplate from. Do I make this change everywhere, or only here?

tegefaulkes: You can look into updating all the places we use it. I think there are only 4 spots but double check.

CMCDragonkai:
Status of this? If @brynblack is not available, see if you can do the work to get it published.

aryanjassal:
> Status of this? If @brynblack is not available, see if you can do the work to get it published.

Well, the `js-rpc` was merged and released by Brian, so this isn't an issue anymore. I can't do a release as I don't have the relevant secrets or knowledge of making a release.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -1,100 +1,120 @@
...
-        secretRemoveMessage.secretName,
-      ]);
+    // Extract the header message from the iterator
+    const headerMessage:
+      | SecretsRemoveHeaderMessage
+      | SecretIdentifierMessageTagged = (await input.next()).value;
+    if (
+      headerMessage == null ||
+      headerMessage.type !== 'VaultNamesHeaderMessage'
+    ) {
```
tegefaulkes: Don't do this, you need to use the `nextValue.done` to see if it has ended. If it ended early that should be a different ending early error as well.

src/client/handlers/VaultsSecretsRemove.ts
```diff
@@ -1,100 +1,120 @@
...
+        for (let i = 0; i < efses.length; i++) {
+          vaultMap.set(headerMessage!.vaultNames[i], efses[i]);
+        }
+        for await (const message of input) {
+          // Ignoring any header messages
+          if (message.type === 'VaultNamesHeaderMessage') {
+            throw new clientErrors.ErrorClientInvalidHeader(
+              'The header message cannot be sent multiple times',
+            );
+          }
```
tegefaulkes: This should be a protocol error. Don't reuse the invalid header error. It's confusing since the fact that its a header here is the problem.

src/git/utils.ts
```diff
@@ -218,7 +218,7 @@ async function listObjects({
         }
         return;
       default:
-        utils.never();
+        utils.never('Invalid type');
```
tegefaulkes: Include what the type was that is invalid. Also include what the allowed types are in the message.

src/vaults/VaultInternal.ts
```diff
@@ -536,6 +537,86 @@ class VaultInternal {
...
+              // After doing mutation we need to commit the new history
+              await this.createCommit();
+            } catch (e_) {
+              e = e_;
+            }
+            // This would happen if an error was caught inside the catch block
+            if (e != null) {
+              // Error implies dirty state
+              await this.cleanWorkingDirectory();
+            }
```
tegefaulkes: Why not just put the `await this.cleanWorkingDirectory();`  inside the catch block then?

tegefaulkes (changes requested):

src/client/handlers/VaultsSecretsRemove.ts
aryanjassal: What error could this be, then? Currently we have `ErrorClientInvalidHeader` for invalid header messages. I will be making a `ErrorClientInvalidProtocol` for sending header messages when a content message is expected. Do I make another error `ErrorClientClosedEarly` for this case when the stream is closed before any data was sent?

src/client/handlers/VaultsSecretsRemove.ts
aryanjassal: I also forgot to add tests for the new header message protocol we are using, so I will add them too before merging.

src/vaults/VaultInternal.ts
aryanjassal: Yes, I can do that, but the error re-assignment (`e = e_`) still needs to be done, as we will be passing any error that happens straight to `release(e)`

src/client/handlers/VaultsSecretsRemove.ts
tegefaulkes: Make it `ErrorClientProtocolError` for both and differentiate using the error message.

tegefaulkes (approved):

PR #870, opened 2025-02-07
feature/ENG-527 Adding agent RPC handlers for getting sigchain and audit data.

### LINKED ISSUE #865: Adding agent RPC handlers for getting sigchain and audit data.
### Specification

This is a simple issue. We just need to make sure that the agent service provides RPC methods for getting the following.
1. Getting audit event data.
2. Getting sigchain data.

I believe we already have a handler for getting the sigchain data. This is used by the Polykey `Discovery` domain for constructing the gestalts. We need to add a method for getting the audit event data. This handler will need to take `seek`, `seekTo,` and `limit` parameters to enable basic paging. This should enable the ability to get all the audit data for a node.

Ideally only a node with a valid `NetworkAuthorityClaim` should be able to get this data from a node. But that may need to be implemented later down the line when authority claims have been properly implemented.
 
### Tasks

1. Ensure we have an agent handler for getting audit data.
2. Ensure we have an agent handler for getting sigchain data.
3. Ensure we have tests testing the functionality of both handlers. 
4. Ensure that cancellablility is properly implemented for both handlers.

CDeltakai:
### Issues Fixed

* Fixes https://github.com/MatrixAI/Polykey/issues/865

### Tasks

- [x] 1. Ensure we have an agent handler for getting audit data.
- [x] 2. Ensure we have an agent handler for getting sigchain data.
- [x] 3. Ensure we have tests testing the functionality of both handlers.
- [x] 4. Ensure that cancellablility is properly implemented for both handlers.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
Hows the progress on this @CDeltakai ?

CMCDragonkai:
So this gives us the possibility giving us sigchain related commands so we can finally inspect the sigchain?

src/network/utils.ts
```diff
@@ -440,6 +440,7 @@ async function resolveHostnames(
 // TODO: review and fix the `toError` and `fromError` code here.
 //  Right now it's very basic and need fleshing out.
 function fromError(error: any) {
+  console.error(error);
```
aryanjassal: We don't leave any console statements in the code. Here, it was probably for debugging, so you can remove it. The linting process (which you can run locally by running `npm run lint` in the project) will detect this and warn you.

If you really need to log something out to the user, we instead use a logger to do so.

tests/nodes/agent/handlers/nodesAuditsGet.test.ts
```diff
@@ -0,0 +1,443 @@
...
+      },
+    }));
+  }
+
+  test('should get audit events', async () => {
+    // Generate valid AuditEventIds
+    const mockAuditEvents = generateMockAuditEvents(
+      100,
+      callProtectedGenerateAuditEventId,
+    );
```
aryanjassal: This is not how we use fast-check to generate mock data. Usually, we create an arbitrary which returns the mock data, then we use the arbitrary in jest to generate new data each time. It allows for higher customisability and even predictability when needed.

```ts
import { test } from '@fast-check/jest';
import fc from 'fast-check';

describe('example', () => {
  const randomNumberArb = fc.string({ min: 0, max: 65535 });

  test.prop([randomNumberArb])('using arbitraries', (num) => {
    // `num` will always be randomly generated now
  });
});
```

Using `test.prop` has other benefits, like allowing us to reduce or increase number of runs by setting `numRuns` to a value. It also allows for providing a seed to guarantee reproducible results.

```ts
test.prop([randomNumberArb], { numRuns: 1, seed: 1234 })('using arbitraries', (num) => {
  // `num` will always be the same value
});
```

In your test, you need to stop sampling a fast check arbitrary and instead create a variable which stores it. Then, use it as I have demonstrated in the examples. We can have a discussion about this if you're still unsure how to use fast-check to write tests.

tests/nodes/agent/handlers/nodesClaimsGet.test.ts
```diff
@@ -189,13 +188,155 @@ describe('nodesClaimsGet', () => {
...
+    // Verify that chainIds are in ascending order
+    let isSorted = true;
+    for (let i = 0; i < chainIds.length - 1; i++) {
+      if (chainIds[i] > chainIds[i + 1]) {
+        isSorted = false;
+        break;
+      }
+    }
+    expect(isSorted).toBe(true);
+  });
```
aryanjassal: I feel like writing two tests which are largely similar like this might get unmaintainable in the future, and is just a pain to read as well. Look into this to merge the tests into a single method.

https://jestjs.io/docs/api#testeachtablename-fn-timeout

This has been done in Polykey CLI like this:
https://github.com/MatrixAI/Polykey-CLI/blob/staging/tests/secrets/env.test.ts#L927-L955

aryanjassal (changes requested):
Just some minor changes, otherwise looks good.

tests/nodes/agent/handlers/nodesAuditsGet.test.ts
CDeltakai: I've taken a look at the other tests in the folder nodes/agent/handlers and it seems like none of the other tests within the folder use property-based testing using test.prop. Instead, the other tests also seem to just use scenario-based testing like I am doing right now.

Whether that's intentional or not I'm not sure but if I am to change the testing manner in nodesAuditsGet.test.ts to be property-based then it would seem like the rest of the tests would also need to follow suit.

CDeltakai: I've decided to hold off on making the suggested changes until it's confirmed that it's just the nodesAuditsGet.test.ts that need to be converted to property-based testing or that all the other tests in this folder also need to be made property-based.

However, some research tells me that the tests in nodes/agent/handlers are end-to-end tests in which property-based tests are unsuited for for a number of reasons.

aryanjassal: The reason other tests aren't property-based is because they don't use fast check. Only tests which use fast check need to use `test.prop`. As no other test is currently using fast-check, so you can only update the tests which are using it.

If you weren't using fast-check to generate mock data, you wouldn't need to use prop. Mixing of both `test.prop` and regular `test` can be seen in `tests/client/handler/vaults.test.ts`. There are probably other places this is done as well, but I'm most familiar with that file.

Note that you need to import it from `@fast-check/jest`, otherwise `test.prop` would be undefined (see top of thread).

tests/nodes/agent/handlers/nodesAuditsGet.test.ts
```diff
@@ -0,0 +1,465 @@
...
+          remoteNodeId: fc.string({ minLength: 1, maxLength: 20 }),
+          remoteHost: fc.ipV4(),
+          remotePort: fc.nat({ max: 65535 }),
+        }),
+        { minLength: 1, maxLength: 100 },
+      ),
+      fc.integer({ min: 1, max: 100 }),
+    ],
+    { numRuns: 3 },
+  )(
```
aryanjassal: I typically leave the number of runs as default (which is 100 iirc) or 1 if I just need randomised data but still need to run the test once. We only change this if we have a valid reason. If there is a valid reason for this, then this is fine.

Also, I can see that the arbitrary is being reused a lot with some minor changes. You can abstract it away into a utility function and pass it parameters normally.

```ts
// An arbitrary which can take a value to change behaviour
const randomEventsArb = (minLength: number) => fc
  .array(
    fc.record({
      remoteNodeId: fc.string({ minLength: 1, maxLength: 20 }),
      remoteHost: fc.ipV4(),
      remotePort: fc.nat({ max: 65535 }),
    }),
    { minLength: minLength, maxLength: 100 },
  );

// An arbitrary which doesn't need customisation
const randomEventsBasicArb = fc
  .array(
    fc.record({
      remoteNodeId: fc.string({ minLength: 1, maxLength: 20 }),
      remoteHost: fc.ipV4(),
      remotePort: fc.nat({ max: 65535 }),
    }),
    { minLength: 1, maxLength: 100 },
  );

// Test here
test.prop([randomEventsArb(1)])('test', () => {});

// Note you don't need to call the basic one as it's just a value
test.prop([randomEventsBasicArb])('test', () => {});

// If the arbs are simple enough, you can define them where relevant
test.prop([
  randomEventsBasicArb,
  fc.integer({min: 0, max: 65535 }),
  ])
  (
    'test', 
    () => {}
  );
```

Usually, these arbitraries are defined in a utils.ts file in the testing directory. Take a look at this from Polykey.

Using arbitraries in tests: https://github.com/MatrixAI/Polykey/blob/8e14922f8fe084f8ee406e11fe06c6879c5821e5/tests/client/handlers/vaults.test.ts#L424-L426
Defining arbitraries: https://github.com/MatrixAI/Polykey/blob/8e14922f8fe084f8ee406e11fe06c6879c5821e5/tests/utils/fastCheck.ts#L43-L59

Make this change across all fast-check tests that you did. Feel free to ask me after you make one change so you know what you are doing is correct.

CDeltakai:
It has been clarified that all seek parameters in relation to pagination should be by default **inclusive**, not exclusive. I'll have to go back and make changes to any RPC handlers with the incorrect assumption.

CDeltakai:
There is a test failure in Discovery.test.ts that for the moment can't be easily resolved. I've commented out the offending elements for now but will need to be addressed later.

PR #872, opened 2025-02-10
Adding cancellation to background handlers to prevent agent from being held open

### LINKED ISSUE #185: `polykey agent stop` command not terminating properly
### Describe the bug

Looking through the logs, looks like the hook for shutting down never triggered. The log cuts out mid line like the process was terminated. That must've been when you had to manually kill it.

This might be a mac specific problem.

### To Reproduce

1. Run the agent
2. Wait for a few minutes
3. Run `polykey agent stop`

Test on Linux version that same way we got the error on Mac to see if it's an OS isolated incident or not.

### Expected behavior

We handle most signals to trigger stopping the agent. The only thing that should really kill it like this is a `SIGKILL` signal.

### Platform

* Device: Mac
* OS: Sonoma 14 Terminal Version 2.14
* Version: `polykey-cli-0.3.1-darwin-universal`

### Additional context

* Polykey logs can be found here: [**Polykey-CLI#183**](https://github.com/MatrixAI/Polykey-CLI/issues/183)
* Propagated problem of connections being leaked causing push error flow being discussed here: [**Polykey-CLI#198**](https://github.com/MatrixAI/Polykey-CLI/issues/198)

### Notify maintainers

@tegefaulkes

aryanjassal:
### Description

Often, some background task handlers hold the agent open when we are trying to shutdown. The only solution is to send a SIGKILL to the process to forcefully shut it down. This should not happen.

### Issues Fixed

* Fixes https://github.com/MatrixAI/Polykey-CLI/issues/185 (FIX ENG-322)

### Tasks

- [x] 1. Add cancellation to `Discovery.discoverVertexHandler`
- [x] 2. Add cancellation to `NodeManager.syncNodeGraphHandler`
- [x] 3. Add cancellation to `NodeManager.checkConnectionsHandler`
- [x] 4. Add cancellation to `NodeManager.refreshBucketHandler`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/discovery/Discovery.ts
+++ b/src/discovery/Discovery.ts
@@ -153,13 +153,11 @@ class Discovery {
     parent: GestaltIdEncoded | null,
   ) => {
     try {
-      this.logger.error('out here 1')
       await this.processVertex(
         vertex,
         lastProcessedCutoffTime ?? undefined,
         ctx,
       );
-      this.logger.error('out here 2')
       this.dispatchEvent(
         new discoveryEvents.EventDiscoveryVertexProcessed({
           detail: {
@@ -171,23 +169,18 @@ class Discovery {
     } catch (e) {
       // We need to reschedule if the task was cancelled due to discovery domain stopping
       if (e === discoveryStoppingTaskReason) {
-        this.logger.error('out here 3')
         // We need to recreate the task for the vertex
         const vertexId = gestaltsUtils.decodeGestaltId(vertex);
         if (vertexId == null) {
           never(`failed to decode vertex GestaltId "${vertex}"`);
         }
-        this.logger.error('out here 4')
         await this.scheduleDiscoveryForVertex(
           vertexId,
           undefined,
           undefined,
           gestaltsUtils.decodeGestaltId(parent ?? undefined),
           true,
-          undefined,
-          ctx,
         );
-        this.logger.error('out here 5')
         return;
       }
       // Aborting a duplicate task is not an error
@@ -213,7 +206,9 @@ class Discovery {
   /**
    * This handler is run periodically to check if nodes are ready to be rediscovered
    */
-  protected checkRediscoveryHandler: TaskHandler = async (ctx: ContextTimed) => {
+  protected checkRediscoveryHandler: TaskHandler = async (
+    ctx: ContextTimed,
+  ) => {
     await this.checkRediscovery(
       Date.now() - this.rediscoverVertexThresholdTime,
       undefined,
@@ -337,26 +332,15 @@ class Discovery {
   /**
    * Queues a node for discovery. Internally calls `pushKeyToDiscoveryQueue`.
    */
+  @ready(new discoveryErrors.ErrorDiscoveryNotRunning())
   public async queueDiscoveryByNode(
     nodeId: NodeId,
     lastProcessedCutoffTime?: number,
-    ctx?: Partial<ContextTimedInput>,
-  ): Promise<void>;
-  @ready(new discoveryErrors.ErrorDiscoveryNotRunning())
-  @timedCancellable(true)
-  public async queueDiscoveryByNode(
-    nodeId: NodeId,
-    lastProcessedCutoffTime: number | undefined,
-    @context ctx: ContextTimed,
   ): Promise<void> {
     await this.scheduleDiscoveryForVertex(
       ['node', nodeId],
       undefined,
       lastProcessedCutoffTime,
-      undefined,
-      undefined,
-      undefined,
-      ctx,
     );
   }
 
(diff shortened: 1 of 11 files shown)


aryanjassal:
There were some weird issues happening when I attempt to link Polykey's dist directory to Polykey CLI's `node_modules` for testing the changes. This was resolved by simply copying the `dist` for Polykey instead of linking it. I believe this was due to the libraries not aligning when using linking.

src/discovery/Discovery.ts
tegefaulkes, line 1: Any functions that just modify the database doesn't need cancellation. So the schedule methods don't need cancellation. In fact cancelling these like this will likely cause problems.

In fact, if a method is just getting data or modifying the database then we don't need to add cancellation to it. These operations will very rarely take a long time. If they used a transaction lock however we'd want to be able to cancel that.

src/discovery/Discovery.ts
```diff
@@ -1034,10 +1117,16 @@ class Discovery {
   public async checkRediscovery(
     lastProcessedCutoffTime: number,
     tran?: DBTransaction,
+    ctx?: Partial<ContextTimedInput>,
+  ): Promise<void>;
+  public async checkRediscovery(
+    lastProcessedCutoffTime: number,
```
tegefaulkes: Missing decorator here?

src/discovery/Discovery.ts
tegefaulkes: You should make sure that any tran.lock() calls take the CTX as well.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1745,19 +1784,39 @@ class NodeConnectionManager {
...
+    nodeId: NodeId,
+    @context ctx: ContextTimed,
+  ): Promise<void> {
     const targetNodeIdString = nodeId.toString() as NodeIdString;
     const connectionsEntry = this.connections.get(targetNodeIdString);
     if (connectionsEntry == null) {
       throw new nodesErrors.ErrorNodeConnectionManagerConnectionNotFound();
     }
+    const { p: abortP, rejectP: triggerAbort } = utils.promise<void>();
+    const abortHandler = () => {
```
tegefaulkes: If a Promise only can reject then it's type should be `never`. This will give better type checking.

Also when using deconstructed promise the namiming is consistently `{p: xP, rejectP: rejectXP, resolveP: resolveXP} = ...`

src/nodes/NodeConnectionManager.ts
```diff
@@ -1745,19 +1784,39 @@ class NodeConnectionManager {
...
+      triggerAbort(ctx.signal.reason);
+    };
+    if (ctx.signal.aborted) {
+      triggerAbort(ctx.signal.reason);
+    } else {
```
tegefaulkes: Check if its aborted and then just throw immediately. Then you can create the promise, abort handler and hook it all up. It should come out cleaner that way.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1745,19 +1784,39 @@ class NodeConnectionManager {
...
+      ctx.signal.addEventListener('abort', abortHandler, { once: true });
+    }
     try {
-      return await connectionsEntry.authenticatedP;
+      return await Promise.race([connectionsEntry.authenticatedP, abortP]);
```
tegefaulkes, line 1822: Just a note, the reason `abortP` promise type should be `never` is so that this `Promise.race` can only return the `authenticatedP` type. It's *fine* here because both promises are void types. But if `authenticatedP` was any other type then the return type would complain here.

Basically the `never` type  means you can never pass this point unless it's an error. The `utils.never()` does this too. It throws an error and the return type is never because it never returns. This means that any code after it *should* be dead code in the eyes of the typescript compiler.

tegefaulkes (commented):
You've added `@timedCancellable(true)` all over the place. Keep in mind that without a default timeout provided then it will default to infinite timeout. In most cases we don't want this. Check what the existing `@timedCancellable` usage for each file is doing.

src/discovery/Discovery.ts
aryanjassal: A fair amount of calls utilise `tran.lock()`, including the ones in other domains like `vaults`, etc.

For this PR, I'll focus only on the discovery domain and the nodes domain. I'll make a note to fix other domains later when I next work on it.

aryanjassal: This is introducing more issues than it's solving, so for the time being, this will not be implemented. Perhaps in the future.

CMCDragonkai: You want to add an issue? Or I think there's existing issue to this.

Also I'd like you learn from @Sooryasanand regarding durable objects pattern used in CF. It's a useful to share notes on how a task manager is created from state in the DB and a durable object alarm.

PR #883, opened 2025-04-01
Fixing leaked RPC authentication error

### LINKED ISSUE #376: Leaked RPC authentication error in polykey agent
### Describe the bug

After running my agent for a while, it eventually crashed with the following log message.
```
Mar 24 00:38:30 matrix-dell-34xx-zenith polykey[4792]: ErrorNodeConnectionManagerRPCDenied: RPC call was denied due to being unauthenticated
```
This indicates that errors are leaking from their context and crashing the entire agent, which shouldn't happen.

### To Reproduce

1. Run the agent for a while
2. Observe error

### Expected behavior

The agent shouldn't crash. It should either be handled internally, or printed to stdout/stderr as a log message.

### Platform

 - Device: Dell Precision 3480
 - OS: NixOS
 - Version: [e.g. 22]

### Notify maintainers

@tegefaulkes

aryanjassal:
### Description

Sometimes, it is possible for a system to finish authentication before another system, and it can start sending RPC messages before the other system has fully processed the authentication. This will crash the program. This PR aims to resolve the crash.

### Issues Fixed

* Fixes https://github.com/MatrixAI/Polykey-CLI/issues/376 ( REF ENG-560)

### Tasks

- [x] 1. Investigate three-step authentication and duplex rework approaches
- [x] 2. Rework network authentication to apply these changes

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


aryanjassal:
So far I have made some changes which should help the resources to be freed upon an abort. Basically, this should again help the agent to shut down without holding itself open for too long. I will merge these changes alongside the ones in the issue.

aryanjassal:
Theoretically, reworking network authentication to be duplex streams seems to be the more reliable approach here. I will rework the network authentication step for this.

Sometimes, if the authentication is rejected, then the node will not re-attempt any connections at all. There needs to be retry limit and maybe some nodes like the seednode can have a perpetual re-authentication attempt. I will make an issue tracking this spec, and after the rework, might just implement it in this PR as well.

CMCDragonkai:
We are still using unary calls?

aryanjassal:
I'm still reading and understanding the code, so my understanding might not be complete yet.

Yes. At the time of adding a new connection, both nodes make a RPC call to the other node with a unary call to the authentication handler, which triggers the reverse authentication on the receiver node. If both nodes return a successful reverse authentication response, then the connection becomes authenticated.

image

The trigger for forward authentication happens simultaneously when the connection gets added. However, they both can end at different times. This can lead to a race condition where the connection is considered authenticated from a node, and sends RPC requests to the node, but the other node has not yet completed the authorisation, leading to this issue.

With the duplex streaming rework, this would be more like a handshake instead, and subsequent code will only run after the authentication has either been successful or has failed.

CMCDragonkai:
You should review your understanding with @tegefaulkes first.

aryanjassal:
image

This is my plan on implementing the duplex stream for authentication. In the RPC method for authentication, the connecting side will call this. It will give a forward auth message token, which the other node will apply the reverse auth to verify the authorisation. Then, after accepting, the node will  return another token from the other node which will be reverse-authenticated by the current node. Once both nodes have returned a success authentication result, then the connection will be authenticated.

This prevents any race conditions which can happen from state desync, as the RPC method can only resolve once both the sides have authenticated their connections.

CMCDragonkai:
> image
> 
> This is my plan on implementing the duplex stream for authentication. In the RPC method for authentication, the connecting side will call this. It will give a forward auth message token, which the other node will apply the reverse auth to verify the authorisation. Then, after accepting, the node will return another token from the other node which will be reverse-authenticated by the current node. Once both nodes have returned a success authentication result, then the connection will be authenticated.
> 
> This prevents any race conditions which can happen from state desync, as the RPC method can only resolve once both the sides have authenticated their connections.

Can this be done more optimally? Is there a more efficient way to achieve this with minimal back and forth? Is there some crypto tricks here to use - including taking advantage of context of the comms?

aryanjassal:
> Can this be done more optimally? Is there a more efficient way to achieve this with minimal back and forth? Is there some crypto tricks here to use - including taking advantage of context of the comms?

What do you mean by crypto tricks? I asked brian and even he was unsure about how we can do this more optimally. At minimum, we need the following steps in the caller side:
- Send forward auth message
- Send reverse auth message
- Ack and close

The RPC handler also does similar steps.
- Send reverse auth message
- Send forward auth message
- Ack and close

This is the minimum requirement of this communication. How else would you propose we handle this?

aryanjassal:
https://github.com/MatrixAI/Polykey/actions/runs/14396425288/job/40372816693

@brynblack the linting job fails here, but the workflow still passes. This shouldn't happen.

aryanjassal:
This PR is otherwise done. I have asked Brian to do a final review before merging it.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1808,21 +1904,33 @@ class NodeConnectionManager {
...
-    ctx.signal.throwIfAborted();
     const targetNodeIdString = nodeId.toString() as NodeIdString;
     const connectionsEntry = this.connections.get(targetNodeIdString);
     if (connectionsEntry == null) {
       throw new nodesErrors.ErrorNodeConnectionManagerConnectionNotFound();
     }
+    // if (!connectionsEntry.authenticateComplete) {
+    //   this.initiateForwardAuthenticate(nodeId);
+    // }
+    // return await utils.valueOrTimeout(connectionsEntry.authenticatedP, ctx);
```
CMCDragonkai: Commented out code?

src/utils/utils.ts
```diff
@@ -547,6 +547,29 @@ async function importFS(fs?: FileSystem): Promise<FileSystem> {
...
+  try {
+    return await Promise.race([prom, abortP]);
+  } catch (e) {
+    Error.captureStackTrace(e);
+    throw e;
+  } finally {
+    ctx.signal.removeEventListener('abort', abortHandler);
+  }
+}
+
```
CMCDragonkai: This needs better documentation and/or naming.

src/nodes/agent/errors.ts
aryanjassal, line 1: I can see that this file does not follow the standard error naming scheme. It extends `ErrorAgent` but the error name starts with `ErrorNodes`. I'm not sure if this is intentional or not, so I left it to discuss over the review.

src/nodes/NodeConnection.ts
```diff
@@ -169,12 +169,9 @@ class NodeConnection {
...
       connectionKeepAliveIntervalTime,
-      connectionKeepAliveTimeoutTime = config.defaultsSystem
-        .nodesConnectionIdleTimeoutTimeMin,
-      connectionInitialMaxStreamsBidi = config.defaultsSystem
-        .nodesConnectionInitialMaxStreamsBidi,
-      connectionInitialMaxStreamsUni = config.defaultsSystem
-        .nodesConnectionInitialMaxStreamsUni,
+      connectionKeepAliveTimeoutTime,
+      connectionInitialMaxStreamsBidi,
+      connectionInitialMaxStreamsUni,
```
tegefaulkes, line 174: Why are we removing the defaults here?

src/nodes/NodeConnectionManager.ts
```diff
@@ -1659,14 +1660,64 @@ class NodeConnectionManager {
       throw new nodesErrors.ErrorNodeConnectionManagerConnectionNotFound();
     }
     // Need to make an authenticate request here. Get the connection and RPC.
+    let reader: ReadableStreamDefaultReader | undefined;
+    let writer: WritableStreamDefaultWriter | undefined;
```
tegefaulkes: You're throwing away all type checking with this. Be sure to apply the proper type with `ReadableStreamDefaultReader<T>`. The default generic assumes `any`.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1659,14 +1660,64 @@ class NodeConnectionManager {
...
+
+        // Write the forward authentication message from this node
+        await writer.write(authenticateMessage);
+        const forwardMessageResultPair = await utils.valueOrTimeout(
+          reader.read(),
+          ctx,
+        );
+        const forwardMessageResult:
+          | SuccessMessage
+          | NodesAuthenticateConnectionMessage = forwardMessageResultPair.value;
```
tegefaulkes: It's really weird you have to apply the type here. Likely because you're not properly defining the types above with the `let` statement.

Also, with a reader you should *ALWAYS* check the `done` property. In this case you're assuming that `value` is defined here which isn't true. So always check that the done and if the stream finished early then throw an explicit protocol error.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1677,6 +1728,9 @@ class NodeConnectionManager {
       connectionsEntry.authenticatedForward = AuthenticatingState.FAIL;
       this.authenticateFail(targetNodeIdString, err);
       return;
+    } finally {
+      await writer?.abort();
+      await reader?.cancel();
```
tegefaulkes: Does this cause a problem if the writer and reader are already completed? I think cancellation should only be done in the case of an error.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1752,8 +1806,12 @@ class NodeConnectionManager {
     }
     const existingAuthenticate =
       this.activeForwardAuthenticateCalls.get(nodeIdString);
+
     // If it exists in the map then we don't need to start one and can just return
     if (existingAuthenticate != null) return;
+    // TODO: retry if fail
+    // create retry authenticatoin on new connections if the existing connection
+    // had a failed authentication. put a limit - soemwhere around 3.
```
tegefaulkes: Clean up your TODO and notes.

src/nodes/NodeConnectionManager.ts
tegefaulkes, line 1: Don't forget to update the docblocks when updating chunks of code. The overall functionality of authentication has changed so their docblocks should be updated.

src/nodes/agent/handlers/NodesAuthenticateConnection.ts
tegefaulkes, line 1: Didn't I mention to put all of this logic into a public handle function in the `NodeConnectionManager`? In fact that is what the `handleAuthentication` method used to be. The logic here should be the `handleAuthentication` method. If you need a separate function that handles a single authentication step then make a new protected method for that.

src/nodes/NodeConnectionManager.ts
```diff
@@ -1775,6 +1831,46 @@ class NodeConnectionManager {
...
+      if (
+        connectionEntry.authenticatedReverse === AuthenticatingState.SUCCESS
+      ) {
+        this.authenticateSuccess(requestingNodeIdString);
+      }
+    } else {
+      connectionEntry.authenticatedReverse = AuthenticatingState.FAIL;
+      this.authenticateFail(requestingNodeIdString, new Error('temp'));
+    }
+  }
```
tegefaulkes: temp error? I'm seeing a few first blush simple review fixes. Did you review your work before asking me to look at it? You should've caught this already before asking for review.

src/nodes/agent/errors.ts
```diff
@@ -8,6 +8,7 @@ class ErrorAgentNodeIdMissing<T> extends ErrorAgent<T> {
   exitCode = sysexits.UNAVAILABLE;
 }
 
+// NOTE: NEEDS REVIEW
 class ErrorNodesConnectionSignalRequestVerificationFailed<
```
tegefaulkes: Yeah this is a little odd. This error should be within the `nodes/errors.ts`. I feel like this `agent/errors.ts` is a vestigial file from back when we combined some domains. It probably should be removed.

tests/nodes/NodeConnectionManager.test.ts
```diff
@@ -950,6 +950,13 @@ describe(`${NodeConnectionManager.name}`, () => {
         ),
       );
 
+      // Disabling authentication function
+      const initiateForwardAuthenticationSpy = jest
+        .spyOn(ncmLocal.nodeConnectionManager, 'initiateForwardAuthenticate')
+        .mockImplementation(() => {
+          // Do nothing
+        });
+
```
tegefaulkes, line 1092: Did you re-base on the ESM migration? How jest works with mocks  has changed slightly. I get the feeling this should complain about types now.

tests/nodes/NodeConnectionManager.test.ts
```diff
@@ -966,13 +973,17 @@ describe(`${NodeConnectionManager.name}`, () => {
         connection?.connection.rpcClient.unaryCaller('dummyMethod', {}),
       ).rejects.toThrow(nodesErrors.ErrorNodeConnectionManagerRPCDenied);
 
+      // Restore the original authentication functionality
+      initiateForwardAuthenticationSpy.mockRestore();
+
```
tegefaulkes, line 1111: Restoring a mock should always be done within the beforeEach first thing. You should call `jest.restoreAllMocks()` or whatever it's called to prevent mocks cross contaminating between tests.

tegefaulkes (changes requested):

src/nodes/agent/handlers/NodesAuthenticateConnection.ts
aryanjassal: I did this because there needs to be a back-and-forth communication here. I felt that creating an async generator public method to deal with this back and forth felt a bit weird and overkill. I'll make the change to use AsyncGenerators for this if my initial approach was incorrect.

src/nodes/NodeConnectionManager.ts
aryanjassal: I was actually going to push up a commit for this but just realised that it doesn't work as intended, so I didn't do that just yet. I should be almost done with that tho.

src/nodes/NodeConnectionManager.ts
aryanjassal: I did go through it once before, I might've just missed this on accident.

tests/nodes/NodeConnectionManager.test.ts
aryanjassal: I'll also do that in the `beforeEach`, but this mock restore needs to be here as the first mock prevents the authentication step from running, as the authentication is triggered at the moment a connection is created. To confirm it will trigger in subsequent connections, I needed to restore the mock before the rest of the code could run.

src/nodes/NodeConnectionManager.ts
aryanjassal: The abort and cancel are idempotent. However, `writer.abort()` sends an error signal in the writable stream where `writer.close()` just gracefully closes it. I think I should switch to `writer.close()` for graceful shutdowns and `writer.abort()` during errors. Doing `reader.cancel()` means we can discard all stored data, but `reader.close()` means we still might reopen the stream for more data. Leaving `reader.cancel()` seems fine to me in this case.

**MDN Docs Reference**
- https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel
- https://developer.mozilla.org/en-US/docs/Web/API/WritableStream/abort
- https://developer.mozilla.org/en-US/docs/Web/API/WritableStream/close

src/nodes/NodeConnection.ts
aryanjassal: It probably doesn't show up in your IDE but these lines are bright red for me. The error message says that default values for parameters can only be set in the base implementation, not in any overloads. I made sure that the defaults are there in the base definition and everything works as intended.

src/nodes/agent/errors.ts
aryanjassal: These errors relate to the RPC handlers in the nodes domain, which is why I thought it was intentional. In this case, I'll remove it and merge it into the node errorrs.

tests/nodes/NodeConnectionManager.test.ts
aryanjassal: I did rebase on staging with the ESM migration, and I did it again when you left the comment. My repo is up to date, and the linter isn't really throwing any errors. The tests also don't break typescript building so we can't rely on that either.

I'll look up the mockimplementation documentation of the version we are currently using to make sure.

aryanjassal: https://jestjs.io/docs/jest-object#jestspyonobject-methodname

The note in this section agrees with my approach, so I believe my approach should be fine.

src/nodes/agent/errors.ts
aryanjassal: I actually feel like the separation of RPC-specific errors and other node errors was intentional, so I am keeping it that way. I'll rename the errors so it makes more sense in this context.

src/nodes/NodeConnectionManager.ts
aryanjassal: I went through the entire `NodeConnectionManager` file and have updated the documentation wherever I could.

src/utils/utils.ts
aryanjassal: I've changed the name to `resultOrAbort` and added documentation explaining its usage. It's basically used to race promises against a context where PromiseCancellable can't be returned, like `reader.read()`, to make sure it would still respect cancellation.

src/nodes/NodeConnectionManager.ts
aryanjassal: Had to get some feedback, removed it now.

src/nodes/utils.ts
```diff
@@ -18,6 +18,7 @@ import type {
   NodesAuthenticateConnectionMessageBasicPublic,
   NodesAuthenticateConnectionMessageNone,
 } from './agent/types.js';
+import type { ContextTimed } from '@matrixai/contexts';
```
aryanjassal: The import order seems incorrect here. Go through the file and fix it.

tests/nodes/NodeConnectionManager.test.ts
```diff
@@ -928,6 +929,138 @@ describe(`${NodeConnectionManager.name}`, () => {
         nodesErrors.ErrorNodeManagerAuthenticationFailed,
       );
     });
+    test('can reauthenticate if previous reverse auth failed', async () => {
```
aryanjassal, line 932: This test takes over 5 seconds to resolve. That is also approximately the default ctx timer duration. The other test after this takes 200ms to resolve. This tells me that that a resource is probably not being freed properly.

CMCDragonkai:
Is this before or after esm migration?

aryanjassal:
> Is this before or after esm migration?

Do you mean this issue or the rpc version bump? This issue was encountered prior to ESM migration. The RPC version bump is after ESM.

src/nodes/NodeManager.ts
```diff
@@ -132,15 +126,7 @@ class NodeManager {
...
-    // Don't use defaults like this
-    // if a default is to be used
-    // provide it directly
-
-    await this.refreshBucket(
-      bucketIndex,
-      this.connectionConnectTimeoutTime,
-      ctx,
-    );
+    await this.refreshBucket(bucketIndex, undefined, ctx);
```
tegefaulkes, line 129: Why was the timeout parameter removed here?

src/nodes/agent/errors.ts
tegefaulkes, line 1: This whole errors file can be rolled into the nodes domain errors file. I think it existed before handlers were merged into the nodes domain. I think it no longer needs to be separate. The Error name should be more along the lines of `failing to verify connection identity`.

src/utils/utils.ts
```diff
@@ -547,6 +547,42 @@ async function importFS(fs?: FileSystem): Promise<FileSystem> {
...
+ * rejects with the reason. Otherwise, the original value is returned upon
+ * resolving.
+ * @param prom The promise to resolve
+ * @param ctx The ctx with which to race the promise against
+ * @returns A promise which resolves to the prom value or errors if ctx aborts
+ */
+async function resultOrAbort<T>(
+  prom: Promise<T>,
+  ctx: ContextTimed,
+): Promise<T> {
```
tegefaulkes: I feel naming anything `XorY` is really bad naming. This races a promise with a abort signal so why not call it `raceSignal` and have it take a promise and an `AbortSignal` as the input?

tegefaulkes (approved):
small stuff but otherwise good.

src/nodes/NodeManager.ts
aryanjassal: It was removed from here as it is already the default value for the `refreshBucket` method, and the comment said to better handle default values, so I assumed that it has already been resolved, so I updated this chunk of code.

src/utils/utils.ts
aryanjassal: Yeah, I was agonising over the naming for a bit lol. I came up with `raceProm` or `raceContext` and that sounded a bit weird. Doing it with an `AbortSignal` and calling it `raceSignal` sounds better, so I'll use this instead.

src/nodes/agent/errors.ts
aryanjassal: Added `ErrorNodeConnectionInvalidIdentity` to `nodes/errors.ts` and removed `nodes/agent/errors.ts`.

aryanjassal:
I have addressed all review comments and added all the required features. I have also rebased it to latest staging, so this PR is now basically done. After the checks pass, this can be merged.

PR #906, opened 2025-06-16
Feature eng 370 network segregationauthentication token logic

### LINKED ISSUE #779: Network Segregation/Authentication Token Logic

tegefaulkes:
### Description

This PR makes sure that all the logic pertaining to creating, verifying and RPC handlers relating to the network claims is added.

### Issues Fixed

* Fixes #779
* REF ENG-370

### Tasks

- [x] 1. Fill out tasks

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


src/gestalts/types.ts
```diff
@@ -14,7 +14,7 @@ import type {
 } from '../claims/payloads/index.js';
 import type { ProviderPaginationToken } from '../identities/types.js';
 
-const gestaltActions = ['notify', 'scan', 'claim'] as const;
+const gestaltActions = ['notify', 'scan', 'claim', 'join'] as const;
```
tegefaulkes, line 18: I'm calling the permission to join a network `join`. Do we want a more descriptive name for this @CMCDragonkai ?

src/nodes/NodeManager.ts
```diff
@@ -1535,9 +1933,111 @@ class NodeManager<Manifest extends AgentClientManifestNodeManager> {
...
+          !fullySignedToken.verifyWithPublicKey(
+            this.keyRing.keyPair.publicKey,
+          ) ||
+          !fullySignedToken.verifyWithPublicKey(requestingNodePublicKey)
+        ) {
+          throw new claimsErrors.ErrorDoublySignedClaimVerificationFailed();
+        }
+        // TODO: verify network and authority claim is correct
+        // Ending the stream
+        return fullySignedToken;
```
tegefaulkes: Lingering TODO

src/nodes/NodeManager.ts
```diff
@@ -1598,8 +2098,9 @@ class NodeManager<Manifest extends AgentClientManifestNodeManager> {
     let success = false;
     for await (const [_, claim] of this.sigchain.getSignedClaims({})) {
       try {
-        assertClaimNetworkAccess(claim.payload);
+        claimNetworkAccessUtils.assertClaimNetworkAccess(claim.payload);
       } catch {
+        // FIXME: check error type.
         continue;
```
tegefaulkes: Lingering fixme

src/nodes/NodeManager.ts
```diff
@@ -1609,8 +2110,11 @@ class NodeManager<Manifest extends AgentClientManifestNodeManager> {
...
       );
       try {
-        assertClaimNetworkAuthority(tokenNetworkAuthority.payload);
+        claimNetworkAuthorityUtils.assertClaimNetworkAuthority(
+          tokenNetworkAuthority.payload,
+        );
       } catch {
+        // FIXME: check error type.
         continue;
       }
```
tegefaulkes: Lingering fixme

tegefaulkes:
Alright, all good to go. Once this is merged the next steps are.
1. Updating the agent and seed nodes to support public networks like `mainnet.polykey.com` and `testnet.polykey.com` using the `ClaimNetworkAccess` to authenticate.
2. Integrate the changes into PKE to create the first proper private network.
3. More robust authentication logic.

PR #908, opened 2025-06-17
Added token verification and signing for login

aryanjassal:
### Description

This PR implements a protocol for taking in a signed JWT, verifying it, then adding additional data and returning the data alongside the original token back to the caller.

This is used in a protocol to securely provide identity proof to a platform. This is not used to access any protected resources like granting RPC access.

### Issues Fixed

* Addresses https://github.com/MatrixAI/Polykey-CLI/issues/408

### Tasks

- [x] 1. Create `AuthSignToken` RPC methods
- [ ] 2. Add tests confirming this behaviour

### Final checklist

* [ ] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/AuthSignToken.ts
+++ b/src/client/handlers/AuthSignToken.ts
@@ -28,11 +28,19 @@ class AuthSignToken extends UnaryHandler<
     // Get and verify incoming node
     const inputToken = { payload: input.payload, signatures: input.signatures };
     const incomingToken = Token.fromEncoded<IdentityRequestData>(inputToken);
+    if (!('publicKey' in incomingToken.payload)) {
+      throw new clientErrors.ErrorAuthenticationInvalidToken(
+        'Input token does not contain public key',
+      );
+    }
     const incomingPublicKey = Buffer.from(
       incomingToken.payload.publicKey,
+      'base64url',
     ) as PublicKey;
     if (!incomingToken.verifyWithPublicKey(incomingPublicKey)) {
-      throw new clientErrors.ErrorAuthenticationInvalidToken();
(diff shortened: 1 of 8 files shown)


src/client/handlers/AuthSignToken.ts
```diff
@@ -0,0 +1,63 @@
...
+    input: ClientRPCRequestParams<TokenIdentityRequest>,
+  ): Promise<TokenIdentityResponse> => {
+    const { keyRing }: { keyRing: KeyRing } = this.container;
+
+    // Get and verify incoming node
+    const inputToken = { payload: input.payload, signatures: input.signatures };
+    const incomingToken = Token.fromEncoded<TokenIdentityRequest>(inputToken);
+    const inputPayload = tokensUtils.parseTokenPayload<IdentityRequestData>(
+      inputToken.payload,
+    );
```
tegefaulkes: `Token.fromEncoded` does all of the parsing for you. There's no need to parse the payload again.

tegefaulkes (changes requested):
Small thing to point out but otherwise fine.

src/client/handlers/AuthSignToken.ts
aryanjassal: The type was a bit mangled as I was using the wrong one here, so I've fixed that now.

aryanjassal:
This PR looks good to go now. Merging.

aryanjassal:
As discussed with roger, the functionality will be merged first to avoid blocking PKE work, then I will write tests in another PR later.

PR #918, opened 2025-07-01
`vaultsSecretsEnv` no longer cancels on missing secret

aryanjassal:
### Description

Previously, `vaultsSecretsEnv` used to cancel the RPC if an error was encountered, disallowing any new secrets to be obtained. This PR will instead return errors as messages, so it is upto the client to parse the message instead of an error being thrown by the RPC.

### Issues Fixed

* Linked to https://github.com/MatrixAI/Polykey-CLI/issues/429

### Tasks

- [x] 1. Make `vaultsSecretsEnv` continue on error/failure
- [x] 2. Update tests to align with behaviour change

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/client/handlers/VaultsSecretsEnv.ts
+++ b/src/client/handlers/VaultsSecretsEnv.ts
@@ -5,12 +5,11 @@ import type {
   ClientRPCRequestParams,
   ClientRPCResponseResult,
   SecretIdentifierMessage,
-  SecretContentMessage,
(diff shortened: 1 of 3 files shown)


tegefaulkes (approved):
Looks fine, nothing stands out.

Issue #923, opened 2025-07-10
Allow agent to join a network using network access tokens


aryanjassal:
### Specification

The Polykey agent needs a way to join a network. The target network can either be a public or a private one. To do this, the agent will perform the following steps:

1. Perform DNS resolution to get details of the target seednode. An agent can connect to a seednode within a private network even while being unauthenticated.
2. Attempt to switch the currently-active network to the target network's seednode
3. If switching worked, then a network access claim already existed. If switching failed, attempt to create a new network access claim.

The network claiming and switching should automatically handle stuff like switching authentication and connection code to align with the new network.

The network join can be triggered at any time in the agent's lifecycle. The two main candidates for this are the `polykey nodes join` command, which can allow a running agent to connect to a said network, or `polykey agent start --network` to start the agent by attempting connection to the provided network.

### Additional context

A dummy private network has not been created yet. Thus, tests need to be written for testing the validity of the new logic. The test called `a node can join a public network without permissions` under `NodeManager.test.ts` contains logic for attempting connection to a private seednode and issuing network access and authority tokens. It can be used as a template for writing tests for this issue. The tests will need to mock DNS resolution, as the seednode is running locally.

### Tasks

1. Update the `NodeManager.claimNetwork` function to check for an existing claim first instead of creating duplicate access tokens
2. Update the `syncNodeGraph` and/or `syncNodeGraphHandler` to add this step
3. Create a RPC handler for the network joining logic

tegefaulkes:
This doesn't need to be a separate issue. The only change needed here is to have the `claimNetwork` function check if it already has a claim for that network and just return or throw an error if it does.<br><br>There's no need for a separate RPC handler for switching the network, in fact the `polykey network join` would do this anyway.

aryanjassal:
I created this issue to lay the groundwork for connecting to a new network. So, this issue involves `claimNetwork` checking the claim before attempting a claim, incorporates the network joining when starting up, and a RPC handler for doing the same.

Did I misunderstand the requirements of the feature?

PR #924, opened 2025-07-11
Allow agent to claim custom networks

aryanjassal:
### Description

The Polykey agent needs a way to join a network. The target network can either be a public or a private one. he network join can be triggered at any time in the agent's lifecycle. First, when starting, the `--network` flag should be able to change the network to one created by PKE as well as the testnet and mainnet. The second should be a RPC handler which will allow the network switching logic to happen while the agent is still running.

### Issues Fixed

* Addresses #923 (REF ENG-643)

### Tasks

- [x] 1. Update network access claim to prevent making duplicate claims for the same network
- [x] 2. Update `syncNodeGraph` to claim the network if a claim does not exist, otherwise switch the network to the target network
- [x] 3. Create a RPC handler wrapping this network switching behaviour

### Final checklist

* [ ] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -1,6 +1,6 @@
 import path from 'node:path';
 import url from 'node:url';
-import tsconfigJSON from './tsconfig.json' assert { type: "json" };
+import tsconfigJSON from './tsconfig.json' with { type: "json" };
 
 const projectPath = path.dirname(url.fileURLToPath(import.meta.url));
 
--- a/src/PolykeyAgent.ts
+++ b/src/PolykeyAgent.ts
@@ -774,6 +774,7 @@ class PolykeyAgent {
         });
         const initialNodesShortlist = initialNodes.slice(0, 3);
         await this.nodeManager.syncNodeGraph(
+          options.network,
           initialNodesShortlist,
           undefined,
           false,
--- a/src/client/callers/index.ts
+++ b/src/client/callers/index.ts
@@ -46,6 +46,7 @@ import nodesFind from './nodesFind.js';
 import nodesGetAll from './nodesGetAll.js';
 import nodesListConnections from './nodesListConnections.js';
 import nodesPing from './nodesPing.js';
+import nodesSyncGraph from './nodesSyncGraph.js';
 import notificationsInboxClear from './notificationsInboxClear.js';
 import notificationsInboxRead from './notificationsInboxRead.js';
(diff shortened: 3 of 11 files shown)


aryanjassal:
After discussing this issue in our sprint, we have highlighted some behavioural questions regarding network switching. This issue is only relevant for when the network is switched in the middle of the agent's lifecycle.

If we start the agent on a network, then we can start with creating new connections to the nodes in that network. There is no issues in this case. However, if we are switching from a network to another one, then how should the agent behave? The agent will need to end old connections and start new connections in the new network. What if a peer has a stream open to our node? If we choose to reject new connections and wait for the previous stream to end, then wouldn't it also suffer from similar issues that sometimes the agent shutdown suffers with, when some tasks or streams just refuse to cancel?

The tests might be able to catch some of these issues. We also need to plan for the behaviour of this function properly to avoid some of these hidden issues or make the system more brittle.

aryanjassal:
Brian and I ran into another question when working on this PR. The `network` at the startup is provided by an option in the polykey agent. However, this option has a type of `string | undefined`. The network is expected to always be a string. Thus, how do we want to handle an undefined network in this case? Do we disable the network claiming? Or do we prevent undefined values and instead force this to be a valid string, throwing if it is undefined when the agent is starting up?

Currently, I have just disabled the network claiming logic when the network is undefined, but I feel that this is not the optimal behaviour. I believe raising an error when starting the agent if the network is unpopulated seems to be the optimal solution here.

aryanjassal:
This PR is ready for merging. I will start working on the PKCLI side of this issue now before merging them both.

CMCDragonkai:
> Brian and I ran into another question when working on this PR. The `network` at the startup is provided by an option in the polykey agent. However, this option has a type of `string | undefined`. The network is expected to always be a string. Thus, how do we want to handle an undefined network in this case? Do we disable the network claiming? Or do we prevent undefined values and instead force this to be a valid string, throwing if it is undefined when the agent is starting up?
> 
> Currently, I have just disabled the network claiming logic when the network is undefined, but I feel that this is not the optimal behaviour. I believe raising an error when starting the agent if the network is unpopulated seems to be the optimal solution here.

Why is the network allowed to be undefined? Perhaps it means there is no network it's part of. In which case claiming a network means setting that network option.

